PR #106, opened 2022-04-28
[In progress] Server-side feed computation

jahooma:


common/contract.ts
```diff
@@ -20,7 +20,8 @@ export type FullContract<
   visibility: 'public' | 'unlisted'
 
   createdTime: number // Milliseconds since epoch
-  lastUpdatedTime: number // If the question or description was changed
+  lastBetTime?: number
```
akrolsmir, line 23: I like that we're tracking both "last bet time" and "last comment time" now (whereas before we just ignored it completely), but I still feel like we also want a single LastUpdated number that reflects "hey, something about this market has changed"; inclusive of bets, comments, and edits.

LastUpdated could be used for:
- "Last updated" sort in Manifold
- SEO (when the market was last updated)
- [API](https://www.notion.so/manifoldmarkets/API-Store-last-modified-time-in-contract-and-then-add-to-markets-5eefbc2d88bd479cbf635161987ff373) (for bot writers to decide whether to pull info from a market again)

I guess we could have all three fields (lastUpdated, lastBet, lastComment), if we need to pull them apart for the Feed.

jahooma: I guess I don't see lastUpdated as that meaningful. We don't use the field at all today. I think we should remove it until we actually want to use it somewhere.

If you feel strongly though, I can add it as a third field.

akrolsmir: Yes, I do feel strongly about it - please add it in! I think our API and SEO will be much better, plus the sort option will also be nice. The sort option would be where we'd use it immediately fwiw.

akrolsmir: And to start, we's just need to make sure to write to it anytime we track bets or comments. We can go back and add in other kinds of edits (description, close time, etc) later if we want, that's less important.

jahooma: Done. Though it doesn't make that much sense to me to lump every kind of update together.

jahooma:
Ok, I think there are no more blockers :-)

PR #107, opened 2022-04-28
Free daily market

IanPhilips:
Users get a free daily market and confetti when they create new markets! The confetti works for ALL owned, newly created markets - not just the free ones.

### DIFF
--- a/web/lib/firebase/contracts.ts
+++ b/web/lib/firebase/contracts.ts
@@ -121,6 +121,19 @@ export async function listAllContracts(): Promise<Contract[]> {
   return snapshot.docs.map((doc) => doc.data() as Contract)
 }
 
+export async function userHasCreatedContractToday(
+  userId: string
+): Promise<boolean> {
+  // uses utc time like the server
+  const todayAtMidnight = dayjs.utc().startOf('day').valueOf()
+  return listContracts(userId).then((contracts) => {
+    const todayContracts = contracts.filter((contract) => {
+      return contract.createdTime > todayAtMidnight
+    })
+    return todayContracts.length === 0
+  })
+}
+
 export function listenForContracts(
   setContracts: (contracts: Contract[]) => void
 ) {
--- a/web/lib/firebase/contracts.ts
+++ b/web/lib/firebase/contracts.ts
@@ -121,19 +121,6 @@ export async function listAllContracts(): Promise<Contract[]> {
   return snapshot.docs.map((doc) => doc.data() as Contract)
 }
 
-export async function userHasCreatedContractToday(
-  userId: string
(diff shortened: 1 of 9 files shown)


web/lib/firebase/contracts.ts
```diff
@@ -121,6 +121,19 @@ export async function listAllContracts(): Promise<Contract[]> {
   return snapshot.docs.map((doc) => doc.data() as Contract)
 }
 
+export async function userHasCreatedContractToday(
+  userId: string
+): Promise<boolean> {
+  // uses utc time like the server
+  const todayAtMidnight = dayjs.utc().startOf('day').valueOf()
+  return listContracts(userId).then((contracts) => {
```
jahooma, line 129: You could switch this to async/await style, e.g.
```
const contracts = await listContracts(userId)
...
```

web/lib/firebase/contracts.ts
```diff
@@ -121,6 +121,19 @@ export async function listAllContracts(): Promise<Contract[]> {
   return snapshot.docs.map((doc) => doc.data() as Contract)
 }
 
+export async function userHasCreatedContractToday(
+  userId: string
+): Promise<boolean> {
+  // uses utc time like the server
```
jahooma, line 127: Nice attention to detail!

web/pages/[username]/[contractSlug].tsx
```diff
@@ -119,6 +122,17 @@ export function ContractPageContent(props: FirstArgument<typeof ContractPage>) {
 
   return (
     <Page rightSidebar={rightSidebar}>
+      {user &&
+        contract.creatorId === user.id &&
+        Date.now() - contract.createdTime < 10 * 1000 && (
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:127]: After react renders, it won't necessarily know to render again when this condition changes. A more robust alternative would be to set a timeout that calls a set state. E.g. https://usehooks-ts.com/react-hook/use-timeout

web/lib/firebase/contracts.ts
jahooma: A very small style nit is to use capital letters at the start of a comment. At Google we even ended with a period, though I wouldn't enforce that.

jahooma (changes requested):
Awesome! This looks pretty good. Just a few comments.

web/pages/[username]/[contractSlug].tsx
Comment (IanPhilips) [reply web/pages/[username]/[contractSlug].tsx:127]: Hmmm the logic should check against a separate time, not the time since it loaded.

jahooma (approved):
Awesome, looks good. Thanks Ian!

PR #110, opened 2022-04-29
Show comments position

IanPhilips:
If a user has made a recent bet, they'll be prompted to comment on it. Otherwise, they'll be commenting with their position.

Ungrouped bets per user request in discord in the bets tab.

Added a tooltip for free response answer labels when a user comments as they can be unreadable when long.

### DIFF
--- a/web/components/feed/activity-items.ts
+++ b/web/components/feed/activity-items.ts
@@ -31,6 +31,8 @@ type BaseActivityItem = {
 
 export type CommentInputItem = BaseActivityItem & {
   type: 'commentInput'
+  bets: Bet[]
+  comments: Comment[]
 }
 
 export type DescriptionItem = BaseActivityItem & {
@@ -48,12 +50,13 @@ export type BetItem = BaseActivityItem & {
   bet: Bet
   hideOutcome: boolean
   smallAvatar: boolean
+  hideComment?: boolean
 }
 
 export type CommentItem = BaseActivityItem & {
   type: 'comment'
   comment: Comment
-  bet: Bet | undefined
+  betsBySameUser: Bet[]
   hideOutcome: boolean
   truncate: boolean
   smallAvatar: boolean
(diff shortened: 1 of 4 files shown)


web/components/feed/feed-items.tsx
```diff
@@ -244,18 +311,68 @@ export function CommentInput(props: { contract: Contract }) {
...
+  noFloorShares = Math.floor(noShares)
+
+  const userPosition = yesFloorShares || noFloorShares
+  const { saleValue } = calculateCpmmSale(
+    contract as FullContract<CPMM, Binary>,
+    yesShares || noShares,
+    yesFloorShares > noFloorShares ? 'YES' : 'NO'
+  )
+  const userPositionMoney = formatMoney(Math.abs(saleValue))
+  return [userPosition, userPositionMoney, yesFloorShares, noFloorShares]
```
jahooma, line 360: Style nit: For many values, it's better to return an object (using '{}') than a tuple.

web/components/feed/feed-items.tsx
```diff
@@ -206,14 +237,50 @@ export function CommentInput(props: { contract: Contract }) {
     setComment('')
   }
 
+  // Should this be oldest bet or most recent bet?
```
jahooma, line 240: I like most recent! Feel free to leave the comment though.

web/components/feed/activity-items.ts
```diff
@@ -308,6 +311,37 @@ function groupBetsAndComments(
...
+    return bets.reduce((acc, bet) => {
+      const userId = bet.userId
+      if (!acc[userId]) {
+        acc[userId] = []
+      }
+      acc[userId].push(bet)
+      return acc
+    }, {} as { [userId: string]: Bet[] })
+  }
+  const betsByUserId = mapBetsByUserId(bets)
```
jahooma, line 329: One suggestion is to use `_.groupBy(bets, bet => bet.userId)`. It's a bit easier!

jahooma (approved):
Sweet! Thanks for all this — no major complaints :-)

One overall comment is that it's a bit better to break up PR's if you can into independent pieces. Np though if it would be too much extra work. I've been there.

IanPhilips:
I think I had too many commits here because I tried too many different ways of doing things with some of them looking rather bad. Should've made more mocks before coding it, I was too confident it would look good!

PR #113, opened 2022-04-30
Fix HTTP charity logo links

mqp:
They seem to have been HTTP for no particular reason, so now they are HTTPS.


akrolsmir:
Good eye, thanks!

PR #114, opened 2022-04-30
Manalink: Send mana to anyone via link

akrolsmir:
See proposal: https://manifoldmarkets.notion.site/External-Bounty-Links-Manalinks-b6a1f196d04646bfb05ecb90d9f023e3


Gurkenglas:
> If a new user signs up to claim this, sender + new user both get a bonus e.g. M$ 200

That recovers some of the ick. Consider letting the visitor using the link bet without having made an account. The point was that you don't need to fear account spam to farm signup bonuses when all the mana comes from users.

common/manalink.ts
```diff
@@ -0,0 +1,27 @@
+export type Manalink = {
+  // The link to send: https://manifold.markets/send/{slug}
+  // Also functions as the unique identifier for the link.
+  slug: string
+
+  // Note: we assume both fromId and toId are of SourceType 'USER'
+  fromId: string
```
jahooma, line 7: How would we cross reference txns to mana links? Maybe the txn will have the id / slug of the manalink in a data field.

jahooma (commented):
Cool idea!

common/manalink.ts
```diff
@@ -0,0 +1,27 @@
+export type Manalink = {
```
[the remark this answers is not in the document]
akrolsmir: Eh, I was hoping to coin a new word "Manalink". It's an internal name anyways

akrolsmir:
@Gurkenglas - The idea to create an anonymous account is pretty interesting, not one I'd considered before.

But not sure how much work that'd be to handle, plus we kinda do want people to be signing up... Otherwise we're open to e.g. assassination markets where someone can move large amounts of M$ and there's no way for us to trace it.

Re: Expiration dates: from the manalink creator side they also don't want to have this possibility of the link getting claimed eg 3 years into the future. Think of it like the close date on a market.

common/manalink.ts
```diff
@@ -0,0 +1,27 @@
...
+  // The link to send: https://manifold.markets/send/{slug}
+  // Also functions as the unique identifier for the link.
+  slug: string
```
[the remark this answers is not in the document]
akrolsmir: The document key is going to be the slug/id as well!

web/components/manalink-card.tsx
```diff
@@ -0,0 +1,69 @@
...
+      className={clsx(
+        className,
+        'min-h-20 group flex flex-col rounded-xl bg-gradient-to-br from-indigo-200 via-indigo-400 to-indigo-800 shadow-lg transition-all'
+      )}
+    >
+      <Col className="mx-4 mt-2 -mb-4 text-right text-sm text-gray-100">
+        <div>
+          {maxUses != null
+            ? `${maxUses - uses}/${maxUses} uses left`
+            : `Infinite use`}
```
sipec, line 35: copy: "Unlimited use" ?

web/pages/link.tsx
```diff
@@ -35,40 +50,105 @@ export default function LinkPage() {
...
+                  : ''
+              }
+              className="input"
+              type="datetime-local"
+              onChange={(e) => {
+                setNewManalink((m) => {
+                  console.log(e.target.value)
+                  console.log(
+                    dayjs(e.target.value, 'YYYY-MM-DDTHH:mm').valueOf()
+                  )
```
sipec: delete this.

personally, I like to type `const p = (x: any) => (console.log(x), x)` at the top of files I'm debugging, then I can just use `p` to print expressions in the middle of statements without having to copy them onto a new line.

sipec (commented):
Just reviewed mqp's changes. Seem pretty good overall - I especially like the LinksTable refactor.

functions/src/transact.ts
```diff
@@ -40,60 +32,53 @@ export const transact = functions
...
+    }
+    const toUser = toSnap.data() as User
+    fbTransaction.update(toDoc, {
+      balance: toUser.balance + amount,
+      totalDeposits: toUser.totalDeposits + amount,
     })
+  }
+
+  const newTxnDoc = firestore.collection(`txns/`).doc()
+  const txn: Txn = { id: newTxnDoc.id, createdTime: Date.now(), ...data }
```
sipec, line 74: I love this simplification!

web/lib/firebase/txns.ts
```diff
@@ -39,3 +41,29 @@ export function listenForTipTxns(
...
+      txnCollection,
+      where('toId', '==', userId),
+      where('category', '==', 'MANALINK'),
+      orderBy('createdTime', 'desc')
+    )
+    listenForValues(fromQuery, setFromTxns)
+    listenForValues(toQuery, setToTxns)
+  }, [userId])
+
+  return sortBy([...fromTxns, ...toTxns], 'createdTime').reverse()
```
sipec, line 68: maybe `orderBy([...fromTxns, ...toTxns], ['createdTime'], ['desc']`

web/pages/link.tsx
```diff
@@ -0,0 +1,322 @@
...
+        <form
+          className="my-5"
+          onSubmit={async (e) => {
+            e.preventDefault()
+            await createManalink({
+              fromId: user.id,
+              amount: newManalink.amount,
+              expiresTime: newManalink.expiresTime,
+              maxUses: newManalink.maxUses,
+              message: newManalink.message,
```
sipec, line 73: maybe `createManaLink({ fromId: user.id, ...newManalink })` ?

web/pages/link.tsx
```diff
@@ -0,0 +1,322 @@
...
+            <div className="form-control flex-auto">
+              <label className="label">Amount</label>
+              <input
+                className="input"
+                type="number"
+                value={newManalink.amount}
+                onChange={(e) =>
+                  setNewManalink((m) => {
+                    return { ...m, amount: parseInt(e.target.value) }
+                  })
```
sipec, line 87: [Not important] Since all the ways of updating a mana link follow this structure, a neat pattern is to replace the useState with a simple useReducer:
```
const [newManaLink, updateNewManaLink] = useReducer((m, update) => ({ ...m, ...update }), (default value here))
```
then on these lines you just update like `updateNewManaLink({ amount: parseInt(e.target.value) })`

sipec (approved):
... and now I've reviewed the rest of it.
I would like another pair of eyes on the firebase stuff though

web/pages/link.tsx
mqp: There's actually another field I don't want in there, so this isn't quite right.

mqp:
Thanks for the feedback @sipec. This is officially good to merge in my book, it doesn't look totally crappy, I tested stuff, and I will merge at my earliest convenience when I can take care of updating security rules & functions & etc.

akrolsmir (commented):
Great work, Marshall! I think the main missing thing from a product perspective is the ability to expire a link - did we decide this wasn't as important for v0?

Other things I'm jotting down for follow-up work:
**Create tab**
- [ ] Expires At => Expires in dropdown that has "1 day / 1 week / never" options ala Discord 
image

- [ ] Maybe also the same for max uses? less sure 
image

**Links tab**
- [ ] Copy text tweaks: "Claimant" to "Claimed by", "Outstanding links" to "Unclaimed links"
- [ ] Created but claimed links should be visible somewhere (eg a new "Claimed links" tab)

**Claim page**
- [ ] Narrow the width of the preview, to match the claim page
- [ ] If there's a message included, also display who the sender is on the claim page. Maybe include their avatar on the bottom right before the message eg "😄 Austin: for fixing our graphs"
- [ ] Claimed links should be more obviously not claimable - disable/hide claim button; change text

mqp:
> I think the main missing thing from a product perspective is the ability to expire or delete your own link - did we decide this wasn't as important for v0?

I think it's important to get this out so that the branch doesn't keep going out of date and so that David can reimburse survey participants. Perhaps we should add more features for managing your own links before we expose the link creation UI in the main navigation.

PR #117, opened 2022-05-02
CPM: sort charities by amount raised

sipec:


sipec:
hmmm.. the sort isn't working on the preview even though it works on local

web/lib/firebase/txns.ts
```diff
@@ -21,3 +21,10 @@ export function listenForCharityTxns(
 ) {
   return listenForValues<Txn>(getCharityQuery(charityId), setTxns)
 }
+
+const getCharitiesQuery = () =>
```
jahooma, line 25: Nit: this doesn't need to be a function: `const charitiesQuery = query(...)`

web/pages/charity/index.tsx
```diff
@@ -3,26 +3,41 @@ import { useState, useMemo } from 'react'
...
 import { Title } from '../../components/title'
-
-const charities = charityList.map((charity) => ({
-  ...charity,
-  raised: 4001,
-}))
+import { useAllCharityTxns } from '../../hooks/use-charity-txns'
 
 export default function Charity() {
+  const allCharityTxn = useAllCharityTxns()
```
jahooma, line 12: I suppose we could even pass the raised amount into the cards, instead of re-fetching the txns in each card.

But really it doesn't matter. This works fine!

web/pages/charity/index.tsx
```diff
@@ -3,26 +3,41 @@ import { useState, useMemo } from 'react'
...
+      return {
+        ...charity,
+        raised: _.sumBy(txns, (txn) => txn.amount),
+      }
+    })
+  }, [allCharityTxn])
+
+  const sortedCharities = _.sortBy(charities, [
+    (charity) => (charity.tags?.includes('Featured') ? 0 : 1),
+    (charity) => -charity.raised,
```
jahooma, line 25: Whoa, I didn't know you could apply a series of sorts using `_.sortBy`!

web/pages/charity/index.tsx
```diff
@@ -3,26 +3,41 @@ import { useState, useMemo } from 'react'
...
+import { useAllCharityTxns } from '../../hooks/use-charity-txns'
 
 export default function Charity() {
+  const allCharityTxn = useAllCharityTxns()
+  const charities = useMemo(() => {
+    return charityList.map((charity) => {
+      const txns = allCharityTxn.filter((txn) => txn.toId === charity.id)
+      return {
+        ...charity,
+        raised: _.sumBy(txns, (txn) => txn.amount),
```
jahooma, line 18: I'd prefer to store raised in a separate map (from charity id to amount raised) instead of creating modified charity objects. But this works so fine to keep.

jahooma (approved):
Sweet!

sipec:
force pushed because I didn't realized this was reviewed lool

PR #119, opened 2022-05-02
condense comment logic in one component

IanPhilips:
this reduces comment input logic complexity, only one component handles comment logic now.

### DIFF
--- a/web/components/feed/feed-items.tsx
+++ b/web/components/feed/feed-items.tsx
@@ -227,35 +227,27 @@ export function CommentInput(props: {
   const user = useUser()
   const [comment, setComment] = useState('')
 
-  async function submitComment() {
-    if (!comment) return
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Nice, I like the simplification!

I also wonder whether we should move to only showing your position and not your recent bet amount...

IanPhilips:
This allows your written comment to persist on to a bet so you don't have to rewrite it if you wrote a comment, then bet, then wanted to add that comment on to the bet.

IanPhilips:
Hmmm I personally like to highlight that my recent position has changed when I attach it to a bet

jahooma:
> Hmmm I personally like to highlight that my recent position has changed when I attach it to a bet

Good point! Maybe it should always show your current position, but in this case show the diff? Something to think about.

PR #120, opened 2022-05-02
Move all charity images to imgur + Next.js

sipec:
Move our charity images froom `` to the Next.js `<Image>` components for better caching, responsiveness on mobile, and lazy loading. No more empty images!

Downsides:
1. replacing an image in a commit will take some time to propagate to clients
2. you have to the image host sites in next.config.js, so it's harder to use images from random websites. This is why I rehosted everything on imgur.

All the images are currently in this album: https://imgur.com/a/WSM6nKC
but it doesn't matter what imgur album we use. We don't need to use imgur either, Next works with everything.


jahooma (approved):
Nice work, Sinclair!

PR #121, opened 2022-05-03
Threaded free response comments & general comments sections

IanPhilips:
This adds threading back to free response answers, now in the comments section. I'm unsure whether the answer sections in the comments section should have bet buttons or not. 

I also added a general comments section at the bottom of FR questions for general comments.


web/components/feed/activity-items.ts
```diff
@@ -1,6 +1,6 @@
 import _ from 'lodash'
 
-import { Answer } from '../../../common/answer'
+import { Answer, getNoneAnswer } from '../../../common/answer'
```
jahooma, line 3: Unused?

web/components/feed/feed-items.tsx
```diff
@@ -560,12 +584,11 @@ export function FeedQuestion(props: {
...
-function canCommentOnBet(bet: Bet, createdTime: number, user?: User | null) {
-  const isSelf = user?.id === bet.userId
+function canCommentOnBet(bet: Bet, user?: User | null) {
+  const { userId, createdTime, isRedemption } = bet
+  const isSelf = user?.id === userId
   // You can comment if your bet was posted in the last hour
-  return (
-    !bet.isRedemption && isSelf && Date.now() - createdTime < 60 * 60 * 1000
-  )
+  return !isRedemption && isSelf && Date.now() - createdTime < 60 * 60 * 1000
```
jahooma, line 591: Comment for later followup: possibly filter out redemptions earlier. Maybe useBets hook defaults to filtering out redemptions

web/components/feed/feed-items.tsx
```diff
@@ -834,6 +857,37 @@ function FeedAnswerGroup(props: {
   )
 }
 
+function FeedGeneralComments(props: {
```
jahooma, line 860: Possibly move into contract tabs as a regular component that has ContractActivity as a child

jahooma (changes requested):
Awesome! Another step forward for free response comments.

jahooma (commented):
Oh and another thing discussed: consider only showing comment button & avatar when you focus the add a comment input

IanPhilips:
Note: this forces all comments on bets on answers to be replies to that answer rather than a general comment. I think I could allow them to comment on a bet in the general comments section too if we want.

web/components/feed/feed-items.tsx
```diff
@@ -298,35 +292,46 @@ export function CommentInput(props: {
...
+              <div className="mt-2">
+                <Textarea
+                  value={comment}
+                  onChange={(e) => setComment(e.target.value)}
+                  className="textarea textarea-bordered w-full resize-none"
+                  placeholder="Add a comment..."
+                  autoFocus={true}
+                  rows={answerOutcome == undefined || focused ? 3 : 1}
+                  onFocus={() => setFocused(true)}
+                  onBlur={() => !comment && setFocused(false)}
```
jahooma, line 305: As discussed, might want to make this simply onBlur={() => setFocused(false)} and then add above an OR condition to render the text area if `comment` has text.

jahooma (approved):
Nice! Thanks for making the suggested changes!

PR #122, opened 2022-05-03
Pre-load charity order to prevent "jump"

sipec:


jahooma (approved):
Nice, this is a good idea! I should have thought of it.

A further step would be to compute an object of charityId to amount raised in getStaticProps, and pass the amount raised into each card.

PR #127, opened 2022-05-04
Add script to denormalize avatars into other docs

mqp:
It seems that this was never actually fixed up for old contracts. Now all contracts have a correct `creatorAvatarUrl`. (It turns out that all comments already did have one -- I was just suspicious.)

I already ran this on dev/prod but I would be interested in anyone taking a glance at it since I have zero experience writing either Typescript or writing code against Firestore.

EDIT: Also answers already had correct avatar URLs.

### DIFF
--- a/functions/src/scripts/denormalize-avatar-urls.ts
+++ b/functions/src/scripts/denormalize-avatar-urls.ts
@@ -1,0 +1,101 @@
+// Script for lining up users and contracts/comments to make sure the denormalized avatar URLs in the contracts and
+// comments match the user avatar URLs.
+
+import * as admin from 'firebase-admin'
+import { initAdmin } from './script-init'
+import {
+  DocumentCorrespondence,
+  findDiffs,
+  describeDiff,
+  applyDiff,
+} from './denormalize'
+import { DocumentSnapshot, Transaction } from 'firebase-admin/firestore'
+
+initAdmin()
+const firestore = admin.firestore()
+
+async function getUsersById(
+  transaction: Transaction
+): Promise<Map<string, DocumentSnapshot>> {
+  const results = new Map()
+  const users = await transaction.get(firestore.collection('users'))
+  users.forEach((doc) => {
+    results.set(doc.get('id'), doc)
+  })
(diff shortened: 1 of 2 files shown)


functions/src/scripts/denormalize-avatar-urls.ts
```diff
@@ -0,0 +1,133 @@
...
+  applyDiff,
+} from './denormalize'
+import { DocumentSnapshot, Transaction } from 'firebase-admin/firestore'
+
+initAdmin()
+const firestore = admin.firestore()
+
+async function getUsersById(
+  transaction: Transaction
+): Promise<Map<string, DocumentSnapshot>> {
```
jahooma, line 19: You typically don't need to specify return types.

In this case it probably can't infer the return type, but that's because your new Map() is missing type parameters:

`const results = new Map<string, DocumentSnapshot>()`

functions/src/scripts/denormalize-avatar-urls.ts
```diff
@@ -0,0 +1,133 @@
...
+async function getCommentsByUserId(
+  transaction: Transaction
+): Promise<Map<string, DocumentSnapshot[]>> {
+  let n = 0
+  let results = new Map()
+  let comments = await transaction.get(firestore.collectionGroup('comments'))
+  comments.forEach((doc) => {
+    let userId = doc.get('userId')
+    let userComments = results.get(userId) || []
+    userComments.push(doc)
```
jahooma, line 55: Nit: prefer `const` over `let` if the variable is not reassigned

functions/src/scripts/denormalize.ts
```diff
@@ -0,0 +1,48 @@
+// Helper functions for maintaining the relationship between fields in one set of documents and denormalized copies in
```
jahooma, line 1: I like having helpers that abstract out the work to be done, but on the other hand, it's often simpler to follow if there're fewer layers of abstraction.

I'm not sure this is better than having one function for each of contracts, comments, and answers that updates the applicable documents for a single user (and then called for every user).

PR #128, opened 2022-05-04
Represent DB avatar URLs as non-null

mqp:
Since #127 null avatar URLs are no longer a thing in production. They are a thing in some super broken contracts on dev that are literally associated with deleted users, but those contracts deserve to be broken anyway, and if you are mad then you should delete them (or poke the database to associate them with real users).

### DIFF
--- a/common/answer.ts
+++ b/common/answer.ts
@@ -9,7 +9,7 @@ export type Answer = {
   userId: string
   username: string
   name: string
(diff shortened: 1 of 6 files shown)


jahooma (approved):
This is great! Glad we could clean up the cases where avatarUrl was missing.

Thanks, Marshall!

PR #129, opened 2022-05-04
Fix up tiny cases where user should not be nullable

mqp:
These are cases where the logged in user could never be `null` and it's just a matter of telling the type system so.


mqp:
There's some complexity to this because of an interaction with the server rendering that I don't 100% understand, so never mind  right now.

PR #131, opened 2022-05-04
Improve free response answer ux

IanPhilips:
This improves the post resolution free response UX, hiding the number of the answer, and showing options that weren't chosen while keeping the highlighted chosen answer visible.

### DIFF
--- a/web/components/answers/answer-item.tsx
+++ b/web/components/answers/answer-item.tsx
@@ -68,7 +68,7 @@ export function AnswerItem(props: {
             </Row>
           </SiteLink>
           {/* TODO: Show total pool? */}
-          <div className="text-base">#{number}</div>
+          <div className="text-base">{showChoice && '#' + number}</div>
         </Row>
       </Col>
 
--- a/web/components/answers/answer-item.tsx
+++ b/web/components/answers/answer-item.tsx
@@ -68,7 +68,7 @@ export function AnswerItem(props: {
             </Row>
           </SiteLink>
           {/* TODO: Show total pool? */}
-          {/*<div className="text-base">#{number}</div>*/}
+          <div className="text-base">#{number}</div>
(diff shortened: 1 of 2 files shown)


web/components/answers/answer-item.tsx
```diff
@@ -68,7 +68,7 @@ export function AnswerItem(props: {
             </Row>
           </SiteLink>
           {/* TODO: Show total pool? */}
-          <div className="text-base">#{number}</div>
+          {/*<div className="text-base">#{number}</div>*/}
```
jahooma, line 71: One reason to still have this is so that the creator can prefer earlier answers when they resolve the market. Maybe we could show the number if `showChoice` is defined?

web/components/answers/answers-panel.tsx
```diff
@@ -52,7 +52,11 @@ export function AnswersPanel(props: {
 
   const chosenTotal = _.sum(Object.values(chosenAnswers))
 
-  const answerItems = getAnswers(contract, user)
+  const answerItems = getAnswers(
+    contract,
+    user,
+    winningAnswers.length > 0 ? winningAnswers.map((a) => a.id) : []
```
jahooma, line 58: Why not just filter the answers to exclude winning answers, instead of modifying getAnswers to take another param?

jahooma (changes requested):
Sweet, thanks for improving this case!

web/components/answers/answers-panel.tsx
IanPhilips: because this is so easy!

web/components/answers/answer-item.tsx
IanPhilips: Added it back in. I'm not sure if users understand what the number means but maybe they do!

PR #133, opened 2022-05-04
Kill warmup spam

mqp:
I am skeptical whether it is good to make these warmup requests once. Perhaps it is. But it definitely isn't good to make them dozens of times on page load (e.g. on the portfolio page in the `sellBet` case, or all over the place in the `createUser` case.)


mantikoros (approved):
Throttling is a good idea. In general, the more users that are active on our site, the less important the warmup functions are. (The Firebase team also made an announcement several months ago that they were working to improve function startup times and that the changes would be ready by June, but we'll see...)

PR #144, opened 2022-05-05
Twiddle name of EA LTFF

mqp:
I noticed that searching "EA" on the charity page doesn't bring this up, which is kind of ridiculous, so I fixed up the name.


jahooma:
It's a little dumb how this works, but the name is used to generate the charity id. If you change the name, it will disassociate all previous transactions, which is bad. This issue would have to be fixed first.

mqp:
Yeah, that sucks...I don't really want to deal with this right now so I will just write an issue and leave it be.

PR #147, opened 2022-05-07
Configure compiler to allow absolute imports from root

mqp:
I got really fed up with the `../../../../../common/foo` business so I fixed it. Now you can import stuff from the root of the repository, e.g. `common/foo` or `web/lib/firebase/bar`.

Fixed up the API endpoint imports as a proof of concept. If you guys like this style maybe we should find/replace it in the rest of the codebase later.


akrolsmir (approved):
Wait this looks amazing!!!

jahooma:
Agreed! This style is a lot clearer.

mqp:
Cool, I will fix up the other projects and random imports to be like this soon, then.

PR #149, opened 2022-05-08
Upgrade eslint, fix eslint warnings

mqp:
This fixes a handful of warnings that were flying out of the build. I don't think they were important problems, I am just cleaning things up. The GA script loading will happen at a somewhat later stage in page loading now (see https://nextjs.org/docs/basic-features/script docs), so I guess I prefer this result, but really I am just doing it because when in Next.js we might as well do as the Next.jsians do. All moral persons use an adblocker so it's irrelevant anyway.

The eslint upgrade is because eslint was screaming that it didn't support our version of Typescript, and it was producing a spurious warning about custom fonts which was fixed in more recent eslint versions.


akrolsmir (approved):
LGTM!

I'm hoping nothing bad happens when we add in things to the deps array, sometimes React does weird things on new deps being added. (This is where Vue is so much better fwiw)

web/pages/[username]/[contractSlug].tsx
```diff
@@ -227,7 +227,7 @@ function ContractLeaderboard(props: { contract: Contract; bets: Bet[] }) {
         setUsers(sortedUsers)
       })
     }
-  }, [])
+  }, [top5Ids, userProfits])
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:230]: As Austin is cautioning, this addition to the useEffect dependencies is not good.

top5Ids and userProfits are new objects on each render, so it will trigger a the effect every time, which will call our listUsers api, which will set state, which will trigger another render...

I think this is the only deps that's a problem in this change, but I would be very careful in the future.

jahooma (changes requested):

web/pages/[username]/[contractSlug].tsx
Comment (mqp) [reply web/pages/[username]/[contractSlug].tsx:230]: Wow, I guess I was confused, I assumed the dependencies were doing some kind of deep equality check instead of reference equality...I will fix this up to not be wrong. This is my first project using React hooks so I have some things to learn :/

Comment (akrolsmir) [reply web/pages/[username]/[contractSlug].tsx:230]: Thanks for the catch, James!

@mqp can we add a comment here that surpresses the ESLint warning and tells other code viewers to be very careful while editing this line (maybe referencing this comment thread?) Thanks!

mqp:
@jahooma Can you take another look at how I rephrased the leaderboard code? This seems like the most natural way to cooperate with React here. I'm curious whether you would have done it like this, or done something else, since I want to quickly become competent at writing hook-based React idiomatically.

mqp:
(Now that I understand what's going on better, I'm also unhappy with the folds one, and will change that somehow too.)

mqp:
OK, I'm happy with both of those now.

web/pages/[username]/[contractSlug].tsx
```diff
@@ -203,31 +209,25 @@ export function ContractPageContent(props: FirstArgument<typeof ContractPage>) {
...
+      .map(([id]) => id)
+
     if (top5Ids.length > 0) {
       listUsers(top5Ids).then((users) => {
         const sortedUsers = _.sortBy(users, (user) => -userProfits[user.id])
         setUsers(sortedUsers)
       })
     }
-  }, [])
+  }, [userProfits])
```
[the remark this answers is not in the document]
Comment (jahooma) [reply web/pages/[username]/[contractSlug].tsx:230]: See also my above comment if you wanted to optimize away this re-fetching.

jahooma (approved):
LGTM! Thanks Marshall!

Sorry for the delay, I had forgotten about this PR.

web/pages/[username]/[contractSlug].tsx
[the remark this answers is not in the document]
Comment (mqp) [reply web/pages/[username]/[contractSlug].tsx:230]: Hmm...you're right. I'm still not super happy. And I'm not happy with the `useMemo` fix, because it's spooky action at a distance (the reason we have a `useMemo` up here, is that down somewhere in the component tree, where you're not looking right now, something has a `useEffect` taking this as a dependency...)

But it works OK (the parent component is not getting re-rendered much) so in the interest of time I will merge this for now and maybe think about it later.

Thanks a lot for your feedback, this is extremely helpful for improving my ability to produce high-quality React programs involving hooks.

PR #150, opened 2022-05-08
Show a preview link for markets urls on hover

akrolsmir:
Some quick questions/notes
- Should we be importing this as a npm package rather than as a <script> embed?
- Can link-summoner support Metaculus previews too?
- Right now, charity pages eg `https://manifold.markets/charity/nonlinear` render badly; where do we go to remove these from link-summoner?
- How should indicate that a link is hoverable? 
Technically: Vlad has [some CSS that does it](https://github.com/Stvad/link-summoner#styling), but I'm not sure global CSS is good/compatible with Tailwind
VisD: I think the τ symbol looks kinda weird, might try something else haha

I'd also be happy to just merge this code as is (it's already a huge improvement!) and follow up the above in a separate PR


Stvad:
> Should we be importing this as a npm package rather than as a <script> embed?

probably yes

> charity pages

https://github.com/Stvad/link-summoner/blob/master/src/rendering/manifold-renderer.tsx#L30 defines the regex that matches what Manifold pages get a preview

> How should indicate that a link is hoverable?

I would be really surprised if having global css won't work.   
You can do whatever styling `τ` is just an example :)   
https://gwern.net has type-specific icons for each link type, but I don't think the summoner supports enough descreete types yet to justify going there (though would be good to have an additional class based on the type)

> Can link-summoner support Metaculus previews too?

Yes, the simples option would be to whitelist it here https://github.com/Stvad/link-summoner/blob/master/src/rendering/iframe-renderer.tsx#L893   
Probably a better option is to add a special handler to show embeds instead of full metaculus links though (similar to [manifold handler](https://github.com/Stvad/link-summoner/blob/master/src/rendering/manifold-renderer.tsx)

Stvad:
Also, I've been thinking about makign the summoner more configurable - lmk what things would be helpful to configure

Stvad:
Also, another question: right now the live previews are not rendered for markets on the main/discovery pages bc the links to the markets are on a different domain and don't match the regex. But when you merge it they will get a preview as-well.   

So just wanted to make sure if that's intended, if not - you maybe only want to load the script on market pages/etc and not everywhere?

web/pages/_document.tsx
```diff
@@ -32,6 +32,10 @@ export default function Document() {
           `,
           }}
         />
+        <script
+          src="https://cdn.jsdelivr.net/npm/link-summoner@1.0.2/dist/browser.min.js"
+          type="module"
+        ></script>
```
mqp, line 38: I can tell you that `yarn lint` is gonna complain you aren't using Next's `Script` component: 

https://nextjs.org/docs/basic-features/script

But I agree with Stvad that I don't know why this shouldn't be bundled.

Stvad:
pushed an updated regex to avoid matching on `charity` haven't released the new version yet though.

web/pages/_document.tsx
akrolsmir: Dumb question, how do we bundle it? @Stvad maybe worth adding instructions to the README?

I was going to try `yarn add link-summoner` and then... try `import "link-summoner"` from the tsx files where we want popups to appear??

Stvad: That sounds about right (may need to add specific subpath, but probably not). will add instructions Tomorrow/Thu

akrolsmir: Hrm yeah I tried the naive thing I described with `import 'link-summoner'` in _app.tsx, and I just got this: 

image

Not sure if this is a NextJS specific issue (where `document` is not available on server-side rendered pages, and thus causes an error), or I'm doing something wrong with the import.

Stvad: Tested things on my side (pushed a bundling example in https://github.com/Stvad/link-summoner/tree/master/examples/bundling) 

Pretty sure it's Next.js SSR stuff. 

I was planning on having a version that you would use a function (allowing to supply additional parameters, etc). I think that would allow you to invoke it only if `document` exists.

Stvad: with version 1.1.0 this should work as follows:

```
import {initPreviews} from "link-summoner"

// can be wrapped into if(typeof document !== undefined)
initPreviews()
```

yet to properly update docs

Stvad: Updated the docs: https://github.com/Stvad/link-summoner#whitelisting-more-websitesadding-custom-preview-types

akrolsmir: Hm, this doesn't seem to work in NextJS because of initShadowRoot(): https://github.com/Stvad/link-summoner/issues/3

akrolsmir: (Note to self: Vlad fixed this now!)

akrolsmir:
Now blocked on https://github.com/Stvad/link-summoner/issues/4

mqp:
@akrolsmir what's the status here? Should we merge this? Did it turn out to be unexpectedly hard?

My 2c are that I am not very fired up about this feature because it's anti-web-like -- if I want to look at something briefly, I click the link and then go back, or I open the link in a new tab -- and because the implementation seems likely hacky (e.g. it doesn't work on mobile, people don't really expect something expensive and potentially slow to happen on hover.) IMO, if it isn't lightning fast, good, and doesn't only happen when the user really wants it, then it's going to be a bad experience. But I wouldn't mind trying it out.

sipec:
agree with mqp here about this being anti-web (side note: my ipad does link hover previews natively, and if you go to the page in a new tab its already loaded)

mostly closing because its been open this long and I don't think it would be hard to roll our own "market links with previews" natively

PR #151, opened 2022-05-08
Adding "Most uncertain" sort for markets

antimatter15:
When looking  at different markets, it would be helpful to be able to sort by "most uncertain" as those are  the markets where a user might be most able to contribute some marginal insight.

### DIFF
--- a/web/components/contract/contracts-list.tsx
+++ b/web/components/contract/contracts-list.tsx
@@ -275,6 +275,8 @@ export function SearchableGrid(props: {
     matches.sort((a, b) => b.volume7Days - a.volume7Days)
   } else if (sort === 'most-likely') {
     matches = _.sortBy(matches, (contract) => -getBinaryProb(contract))
+  } else if (sort === 'most-uncertain') {
+    matches = _.sortBy(matches, (contract) => Math.abs(getBinaryProb(contract) - 0.5))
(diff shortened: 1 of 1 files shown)


akrolsmir:
HI @antimatter15 ! Thanks for this suggesting this - finding markets where a user can contribute most makes a ton of sense.

We're actually pretty likely (80%+) to remove our "most and least likely" sorts from the UI, though, for a couple reason:
1. We're switching over to Algolia Search https://github.com/manifoldmarkets/manifold/pull/136 which makes supporting new sort options a fair bit trickier, technically
2. Every additional sort option adds extra UI complexity to the end user, so we should be a bit skeptical of having a bunch.

And if we don't have most/least likely, I think we wouldn't want "most uncertain" either. Would love to hear your/the rest of the community's thoughts though!

akrolsmir:
In true Manifold fashion, a prediction market on which sort options we'll support: https://manifold.markets/Austin/which-market-sorting-options-will-m

sipec:
This is interesting, but I'm not sure if closeness to 50% is actually what you're looking for. You might want to bet against 90% markets that are obviously overconfident because of higher returns, for instance.

What you really want is to look at markets with low volume I think? Or markets that fit your interests?

PR #156, opened 2022-05-09
More absolute imports

mqp:
Following up to #147 -- since my change was popular, I applied it to all the top-level reusable modules:

- All `../../../common/foo` in `functions` became `common/foo`
- All `../../../common/foo` in `web` became `common/foo`
- All `../../../lib/foo` in `web` became `web/lib/foo`
- All `../../../hooks/foo` in `web` became `web/hooks/foo`
- All `../../../components/foo` in `web` became `web/components/foo`


web/components/nav/sidebar.tsx
```diff
@@ -46,8 +46,16 @@ const signedOutNavigation = [
...
+  {
+    name: 'Discord',
+    href: 'https://discord.gg/eHQBNBqXuh',
+    icon: IconFromUrl('/discord-logo.svg'),
+  },
+  {
+    name: 'Twitter',
+    href: 'https://twitter.com/ManifoldMarkets',
+    icon: IconFromUrl('/twitter-logo.svg'),
+  },
```
mqp, line 58: Prettier randomly snuck this change in, but it's what prettier wants to do, so I may as well leave it in.

PR #157, opened 2022-05-09
Free response markets with investment cap

IanPhilips:
Allow for Free Response markets to have a maximum M$ per user in a question. There is no UI for a market creator yet, just a slot in the contract model and a check on the place-bet side.


akrolsmir:
Hm, what's the rationale for instituting a M$ cap?

IanPhilips:
It would allow for FR questions to be polls rather than default to whale bait if the resolution criteria is more loose, along the lines of: what should this thing be? I could see the counterargument be: we need to educate people to refine their resolution criteria, but I could also see polls being useful too.

jahooma (approved):
As discussed in standup, we want to trial this cap which unlocks a new style of free response markets more akin to voting.

There's no user-facing change, except if this is enabled by setting the mana limit in the database.

We plan to remove this change if there's not much usage, etc.

Code looks good! Just make sure to merge in main and fix the conflicts.

PR #168, opened 2022-05-10
Make absolute imports work with functions project

mqp:
Now you can do `import { foo } from "common/foo"` in the functions code and neither Typescript nor the Firebase deploy will complain.

It's kind of a silly hack, but I know you guys like being able to type `common/foo`, so why not. Just don't go look what `module-alias` does :-)


jahooma (approved):
Awesome, will be good to have the same shorter import style in functions as web! Seems worth a minor hack.

PR #170, opened 2022-05-10
Show every user's bets on their profile

akrolsmir:
image

Somewhat controversial to make all these bets public, but I think this is actually correct, in the context of radical transparency, and pragmatically all this data was available anyways. Hopefully encourages people to bet more, too -- more betting is one thing we're kind of constrained on atm.

Some proposals/ways to move forward:

P1: Ship this as-is -- it's not revealing anything that wasn't already query-able via Firebase command before
P1.5: Because this is making public things that weren't previously public, give everyone a chance to delete their account or any specific bets they didn't want to expose to the world

P2: Only you get to see this on your portfolio
P2.1: You can opt in to showing this to the world
P2.5: Show total bet count and total profit/loss for everybody else

P3: Get rid of the Portfolio page in the sidebar, since it's duplicated here

P4: Make bettor identities in the "bets" tab on each market public

P5: Make bettor identities public in the API

P6: Make all txn/transaction history public, everywhere all the time

### DIFF
--- a/web/components/contract/contracts-list.tsx
+++ b/web/components/contract/contracts-list.tsx
@@ -19,7 +19,6 @@ export function ContractsGrid(props: {
   const isBottomVisible = useIsVisible(elem)
 
   useEffect(() => {
-    console.log({ isBottomVisible, hasMore })
     if (isBottomVisible) {
       loadMore()
     }
--- a/web/components/nav/nav-bar.tsx
+++ b/web/components/nav/nav-bar.tsx
@@ -47,7 +47,7 @@ export function BottomNavBar() {
       )}
 
       {user !== null && (
-        <Link href={`${user}?tab=Bets`}>
+        <Link href={`${user}/bets`}>
           <a className="block w-full py-1 px-3 text-center hover:bg-indigo-200 hover:text-indigo-700">
             <PresentationChartLineIcon
               className="my-1 mx-auto h-6 w-6"
--- a/web/components/nav/nav-bar.tsx
+++ b/web/components/nav/nav-bar.tsx
@@ -47,7 +47,7 @@ export function BottomNavBar() {
       )}
 
       {user !== null && (
-        <Link href="/portfolio">
+        <Link href={`${user}?tab=Bets`}>
           <a className="block w-full py-1 px-3 text-center hover:bg-indigo-200 hover:text-indigo-700">
             <PresentationChartLineIcon
               className="my-1 mx-auto h-6 w-6"
--- a/web/components/nav/sidebar.tsx
+++ b/web/components/nav/sidebar.tsx
@@ -28,12 +28,18 @@ function IconFromUrl(url: string): React.ComponentType<{ className?: string }> {
   }
 }
 
-const navigation = [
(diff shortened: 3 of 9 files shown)


jahooma:
Cool! Showing other users' profit has many beneficial effects. Showing their bets is more a mixed bag, although I can see it as a path to a more viral social app.

I guess I'm leaning toward "let's try it!" and see if we get a negative reaction or if it just increases engagement. Ditto for revealing bets on the market page (P4).

The friendliest way to do it may be to let people click a button to opt out of showing their bets, and email them e.g. 1 week before it goes live with that opt out option.

sipec:
maybe we could just hide bets (unless commented) on any market in the personal fold
then we go around and tag all the #personal questions manually, just this once

I'm hesitant to roll this out without a robust implementation of personal / private markets. I think it hurts info aggregation in markets like "is everyone at this party going to have fun", "will I stay married" if people can't anonymously trade in these markets. These are exactly the kind of markets we're supposed to be better than other platforms at!

sipec:
more polemically, I've been seriously thinking about making the market "Will Genesis House have any sexual assaults that I come to learn of in 2022" because I know some friends who have highly relevant knowledge, and other friends who would want to know that knowledge, but I myself and uncertain about the truth and do not have the social capital or time necessary to determine whether I should warn my friends or not.

There is literally no other info aggregation system, other than anonymous bets, that I think can actually solve this category of social problem. Legal systems are not up to the task. Twitter isn't up to the task. The informal networks of social capital that transact on whispers of gossip are not up to the task.

akrolsmir:
We discussed this a bit in standup today - I do think having private markets and private bets on Manifold would be good, but the reality is that we support neither at the moment.

I think implementing private markets and bets could be something we go off and do (especially if you're interested in making it happen, Sinclair!) -- other people who would be interested include markets for corporations or nonprofits with sensitive data. But I don't think we should block this feature on waiting for those to be implemented, because it'll take some amount of work to get private markets/bets technically correct.

(I worked a bit on the "off the record" feature of Hangouts Chat, and that was a nightmare to do correctly.)

jahooma:
I actually don't think it would be too hard to make bets private. We just need to not fetch the bets collection client-side, except for your own bets.

We'd need to call some api which gets bets stripped of their userId. Could be the same as the external API. Potentially we could mark some bets as revealed if you made a comment with it.

akrolsmir:
Hm, but then are we committing to keeping user balances private forever as well? What about leaderboards?

Because if we're not, then one theoretical attack slash thing that will happen in the future is someone scraping all our markets for activity, and matching up bets made with balance changes. Yes, I get that there's a huge difference in UX between having something be "natively supported" vs "possible", but on the other hand we should also be ready for worst-case attacks if we're going to run a trustworthy financial system.

I really think we should look to what crypto does -- again, crypto takes money transfer very seriously, and we... have not been, but we need to (cf double-entry accounting aka `txn`s only being introduced 4 months after launch). The full transaction history of the major blockchains are fully public/pseudonymous, and this auditability is a core feature; in principle, we should support the same.

If we want to support private trades, I think we should do something like allow for burner wallets, not default transaction history to private.

PR #173, opened 2022-05-10
Feed container markup cleanup

mqp:
It seemed like all these wrappers and classes served no purpose.

There is one tiny user-visible change, which is that on the home page activity feed, on small screens, there is a little less right-padding on each feed item. I actually think this change is a bug fix; it means that the right and left margins are the same size, and it means that the right UI elements are right-aligned with e.g. the "create market" button at the top. So I am keeping this change.

I tested on all my browsers. I don't have an iPhone to test with. But the existence of the wrappers made no sense to me, so I don't feel it is likely to break anything.

### DIFF
--- a/web/components/answers/answers-panel.tsx
+++ b/web/components/answers/answers-panel.tsx
@@ -111,7 +111,7 @@ export function AnswersPanel(props: {
         <FeedItems
           contract={contract}
           items={answerItems}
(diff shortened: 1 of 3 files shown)


jahooma (approved):
Thanks for the code cleanup, Marshall!

PR #178, opened 2022-05-11
Clean up extra wrapper and CSS on BetRow

mqp:
Someday I am going to stop finding random chunks of markup and CSS to delete, but today is not that day. No user-visible changes.


jahooma (approved):
LGTM!

PR #179, opened 2022-05-11
Print out build info in console

mqp:
I noticed that when I do a deploy and I want to test to make sure everything is OK, I sit around looking at the site thinking, are my changes live yet? Now on page load it will print a helpful message in the console making it super clear what is the live code, so I don't have to wonder whether or not I got the new one.

image

### DIFF
--- a/web/pages/_app.tsx
+++ b/web/pages/_app.tsx
@@ -1,22 +1,27 @@
 import 'tailwindcss/tailwind.css'
 import type { AppProps } from 'next/app'
+import { useEffect } from 'react'
 import Head from 'next/head'
 import Script from 'next/script'
 import { usePreserveScroll } from 'web/hooks/use-preserve-scroll'
 
 function printBuildInfo() {
-  // undefined if e.g. dev server
+  // These are undefined if e.g. dev server
   if (process.env.NEXT_PUBLIC_VERCEL_ENV) {
     let env = process.env.NEXT_PUBLIC_VERCEL_ENV
-    let sha = process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA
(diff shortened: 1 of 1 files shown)


web/pages/_app.tsx
```diff
@@ -4,7 +4,18 @@ import Head from 'next/head'
 import Script from 'next/script'
 import { usePreserveScroll } from 'web/hooks/use-preserve-scroll'
 
+function printBuildInfo() {
+  // undefined if e.g. dev server
+  if (process.env.NEXT_PUBLIC_VERCEL_ENV) {
```
jahooma, line 9: Would it be better to put this in a useEffect(() => { ... } ,[]) so that it only runs once?

jahooma (approved):
Awesome! This is a cool idea that could help us debug!

akrolsmir (approved):
Agreed, this is super cool!

Might suggest formatting it as a link to a Github commit eg `https://github.com/manifoldmarkets/manifold/commit/948d87822298bf95a30bf0c0e77e00354c11061b` so you can just click on the link to see what commit it's pointing too

web/pages/_app.tsx
mqp: I put it here partly because this seemed like the most obvious place where it would run exactly once per page -- do you suspect it could run more than once? If so, sure, I will add that.

mqp:
> Agreed, this is super cool!
> 
> Might suggest formatting it as a link to a Github commit eg `https://github.com/manifoldmarkets/manifold/commit/948d87822298bf95a30bf0c0e77e00354c11061b` so you can just click on the link to see what commit it's pointing too

Good suggestion!

image

PR #181, opened 2022-05-11
Tell eslint to run on the hooks directory

mqp:
Before, it was just not running on any of the stuff in `hooks`, because the default Next config is to only run on `lib`, `components`, and `pages`.

Of course there are a bunch of warnings that fly out, which I guess we can deal with at our leisure.


mqp:
I guess Vercel gets too mad to deploy if there are literal errors coming out of the linter. I will close this until we have time to fix them up.

PR #182, opened 2022-05-11
Allow users to generate an API key in their profile

mqp:
_**Don't merge this before deploying the associated Firebase rules to prod or it won't work.**_ (They are already deployed to dev.)

The key doesn't do anything yet. I could hide it from the UI for now, but why bother? Let people think about it.

The key is stored in `/private-users/{uid}/apiKey`.

### DIFF
--- a/common/user.ts
+++ b/common/user.ts
@@ -35,4 +35,5 @@ export type PrivateUser = {
   unsubscribedFromGenericEmails?: boolean
   initialDeviceToken?: string
   initialIpAddress?: string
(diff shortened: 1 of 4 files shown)


akrolsmir (approved):
Awesome, LGTM! Love how simple this is.

One small thing: could we put it at the bottom of the profile page (below Balance?) Since most of our users aren't programmers and won't ever use the API key, we should prioritize things they'll care more about (eg Bio, Twitter URL, etc)

PR #186, opened 2022-05-11
restyle home page (separation between markets)

sipec:
Replace ugly line with:

image

### DIFF
--- a/web/components/feed/activity-feed.tsx
+++ b/web/components/feed/activity-feed.tsx
@@ -20,11 +20,11 @@ export function ActivityFeed(props: {
   const user = useUser()
 
   return (
-    <Col className="divide-y divide-gray-300 bg-white">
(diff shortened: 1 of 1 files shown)


akrolsmir (approved):
Eh I didn't hate the old version (but I'm biased, I implemented it). Wondering - maybe we just wanted to use `divide-gray-50` instead of `divide-gray-300` and then it looks good?

Anyways, this version looks fine too!

mantikoros:
I think just making the line lighter would look best.

jahooma:
I like the gap better. Was never much a fan of the border line. It's also what Facebook does.

sipec:
I now see the wisdom of pushing to main directly - prevents bikeshedding

side note: probably the "your trades" page should also follow the same style

PR #197, opened 2022-05-12
Show category on market card

jahooma:


jahooma:
Hmmm, I'm not opposed, just let me fix the margin

PR #204, opened 2022-05-12
Bet embed

sipec:


web/pages/embed/[username]/[contractSlug].tsx
```diff
@@ -115,7 +115,10 @@ function ContractEmbed(props: { contract: Contract; bets: Bet[] }) {
 
           {isBinary && (
             <Row className="items-center gap-4">
-              <BetRow contract={contract as BinaryContract} />
+              <BetRow
+                contract={contract as BinaryContract}
+                betPanelClassName="scale-75"
```
Comment (jahooma) [inline web/pages/embed/[username]/[contractSlug].tsx:120]: Are you sure we need to scale this down? It makes the font small and is inconsistent. Plus, if you have shares already, you can still access the sell panel, which is not scaled.

I would advise finding another solution if it needs to be smaller. But the panel is designed to work on smaller mobile devices, so I would think it should be good?

jahooma (changes requested):
Very nice! Simple solution, clean code.

sipec:
this isn't the best style, but it is the 80/20 and I think the v2 after this should be slide to bet instead of a prettier panel.
I'm just gonna merge it.

PR #205, opened 2022-05-12
Fix random errors

mqp:
This fixes some stuff that React was screaming about, but which probably doesn't matter except for spamming the console.


jahooma (approved):
LGTM!

PR #210, opened 2022-05-13
Swap values with structured assignment rather than temporary

Sjlver:
Saves two lines and a variable.

Also, a desperate attempt to save my Mana in https://manifold.markets/ManifoldMarkets/if-we-open-source-our-frontend-code

### DIFF
--- a/common/util/random.ts
+++ b/common/util/random.ts
@@ -42,8 +42,6 @@ export function createRNG(seed: string) {
 export const shuffle = (array: any[], rand: () => number) => {
   for (let i = 0; i < array.length; i++) {
     const swapIndex = Math.floor(rand() * (array.length - i))
-    const temp = array[i]
-    array[i] = array[swapIndex]
-    array[swapIndex] = temp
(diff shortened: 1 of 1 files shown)


akrolsmir:
Hrm, seems to have generated a typescript build error of some kind kind here?

I appreciate the attempt at market manipulation! I'd probably lump together your 4 outgoing PRs as a single "significant contribution" for the purposes of the market 😝

(though - small PRs are great, thanks for splitting these up)

mqp:
You need a semicolon because the second line is otherwise ambiguous between a new statement and an indexing expression.

Sjlver:
Thanks for finding the semicolon problem. Fixed now.

Note: for some reason, `prettier` wants the semicolon to go to the beginning of the line.

> I appreciate the attempt at market manipulation! I'd probably lump together your 4 outgoing PRs as a single "significant contribution" for the purposes of the market 😝

Fair enough ;-)

jahooma:
This code is not an improvement. Pull requests like this also cost attention.

Despite the good intentions, I don't think this PR should have been merged in.

PR #211, opened 2022-05-13
Clean the user's display name on update.

Sjlver:
The user's display name should always be clean (see for example
functions/src/create-user.ts). However, change-user-info.ts does not
enforce this, thus potentially allowing a malicious user to change their
name to something that doesn't satisfy the rules for clean display
names.

Note: this cannot happen currently because all callers (in profile.tsx)
clean the name. However, doing it here is good defense in depth
(similar to how the userName is cleaned).

### DIFF
--- a/functions/src/change-user-info.ts
+++ b/functions/src/change-user-info.ts
@@ -5,7 +5,7 @@ import { getUser } from './utils'
 import { Contract } from 'common/contract'
 import { Comment } from 'common/comment'
 import { User } from 'common/user'
-import { cleanUsername } from 'common/util/clean-username'
+import { cleanUsername, cleanDisplayName } from 'common/util/clean-username'
 import { removeUndefinedProps } from 'common/util/object'
 import { Answer } from 'common/answer'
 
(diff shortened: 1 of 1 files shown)


Sjlver:
Note: I tried to find testcases for this but couldn't... are there any?

mqp:
I agree with this change, it makes zero sense to store unclean display names in the DB. However, the one thing I don't like, is that there may still be unclean display names in the DB -- the old ones. Once this change is in, that will be surprising and be a source of bugs if someone later concludes that it's safe to stop sanitizing the names on the client.

So I advocate that before merging this, someone should check the DB to ensure that there aren't any unclean names, and sanitize any that are unclean.

Sjlver:
@mqp Thanks for the feedback. I don't think I have any possibility to check the existing DB, right?

Anyway, currently all callers do clean the name, so this is more of a "defense in depth" PR. Of course, there could still be "unclean" display names from even earlier in the development. But the code in profile.tsx has been in place for at least 3 months (https://github.com/Sjlver/manifold/commit/03f36cf95471674dcfe752d0ea2229791f96185e#diff-8dcc2ae3046243f160e2021674f7f43ec40096a3316b8ea24ec050b486054176R39)

sipec (approved):

sipec:
I don't think doing the db name sweep should block this

mqp:
I am going to deploy this and make sure existing names are clean so we can stop thinking about it.

mqp:
See #543.

PR #212, opened 2022-05-13
Update mantic.markets -> manifold.markets

Sjlver:
The other references to "mantic" that I could find look legit, but the
referrer in stripe.ts should probably be updated.


akrolsmir:
Nice catch, thanks!

PR #213, opened 2022-05-13
Better random

Sjlver:
- Generate random strings in a cryptographically safe way
- Ensure there are no collisions when using only few bits of randomness

### DIFF
--- a/common/util/random.ts
+++ b/common/util/random.ts
@@ -1,7 +1,11 @@
-export const randomString = (length = 12) =>
-  Math.random()
-    .toString(16)
-    .substring(2, length + 2)
+// Returns a cryptographically random hexadecimal string of length `length`
+// (thus containing 4*`length` bits of entropy).
+export const randomString = (length = 12) => {
+  const bytes = new Uint8Array(Math.ceil(length / 2))
+  crypto.getRandomValues(bytes)
+  const hex = bytes.reduce((s, b) => s + ('0' + b.toString(16)).slice(-2), '')
+  return hex.substring(0, length)
+}
 
 export function genHash(str: string) {
   // xmur3
--- a/functions/src/create-user.ts
+++ b/functions/src/create-user.ts
@@ -39,8 +39,7 @@ export const createUser = functions
     const name = cleanDisplayName(rawName)
     let username = cleanUsername(name)
 
-    const sameNameUser = await getUserByUsername(username)
-    if (sameNameUser) {
+    while (await getUserByUsername(username)) {
       username += randomString(4)
     }
 


common/util/random.ts
```diff
@@ -1,7 +1,13 @@
...
-  Math.random()
-    .toString(16)
-    .substring(2, length + 2)
+// Returns a cryptographically random hexadecimal string of length `length`
+// (thus containing 4*`length` bits of entropy).
+export const randomString = (length = 12) => {
+  const bytes = new Uint8Array(Math.ceil(length / 2))
+  crypto.getRandomValues(bytes)
+  const hex = Array.prototype.map
+    .call(bytes, (b) => ('0' + (b & 0xff).toString(16)).slice(-2))
```
mqp, line 7: This all makes sense to me except why isn't `b & 0xff` a no-op?

I would have written `bytes.reduce((s, b) => s += ('0' + b.toString(16)).slice(-2), '')`.

mqp:
I don't think it's important that the random string is high quality, but the new code is certainly simpler than the old weird approach of "generate some mysterious amount of randomness and hexify it", and the collision thing is a bad bug, so thanks for fixing it. If you comment on the mysterious-to-me `b & 0xff` I will merge it.

Sjlver:
I've removed the `& 0xff` since I couldn't find what it was for. Given that it's operating on Uint8Array entries, it should indeed be a no-op. To my shame, I admit that that part of the code was copied from StackOverflow.

I tested using `reduce`, and this indeed also work:

```typescript
bytes.reduce((s, b) => s + ('0' + b.toString(16)).slice(-2), '')
```

(note no `+=` is needed.) I find map+join slightly clearer. My intuition was that it would also be faster... but apparently no:

```typescript
const usingMap = bytes => Array.prototype.map
    .call(bytes, (b) => ('0' + b.toString(16)).slice(-2))
    .join('');
const usingReduce = bytes => bytes.reduce((s, b) => s + ('0' + b.toString(16)).slice(-2), '');

bytes = new Uint8Array(65000);
crypto.getRandomValues(bytes);
let timeForMap = timeForReduce = 0;
for (let i = 0; i < 1000; ++i) {
  const d0 = new Date();
  const a = usingMap(bytes);
  const d1 = new Date();
  const b = usingReduce(bytes);
  const d2 = new Date();
  if (a !== b) throw "oops";
  timeForMap += d1 - d0;
  timeForReduce += d2 - d1;
}
console.log(`map took {timeForMap} ms, reduce took {timeForReduce} ms.`);
```

Prints `map took 12592 ms, reduce took 5638 ms.` on my machine.

Thus changed to using reduce.

mqp:
OK, this looks good and I'll merge it in a bit when I am available to deploy the cloud function change.

jahooma:
Unfortunately this had to be reverted. The `crypto` api works in modern browsers, but not in node. Code in the `common` directory needs to be able to run on the client and server.

Sjlver:
Apparently crypto.getRandomValues was added in node v15.0.0; do you have to use an older version?

https://nodejs.org/api/webcrypto.html#cryptogetrandomvaluestypedarray

Issue #191, opened 2022-05-11
Some layout stuff on the page is just rendered twice


mqp:
Specifically:

- The main sidebar React component is rendered even on small widths, and then hidden with CSS; when you click the profile button to have the sidebar slide out from the left, it's actually a second, new sidebar.
- The "right sidebar" (e.g. the charity pages have this) is rendered twice at all widths, and then CSS decides whether to show it on the right or on the bottom.

These sidebars are not trivial components so this is obviously silly if not necessary.

At least in the case of the "right sidebar", it should be easy to just render this once and use CSS to control the position (e.g. flex wrap it down.) For the "main sidebar" it may or may not be easy, because I don't know how the slider component we're using works. The "main sidebar" also has different options on mobile.

mqp:
A related funny bug is that if you have a small window, you open the sidebar, and then you enlarge your window, ta-da! You have two sidebars!

mqp:
The main sidebar part of this issue was easy to fix after all (#225) so only the right sidebar part remains.

PR #225, opened 2022-05-14
Make single sidebar slide in and out instead of dialog hack

mqp:
Fixes the main-sidebar part of #191. Previously, the main sidebar was stuck in the flow to the left of the content, and when you were in mobile mode, it would be hidden with a media query, and then when you clicked the profile button, a whole new sidebar would be generated as a headlessui `Dialog` and dropped into place.

Now the sidebar basically is made the way you might expect; it uses media queries to decide whether to be a fixed-thing-on-the-left or a sticky-thing-not-as-far-on-the-left, and no `Dialog` is involved, although I had to reinvent an overlay component to replace it.

There are only small user-visible changes:

- Fixes the weird bugs where if you open the sidebar and then grow the browser window you will be in some kind of big-layout-but-with-open-mobile-sidebar mode.
- When you close the mobile sidebar, the background no longer does an animated fade-in; it just goes away. This is simpler but also intentional. During the fade-in you still couldn't use any of the UI on the page, and I really dislike implementing animations that block the user from doing anything while they are going on. Feel free to fight with me about this if you really want to.
- When the mobile sidebar is open, you can still scroll the page with e.g. scroll wheel. This seems fine.
- When the mobile sidebar is open, hitting a key doesn't close it. This seems fine.
- Very small fixes to some spacing to look slightly better to me.
- Fixes the issue where the sidebar moves slightly when you start scrolling down and it "sticks".

### DIFF
--- a/web/components/nav/nav-bar.tsx
+++ b/web/components/nav/nav-bar.tsx
@@ -5,18 +5,14 @@ import {
   MenuAlt3Icon,
   PresentationChartLineIcon,
   SearchIcon,
-  XIcon,
 } from '@heroicons/react/outline'
-import { Transition, Dialog } from '@headlessui/react'
-import { useState, Fragment } from 'react'
-import Sidebar from './sidebar'
 import { useUser } from 'web/hooks/use-user'
 import { formatMoney } from 'common/util/format'
 import { Avatar } from '../avatar'
 
 // From https://codepen.io/chris__sev/pen/QWGvYbL
-export function BottomNavBar() {
-  const [sidebarOpen, setSidebarOpen] = useState(false)
+export function BottomNavBar(props: { toggleSidebar: () => void }) {
+  const { toggleSidebar } = props
 
   const user = useUser()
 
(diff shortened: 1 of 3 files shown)


mqp:
P.S. if you have an iPhone it would be worth glancing at this to make sure it is working well. I typically only test on desktop and Android Firefox + Chrome.

mqp:
If you just like the animated fade and the scrolling prevention, I can easily add that back in, and I'll take a look at what is causing this strange layout in your Chrome testing. If those were fixed, would you agree that this is an improvement?

jahooma:
I think it could be an improvement with those fixes. I'm a bit suspicious there could be more small problems by reimplementing our own modal though.

It would be nice to have standard components that we reuse. For example, we do already have a `modal.tsx` component used in quite a few places.

mqp:
It looks to me like the "modal" component is a very specific kind of "centered modal dialog", which isn't really very appropriate to this case, but I will see if I can reuse the HeadlessUI "dialog overlay" technology for use here, which would be an appropriate reuse if possible.

mqp:
...I say that, but I can't reproduce the problem you screenshotted on Chrome...

mqp:
I implemented some reasonable change, but with your Chrome issue, and I am also seeing suspicious iOS-specific behavior being talked about when it comes to the "prevent scrolling" business, I am willing to just can this PR for now.

PR #227, opened 2022-05-15
Implement `onRequest` versions of `createContract`, `placeBet` functions

mqp:
_**Function status: these replacement functions are currently deployed on dev for testing, and not yet on production.**_

This enables these functions to serve future public API requests. Future public APIs will be converted in a similar fashion when desired.

The design of this change is to be backwards-compatible with the existing `onCall` machinery, so deploying these new versions of the functions will require no client changes. We will break compatibility to make necessary improvements (e.g. real HTTP status codes) in future steps. The most important thing about the new versions is that they now accept API keys (see #182) as an alternative authentication mechanism to the Firebase JWTs. No intentional changes were made to the function behavior except related to authentication.

### DIFF
--- a/web/components/bet-panel.tsx
+++ b/web/components/bet-panel.tsx
@@ -215,7 +215,7 @@ function BuyPanel(props: {
 
   useEffect(() => {
     // warm up cloud function
-    placeBet({}).catch()
+    placeBet({}).catch(() => {})
   }, [])
 
   useEffect(() => {
--- a/functions/package.json
+++ b/functions/package.json
@@ -20,6 +20,7 @@
   "main": "lib/functions/src/index.js",
   "dependencies": {
     "@react-query-firebase/firestore": "0.4.2",
+    "cors": "2.8.5",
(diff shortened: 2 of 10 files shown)


web/components/bet-panel.tsx
```diff
@@ -215,7 +215,7 @@ function BuyPanel(props: {
 
   useEffect(() => {
     // warm up cloud function
-    placeBet({}).catch()
```
mqp, line 218: These changes aren't strictly necessary for the operation of this PR, but I put them in anyway since it was convenient. All these warmup functions would actually blow up if an error happened (e.g. the function returns a 500), because `catch()` is a no-op -- you need an actual handler function in order for anything to get caught.

mantikoros: Good catch (pun intended).

mantikoros (approved):
Looks good to me.

mantikoros:
If you've tested this sufficiently and everything works, feel free to merge it in.

PR #232, opened 2022-05-15
Clean up some stuff with SellPanel and AmountInput

mqp:
Previously, I felt this code was fishy because all the logic in `SellAmountInput` was responsible for producing the first row of "sale proceeds" UI, which was clearly not conceptually tied to the input component -- it was supposed to be part of the `SellPanel` above it. By moving this logic up, that markup got simpler, and `SellAmountInput` basically turned into nothing.

What do you think?

In the future, I think this could be further improved to throw out more code and unify some of the validation and error-handling logic between buy and sell components (which is subtly different in ways that I think are as likely accidental as intentional.)

There should be no user-visible changes.

### DIFF
--- a/web/components/amount-input.tsx
+++ b/web/components/amount-input.tsx
@@ -1,10 +1,8 @@
 import clsx from 'clsx'
-import _ from 'lodash'
 import { useUser } from 'web/hooks/use-user'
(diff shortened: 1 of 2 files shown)


jahooma (approved):
Ok, great! Looks like a strict improvement.

PR #233, opened 2022-05-16
Change lodash stuff so that it can be tree-shaken out of build

mqp:
Previously:
- Everyone did `import _ as lodash` everywhere.
- As a result all of lodash got imported into our production bundle, even though we only use like a dozen functions from it. (Actually for some reason I don't understand, it looked like both numerous individual lodash modules and also all of lodash got separately imported into different chunks.)

Now:
- I changed everything to do `import { sortBy, sum, ... } from 'lodash'`. (I also got rid of a few unnecessary lodash uses. Take note of the existence of `Object.entries` and `Object.fromEntries`, which can be used instead of `_.fromPairs` and `_.toPairs`.)
- Furthermore, I configured the build to transform that into `import sortBy from 'lodash/sortby'` and so on.
- Therefore, only a bit of lodash is now in our bundle.

This kills a few dozen kilobytes of Javascript that have to be loaded prior to render on basically every page so it's definitely worth doing. But if people start `import _`ing again, it will stop working. So don't do that!

Next up for "make tree shaking actually work" is the Firebase library, or maybe the icon packages.

### DIFF
--- a/common/.eslintrc.js
+++ b/common/.eslintrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+  plugins: ['lodash'],
   extends: ['eslint:recommended'],
   env: {
     browser: true,
@@ -14,5 +15,6 @@ module.exports = {
   rules: {
     'no-unused-vars': 'off',
     'no-constant-condition': ['error', { checkLoops: false }],
+    'lodash/import-scope': [2, 'member'],
   },
 }
--- a/common/antes.ts
+++ b/common/antes.ts
@@ -1,3 +1,4 @@
+import { range } from 'lodash'
 import { Bet, NumericBet } from './bet'
 import { getDpmProbability, getValueFromBucket } from './calculate-dpm'
 import {
@@ -11,7 +12,6 @@ import {
 import { User } from './user'
 import { LiquidityProvision } from './liquidity-provision'
 import { noFees } from './fees'
-import * as _ from 'lodash'
 
 export const FIXED_ANTE = 100
 
@@ -127,11 +127,11 @@ export function getNumericAnte(
   const betShares = Math.sqrt(ante ** 2 / bucketCount)
 
   const allOutcomeShares = Object.fromEntries(
-    _.range(0, bucketCount).map((_, i) => [i, betShares])
+    range(0, bucketCount).map((_, i) => [i, betShares])
   )
 
   const allBetAmounts = Object.fromEntries(
-    _.range(0, bucketCount).map((_, i) => [i, betAnte])
+    range(0, bucketCount).map((_, i) => [i, betAnte])
   )
 
   const anteBet: NumericBet = {
--- a/common/calculate-cpmm.ts
+++ b/common/calculate-cpmm.ts
@@ -1,4 +1,4 @@
-import * as _ from 'lodash'
+import { sum, groupBy, mapValues, sumBy } from 'lodash'
 
 import { Binary, CPMM, FullContract } from './contract'
 import { CREATOR_FEE, Fees, LIQUIDITY_FEE, noFees, PLATFORM_FEE } from './fees'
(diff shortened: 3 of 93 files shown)


akrolsmir:
Hrm, I'm a bit more skeptical of this one - I see the win for tree-shaking, but I wouldn't do it if it comes at a cost of code readability.

Readability is somewhat fuzzy/subjective, but IMO seeing the underline import makes it easy to understand "oh and this is a lodash function).

I wonder if we could just proxy the lodash functions into our own lodash library (eg `__.js`) and then have people do `__.groupBy(...)`, so we get both readability and treeshaking benefits?

(It's also kind of surprising to me that treeshaking can't already remove unused functions out of lodash through an underscore import)

@jahooma thoughts?

jahooma:
While I agree with Austin, I also think there could be a small benefit to readability from not having the underscore, so it's a wash. Another downside is you cannot just type `_.` and see the list of possible functions.

What I don't like with this change is now we have to monitor contributions to make sure they don't do `import _ from 'lodash'`.

Similar to what Austin was saying, maybe we could make a file in `/common` called _.ts, which exports all the lodash functions we use. Then `/web` and `/functions` would remove lodash as a dependency. It would be natural typescript from that point to import util functions.

Actually, we already have `common/util/array.ts`. Could we just export all the lodash functions related to arrays from this file, and do the same for `common/util/object.ts`?

mqp:
Cool, I totally agree with "just have the utility functions ourself and stop taking lodash as a dependency", that's what I would do if I were in charge. I'm happy to implement that approach, since you guys both like it.

(My feeling about lodash is that it made sense to me back when writing everyday Javascript was hard because of browser compatibility concerns, but in 2022, it makes little sense. I am perfectly capable of writing `groupBy` and `sumBy` and I don't need a giant library to hand them to me.)

mqp:
> (It's also kind of surprising to me that treeshaking can't already remove unused functions out of lodash through an underscore import)

This is going to be generally true about all JS, so update accordingly. These Javascript build tools typically aren't doing complicated program flow analysis stuff to try to eliminate dead code -- maybe Closure Compiler did that, but nobody uses it anymore. They are just handling the simple case of "if a package exposes many ES modules, the modules have no side effects, and you import one of them, don't pull in all the others." Anything more complicated and it doesn't work.

mqp:
But aren't you equally concerned that someone will write `import _ from 'lodash'` in that scenario, in order to use whatever they are thinking of, instead of putting it in our utility module? There won't be any error.

mqp:
@jahooma I'm not 100% sure how all the pieces work together because I haven't used Yarn workspaces before, but it looks to me like even if we remove lodash from `web/package.json`, there won't be any obvious compiler or linter errors flying out when we `import` from it in the web code. I think if your goal is to stop anyone from committing code with `import _ from 'lodash'`, we have something like three options:

1. Just explain to people not to do it and try to catch it if they do.
2. Remove lodash as a thing that exists in our `node_modules` folder at all, including as a transitive dependency.
3. [Configure eslint to check for this](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md) and then actually look at eslint's output ever and take any action based on it.

sipec:
hot take:
```javascript
import { sum as _sum } from 'lodash'
```

sipec:
cold take:
[stack overflow](https://stackoverflow.com/a/62860462) claims that webpack can tree shake lodash as is but you probably need to import from `lodash-es` instead. (Next.js should be using webpack 5 which can tree shake some common js packages maybe, but lodash-es should def work)

mqp:
> [stack overflow](https://stackoverflow.com/a/62860462) claims that webpack can tree shake lodash as is but you probably need to import from lodash-es instead. (Next.js should be using webpack 5 which can tree shake some common js packages maybe, but lodash-es should def work)

I tried this and it maybe worked slightly, but it eliminated much less code. Perhaps we should also be using `lodash-es`, but it doesn't solve this problem by itself. Patches welcome if you disagree.

jahooma:
Hmm, if removing lodash from web/package.json doesn't prevent you from importing lodash, then I'm not sure I like any of these solutions.

One more solution would be to export some lodash methods from common, and then have a norm that you never import lodash directly, you just import from our util files. That norm might be easier than "make sure you don't import the whole package of lodash, but do import the functions in it".

But maybe we should just take no action. Even adding `lodash-es` seems like it'll cause some build problems at some point. Not sure if this is worth the effort.

mqp:
Why don't you like using eslint to enforce making appropriate imports, and why don't you like replacing lodash with our own utility file? I feel like both of those solutions are basically all upside and no downside.

jahooma:
Ok, I guess the eslint idea is not so bad. I'm on board. I prefer it to re-implementing all the lodash functions.

mqp:
OK, the eslint check is in, so I'm going to merge this.

PR #234, opened 2022-05-16
Upgrade HeadlessUI to 1.6.1

mqp:
This is relevant to us because tree shaking was busted in our version of HeadlessUI, and now it's [fixed](https://github.com/tailwindlabs/headlessui/pull/1247), and we use very few HeadlessUI components. So we will save a few tens of kilobytes of bundle size.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -17,7 +17,7 @@
     "postbuild": "next-sitemap"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


akrolsmir (approved):
LGTM, thanks for always proactively working to keep our site lean!

PR #235, opened 2022-05-16
Redesign the contract card

akrolsmir:
image

image

### DIFF
--- a/web/components/contract/contract-card.tsx
+++ b/web/components/contract/contract-card.tsx
@@ -6,6 +6,7 @@ import {
   Contract,
   contractPath,
   getBinaryProbPercent,
+  getBinaryProb,
 } from 'web/lib/firebase/contracts'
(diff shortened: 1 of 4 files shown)


jahooma:
I like this!

Very minor suggestions:
- Maybe limit to showing one category so it doesn't look cluttered
- Darken the font for the creator name to make it more prominent / distinct from category and amount bet
- I'm not a fan of the '#' in the category

akrolsmir:
> I like this!
> 
> Very minor suggestions:
> 
> * Maybe limit to showing one category so it doesn't look cluttered
> * Darken the font for the creator name to make it more prominent / distinct from category and amount bet
> * I'm not a fan of the '#' in the category

Thanks!
- Limited to one category (sorted by the first one in CATEGORIES_LIST)
- Lightened category & amount bet
- Keeping #hashtags though (think you're wrong about this one 😛 )

PR #239, opened 2022-05-16
Enable url linking to comments and a copy to clipboard function

IanPhilips:
Added a link icon to the created time bit on a comment that, when tapped, copies the comment url to the user's clipboard and shows a toast that the url was copied. 

The comment's bg, when navigated to, will highlight green for a few seconds and then fade to the normal bg color.


IanPhilips:
Interested in your visD feedback Austin!

web/components/feed/feed-items.tsx
```diff
@@ -230,7 +253,12 @@ export function FeedComment(props: {
     )
 
   return (
-    <>
+    <Row
+      className={clsx(
+        'flex space-x-3 transition-all duration-1000',
+        highlighted ? `bg-primary/[0.3] rounded` : ''
```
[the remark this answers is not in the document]
akrolsmir: Also not sure how you do the scrolling - but would be nice to have a bit of whitespace padding above the highlighted comment when you open it in a new tab.

(Eg 0.5rem of whitespace above the top of the green box, when you first navigate to a comment link)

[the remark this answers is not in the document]
akrolsmir: If either of these are hard, though, don't worry too much about it!

akrolsmir (approved):
This looks super lovely, can't wait to use it!

Major suggestion: Could we also link to answers in a Free Response market? Because it looks kind of like a comment too, so it's weird you can't link to it.

Visually, you could just add the link icon after "answered", or you could add the date after answered as it looks today; I think either might work fine.

(Also, this could totally be done in a follow up PR!)

web/components/feed/feed-items.tsx
```diff
@@ -221,6 +223,27 @@ export function FeedComment(props: {
...
+    }
+  }, [router.asPath])
+
+  function copyLink(event: React.MouseEvent<HTMLAnchorElement, MouseEvent>) {
+    event.preventDefault()
+    let currentLocation = window.location.href
+    if (currentLocation.includes('#')) {
+      currentLocation = currentLocation.split('#')[0]
+    }
+    navigator.clipboard.writeText(`${currentLocation}#${comment.id}`)
```
jahooma, line 242: You should use `copyToClipboard` the library function (in copy.ts), which handles weird cases like Firefox on Linux.

web/components/feed/feed-items.tsx
```diff
@@ -271,7 +299,37 @@ export function FeedComment(props: {
...
+              passHref={true}
+            >
+              <a
+                onClick={(event) => copyLink(event)}
+                className={'cursor-pointer hover:underline'}
+              >
+                <span className="ml-2 whitespace-nowrap text-gray-400">
+                  {fromNow(createdTime)}
+                  {showToast && (
+                    <div
```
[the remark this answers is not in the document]
jahooma: +1. I think it's the CopyLinkButton and ShareEmbedButton components

jahooma (commented):
Nice work!

One minor complaint I have is that it underlines the date when you mouse over the link icon, and so it makes it less clear what is happening. (Plus you can click the "2 months ago" to copy the link, which doesn't make sense.)
Screen Shot 2022-05-16 at 7 16 10 PM

IanPhilips:
Thanks for the the pointer @jahooma! Austin and I found that it is standard to be able to click a timestamp to navigate it although I never knew this nor used it before. So this is a hybrid of that function. Do you think the user should just be able to click the link icon? Personally I like the share button that reddit has as well.

akrolsmir:
@jahooma clicking on date to get a link is actually quite standard, it exists here on Github and Reddit for example.

Also adding the date makes the click target much larger, so I'm very strongly in favor of keeping it as is

web/components/feed/feed-items.tsx
```diff
@@ -221,6 +226,31 @@ export function FeedComment(props: {
...
+      setHighlighted(true)
+    }
+  }, [router.asPath])
+
+  function copyLinkToComment(
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>
+  ) {
+    event.preventDefault()
+
+    let currentLocation = window.location.href.includes('/home?u=')
```
jahooma, line 243: Nice detail!

web/components/toast-clipboard.tsx
```diff
@@ -0,0 +1,21 @@
+import { ClipboardCopyIcon } from '@heroicons/react/outline'
+import React from 'react'
+import clsx from 'clsx'
+import { Row } from 'web/components/layout/row'
+
+export function ToastClipboard(props: { className?: string }) {
```
jahooma, line 6: Thanks!

jahooma (approved):
Ok cool, I'm convinced that it's a standard. Approved!

Although, maybe it's the timestamp tooltip that makes it seem more confusing here? Or that the underline doesn't extend through the link icon. In any case, it's fairly clear

web/components/feed/feed-items.tsx
akrolsmir: 😠 we need to get rid of the `/home?u=` thing

IanPhilips:
Checking out adding links to the answer comments rn

jahooma:
I like the background color change on hover! Makes more sense now

PR #241, opened 2022-05-17
Rewrite client for new public APIs to use `fetch` instead of callables

mqp:
Followup to #227. Again, not changing the contract of the API; just doing all this plumbing work first. After this, now that we are in control, we will start fixing up the actual API contract on the functions and client to be something we endorse for public consumption.

Should be no user-visible changes.


mqp:
@mantikoros Any thoughts?

web/components/answers/answer-bet-panel.tsx
```diff
@@ -52,22 +52,26 @@ export function AnswerBetPanel(props: {
...
-      setIsSubmitting(false)
-    }
+    })
+      .then((r) => {
+        console.log('placed bet. Result:', r)
+        setIsSubmitting(false)
+        setBetAmount(undefined)
+        props.closePanel()
+      })
+      .catch((e) => {
```
mantikoros, line 66: I like the more explicit / less hacky error handling.

mantikoros (approved):
LGTM. Feel free to merge in if you've made sure everything works.

web/components/answers/answer-bet-panel.tsx
mqp: Good! Me too.

Issue #206, opened 2022-05-12
Sign in to comment button overlapping input


jahooma:
<img width="936" alt="Screen Shot 2022-05-12 at 7 09 28 PM" src="https://user-images.githubusercontent.com/1593821/168181758-a402d922-0d5c-41c6-aa5c-799e23e4177d.png">

Issue #228, opened 2022-05-15
https://manifold


manifoldbot:
markets/Gurkenglas/suggest-features shows every comment-on-an-answer an additional time under General Comments
https://discord.com/channels/915138780216823849/943628693958459462/975383868859367424

— Gurkenglas, from Discord

Issue #207, opened 2022-05-13
i did two trades and then two comments


manifoldbot:
my second comment got labelled with the first trade which seems wrong? https://manifold.markets/jack/will-instacart-file-to-ipo-in-2022
https://discord.com/channels/915138780216823849/943628693958459462/974470306179743754

— jack, from Discord

description: 88% chance. Resolves YES if in 2022 Instacart files to go public (e.g. files S-1). Going public may include IPO, direct listing, SPAC, etc.

Context: Instacart is one of the 10 most valuable startups according to https://www.cbinsights.com/research-unicorn-companies, but cut its valuation 40% from $39B to $24B amidst the recent tech selloff, whi...
[...]

Issue #199, opened 2022-05-12
all old comments on this market appear twice:


manifoldbot:
https://manifold.markets/BoltonBailey/what-will-be-the-next-country-to-jo
https://discord.com/channels/915138780216823849/943628693958459462/974364423995138079

— Yev, from Discord

description: If no new country joins NATO before February 23, 2032, this market resolves to N/A.
thumbnail: {'url': 'https://manifold-og-image.vercel.app/m.png?question=What%20will%20be%20the%20next%20country%20to%20join%20NATO%3F&metadata=Feb%2023%20%E2%80%A2%20Closes%20Feb%2024%2C%205%3A59am%20%E2%80%A2%20M%241%2C057%20bet%20%E2%80%A2%20%23World%20%23Geopolitics%20%23NATO&creatorName=Bolton%20Bailey&creatorAvatarUrl=https%3A%2F%2Ffirebasestorage.googleapis.com%2Fv0%2Fb%2Fmantic-markets.appspot.com%2Fo%2Fuser-images%252FBoltonBailey%252Fsandpile.gif%3Falt%3Dmedia%26token%3D0ee5d4ad-984f-439d-a636-412c606f9103&creatorUsername=BoltonBailey', 'proxy_url': 'https://images-ext-1.discordapp.net/external/Pomk3LWqUbQIWSFCVbyR0VXaKZKdPlObJZR0UuZhjes/%3Fquestion%3DWhat%2520will%2520be%2520the%2520next%2520country%2520to%2520join%2520NATO%253F%26metadata%3DFeb%252023%2520%25E2%2580%25A2%2520Closes%2520Feb%252024%252C%25205%253A59am%2520%25E2%2580%25A2%2520M%25241%252C057%2520bet%2520%25E2%2580%25A2%2520%2523World%2520%2523Geopolitics%2520%2523NATO%26creatorName%3DBolton%2520Bailey%26creatorAvatarUrl%3Dhttps%253A%252F%252Ffirebasestorage.googleapis.com%252Fv0%252Fb%252Fmantic-markets.appspot.com%252Fo%252Fuser-images%25252FBoltonBailey%25252Fsandpile.gif%253Falt%253Dmedia%2526token%253D0ee5d4ad-984f-439d-a636-412c606f9103%26creatorUsername%3DBoltonBailey/https/manifold-og-image.vercel.app/m.png', 'width': 1200, 'height': 630}
title: What will be the next country to join NATO?
type: article
[...]

PR #246, opened 2022-05-17
Comments ux improvements and bugfixes

### LINKED ISSUE #237: text is too light when composing a comment
https://discord.com/channels/915138780216823849/943628693958459462/975838287199633478

— palebone, from Discord

IanPhilips:
Fixes #237 #228  #207 #199

Also moves from showing peoples' value of mana given that they sell it to simply showing which choice they're most invested in.

WIP - going to fix that overlapping sign in to comment button too


IanPhilips:
Fixed #206

IanPhilips:
The changes are actually pretty simple but due to the refactoring appear large.

web/components/feed/activity-items.ts
```diff
@@ -304,8 +302,10 @@ function getAnswerAndCommentInputGroups(
         answer,
         items,
         user,
-        betsByCurrentUser: answerBets.filter((bet) => bet.userId === user?.id),
-        comments: answerComments,
+        betsByCurrentUser: betsByCurrentUser,
```
jahooma, line 305: nit: could just be
`betsByCurrentUser,`

web/components/feed/activity-items.ts
```diff
@@ -520,6 +520,13 @@ export function getRecentContractActivityItems(
   return [questionItem, ...items]
 }
 
+function commentIsGeneralComment(comment: Comment, contract: Contract) {
+  return (
+    comment.answerOutcome === undefined &&
+    (contract.outcomeType !== 'BINARY' ? comment.betId === undefined : true)
```
jahooma, line 526: Stephen is adding numerical markets soon. Might be better to check if it is equal to 'FREE_RESPONSE' instead.

web/components/feed/feed-answer-comment-group.tsx
```diff
@@ -28,23 +31,24 @@ export function FeedAnswerCommentGroup(props: {
...
+    commentsByCurrentUser ?? [],
     user,
     answer.number + ''
   )
   const prob = getDpmOutcomeProbability(contract.totalShares, answer.id)
   const probPercent = formatPercent(prob)
   const [open, setOpen] = useState(false)
   const [showReply, setShowReply] = useState(false)
-  const isFreeResponseContractPage = comments
+  const isFreeResponseContractPage = commentsByCurrentUser
```
jahooma, line 51: nit: If the variable starts with `is` one might assume it is a boolean. You could assign to `!!commentsByCurrentUser` to  convert it.

web/components/feed/feed-comments.tsx
```diff
@@ -1,35 +1,487 @@
...
+                    />
+                  </>
+                </>
+              )}
+            </div>
+
+            <Row className="grid grid-cols-8 gap-1.5 text-gray-700">
+              <Col className={'col-span-4 sm:col-span-6'}>
+                <Textarea
+                  ref={(ref: HTMLTextAreaElement) => setRef?.(ref)}
```
jahooma, line 329: I know this is just moved code, but I think this could be `ref={setRef}`

web/components/feed/feed-comments.tsx
```diff
@@ -1,35 +1,487 @@
...
+    setComment(replacement + comment.replace(replacement, ''))
+  }, [user, replyToUsername])
+
+  async function submitComment(betId: string | undefined) {
+    if (!user) {
+      return await firebaseLogin()
+    }
+    if (!comment) return
+
+    // Update state asap to avoid double submission.
```
jahooma, line 274: Another solution FYI would be to have an isSubmitting boolean, and call setIsSubmitting before and after the async call. Then disable submits while isSubmitting is true.

This solution works ok though, unless we need to handle errors or want the visual feedback of showing the comment button disabled on submit.

web/components/feed/feed-comments.tsx
```diff
@@ -1,35 +1,487 @@
...
+  const emptyReturn = {
+    userPosition: 0,
+    outcome: '',
+  }
+  const previousBets = bets.filter(
+    (prevBet) => prevBet.createdTime < createdTime && !prevBet.isAnte
+  )
+
+  if (contract.outcomeType !== 'BINARY') {
+    const answerCounts: any = {}
```
jahooma, line 448: `const answerCounts: { [outcome: string]: number } = {}`

web/components/feed/feed-comments.tsx
```diff
@@ -1,35 +1,487 @@
...
+    for (const bet of previousBets) {
+      if (bet.outcome) {
+        if (!answerCounts[bet.outcome]) {
+          answerCounts[bet.outcome] = bet.amount
+        } else {
+          answerCounts[bet.outcome] += bet.amount
+        }
+      }
+    }
+    const majorityAnswer = Object.keys(answerCounts).reduce(
```
jahooma, line 458: This is nice. Could be `_.maxBy(Object.keys(answerCounts), outcome => answerCounts[outcome])`

jahooma (approved):
Cool! I like the idea of splitting out some of the feed components into separate files since there is so much code.

I came up with a bunch of really minor suggestions for the code being moved, none are that important!

akrolsmir:
image

Hm, "Austin was betting on YES" is quite long, I'd prefer something succinct like "Austin held YES".

Also if I believe it's 30% and I bet YES at 10%, then it goes to 50%, the message kind of implies I'm still long YES even though I may actually prefer to sell now. Maybe the thing is to surface the last thing they purchased? E.g. "Austin held YES at 10%" might be the actual thing to demonstrate

> The changes are actually pretty simple but due to the refactoring appear large.

Not an issue for this PR, but in the future please consider splitting up changes and refactoring into two PRs for easy readability. (The refactor could just be a straight commit to `main`) See: https://google.github.io/eng-practices/review/developer/small-cls.html#refactoring

jahooma:
I mostly agree with Austin. "held YES" is better; in the meeting Stephen proposed "is betting YES", which I think is even better because "held" or holding a position is still somewhat technical. It doesn't need to be past tense because it's a record of what was once the present.

Adding the market probability when they commented is not a bad idea, but maybe there's a more systematic way to show how the probability updated, e.g. in a column next to every comment. Something to consider.

Also, I think we should make it standard practice to screenshot the relevant change in a PR as well. It not only makes the PR more clear, but it helps us not forget the visual side.

IanPhilips:
Screen Shot 2022-05-18 at 8 30 13 AM

PR #254, opened 2022-05-18
Show embed view of normal market links in iframes

sipec:
Now, if a user puts a `/user/[contractSlug]` link rather than a `/embed/user/[contractSlug]` into an iframe, it shows the embed view correctly anyways. This change allows you to easily use the "copy link" button to embed into Notion, LW, etc.
(well, it would have worked in LW regardless, because ~they~ I fix the url on their end, but still)

Perhaps we don't even need an embed route?


web/pages/[username]/[contractSlug].tsx
```diff
@@ -36,6 +36,8 @@ import { useWindowSize } from 'web/hooks/use-window-size'
 import Confetti from 'react-confetti'
 import { FeedComment } from 'web/components/feed/feed-comments'
 import { FeedBet } from 'web/components/feed/feed-bets'
+import { inIframe, useInIframe } from 'web/lib/util/embed'
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:39]: inIframe unused?

web/lib/util/embed.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react'
+
+export function inIframe() {
```
jahooma, line 3: Might not need to export if you should always call it from an effect clientside, in which case you would just use the hook below.

web/lib/util/embed.ts
```diff
@@ -0,0 +1,16 @@
...
+  try {
+    return window.self !== window.top
+  } catch (e) {
+    return true
+  }
+}
+
+// use a hook so this calculation happens client side
+export function useInIframe() {
```
jahooma, line 12: You could move this file into the hooks directory, especially if you are only exporting hooks.

jahooma, line 12: I tend to like verbose names, but to me `useIsInIframe` is better. Wait! Just do `useIsIframe`. I think that's best.

jahooma (changes requested):
Awesome idea! As you said, I think we should just use the market url for everything then!

Feel free to create a follow-up PR to remove the embed page and update the url copied from the embed button. Or, just make that change here!

PR #255, opened 2022-05-18
Show prob at time of comment on binary markets

### LINKED ISSUE #252: "Austin is betting NO at 0%“ isn't accurate
I bet NO down to 30% but it looks like it's using the current prob instead of prob at time of bet
https://discord.com/channels/915138780216823849/943628693958459462/976562553616466010
Attached: 
https://cdn.discordapp.com/attachments/943628693958459462/976562552614043718/Screenshot_20220518-150922.png

— akrolsmir, from Discord

thumbnail: {'url': 'https://cdn.discordapp.com/attachments/943628693958459462/976562552614043718/Screenshot_20220518-150922.png', 'proxy_url': 'https://media.discordapp.net/attachments/943628693958459462/976562552614043718/Screenshot_20220518-150922.png', 'width': 1080, 'height': 2400}
type: image
url: https://cdn.discordapp.com/attachments/943628693958459462/976562552614043718/Screenshot_20220518-150922.png

IanPhilips:
The easiest way to do this was just get the soonest bet before the comment and read its probAfter. This doesn't work on fr questions though (because a bet on a different answer may have determined the comment answer's prob) and unless you know of an easy way I'm going to leave it for now as I'm not confident enough in this feature to figure it out.
Fix #252

### DIFF
--- a/web/components/feed/feed-comments.tsx
+++ b/web/components/feed/feed-comments.tsx
@@ -21,7 +21,7 @@ import { Linkify } from 'web/components/linkify'
 import { SiteLink } from 'web/components/site-link'
 import { BetStatusText } from 'web/components/feed/feed-bets'
 import { Col } from 'web/components/layout/col'
-import { getOutcomeProbability, getProbability } from 'common/calculate'
+import { getProbability } from 'common/calculate'
 import { LoadingIndicator } from 'web/components/loading-indicator'
 
 export function FeedCommentThread(props: {


IanPhilips:
Perhaps I should just add probAt to the comment model?

jahooma:
Yeah, I think this needs more spec work. I think probabilities could be shown, but probably in a more systematic way. I'd favor removing the probability for now from free comments.

But if you think this is better, also happy to just approve and move on.

web/components/feed/feed-comments.tsx
```diff
@@ -224,15 +233,17 @@ export function getMostRecentCommentableBet(
...
-  const { contract, outcome } = props
+function CommentStatus(props: {
+  contract: Contract
+  outcome: string
+  prob?: number
+}) {
+  const { contract, outcome, prob } = props
   return (
     <>
       {' betting '}
```
mantikoros, line 244: Maybe use " betting on" instead? When the FR answer is number it looks like you're betting that amount.

web/components/feed/feed-comments.tsx
```diff
@@ -71,6 +65,15 @@ export function FeedCommentThread(props: {
             comment={comment}
             betsBySameUser={betsByUserId[comment.userId] ?? []}
             onReplyClick={scrollAndOpenReplyInput}
+            probAtCreatedTime={
```
mantikoros, line 68: Reasonable-looking hack. Might want a more robust solution in the long-term, or to ax probability displays in the comments entirely. I personally am not sure.

mantikoros (commented):
Reasonable fix for now. If we are showing probabilities on comments, this is the way to go. But I'm still a little ambivalent if that's what we want to do in the long term. Feel free to merge this in now if you want.

IanPhilips:
It's better than what there is now, and what we have now is a bug so I'm going to merge this while I think about it more.

PR #266, opened 2022-05-19
Public API improvements

mqp:
_**Function status: Not yet deployed to dev or prod. The function changes and client changes must be deployed together.**_

Simple changes to make the public API more fit for actual public consumption. Since the changes are very simple and there's little risk of serious complication, there's no upgrade strategy; once I have tested it on dev, I am just gonna break the site for a couple minutes while I deploy the functions and client and they match again.

### DIFF
--- a/functions/src/api.ts
+++ b/functions/src/api.ts
@@ -115,15 +115,13 @@ export const newEndpoint = (methods: [string], fn: Handler) =>
         const allowed = methods.join(', ')
         throw new APIError(405, `This endpoint supports only ${allowed}.`)
       }
-      const data = await fn(req, res)
-      data.status = 'success'
-      res.status(200).json(data)
+      res.status(200).json(await fn(req, res))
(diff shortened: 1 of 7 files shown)


mqp:
Hey @akrolsmir, here's a question for you. Right now, as an implementation detail of our website, we rely on `createContract` returning more or less a fully fleshed out contract object. However, from a public API point of view, I would have tended to be more conservative, and only return stuff that is likely to be kind of stable, rather than just exposing our database document to clients.

Would you say you are comfortable with the "just return basically the whole database document" strategy for now?

akrolsmir:
Hm, that's a good question. I don't feel strongly one way or the other. What things would you like to cut?

I'm also wondering if there's any chance we have clients who depend on those things existing (I think our main API consumer is metaforecast, atm)

akrolsmir (approved):
LGTM (on a very quick skim)

PR #268, opened 2022-05-19
Simplify market creation page

IanPhilips:
This simplifies the wording and choices around probability of an event happening as well as duration of the question. It still allows the advanced user to get more specific, however.
Screen Shot 2022-05-19 at 5 38 52 PM


IanPhilips:
I'm using indigo-600 as the background on the toggles as it offers more contrast with the white text so is easier on the eyes. Indigo-500 looks like it's the standard for colored text, however

jahooma (approved):
Looks pretty good! I like the simpler probability selection and close date. Not letting the user pick the exact close time (only the date) is mostly good.

Just a couple suggestions for you. Use your judgment.

jahooma:
The academics Sharef & Leon suggested the buttons for quickly choosing a close date, just like you've implemented!

They also wanted 6 months and a year buttons. I think at the least you should add a button for a year!

See: https://www.notion.so/manifoldmarkets/Prediction-markets-on-research-academic-papers-94ba4cfe441247d0873d365d6ed8faf2#f69da9b429304bd89dfb3872f65c13de

IanPhilips:
Thanks for the review, James! Adding a year button and/or a custom button to the probability row breaks the row into two rows on my phone and looks quite bad. I'm going to graph the distribution of market durations up until now and see if the choices I added make sense and/or if a year button is worth adding. I like removing the custom button for probabilities as I think clustering the liquidity around those non-poles seems like it will result in a better UX for bettors.

PR #271, opened 2022-05-20
Minor simplification to `placeBet`.

Sjlver:
### Low-priority pull request, feel free to close without comment.

I carefully read through the code of `placeBet`, because this is one
area where bugs might allow me to cheat ;-) I think that the code is
fine and I didn't find any obvious bugs.

That said, there was one confusing use of `Math.abs(amount)`. Yet,
`amount` is a const and we test it in line 21:

`if (amount <= 0 || isNaN(amount) || !isFinite(amount)) throw ...`

Thus, `amount` cannot be negative at this point, and the `Math.abs` is a
no-op. I believe it would be better to remove it, to avoid the
programmer thinking "hmm, I thought this couldn't be negative".


mqp (approved):
This seems clearly correct and thanks for cleaning up the dead code. I will merge and deploy soon.

mqp:
I'm taking care of this as part of #302.

PR #273, opened 2022-05-20
Require a minimum amount of 1 Mana per bet.

Sjlver:
This is a hacky patch for a problem that runs more deeply: the use of
floating-point for Mana calculations, along with the associated rounding
errors.

Consider the following example:

```typescript
const balance = 1000
const bet = 5.6e-14
const newBalance = balance - bet
if (newBalance == balance) {
  alert(`I placed a bet of ${bet} without changing my balance.`)
}
```

This will actually print a message, because floating-point numbers can
only represent so many digits, and thus we get 1000.0 rather than
999.99999999999994.

This is a purely theoretical attack at this point; nobody could create
enough pico-bets to actually affect their balance using this technique.
However, I believe is worth ensuring a minimum bet amount, and might
prevent other problems such as the UI showing messages like "User Foo
bought M0 of YES", which could confuse users.

For a more definite solution, we would probably need to change some
computation to integers, where addition is always commutative and
value-preserving. This is similar to what most financial software does
(e.g., Bitcoin uses 1 Satoshi = 0.00000001 BTC as their unit).

### DIFF
--- a/functions/src/place-bet.ts
+++ b/functions/src/place-bet.ts
@@ -18,7 +18,7 @@ export const placeBet = newEndpoint(['POST'], async (req, _res) => {
   const [bettor, _privateUser] = await lookupUser(await parseCredentials(req))
   const { amount, outcome, contractId, value } = req.body.data || {}
 
-  if (amount <= 0 || isNaN(amount) || !isFinite(amount))
+  if (amount < 1 || isNaN(amount) || !isFinite(amount))
     throw new APIError(400, 'Invalid amount')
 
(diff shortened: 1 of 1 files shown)


mqp (approved):
I like this idea, because even if we had infinite-precision calculations, it's never going to be the case that bets are "free" to us -- people can DoS the servers and UI with tiny bets. Maybe in a perfect world we would handle that in some other way (like a minimum fee on bets) but this seems like a practical thing to do. I will merge and deploy this soon.

mantikoros:
LGTM!

PR #275, opened 2022-05-20
Shrink avatar, paddding, gaps for mobile screens

IanPhilips:
I was testing out the mobile commenting experience and comparing it to reddit. I think there's too much wasted space and the avatars are too big (The avatars for comments and comment inputs are all hard coded to small. I tried making them larger for larger screen sizes but I just think they're too large for all cases). the following screenshots are the same screen height. 
  

new: Screen Shot 2022-05-20 at 10 02 48 AM
old: Screen Shot 2022-05-20 at 10 02 25 AM


akrolsmir (approved):
Neat improvement!

image

Would like a smidgen more (like p-0.5) horizontal padding between avatar and text

image

Looks a lot better than the old comment button! Would shrink the padding between the arrow and the text box.

Also, is there any way to include the button inside the text box (eg with absolute positioning, right: 0 and bottom: 0)? I wonder if that would look better, anyways.

IanPhilips:
I like it! 

Screen Shot 2022-05-20 at 4 46 24 PM
Screen Shot 2022-05-20 at 4 46 03 PM

akrolsmir:
A few more padding thoughts:
image

Avatar could use a smidgen more padding away from the line on the left
Also, are we no longer showing the lines in the replies? I think that makes the avatars look visually unbalanced/free-floating

Actually, how about something like this:
image

----

image

Right side of the box should not go to the edge, I think we want to maintain p-2 around mobile pages

PR #276, opened 2022-05-20
Add countdown timer for daily free market

IanPhilips:
Added a countdown to a user's next free market. Counts down the hours and minutes until both are 0, then counts down the seconds. Keeping the seconds in there the whole time is visually distracting until the fun end when you're about to get the freebie! Not pictured but I did take out the 'daily' part
Screen Shot 2022-05-20 at 3 11 52 PM
Screen Shot 2022-05-20 at 3 12 16 PM


web/hooks/use-has-created-contract-today.ts
```diff
@@ -4,16 +4,20 @@ import { User } from 'common/user'
 
 let sessionCreatedContractToday = true
 
+export function getUtcMidnightToday() {
+  // Uses utc time like the server.
+  const utcTimeString = new Date().toISOString()
+  return new Date(utcTimeString).setUTCHours(0, 0, 0, 0)
```
akrolsmir, line 10: Actually, if we're going to pick a hardcoded time, I might try 4PM UTC! It's noon in ET, and 9am in PT; I think US/UK is our main audience for now, and it might be cool & habit-forming to have the countdown when everyone is awake/online; could imagine everyone logging on at the same time each day and betting on all the new markets haha

akrolsmir (approved):

web/hooks/use-has-created-contract-today.ts
akrolsmir: (On balance, I'm leaning (70%) "user's local midnight" to still be the best design, but it's a lot more work afaict)

PR #277, opened 2022-05-20
Fix up API CORS header processing

mqp:
This needs to be working and also the client API is about to have to generate real CORS headers for `placeBet` and `createContract` APIs, so get it into `common` instead of the `functions` package.

### DIFF
--- a/functions/src/api.ts
+++ b/functions/src/api.ts
@@ -3,6 +3,10 @@ import * as functions from 'firebase-functions'
 import * as Cors from 'cors'
 
 import { User, PrivateUser } from 'common/user'
+import {
+  CORS_ORIGIN_MANIFOLD,
+  CORS_ORIGIN_LOCALHOST,
+} from 'common/envs/constants'
 
 type Request = functions.https.Request
 type Response = functions.Response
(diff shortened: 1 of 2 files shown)


functions/src/api.ts
```diff
@@ -107,7 +108,7 @@ export const applyCors = (req: any, res: any, params: object) => {
 export const newEndpoint = (methods: [string], fn: Handler) =>
   functions.runWith({ minInstances: 1 }).https.onRequest(async (req, res) => {
     await applyCors(req, res, {
-      origins: [CORS_ORIGIN_MANIFOLD, CORS_ORIGIN_LOCALHOST],
```
mqp, line 110: Very sad... Typescript! I needed you!

functions/src/api.ts
```diff
@@ -107,7 +108,7 @@ export const applyCors = (req: any, res: any, params: object) => {
...
+      origin: [CORS_ORIGIN_MANIFOLD, CORS_ORIGIN_LOCALHOST],
```
mqp, line 111: Very sad... Typescript! I needed you!!!

mqp:
OK, I tested this a lot and CORS is really working now, I swear!

PR #280, opened 2022-05-21
Add Vercel routes for public write API

mqp:
Now you can `POST /api/v0/markets` to create a market and `POST /api/v0/bets` to make a bet. These routes will just forward your request to the relevant cloud function and do zero intelligent work on their own.

### DIFF
--- a/common/envs/dev.ts
+++ b/common/envs/dev.ts
@@ -6,6 +6,7 @@ export const DEV_CONFIG: EnvConfig = {
     apiKey: 'AIzaSyBoq3rzUa8Ekyo3ZaTnlycQYPRCA26VpOw',
     authDomain: 'dev-mantic-markets.firebaseapp.com',
     projectId: 'dev-mantic-markets',
+    region: 'us-central1',
     storageBucket: 'dev-mantic-markets.appspot.com',
     messagingSenderId: '134303100058',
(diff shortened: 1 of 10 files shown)


akrolsmir:
This looks super, super cool! I'm going to ask @sipec to be the main reviewer on this, so that we spread out knowledge on how the API works a bit more.

Also, could you document how the write API will work? For this PR, just the parameters etc that will be used to create/bet on markets. Later ofc we'll want to publish a full "how to use the Manifold Write API" documentation (eg where to get token, how to form requests, ideally code snippets you can run in Javascript and Python)

mqp:
> Also, could you document how the write API will work? For this PR, just the parameters etc that will be used to create/bet on markets. Later ofc we'll want to publish a full "how to use the Manifold Write API" documentation (eg where to get token, how to form requests, ideally code snippets you can run in Javascript and Python)

Of course, I am just working through it gradually. Currently I feel that all the "general API machinery" is good to go, so next I want to revisit the specific parameters and outputs of our existing APIs to make sure they make sense, and then I will write documentation and we can announce it.

mqp:
@sipec any comment?

sipec (approved):
This looks pretty good!
(Thanks for being patient 😅 )

PR #288, opened 2022-05-21
Kill Husky, lint-staged, and pretty-quick

mqp:
Now that we have visual feedback on GitHub whenever something we commit is busted, we don't need to enforce this in local development anymore.

(Of course, you will still save your time and energy if you run `prettier` and `eslint` in your editor. I think we all do anyway?)


akrolsmir (approved):
I spent a long time getting this working, but will be happy to see it gone!

PR #291, opened 2022-05-22
Implement quick betting: directly from the market card

akrolsmir:
Okay, this is good to go! Some things that are missing that can be done in follow-ups:
- [ ] Can't bet on Numeric markets yet. I think upbetting should just be a uniform bet on [EV, max], though I'd be okay with a normally distributed bet at (EV + max) / 2, I guess.
- [ ] Can't short on Free Response markets yet. Future work for James~
- [ ] The useSaveShare thing is a bit slow, so there's a lag between betting yes and having the UI reflect it. Plausibly, we should optimistically show the final result, and sync with server later
- [ ] Should batch together multiple clicks by debouncing with a eg 500ms window, rather than spamming the feed with multiple M$10 bets
- [ ] Maybe increase default bet size based on market volume? or on user wealth?


web/components/contract/contract-card.tsx
```diff
@@ -71,80 +87,233 @@ function getColor(contract: Contract) {
...
   showCloseTime?: boolean
   className?: string
 }) {
-  const { contract, showHotVolume, showCloseTime, className } = props
+  const { showHotVolume, showCloseTime, className } = props
+  const [liveUpdate, setLiveUpdate] = useState(false)
+  const [contract, setContract] = useState<Contract>(props.contract)
   const { question, outcomeType } = contract
 
+  // When liveUpdate is true, start keeping this contract in sync
```
akrolsmir, line 241: This mostly works, but ESLint is complaining

```
246:6  Warning: React Hook useEffect has a missing dependency: 'contract'. Either include it or remove the dependency array. You can also do a functional update 'setContract(c => ...)' if you only need 'contract' in the 'setContract' call.
```

Any tips, React wizards?

jahooma: Outside the effect:
```const contractId = contract.id```
then,
```
useEffect(() => {
if (liveUpdate) {
      return listenForContract(contractId, (c) => c && setContract(c))
}, [contractId, liveUpdate])
```
Don't forget to return the listener so that it unsubscribes.

Comment (jahooma) [reply web/components/contract/contract-card.tsx:43]: But also, why not just always live update? Call useContract instead of all this.

(Will do a full review of this PR in a bit.)

Comment (akrolsmir) [reply web/components/contract/contract-card.tsx:43]: Lol I guess I assumed it was extra expensive to be listening to changes on all contract cards. Maybe not!

File: web/components/contract/quick-bet.tsx [hunk 2]
```diff
@@ -0,0 +1,242 @@
...
+}) {
+  const { contract, setLiveUpdate } = props
+
+  const user = useUser()
+  const userBets = useUserContractBets(user?.id, contract.id)
+  const { yesFloorShares, noFloorShares, yesShares, noShares } = useSaveShares(
+    contract as FullContract<CPMM | DPM, Binary>,
+    userBets
+  )
+  // For some reason, Floor Shares are inverted for non-BINARY markets
```
jahooma, line 42: `useSaveShares` is only for binary markets, currently? yes/no floor shares is probably nonsense for non-binary

You could update the hook to save shares for FR and Numeric.

web/components/outcome-label.tsx
```diff
@@ -84,11 +84,12 @@ export function FreeResponseOutcomeLabel(props: {
   if (!chosen) return <AnswerNumberLabel number={resolution} />
   return (
     <FreeResponseAnswerToolTip text={chosen.text}>
-      <AnswerLabel
+      <span className="text-blue-400">TOP</span>
```
jahooma, line 87: Is this for testing? Did you mean to put back the commented out AnswerLabel?

web/components/contract/quick-bet.tsx
```diff
@@ -0,0 +1,242 @@
...
+        'bg-gray-50'
+      )}
+    >
+      {/* Up bet triangle */}
+      <div>
+        <div
+          className="peer absolute top-0 left-0 right-0 h-[50%]"
+          onClick={() => placeQuickBet('UP')}
+        ></div>
+        <div className="mt-2 text-center text-xs text-transparent peer-hover:text-gray-400">
```
jahooma, line 100: I think this should animate opacity on hover.

`transition-opacity ease-in duration-700 opacity-100 peer-hover:opacity-0`

Ditto for down arrow

web/components/contract/contract-card.tsx
jahooma: I don't think it's that expensive! If they scroll down a bit, might fetch ~100 docs. If the markets are very active, could be more.

It's actually nice for the cards to show fresh data, too.

web/components/outcome-label.tsx
jahooma: Might have to rearrange the card for resolved FR markets to show the resolved answer at the bottom of the card.

jahooma (approved):
Pretty neat change! I think some people will love it. Let's try it and see.

Visually, it looks pretty good, although it leaves less space for the question text and is somewhat busier. And FR market cards don't look very distinct — there might be a way to improve that.

jahooma:
bug: Should say `22 expected` instead of just `22`
Screen Shot 2022-05-23 at 4 54 53 PM

web/components/contract/contract-details.tsx
```diff
@@ -26,60 +28,82 @@ import NewContractBadge from '../new-contract-badge'
...
-        <Row className="items-center gap-2">
-          <Avatar
-            username={creatorUsername}
-            avatarUrl={contract.creatorAvatarUrl}
-            size={6}
-          />
-          <UserLink name={creatorName} username={creatorUsername} />
-        </Row>
+    <Row className="items-center gap-3 text-sm text-gray-400">
+      {/* {contract.createdTime % 3 == 1 ? (
```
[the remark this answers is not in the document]
akrolsmir: Done!

web/components/contract/contract-details.tsx
```diff
@@ -26,60 +28,82 @@ import NewContractBadge from '../new-contract-badge'
...
+}
+
+export function AbbrContractDetails(props: {
+  contract: Contract
+  showHotVolume?: boolean
+  showCloseTime?: boolean
+}) {
+  const { contract, showHotVolume, showCloseTime } = props
+  return (
+    <Col className="gap-2">
```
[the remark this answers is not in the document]
akrolsmir: Removed!

web/components/contract/quick-bet.tsx
akrolsmir: Hm, well it's doing _something_, specifically it's useful for figuring out that there's a correct bet. Putting a TODO for now to investigate more in the future.

web/components/outcome-label.tsx
akrolsmir: Think we're already doing this: 
image

Removing the commented code!

web/components/contract/quick-bet.tsx
akrolsmir: Appreciate the tip! I'd love to get a nice animation in there. When I tried this code, though, it seemed to run into weird corner cases where it'd flicker a lot (perhaps due to the way peer opacity works??)

I'll submit for now without the animation, but please please add it in afterwards!

akrolsmir:
> I don't think it's that expensive! If they scroll down a bit, might fetch ~100 docs. If the markets are very active, could be more.

Haha I figured there was some reason we weren't already using useContract. Guess not!

> bug: Should say `22 expected` instead of just `22` Screen Shot 2022-05-23 at 4 54 53 PM

Fixed!

PR #293, opened 2022-05-22
Kill /about page

mqp:
It seems to me this isn't linked from anywhere at all. It contains a copy of some of the stuff on https://docs.manifold.markets, which is what we actually link to from the "about" menu entry.

The reason I bothered to delete it, is because it pulls in some CSS classes from Tailwind with some ridiculously complicated looking selectors that makes `next build` barf out a bunch of warnings, and it seems easier to just delete the problem than debug it.


akrolsmir (approved):
Lgtm, we might resurrect this in some way (specifically, the css code that does styling) as we do plan on having a guide/about page on our site eventually, but for now this is fine

PR #301, opened 2022-05-22
[WIP] Toast counts down for 3 seconds before placing the bet

akrolsmir:
There's some weird timing/React hook error that sometimes allows this thing to bet twice? IDK, my brain is fried trying to understand it.

Also, progress bar that scrolls down might be cooler than a number countdown? See https://www.npmjs.com/package/react-toastify, although the API for that is more unwieldy/less Tailwind-friendly?


web/components/contract/contract-card.tsx
```diff
@@ -346,3 +395,64 @@ export function NumericResolutionOrExpectation(props: {
...
+      setSecondsLeft((seconds) => seconds - 1)
+    }, 1000)
+    return () => clearInterval(interval)
+  }, [])
+
+  if (secondsLeft <= 0) {
+    console.log('finishing')
+    onToastFinish()
+    // return null
+  }
```
jahooma, line 423: Shouldn't this be in a useEffect?

e.g.
```
useEffect(() => {
  if (secondsLeft <= 0) {
    console.log('finishing')
    onToastFinish()
  }
}, [secondsLeft])
```

jahooma: Also, why not have a single setTimeout for the whole duration? You could cancel with clearTimeout on unmount.

jahooma: Oh I guess you render the seconds left...

akrolsmir:
Eh, maybe will pick this up someday.

I'm not sure about the "toast" implementation vs "something changes on the card directly", which would be less intrusive/hard to track.

PR #302, opened 2022-05-23
Serious business API validation & big cleanup of `createContract`, `placeBet`

mqp:
_**Function status: Deployed on dev for testing. Not yet deployed on prod.**_

- Dug up a good validation library ([`zod`](https://github.com/colinhacks/zod)) which will be useful in the future for validating API inputs.
- Applied it to `createContract` and `placeBet`, thereby implementing a variety of input validation that was sort of only mostly-implemented before (and making it work with Typescript properly.)
- Got rid of some unused code and made other refactors that shouldn't affect any behavior.

I still need to test all the different error cases to make sure this validation is working and produces acceptable error messages to API consumers, but I'm putting this out there to get eyes on it.

### DIFF
--- a/common/new-bet.ts
+++ b/common/new-bet.ts
@@ -18,13 +18,13 @@ import {
   Multi,
   NumericContract,
 } from './contract'
-import { User } from './user'
 import { noFees } from './fees'
 import { addObjects } from './util/object'
 import { NUMERIC_FIXED_VAR } from './numeric-constants'
 
+export type CandidateBet<T extends Bet> = Omit<T, 'id' | 'userId'>
 export type BetInfo = {
-  newBet: Bet
+  newBet: CandidateBet<Bet>
   newPool?: { [outcome: string]: number }
   newTotalShares?: { [outcome: string]: number }
   newTotalBets?: { [outcome: string]: number }
@@ -33,12 +33,10 @@ export type BetInfo = {
 }
 
 export const getNewBinaryCpmmBetInfo = (
-  user: User,
   outcome: 'YES' | 'NO',
   amount: number,
   contract: FullContract<CPMM, Binary>,
-  loanAmount: number,
-  newBetId: string
+  loanAmount: number
 ) => {
   const { shares, newPool, newP, fees } = calculateCpmmPurchase(
     contract,
@@ -50,9 +48,7 @@ export const getNewBinaryCpmmBetInfo = (
   const probBefore = getCpmmProbability(pool, p)
   const probAfter = getCpmmProbability(newPool, newP)
 
-  const newBet: Bet = {
-    id: newBetId,
-    userId: user.id,
+  const newBet: CandidateBet<Bet> = {
     contractId: contract.id,
     amount,
     shares,
@@ -71,12 +67,10 @@ export const getNewBinaryCpmmBetInfo = (
 }
 
 export const getNewBinaryDpmBetInfo = (
-  user: User,
   outcome: 'YES' | 'NO',
   amount: number,
   contract: FullContract<DPM, Binary>,
-  loanAmount: number,
-  newBetId: string
+  loanAmount: number
 ) => {
   const { YES: yesPool, NO: noPool } = contract.pool
 
@@ -104,9 +98,7 @@ export const getNewBinaryDpmBetInfo = (
   const probBefore = getDpmProbability(contract.totalShares)
   const probAfter = getDpmProbability(newTotalShares)
 
-  const newBet: Bet = {
-    id: newBetId,
-    userId: user.id,
+  const newBet: CandidateBet<Bet> = {
     contractId: contract.id,
     amount,
     loanAmount,
@@ -122,12 +114,10 @@ export const getNewBinaryDpmBetInfo = (
 }
 
 export const getNewMultiBetInfo = (
-  user: User,
   outcome: string,
   amount: number,
   contract: FullContract<DPM, Multi | FreeResponse>,
-  loanAmount: number,
-  newBetId: string
+  loanAmount: number
 ) => {
   const { pool, totalShares, totalBets } = contract
 
@@ -145,9 +135,7 @@ export const getNewMultiBetInfo = (
   const probBefore = getDpmOutcomeProbability(totalShares, outcome)
   const probAfter = getDpmOutcomeProbability(newTotalShares, outcome)
 
-  const newBet: Bet = {
-    id: newBetId,
-    userId: user.id,
+  const newBet: CandidateBet<Bet> = {
     contractId: contract.id,
     amount,
     loanAmount,
@@ -163,12 +151,10 @@ export const getNewMultiBetInfo = (
 }
 
 export const getNumericBetsInfo = (
-  user: User,
   value: number,
   outcome: string,
   amount: number,
-  contract: NumericContract,
-  newBetId: string
+  contract: NumericContract
 ) => {
   const { pool, totalShares, totalBets } = contract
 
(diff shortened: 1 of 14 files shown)


mqp:
I note that in this branch I changed the max question length and description to be validation errors. That seems like a good idea but I need to go in the client to make sure the UI lines up.

functions/src/api.ts
```diff
@@ -1,28 +1,33 @@
...
 import * as Cors from 'cors'
+import { z } from 'zod'
 
 import { User, PrivateUser } from '../../common/user'
 import {
   CORS_ORIGIN_MANIFOLD,
   CORS_ORIGIN_LOCALHOST,
 } from '../../common/envs/constants'
 
+type Output = Record<string, unknown>
```
mantikoros, line 12: Probably should use a generic type instead of unknown.

functions/src/api.ts
```diff
@@ -1,28 +1,33 @@
...
+type Handler = (req: Request, user: AuthedUser) => Promise<Output>
 type JwtCredentials = { kind: 'jwt'; data: admin.auth.DecodedIdToken }
 type KeyCredentials = { kind: 'key'; data: string }
 type Credentials = JwtCredentials | KeyCredentials
 
 export class APIError {
   code: number
   msg: string
-  constructor(code: number, msg: string) {
+  details: unknown
```
mantikoros, line 24: Details should be either a generic type or any. (Unknown should be used very sparingly in Typescript.)

functions/src/create-answer.ts
```diff
@@ -124,12 +126,6 @@ export const createAnswer = functions.runWith({ minInstances: 1 }).https.onCall(
         volume: volume + amount,
       })
 
-      if (!isFinite(newBalance)) {
```
mantikoros, line 127: I guess it's fine to remove the sanity check if amount is suitably validated.

functions/src/create-contract.ts
```diff
@@ -26,66 +27,42 @@ import {
...
-    throw new APIError(400, 'Invalid tags field')
-
-  tags = (tags || []).map((tag: string) =>
-    tag.toString().slice(0, MAX_TAG_LENGTH)
-  )
+const bodySchema = z.object({
+  question: z.string().min(1).max(MAX_QUESTION_LENGTH),
+  description: z.string().max(MAX_DESCRIPTION_LENGTH),
+  tags: z.array(z.string().min(1).max(MAX_TAG_LENGTH)).optional(),
+  closeTime: z.number(),
```
mantikoros, line 39: Check if closeTime is positive, in the future, and not too far in the future (< 100 years — or whatever you think the lifetime of Manifold is).

functions/src/place-bet.ts
```diff
@@ -1,146 +1,140 @@
...
+
+    const newBetDoc = contractDoc.collection('bets').doc()
+    const {
+      newBet,
+      newPool,
+      newTotalShares,
+      newTotalBets,
+      newTotalLiquidity,
+      newP,
+    } = await (async (): Promise<BetInfo> => {
```
mantikoros, line 64: Might want to extract this logic into a new (named) function

functions/src/create-contract.ts
```diff
@@ -26,66 +27,42 @@ import {
...
-
-  if (!question || typeof question != 'string')
-    throw new APIError(400, 'Missing or invalid question field')
-
-  question = question.slice(0, MAX_QUESTION_LENGTH)
-
-  if (typeof description !== 'string')
-    throw new APIError(400, 'Invalid description field')
-
-  description = description.slice(0, MAX_DESCRIPTION_LENGTH)
```
mantikoros, line 57: I guess we're now rejecting everything instead of doing truncation. Reasonable choice.

mantikoros (commented):
This more principled approach to validation seems like it could be helpful in preventing future errors. Feel free to merge in after testing. 

(Also, the unknown type in Typescript should be used very sparingly.)

jahooma:
Small comment: `unknown` is actually the right type to use if you are receiving data from a client that needs to be validated.

common/new-bet.ts
```diff
@@ -23,6 +23,15 @@ import { noFees } from './fees'
 import { addObjects } from './util/object'
 import { NUMERIC_FIXED_VAR } from './numeric-constants'
 
+export type BetInfo = {
+  newBet: Bet
+  newPool?: object
+  newTotalShares?: object
+  newTotalBets?: object
```
jahooma, line 30: I think object is not as specific as { [outcome: string]: number }. I would switch this for these three fields.

functions/src/create-contract.ts
```diff
@@ -94,58 +71,47 @@ export const createContract = newEndpoint(['POST'], async (req, _res) => {
...
   const contract = getNewContract(
     contractRef.id,
     slug,
-    creator,
+    user,
     question,
     outcomeType,
     description,
-    initialProb,
+    initialProb ?? 0,
```
jahooma, line 100: Shouldn't we not set initialProb for non-binary markets?

functions/src/create-contract.ts
mqp: 👍

functions/src/create-answer.ts
mqp: The `zod` `number` validator enforces finitude.

functions/src/create-contract.ts
```diff
@@ -94,58 +71,47 @@ export const createContract = newEndpoint(['POST'], async (req, _res) => {
...
     ante,
     closeTime,
     tags ?? [],
     NUMERIC_BUCKET_COUNT,
     min ?? 0,
     max ?? 0,
-    manaLimitPerUser ?? 0
+    0
```
jahooma, line 107: I think we should remove the manaLimitPerUser. Most of this feature has already been removed.

functions/src/api.ts
mqp: Can you clarify your criticism of the `unknown` type in these two cases? They seem conceptually correct to me. The intent is for them to represent
- any Javascript object
- which can be literally anything and we don't care what, because we are just a middleman echoing them through to `JSON.stringify`.

jahooma (commented):
Nice cleanup!

functions/src/create-contract.ts
mqp: Then I will!

functions/src/create-contract.ts
mqp: So what happened here is:

- The old code was typing `initialProb` as `any`, even though it could be `undefined`.
- So it passed `undefined` for non-binary markets into `getNewContract`, even though `getNewContract` claims to only take a `number`.
- However, inside `getNewContract`, it only uses `initialProb` or sets it on the contract in binary markets.

So I think this code actually behaves identically in the end to the old code. Of course, it would be better if `getNewContract` was better designed so it could accept `undefined` in cases where it was a binary contract. This should be doable with some type refactoring. But in the meantime, I think I prefer this (passing in a meaningless value that is the right type) to the previous status quo (passing in a meaningless value that is also the wrong type.)

functions/src/create-contract.ts
mqp: Yeah, this seems better for users than silently accepting their overlong things and then truncating them on our end. Especially since it's impossible to edit question text after your market is live.

functions/src/place-bet.ts
mqp: Unfortunately, it's kind of awkward to do that, since each branch relies on, like, validating additional parameters on the request body, and one branch looks up stuff using the Firestore transaction, so it feels really poorly factored to break it out. The broken out function would take like eight random parameters.

mqp:
OK, most feedback was addressed, I will merge this at my convenience.

PR #314, opened 2022-05-24
Update leaderboards.tsx

SirSaltyy:
Renamed top creators header from "market volume" to "market liquidity".


mantikoros:
Hmm... I think I'm going to revert this change. The label was changed from volume to liquidity but not the actual ranking, which is still in terms of volume. We could consider doing rankings by liquidity instead of volume in the future, but for now,  words mean things!

PR #318, opened 2022-05-24
Show just first 3 letters and chosen answer on fr cards

IanPhilips:
Screen Shot 2022-05-24 at 9 28 55 AM
Screen Shot 2022-05-24 at 9 30 33 AM
Screen Shot 2022-05-24 at 9 30 44 AM

### DIFF
--- a/web/components/contract/contract-card.tsx
+++ b/web/components/contract/contract-card.tsx
@@ -183,12 +183,14 @@ export function FreeResponseResolutionOrChance(props: {
       {resolution ? (
         <>
           <div className={clsx('text-base text-gray-500')}>Resolved</div>
-          <FreeResponseOutcomeLabel
-            contract={contract}
-            resolution={resolution}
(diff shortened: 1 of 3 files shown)


IanPhilips:
I'll make the changes suggested by stevo in a bit

IanPhilips:
Screen Shot 2022-05-24 at 1 54 56 PM

Screen Shot 2022-05-24 at 1 55 43 PM

I think I like keeping the resolved title.

mqp:
I know you changed it, but FYI, this is a much worse solution than using CSS `text-overflow` to truncate the answer:
- Characters are different lengths, so this doesn't do a good job accomplishing the actual goal of "print as much answer as can fit in a small box."
- Using `text-overflow` lets the browser do useful things like use the full answer in reader mode or screen readers, and put a tooltip containing the full answer in most situations.

PR #320, opened 2022-05-24
Fix comment links from /activity & add to emails

IanPhilips:
I added the comment id to the market url in the comment email. The variable isn't appropriately named now though, but I'm not sure how to mess with the email template and I'm assuming it's looking for the key: 'marketurl'.

I also just defaulted to always adding the manifold.markets url to the comment link as who knows what page people will be linking from: /home, /activity, etc. On local dev you'll have to replace https://manifold.markets with localhost:3000.

### DIFF
--- a/functions/src/emails.ts
+++ b/functions/src/emails.ts
@@ -262,7 +262,7 @@ export const sendNewCommentEmail = async (
     return
 
   const { question, creatorUsername, slug } = contract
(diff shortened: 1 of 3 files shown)


jahooma (approved):
Sweet! Looks good.

Don't forget to redeploy onCreateComment cloud function

PR #324, opened 2022-05-24
Add tsconfig, more linting to `common` package

mqp:
Just gradually turning on more checks when it's easy and fixing up trivial issues.

There are lots of interesting typing problems related to "contracts that have some properties" which I don't know how to solve yet, so I am not attempting to turn on `no-explicit-any` until I am a better Typescript programmer and I understand the best solution to these problems.

### DIFF
--- a/common/.eslintrc.js
+++ b/common/.eslintrc.js
@@ -9,15 +9,18 @@ module.exports = {
     {
       files: ['**/*.ts'],
       plugins: ['@typescript-eslint'],
+      extends: ['plugin:@typescript-eslint/recommended'],
       parser: '@typescript-eslint/parser',
       parserOptions: {
         tsconfigRootDir: __dirname,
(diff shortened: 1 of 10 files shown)


mqp:
@akrolsmir Just letting you know that this is the sort of thing I am up to in my spare time, since you seemed interested in the idea of "turning on more linting stuff".

ForrestWeiswolf:
> There are lots of interesting typing problems related to "contracts that have some properties" which I don't know how to solve yet, so I am not attempting to turn on no-explicit-any until I am a better Typescript programmer and I understand the best solution to these problems.

I've been looking into a few of those, work so far is on https://github.com/ForrestWeiswolf/manifold/tree/remove-any, but I figured I should hold off on making a PR until this is merged.

jahooma (approved):
Cool.

PR #332, opened 2022-05-25
Move to tailwindui for create market page

IanPhilips:
I added the liklihoods with unsure first as that's what most people use and I like the default options being leftmost but I'm open to other opinions.

Screen Shot 2022-05-25 at 6 52 00 AM

### DIFF
--- a/web/components/choices-toggle-group.tsx
+++ b/web/components/choices-toggle-group.tsx
@@ -1,6 +1,7 @@
 import { Row } from './layout/row'
 import { RadioGroup } from '@headlessui/react'
 import clsx from 'clsx'
+import React from 'react'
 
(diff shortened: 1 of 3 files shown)


IanPhilips:
Actually I'll add a custom prob toggle now that tailwindui is working for me

IanPhilips:
@mantikoros  is it okay that I limit initial probability of binary market creation to be between 5-95%? I know we talked about it but it hasn't been implemented yet.

IanPhilips:
Screen Shot 2022-05-25 at 7 49 26 AM

mantikoros:
Very nice. I think bg-indigo-500 would look nicer for the buttons. 

Restricting the probability is also reasonable.

PR #338, opened 2022-05-25
Handle free response outcomes for yes/no shares

IanPhilips:
Now with free response contracts the quick bet view will show a highlighted up arrow if the user has shares in the top answer. If they don't own shares in the top answer but do in other answers, it'll show the down arrow. For numeric markets it will just show the up arrow if they've bet on the market anywhere. This should probably be whether their bets are majority above or below the displayed value?


akrolsmir (approved):
Cool, this make sense to me! I'd say, don't worry about Numeric markets too much because we're going to replace them pretty soon https://manifold.markets/Austin/will-manifold-change-numeric-market

PR #343, opened 2022-05-26
Add more linting to `web` package

mqp:
Just really normal stuff, importing `React` when we didn't import it, changing a lot of `let`s to `const`s, fixing up typing when we were lazy, etc. Should be no visible changes.


jahooma (approved):
Thanks!

PR #348, opened 2022-05-26
Add quick bet arrows to embedded markets

sipec:


akrolsmir (approved):
Looks great! Two questions:
1. Do we want to show arrows to signed-out users? We don't currently in the main site.
2. The cursor doesn't switch to a pointer on hover here; we should add `cursor-pointer` somewhere to make that work (wherever the total hitbox is for the quick bet, which I intentionally made pretty large)

web/pages/embed/[username]/[contractSlug].tsx
```diff
@@ -114,31 +111,16 @@ function ContractEmbed(props: { contract: Contract; bets: Bet[] }) {
...
-            />
-          )}
-
-          {outcomeType === 'NUMERIC' && (
-            <NumericResolutionOrExpectation
-              contract={contract as NumericContract}
-            />
-          )}
+            )}
+            <QuickBet contract={contract} />
```
Comment (jahooma) [inline web/pages/embed/[username]/[contractSlug].tsx:121]: Binary has the bet row and quick bet? I guess that makes sense

jahooma (commented):
Nice! One other suggestion is to include a screenshot in any PR with user-facing changes. Helps us understand the change, and gives an opportunity to scrutinize the change visually.

jahooma:
One other suggestion: make the quick bet component be to the right of both the question and contract details, since the voting arrows are quite tall.

PR #354, opened 2022-05-27
Notifications

IanPhilips:
Users get notifications:
- When they participated in any contract (i.e. created, bet, provided liqudity, or commented) and then later anyone else comments on the contract, creates a new answer, or updates the contract
- When they created a contract and someone commented on it, or submitted an answer to it
Screen Shot 2022-05-27 at 4 57 31 PM
Screen Shot 2022-05-27 at 4 56 45 PM
Screen Shot 2022-05-27 at 4 55 46 PM

Unfortunate state of affairs on skinny screens with Avatar + Userlink + notification status text:
Screen Shot 2022-05-27 at 5 16 54 PM


web/lib/firebase/notifications.ts
```diff
@@ -0,0 +1,27 @@
...
+import { db } from 'web/lib/firebase/init'
+import { getValues, listenForValues } from 'web/lib/firebase/utils'
+
+function getNotificationsCollection(userId: string, unseenOnly?: boolean) {
+  const notifsCollection = collection(db, `/users/${userId}/notifications`)
+  if (unseenOnly) return query(notifsCollection, where('isSeen', '==', false))
+  return query(notifsCollection)
+}
+
+export function getUnseenNotifications(userId: string) {
```
jahooma, line 12: Should be pretty easy to listen for unseen notifications, right? (Use listenForValues instead of getValues, then create a hook.) Seems way easier and less hacky than polling every 5 seconds.

jahooma:
This is awesome!

Just looking at the screenshots (code review forth-coming!), couple quick suggestions
- The UserLink component uses the `truncate` class, which means it collapses if competing with another flex div for horizontal space. You might try flex-0 to make it not truncate, or add flex-1 to the two children of the Row.
- The reason text is a bit wordy. Try 'answered' instead of 'submitted an answer to'.

web/lib/firebase/notifications.ts
IanPhilips: Is there an extra cost to listening? I figured people don't need them real time so toned down the urgency a bit with a poll. Happy to listen though, already doing so elsewhere

IanPhilips:
I'l probably just link the username manually rather than use Userlink as I think it'll make the most sense for all the text to be in the same div

web/lib/firebase/notifications.ts
mantikoros: Basically no. Firebase listeners are very lightweight/low-cost. There's no reason to do polling instead.

IanPhilips: thanks for the tips!

jahooma (commented):
First pass. This is awesome, Ian! Thanks for building this.

IanPhilips:
Thank you for the first pass review!

functions/src/create-notification.ts
```diff
@@ -0,0 +1,172 @@
...
+      recipientUserIds.forEach((userId) => {
+        if (shouldGetNotification(userId, userToReasonTexts))
+          userToReasonTexts[userId] = {
+            text: `${reasonTextPretext} a question you bet on`,
+            reason,
+          }
+      })
+    }
+
+    const getUsersToNotify = async () => {
```
jahooma, line 141: This is good. I was thinking of it being another top-level function with all the helpers defined in it. Up to you though!

functions/src/on-update-contract.ts
```diff
@@ -0,0 +1,38 @@
...
+  .document('contracts/{contractId}')
+  .onUpdate(async (change, context) => {
+    const contract = change.after.data() as Contract
+    const { eventId } = context
+
+    const contractUpdater = await getUser(contract.creatorId)
+    if (!contractUpdater) throw new Error('Could not find contract updater')
+
+    const previousValue = change.before.data() as Contract
+    if (previousValue.isResolved !== contract.isResolved) {
```
jahooma, line 16: Nice!

jahooma (approved):
LGTM! Wait for one of the other reviewers to take a look, I guess! (Stephen says he will.)

functions/src/create-notification.ts
```diff
@@ -0,0 +1,172 @@
...
+
+export const createNotification = async (
+  sourceId: string,
+  sourceType: notification_source_types,
+  reason: notification_reason_types,
+  sourceContract: Contract,
+  sourceUser: User,
+  idempotencyKey: string
+) => {
+  const shouldGetNotification = (
```
mantikoros, line 28: One suggestion: Move inner functions to top-level scope. I think it improves readability.

common/notification.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+export type notification_source_types =
+  | 'contract'
+  | 'comment'
+  | 'bet'
+  | 'answer'
+  | 'liquidity'
+
+export type notification_reason_types =
+  | 'created'
```
mantikoros, line 25: The idea is to use the 'created' reason type with each of the different source types? That might make it slightly more annoying to query for new comment notifications b/c, I believe, Firestore will require generating a new index.

functions/src/create-notification.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    const getUsersToNotify = async () => {
+      const userToReasonTexts: user_to_reason_texts = {}
+      // The following functions modify the userToReasonTexts object in place.
+      await notifyContractCreator(userToReasonTexts)
+      await notifyOtherAnswerersOnContract(userToReasonTexts)
+      await notifyOtherCommentersOnContract(userToReasonTexts)
+      await notifyOtherBettorsOnContract(userToReasonTexts)
+      return userToReasonTexts
```
mantikoros, line 148: I think it would be better to make all these functions immutable. Something like:
`const userToReasonTexts  = { ...await notifyContractCreator(userToReasonTexts), ...await notifyOtherAnswerersOnContract(userToReasonTexts), ... }`

web/components/notifications-icon.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  useEffect(() => {
+    if (user) return listenForNotifications(user.id, setNotifications, true)
+  }, [user])
+
+  return (
+    <Row className={clsx('justify-center')}>
+      <div className={'relative'}>
+        {notifications && notifications.length > 0 && (
+          <div className="-mt-0.75 absolute ml-3.5 min-w-[15px] rounded-full bg-indigo-500 p-[2px] text-center text-[10px] leading-3 text-white lg:-mt-1 lg:ml-2">
```
mantikoros, line 28: I think the notification circle should be red (probably bg-red-500). This is the standard on most platforms, and red is good at getting your attention.

mantikoros (changes requested):
Great job! Notifications are going to be a huge improvement.

common/notification.ts
IanPhilips: The reason types are for front end generation of the text that the user will see, not for firestore querying - i.e. 'someone commented/edited a comment/deleted a comment etc.' . The query will just use isSeen.

functions/src/create-notification.ts
IanPhilips: True, I knew that tho and am going to refactor them on the next iteration.

IanPhilips:
Thank you for the review Stephen! I'm going to use those suggestions on the next iteration.

PR #355, opened 2022-05-28
Refactor contract typing to be more concise and more correct

mqp:
Prior to this change we had a problem. Right now there are 5 actual kinds of contracts in existence:

```typescript
FullContract<CPMM, Binary>
FullContract<DPM, Binary>
FullContract<DPM, FreeResponse>
FullContract<DPM, Multi>
FullContract<DPM, Numeric>
```
(I guess `Multi` doesn't actually exist anymore, but we can deal with that later -- it's not relevant to the logic of this PR.)

However, Typescript doesn't know that. It thinks there are 8 -- the combination of each outcome type and each mechanism. in my opinion, this causes two problems:

1. If you check to see if something is DPM, you logically know whether it's numeric, and vice versa, but Typescript doesn't know that, so you have to start doing `any` and casts and stuff to get Typescript on the same page.
2. If you do that, and then someone e.g. makes CPMM numeric markets, your code will not produce a type error, but it will explode at runtime. If we typed everything correctly, then making CPMM numeric markets would produce a type error for anyone relying on the fact that numeric === DPM.

```typescript
if (ct.outcomeType === 'NUMERIC') {
  // Current version: Type error. Tempts you to use `any`.
  // Current version if we use `any` and add new numeric mechanism: Looks great then kaboom at runtime!
  // New version: Works great.
  // New version if we add new numeric mechanism: Type error. This code needs to be fixed.
  console.log(ct.totalShares)
}
```

We can fix this problem if we define
```typescript
export type AnyContractType = (CPMM & Binary) | (DPM & AnyOutcomeType)
export type Contract<T extends AnyContractType = AnyContractType> = { ...common fields } & T
```

Now you will do `Contract<CPMM & Binary>` instead of `FullContract<CPMM, Binary>`. You can also do `Contract` to get the 'any possible contract' type, or e.g. `Contract & Binary` to get the 'any possible binary contract' type, or `Contract & CPMM` to get the 'any possible CPMM contract' type (which currently implies binary.) If you do `Contract<CPMM & Multi>` the compiler will yell.

As you can see, this fixes a ton of places in the codebase where we were making unnecessary casts, typing things as `any`, etc.

(P.S. Sometimes people would do `FullContract<any, Binary>` when they meant "binary CPMM or DPMM contracts." This was wrong -- `FullContract<any, Binary>` is defined as `{ FullContract stuff } & any & Binary` which is identical to `any`. They should have been doing `FullContract<CPMM | DPMM, Binary>`.)

### DIFF
--- a/common/add-liquidity.ts
+++ b/common/add-liquidity.ts
@@ -1,12 +1,12 @@
 import { addCpmmLiquidity, getCpmmLiquidity } from './calculate-cpmm'
-import { Binary, Contract, CPMM } from './contract'
+import { CPMMContract } from './contract'
 import { LiquidityProvision } from './liquidity-provision'
 import { User } from './user'
 
 export const getNewLiquidityProvision = (
   user: User,
   amount: number,
-  contract: Contract<CPMM & Binary>,
+  contract: CPMMContract,
   newLiquidityProvisionId: string
 ) => {
   const { pool, p, totalLiquidity } = contract
--- a/common/add-liquidity.ts
+++ b/common/add-liquidity.ts
@@ -1,12 +1,12 @@
 import { addCpmmLiquidity, getCpmmLiquidity } from './calculate-cpmm'
-import { Binary, CPMM, FullContract } from './contract'
+import { Binary, Contract, CPMM } from './contract'
 import { LiquidityProvision } from './liquidity-provision'
 import { User } from './user'
 
 export const getNewLiquidityProvision = (
   user: User,
   amount: number,
-  contract: FullContract<CPMM, Binary>,
+  contract: Contract<CPMM & Binary>,
   newLiquidityProvisionId: string
 ) => {
   const { pool, p, totalLiquidity } = contract
--- a/common/antes.ts
+++ b/common/antes.ts
@@ -1,7 +1,12 @@
 import { range } from 'lodash'
 import { Bet, NumericBet } from './bet'
 import { getDpmProbability, getValueFromBucket } from './calculate-dpm'
-import { Binary, Contract, CPMM, DPM, FreeResponse, Numeric } from './contract'
+import {
+  CPMMBinaryContract,
+  DPMBinaryContract,
+  FreeResponseContract,
+  NumericContract,
+} from './contract'
 import { User } from './user'
 import { LiquidityProvision } from './liquidity-provision'
 import { noFees } from './fees'
@@ -16,7 +21,7 @@ export const HOUSE_LIQUIDITY_PROVIDER_ID = 'IPTOzEqrpkWmEzh6hwvAyY9PqFb2' // @Ma
 
 export function getCpmmInitialLiquidity(
   providerId: string,
-  contract: Contract<CPMM & Binary>,
+  contract: CPMMBinaryContract,
   anteId: string,
   amount: number
 ) {
@@ -40,7 +45,7 @@ export function getCpmmInitialLiquidity(
 
 export function getAnteBets(
   creator: User,
-  contract: Contract<DPM & Binary>,
+  contract: DPMBinaryContract,
   yesAnteId: string,
   noAnteId: string
 ) {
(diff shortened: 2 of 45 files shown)


jahooma:
Cool! I think getting the types precise so that no casting is needed as types narrow in e.g. if statements is great!

On the other hand, I don't think it's very intuitive to use '&' for intersection types as part of regular use.

I would prefer to calculate out all the usable alternatives as single exports:
```
// contract.ts
export type BinaryContract = Contract & Binary
export type CpmmBinaryContract = Contract & Binary & CPMM
...
export type DpmContract = Contract & DPM
```

What do you think?

mqp:
> On the other hand, I don't think it's very intuitive to use '&' for intersection types as part of regular use.

I think it's better to do this, just like I would prefer to write `alpha + beta` rather than define `AlphaBeta = alpha + beta` and refer to `AlphaBeta` everywhere (if you ignore any efficiency concerns.)

In general, if the extra name isn't conveying any information, and it's not more concise, I think it has negative value (no value, minus some value for concealing how something works that would otherwise be visible at a glance.)

However, it seems unimportant, so if you feel strongly about this and others agree then sure.

jahooma:
Well, really we are hiding implementation complexity because it's not obvious how Binary and CPMM types would interact. If we asked Ian to type a contract, it would be much easier for him to import one type than try to figure out how to combine multiple types with an unfamiliar operator. That takes study of the original types (or copying a pattern you might not understand).

Secondly, if we can compute the final value of something, it's better to export that (e.g. exporting HOUR_MS vs writing 1000 * 60 * 60) And, it's more ergonomic to import 1 type, and autocomplete works. So there is a savings of referring to just one symbol even if it's longer.

mqp:
> Well, really we are hiding implementation complexity because it's not obvious how Binary and CPMM types would interact. If we asked Ian to type a contract, it would be much easier for him to import one type than try to figure out how to combine multiple types with an unfamiliar operator. That takes study of the original types (or copying a pattern you might not understand).

I would say that this is fundamental knowledge that I am actively interested in transmitting through its visible use in the codebase. I like to work in an environment where I come across things that are new to me which are then instructive and rewarding for me to learn.

Anyway, I have no interest in expending any effort arguing about it further, if you like the aliases we can just put them.

IanPhilips:
Thanks for working on this Marshall! I'll voice my opinion in support of the aliases as well: they feel a lot more natural to write and understand.

akrolsmir:
FWIW my vote is to punt, because I feel like we'll get to a single generalized implementation of (2 to n) category CPMM, with limit orders/uniswapv3 implemented, that will subsume all these different contract types. And also include perp swaps/equity, haha.

mqp:
> I agree with you but part of the point is it's not obvious a priori. Contract & Binary & CPMM being a correct way to type binary contracts relies on an exact modularization of the types. I could imagine that CPMM Binary contracts would want their own field names specific to that combination, for example.

OK, that's a good point, and I agree that if in the future the contract type and mechanism data aren't so clearly independent, the aliases will make much more sense.

> FWIW my vote is to punt, because I feel like we'll get to a single generalized implementation of (2 to n) category CPMM, with limit orders/uniswapv3 implemented, that will subsume all these different contract types. And also include perp swaps/equity, haha.

If by "punt" you mean don't merge this PR then I have no idea why that is an argument for punting? Having working static types for the contracts in our codebase should make it much easier to change, add, or remove contract implementations later! Maybe I wasn't clear enough -- we basically have no typechecking on half of our code dealing with contracts right now, because it does explicit casts, or uses `any`, or uses stuff like `FullContract<any, Binary>` which is identical to `any`.

But I will also point out that having a unified uber-contract-structure that handles all kinds of contracts doesn't mean we can stop dealing with this, because our database will still be loaded with old contracts, and unless you can cleverly migrate them over, then as long as we ever want to display them or interpret them, we will need these old types.

Anyway -- it sounds like the correct action here is that I will add the aliases you guys like and then merge this.

PR #372, opened 2022-06-01
Allow selling CPMM binary position from bet table

sipec:
For CPMM, it doesn't really make sense to but a sell button on every individual buy bet, so instead I put it in next to the profit.

Individual market:
image

Your bets in profile:
image

Upon clicking, it opens a modal

image


jahooma:
This is a good change that people want, however:
- Sell button is mostly offscreen on mobile.
- Sell button still appears after you've sold all your shares. Clicking the sell button again in this case leads to an infinite loop that freezes the tab.

PR #377, opened 2022-06-01
Give removeUndefinedProps an explicit type

ForrestWeiswolf:


jahooma:
Looks like static analysis is failing in one spot? https://github.com/manifoldmarkets/manifold/runs/6681379906?check_suite_focus=true

mqp:
I am a Typescript beginner, but this seems very wrong to me.

[Here](https://www.typescriptlang.org/play?#code/C4TwDgpgBA6glsAFgewK7AKoDsAmEBmcWEOA0hCAM4A8AKgHxQC8UA3gFBRQDaAClESgBrCsnxRaAXQBcUAKIAPAMYAbVHjp9JAGiipcBIiXrsAvu3ZLkWSsCgAnCAFtkANwjY8hYjl73kYJTMUHRQEArAELhByABGAFYQSsD0ABRx8bK0AJSy8EhomAbeJORUdIxMjBxcVjZ2xADuAPIJsrwAhvbAcB0qFcGs5pxQdbbCFEEsrYnJAHQiVOkJ2VAdQQCC9vYdINSLYhImXPjI9lCpY3aLUIeLlKs1XALiqcvxa0EdWCDZ3IuSKAAQiYLH0XiMOFWTRm-wogJY70+ax+fwBI2GXEcwFQ9iwUBhCWR+RQ6E8hh8ZRoDDMFlAkCgADFkMgAMLWYA7ZKDAQ4WS2exEADmukclGQah61lkWFQTliEHOAB89MVIVBzCoIHZTshZMy2RyuXYWKxebIAEQdWJKC2iiDiyVwaWqiE+DXsLV2VQQb4kA3BRwudzkkq+fyBVK67KWayOiBzFTIIWpAAGACUHRL0M6sLIACSsH1+nAGuZi7NSrBzYDIADKnOFqWyplT2SAA) is an example that fails:

```typescript
type FooContract = { id: string, resolution: number | undefined }
let foo: FooContract = { id: "abc", resolution: undefined }

// type of cleanedFoo is { id: string, resolution: number } which is obviously wrong
let cleanedFoo = removeUndefinedProps(foo)

// blows up at runtime
console.log(`Resolution: ${cleanedFoo.resolution.toString()}`)
```

Generally speaking, I think the existing implementation of `removeUndefinedProps` is almost correct. The only part that seems wrong to me is that it's conflating e.g. `T | undefined` with `T?` -- `removeUndefinedProps` is guaranteed to actually delete any undefined properties, so it's like it's converting their type from `T | undefined` to `T?`.

It's not going to be possible to express "removing the undefined properties" in more detail in the Typescript type system, because Typescript basically doesn't know at compile time whether any given `T | undefined` property is going to be `T` or `undefined`, so it's not going to know at compile time whether any given property is going to be removed.

PR #378, opened 2022-06-01
Add `yarn verify` script

ForrestWeiswolf:
Basically the same static analysis as the github workflow, but in the form of a yarn script for more convenient local use.
If there's already an easy way to do this let me know, but I didn't see anything along these lines in the README or package.json.

### DIFF
--- a/common/package.json
+++ b/common/package.json
@@ -2,7 +2,9 @@
   "name": "common",
   "version": "1.0.0",
   "private": true,
-  "scripts": {},
+  "scripts": {
+    "verify": "(cd .. && yarn verify)"
+  },
   "sideEffects": false,
   "dependencies": {
     "lodash": "4.17.21"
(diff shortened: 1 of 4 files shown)


akrolsmir (approved):
LGTM, thanks!

I'm wondering if we want this script to live on the top level, vs inside of `web/`; mostly because most of our scripts are run from the web directory today, so I might forget that to cd out before calling `yarn verify`.

(Maybe the fix would be to add `"verify": "cd .. && yarn verify"` to web, haha)

I'll leave it up to you @ForrestWeiswolf!

mqp:
I think it would be better if the `verify` scripts in each subdirectory did all the work for that subproject, and then in the root directory it checked everything. To me that would be the ideal version. But I like this better than the status quo. Maybe someone can improve it later.

PR #379, opened 2022-06-01
Fix dependency of useEffect in BetsList

ForrestWeiswolf:
Commit a84014399 is in main and fails lint because a dependency of `useEffect` in `BetsList` changed but what actually gets passed to the `useEffect` didn't.

This fixes it.

### DIFF
--- a/web/components/bets-list.tsx
+++ b/web/components/bets-list.tsx
@@ -82,7 +82,7 @@ export function BetsList(props: { user: User; hideBetsBefore?: number }) {
         disposed = true
       }
     }
-  }, [bets])
+  }, [allBets])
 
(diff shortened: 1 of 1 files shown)


ForrestWeiswolf:
Note: the github workflow runs lint, so it appears that until this is fixed, all PRs that *don't* fix it will have a failure on that step.

jahooma (changes requested):
Uh, don't merge this one! I think using `bets` as the dependency was causing an infinite loop, and @akrolsmir changed it to `allBets`.

I think you could add the lint ignore comment instead, which will suppress the warning.

jahooma (approved):
Thanks @ForrestWeiswolf !

PR #381, opened 2022-06-01
`MULTI` contracts are not a thing

mqp:
So, I don't actually understand the history here. All I see is that in #47, we added some code suggesting that there could be contracts with the `MULTI` outcome type, distinct from the `FREE_RESPONSE` outcome type, which would have an array of strings that were like outcomes, but weren't free response answers. Maybe they were gonna be like free response markets with a fixed list of answers set by the creator? However, we never implemented any actual code for creating these or using them, and they never came to exist. So in that case let's clean up the last tiny bit of stuff that thinks they exist.

### DIFF
--- a/common/contract.ts
+++ b/common/contract.ts
@@ -2,11 +2,11 @@ import { Answer } from './answer'
 import { Fees } from './fees'
 
 export type AnyMechanism = DPM | CPMM
-export type AnyOutcomeType = Binary | Multi | FreeResponse | Numeric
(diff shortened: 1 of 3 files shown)


mqp:
@jahooma you worked on this, so perhaps you can tell me if I am off base here somehow.

jahooma (approved):
That's correct. We'll probably add Multi soon, but also happy to reformulate them then.

PR #382, opened 2022-06-01
Add charities: EA Funds and Founders Pledge funds

jackc271:
Adding some suggested charities


akrolsmir (approved):
Thanks so much, @jackc271!

Ideally, there'd be some way to visually distinguish the different funds with the same logo; maybe we edit the image in photoshop, or overlay the fund name in CSS for funds with duplicate photos...

PR #383, opened 2022-06-01
Notifications ux fixes - wip

IanPhilips:
Screen Shot 2022-06-01 at 10 32 28 AM
Screen Shot 2022-06-01 at 10 34 37 AM


jahooma:
Might want to preserve whitespace in notifications. Could use 'whitespace-pre-line' class. Or use <Linkify> component. Not actually sure if links should be clickable...

Another idea would be to not show whitespace, but actually truncate each message after x number of characters.

PR #389, opened 2022-06-01
Remove more uses of `any`

ForrestWeiswolf:


mqp:
Looks highly reasonable, I will merge soon. By the way, I don't really know what I think of choosing between `JSX.Element` and `ReactElement` and `ReactNode` but clearly they all beat `any`.

ForrestWeiswolf:
> Looks highly reasonable, I will merge soon. By the way, I don't really know what I think of choosing between `JSX.Element` and `ReactElement` and `ReactNode` but clearly they all beat `any`.

Using `JSX.Element` rather than `ReactNode` here is deliberate - it doesn't allow `null` or strings or arrays, and these looked like contexts where an array of elements wouldn't make sense. As far as I recall `ReactElement` and `JSX.Element` should be basically equivalent in this context, though.

PR #390, opened 2022-06-01
Destroy the blue notification bubble

mqp:
The point of notifications is to get ones you want to see and not get ones you don't want to see. Drawing people's attention to notifications they don't want to see is a way to actively piss off users. If this were on a different website that I am not working on I would have instantly made a uBlock rule hiding the "notifications" link in the nav forever.

This feature will be fine once there is a way to cause the notification bubble to only appear for things you want to see.

### DIFF
--- a/web/components/notifications-icon.tsx
+++ b/web/components/notifications-icon.tsx
@@ -21,14 +21,19 @@ export default function NotificationsIcon(props: { className?: string }) {
     if (user) return listenForNotifications(user.id, setNotifications, true)
   }, [user])
 
+  // mqp - kill this until you can control what notifications you get, or
+  // until the defaults aren't so spammy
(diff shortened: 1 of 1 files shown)


jahooma:
Thanks for your feedback! It's good to know that notifications can become annoying.

However, it is common practice, and definitely helps us surface interesting content. It would be better to adjust notifications to not include things you don't want to see.

Let's punt on this for a few days to gauge other users' reactions. You also might get used to it!

sipec:
at least it's not red

IanPhilips:
@mqp are you satisfied with the notification controls enough to close this PR? You can select the 'None' toggle in the notifications settings to stop seeing the blue bubble

mqp:
Nice! I am satisfied.

PR #401, opened 2022-06-02
Refactoring

TrueMilli:
Minor changes which don't affect functionality split from the feature to automatically resolve markets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 .vercel
 node_modules
 yarn-error.log
-.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .vercel
 node_modules
 yarn-error.log
+.vscode
--- a/functions/src/create-contract.ts
+++ b/functions/src/create-contract.ts
@@ -56,11 +56,11 @@ export const createContract = newEndpoint(['POST'], async (req, [user, _]) => {
 
   let min, max, initialProb
   if (outcomeType === 'NUMERIC') {
-    ;({ min, max } = validate(numericSchema, req.body))  // leading ; intentional: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#assignment_separate_from_declaration_2
+    ;({ min, max } = validate(numericSchema, req.body))
     if (max - min <= 0.01) throw new APIError(400, 'Invalid range.')
   }
   if (outcomeType === 'BINARY') {
-    ;({ initialProb } = validate(binarySchema, req.body))  // leading ; intentional: see above
+    ;({ initialProb } = validate(binarySchema, req.body))
   }
 
   // Uses utc time on server:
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@
 .vercel
 node_modules
 yarn-error.log
+.vscode
```
jahooma, line 6: I think we have workspace settings here that could be useful to others? I wouldn't ignore it

functions/src/create-contract.ts
```diff
@@ -56,11 +56,11 @@ export const createContract = newEndpoint(['POST'], async (req, [user, _]) => {
 
   let min, max, initialProb
   if (outcomeType === 'NUMERIC') {
-    ;({ min, max } = validate(numericSchema, req.body))
+    ;({ min, max } = validate(numericSchema, req.body))  // leading ; intentional: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#assignment_separate_from_declaration_2
     if (max - min <= 0.01) throw new APIError(400, 'Invalid range.')
   }
   if (outcomeType === 'BINARY') {
-    ;({ initialProb } = validate(binarySchema, req.body))
+    ;({ initialProb } = validate(binarySchema, req.body))  // leading ; intentional: see abive
```
jahooma, line 63: Typo: in abive => above

jahooma (approved):
Thanks @TrueMilli!

functions/src/create-contract.ts
TrueMilli, line 63:
```suggestion
    ;({ initialProb } = validate(binarySchema, req.body))  // leading ; intentional: see above
```

functions/src/create-contract.ts
```diff
@@ -56,11 +56,11 @@ export const createContract = newEndpoint(['POST'], async (req, [user, _]) => {
 
   let min, max, initialProb
   if (outcomeType === 'NUMERIC') {
-    ;({ min, max } = validate(numericSchema, req.body))
+    ;({ min, max } = validate(numericSchema, req.body))  // leading ; intentional: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#assignment_separate_from_declaration_2
```
sipec, line 59: I kinda dislike these comments.

For one, It's not strictly true. The semicolon is not actually needed here as it's the first line in a block. Prettier adds the semicolon anyways because it adds one in front of every leading `(` and `[`. [(See reasoning here)](https://github.com/prettier/prettier/issues/736#issuecomment-291934981) 

Since we use Prettier, we don't actually need to document weird syntactic style. The autoformatter or the CI linter will catch it if someone tries to change it.

I just don't think these comments actually make the code more understandable.

If this line were really hard to read, it would make more sense to just rewrite it than to leave a comment on why it's written this way.

sipec (changes requested):
this mostly looks good

.gitignore
TrueMilli: I thought it was default to exclude it, but after reading there seem to be good arguments for it.
The debugging settings are in there, so if anyone gets it to work with VS Code it's shared directly.

functions/src/create-contract.ts
jahooma: +1

TrueMilli: Thanks for the detailed explanation, I realized I had only half understood the reasoning.

sipec (approved):

PR #402, opened 2022-06-02
Contract card ui tweaks: consistent market type colors, no underline,…

mantikoros:
… adjust font/border size


mantikoros:
Before:
Screen Shot 2022-06-02 at 6 17 11 PM

After:
Screen Shot 2022-06-02 at 6 17 24 PM

akrolsmir (changes requested):
Thanks for sending the PR! But I basically disagree with all the VisD changes here lol. In order from strongest to weakest opinions:

- Removing underline gives you less feedback about what is actually going to happen when you click on the card
- Decreasing bar width makes the bar look kinda like a bug
- Semibold is great for making the text more emphasized
- Red + Green is more eyecatching on our site - which is what we want. If you don't like red, we should switch the whole color scheme over to something like Kalshi's blue and green instead.

I guess I would ask Alice for her take, as the person I trust most with VisD on the whole team

akrolsmir:
(also the net effect is to make the card view less dense, which I'm not sure is the correct direction)

mattprice516:
Agree, always green bar for % chance of YES (regardless of if that % chance is higher or lower than 50%) is clearer communication at a glance. The other changes don't seem to make much a positive difference IMO, slightly negative if anything.

PR #403, opened 2022-06-02
Notification detail, grouping, and settings control [wip]

IanPhilips:
Give users control over their email notifications in concert with their in-app notifications. Provide more details on why they're seeing a notification. Group notifications on the same question.

## More detailed notifications
The new notifications tab shows the more detailed notification types

 ```
 | 'tagged_user'
  | 'on_users_contract'
  | 'on_contract_with_users_shares_in'
  | 'on_contract_with_users_shares_out'
  | 'on_contract_with_users_answer'
  | 'on_contract_with_users_comment'
  | 'reply_to_users_answer'
  | 'reply_to_users_comment'
```
Screen Shot 2022-06-02 at 5 28 26 PM

## In-app notifications settings:
For now the less toggle would only send notifications if:
- a market you interacted with is updated/closed/resolved by the creator, 
- replies to your comments/answers, 
- comments with your tag,
- comments on your markets
- ? comments on markets you have open bets in ?

The all toggle would include
- the notifications from less, 
- all comments on markets you interacted with
- new answers submitted to markets you interacted with 

## Email notifications settings:
```
if (newValue === 'all') {
      updatePrivateUser(privateUser.id, {
        unsubscribedFromResolutionEmails: false,
        unsubscribedFromCommentEmails: false,
        unsubscribedFromAnswerEmails: false,
      })
    } else if (newValue === 'less') {
      updatePrivateUser(privateUser.id, {
        unsubscribedFromResolutionEmails: false,
        unsubscribedFromCommentEmails: true,
        unsubscribedFromAnswerEmails: true,
      })
    } else if (newValue === 'none') {
      updatePrivateUser(privateUser.id, {
        unsubscribedFromCommentEmails: true,
        unsubscribedFromAnswerEmails: true,
        unsubscribedFromResolutionEmails: true,
      })
    }
```

image

## Grouping
Grouping activity on questions as well to not flood the new notifications tab:

Screen Shot 2022-06-02 at 5 26 54 PM

### CODE REVIEW

**Comment by IanPhilips** [comment]
I'm going to go over the code again tomorrow so don't look too closely at that!

**Comment by akrolsmir** [comment]
Great to see more control over your notifs! Quick thoughts:

- Could we just do grouping the way we do on the User Profile Comments page, where show the text of the comment (maybe truncated), not just the count?
- Let's highlight the user's own name when they're tagged. (Actually, we should do this globally across our entire site)
- Entries need a bit more padding on the left side between the white rectangle and the rest of the card.
- Timestamp is vertically misaligned from from the title
- When "One of the buttons" are selected, let's show which notifications are sent and which are not: 
image
- You can also add more horizontal padding to the buttons to make them look nicer

**Comment by IanPhilips** [comment]
I'm not going to list all the notifications like on the user profile as that will get too big but here's an idea. Each new line is a summary of what happened and I can increase the number of lines shown if you'd prefer.

Screen Shot 2022-06-03 at 8 10 43 AM

File: web/pages/notifications.tsx [hunk 1]
```diff
@@ -47,18 +112,51 @@ export default function Notifications() {
           className={'pb-2 pt-1 '}
           defaultIndex={0}
           tabs={[
+            {
+              title: 'New Notifications',
```
jahooma, line 116: Not a big fan of a separate tab for new vs. all notifications. Bad UX to have to click between them, plus New notifications are very ephemeral since they disappear once seen.

Isn't it standard to have like a colored dot on unseen notifications?

web/pages/notifications.tsx
```diff
@@ -69,60 +167,326 @@ export default function Notifications() {
...
+  })
+  return notifications
+}
+
+function GroupNotifications(
+  notifications: Notification[],
+  hideNotificationIds: string[]
+) {
+  // Because hidden notifications won't be rendered, set them to seen here
+  setNotificationsAsSeen(
```
jahooma, line 190: Weird to have effects like this within a function to group notifications...
Could you move this out?

web/pages/notifications.tsx
```diff
@@ -69,60 +167,326 @@ export default function Notifications() {
...
+    // Group notifications by contract:
+    const groupedNotificationsByContractId = groupBy(
+      notificationGroupsByDay[day],
+      (notification) => {
+        return notification.sourceContractId
+      }
+    )
+
+    // Create a notification group for each contract within each day
+    Object.keys(groupedNotificationsByContractId).forEach((contractId, i) => {
```
jahooma, line 212: Couldn't this be a map that returns the list of NotificationGroups? Prefer a declarative style over an imperative styles with mutation

web/pages/notifications.tsx
```diff
@@ -69,60 +167,326 @@ export default function Notifications() {
...
+        setChoice={(choice) =>
+          changeEmailNotifications(choice as notification_subscribe_types)
+        }
+        className={'col-span-4 w-24'}
+      />
+    </div>
+  )
+}
+
+function GetNotificationSummaryText(
```
jahooma, line 429: Lowercase functions

web/pages/notifications.tsx
```diff
@@ -69,60 +167,326 @@ export default function Notifications() {
...
+    }
+    if (questionUpdates > 0)
+      tempSummaryLines.push(
+        `${questionUpdates} new question update${
+          questionUpdates > 1 ? 's' : ''
+        }`
+      )
+    setActivitySummaryLines(tempSummaryLines)
+  }
+
```
jahooma, line 269: Suspicious that this isn't better solved by embedding the whole text of the notification + css to truncate it to one line

web/pages/notifications.tsx
```diff
@@ -203,19 +564,99 @@ function Notification(props: {
...
+}
+const less_priority_reasons = [
+  'on_contract_with_users_comment',
+  'on_contract_with_users_answer',
+  'on_contract_with_users_shares_out',
+  // Not sure if users will want to see these w/ less:
+  // 'on_contract_with_users_shares_in',
+]
+
+export function GetAppropriateNotifications(
```
jahooma, line 646: Lowercase functions

web/pages/notifications.tsx
```diff
@@ -203,19 +564,99 @@ function Notification(props: {
...
+        reasonText = `answered`
+      else reasonText = `answered`
+      break
     default:
-      return ''
+      reasonText = ''
   }
+  return `${reasonText} ${contract?.question}`
+}
+const less_priority_reasons = [
```
jahooma, line 638: lessPriorityReasons

web/pages/notifications.tsx
```diff
@@ -69,60 +167,326 @@ export default function Notifications() {
...
+          <div className={'line-clamp-4 mt-1 gap-1 whitespace-pre-line'}>
+            {activitySummaryLines.map((line, i) => {
+              return (
+                <div key={line} className={'line-clamp-1'}>
+                  <Linkify text={line} />
+                </div>
+              )
+            })}
+            <div>
+              {notifications.length - numSummaryLines > 0
```
jahooma, line 313: No way to expand the group of notifications?

jahooma (commented):
This is awesome, another big step forward for notifications!

A few overall comments:
- Not thrilled about a separate tab for new notifications, would like to see new notifications in the same list, but highlighted with a colored dot.
- Grouped notifications is a good idea, but I wish you could easily click to expand them to see all. Another idea would be to show all the grouped notifications, but in a dense format like one line each, including the user's name and their comment
- Agree with Austin's point that "All" "Less" and "None" should give more explanation of what kinds of notifications you are getting.

web/pages/notifications.tsx
IanPhilips: I was just thinking about doing this esp after today's standup!

functions/src/on-create-comment.ts
```diff
@@ -71,6 +62,24 @@ export const onCreateComment = functions.firestore
...
+      commentCreator,
+      eventId,
+      comment.replyToCommentId
+        ? 'comment'
+        : comment.answerOutcome
+        ? 'answer'
+        : undefined,
+      comment.replyToCommentId
+        ? comments.find((c) => c.id === comment.replyToCommentId)?.userId
+        : answer?.userId,
```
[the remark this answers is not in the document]
IanPhilips: sounds good

common/notification.ts
```diff
@@ -21,9 +22,19 @@ export type notification_source_types =
...
+
+export type notification_reason_types =
+  | 'tagged_user'
+  | 'on_users_contract'
+  | 'on_contract_with_users_shares_in'
+  | 'on_contract_with_users_shares_out'
+  | 'on_contract_with_users_answer'
+  | 'on_contract_with_users_comment'
+  | 'reply_to_users_answer'
+  | 'reply_to_users_comment'
```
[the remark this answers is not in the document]
IanPhilips: I think it makes it easier to read and understand, the 'on' makes sense with a comment 'on' a users contract, but doesn't with a comment that tags a user. I suppose replies could be comments 'on' your comment though

web/components/notifications-icon.tsx
```diff
@@ -6,28 +6,47 @@ import { Notification } from 'common/notification'
...
+          setNotifications(
+            GetAppropriateNotifications(
+              notifs,
+              privateUser.notificationPreferences
+            )
+          ),
+        true
+      )
+    }
+  }, [privateUser])
```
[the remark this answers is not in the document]
IanPhilips: good idea!

web/pages/notifications.tsx
IanPhilips: good thinking!

web/pages/notifications.tsx
IanPhilips: I couldn't get the dictionary to map so I went with iterating over the keys but now I see you have to map(Dictionary)!

web/pages/notifications.tsx
IanPhilips: I did it this way at a user's request for inbox status of 0 but I can see both sides.

akrolsmir:
+1 for quick expand, that's a UI pattern we can lean more into imo

web/pages/notifications.tsx
IanPhilips: hoooo wee I should've done this to begin with!

web/pages/notifications.tsx
IanPhilips: will do, thanks!

IanPhilips:
Made most of the changes requested except adding what all/less/none mean for notifications. Will add those on Monday & possibly merge the notifications into one screen. Also added a notification for following users.

Screen Shot 2022-06-03 at 5 18 58 PM

After click to expand:

Screen Shot 2022-06-03 at 5 19 17 PM

IanPhilips:
Perhaps I should add a carrot on the right to expand/retract? Right now you can just click anywhere in the grouped notification to expand it. If you click a notification within the group once it's expanded, however, it'll take you to that notification. You can collapse it if you tap on the top part of the group with the title.

jahooma:
Some feedback after trying out the change.
- Resolved notifications should say resolved again: "Resolved NO" instead of just "NO"
- New notifications tab should have a placeholder when empty. E.g. "You have no new notifications!" You could even consider defaulting to the All notifications tab if you have no notifications.
- Re: grouped items. You might add the gray vertical line to the left of all the group items (like in comment replies) to emphasize there are multiple related notifications. On desktop you could add a hover effect that darkens the background of the whole group, which could be a clue for what clicking would do.
- I would show more than two notifications in groups by default, maybe show like 5? They are presented in a compact way which is nice to skim.
- The text of group item notifications seems unnecessarily truncated on desktop. So do questions for other notifications.

jahooma:
Also, this change is really good! Notifications are becoming a lot easier to parse, and controlling the settings is essential.

web/pages/notifications.tsx
```diff
@@ -69,60 +165,310 @@ export default function Notifications() {
...
+        setChoice={(choice) =>
+          changeEmailNotifications(choice as notification_subscribe_types)
+        }
+        className={'col-span-4 w-24'}
+      />
+    </div>
+  )
+}
+
+async function GetNotificationSummaryText(
```
jahooma, line 402: nit: lowercase function names if they are not componets

jahooma (approved):
LGTM!

I think the biggest thing that feels weird is clicking notifications and seeing the empty page of new notifications. I'd fix that with a placeholder before merging.

Otherwise, this is on the right track and I'm excited to see how notifications evolve!

One more suggestion for the future would be to include all the relevant data within a notification so no further fetching is needed. E.g. include the comment text and include the contract's question in the notification document. Should simplify the code and improve the UX.

web/pages/notifications.tsx
IanPhilips: thanks!

functions/src/on-follow-user.ts
```diff
@@ -0,0 +1,28 @@
+import * as functions from 'firebase-functions'
+import { getUser } from './utils'
+import { createNotification } from './create-notification'
+
+export const onFollowUser = functions.firestore
```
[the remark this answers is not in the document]
IanPhilips: now have to make a notification when a follower creates a market eh?

PR #404, opened 2022-06-02
Automated market resolution

TrueMilli:
### This feature will brings automatic resolutions to abandoned markets.

Automatic resolution means that unresolved markets get resolved as 'MKT' (paying out investors according to the current odds) at a specified date. By default that date is 1 week after the market closes but it can be modified by the market creator.

## Features
* Market info panel: The auto resolution time (editable) and the auto resolution 'MKT' is displayed
* Editing close date: The auto resolution time is set to close date + 1 week if it would be less otherwise 
* Contract: The auto resolution time is added
* Firestore rules: The market creator is allowed to change the auto resolution time 
* Market creation: The auto resolution time is set as close date + 1 week
* LiteMarket: Added auto resolution time and auto resolution
* Firebase: Every hour is checked whether markets should be auto resolved
* Firebase: Existing contracts get an auto resolution time (current date + 2 weeks for closed contracts, close date + 1 week otherwise)

## Already done on prod
* Update firebase rules (add autoResolutionTime to update)
* Add a close time manually to old contracts lacking it
* Run add-auto-resolution.ts (/.js) on db

## ToDo prod db
* Deploy updated version of createmarket to db
* Deploy updated version of resolveMarket to db (refactoring)
* Deploy autoResolveMarkets to db
* Reset auto resolution times (initial setup was on 2022-06-10)

## Possible features for later PRs
* eMails 1 day before auto resolve
* Notification that markets auto close now for creators with closed unresolved markets
* Change the "Market Info" interface to look nicer and allow edits of the close date
* Changeable auto resolution type (if requested)
* Removable auto resolution time
* Make close time required

## Untested
* Numeric markets are not tested because they are disabled.


TrueMilli:
Re static analysis: 
I don't understand the prettier complaint, I don't see components/checkbox.tsx and hooks/use-event.ts in the changed files or even the repository.
Can somebody help me out?

functions/src/scripts/add-auto-resolution.ts
```diff
@@ -0,0 +1,58 @@
+// Existing contracts don't have auto resolutions (Time + resolution). Let's add them.
```
sipec, line 1: I take mqp's position that there are particular markets that were intended by their creators to be resolved years into the future, which we'll need to manually fix. I think this is great for the majority of markets though!

sipec (commented):
This looks pretty good!

common/contract.ts
```diff
@@ -31,9 +31,10 @@ export type Contract<T extends AnyContractType = AnyContractType> = {
   closeTime?: number // When no more trading is allowed
 
   isResolved: boolean
-  resolutionTime?: number // When the contract creator resolved the market
+  resolutionTime?: number // When the market is resolved
   resolution?: string
-
+  autoResolution: resolution
+  autoResolutionTime: number // When the market will be resolved automatically
```
mqp, line 37: This seems like a lie because there will be many contracts for which these fields are not set.

functions/src/resolve-market.ts
```diff
@@ -31,140 +35,207 @@ export const resolveMarket = functions
...
+export const autoResolveMarkets = functions.pubsub
+  .schedule('every 1 hours')
+  .onRun(async () => {
+    const contracts = await getValues<Contract>(
+      firestore.collection('contracts')
+    )
+
+    const contractsToResolve = contracts
+      .filter((c) => !c.isResolved)
+      .filter((c) => c.autoResolutionTime < Date.now())
```
mqp, line 59: Why not query for this? Filtering them in memory seems super gross.

functions/src/resolve-market.ts
```diff
@@ -31,140 +35,207 @@ export const resolveMarket = functions
...
-          outcome !== 'CANCEL'
+      if (contract.creatorId !== userId)
+        return { status: 'error', message: 'User not creator of contract' }
+
+      return privateResolveMarket(contract, data, contractDoc)
+    }
+  )
+
+export const autoResolveMarkets = functions.pubsub
+  .schedule('every 1 hours')
```
mqp, line 51: I would be more inclined to run this, say, every minute.

functions/src/resolve-market.ts
```diff
@@ -31,140 +35,207 @@ export const resolveMarket = functions
...
+  contract: Contract,
+  data: {
+    outcome: resolution
+    value?: number
+    probabilityInt?: number
+    resolutions?: { [outcome: string]: number }
+  },
+  contractDoc: admin.firestore.DocumentReference<admin.firestore.DocumentData>,
+  autoResolution = false
+) => {
```
mqp, line 120: This whole file kind of seems like a mess to me, but I am not going to say you have to fix it to merge this. It would just be nice if it were nicer.

functions/src/scripts/add-auto-resolution.ts
```diff
@@ -0,0 +1,58 @@
...
+async function addAutoResolutionToContracts() {
+  console.log('Adding auto resolution to existing contracts')
+
+  const contracts = await getValues<Contract>(firestore.collection('contracts'))
+
+  console.log('Loaded', contracts.length, 'contracts')
+
+  for (const contract of contracts.filter((c) => !c.isResolved)) {
+    addAutoResolutionToContract(contract)
+  }
```
mqp, line 23: I think you meant to do something like a `map` and `Promise.all` here.

functions/src/scripts/add-auto-resolution.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+async function addAutoResolutionToContract(contract: Contract) {
+  const contractRef = firestore.doc(`folds/${contract.id}`)
```
mqp, line 27: Clearly you meant `contracts`.

web/components/contract/contract-info-dialog.tsx
```diff
@@ -147,3 +176,69 @@ const getTweetText = (contract: Contract, isCreator: boolean) => {
...
+
+export function EditableResolutionTime(props: {
+  time: number
+  contract: Contract
+  isCreator: boolean
+}) {
+  const { time, contract, isCreator } = props
+
+  const [isEditing, setIsEditing] = useState(false)
+  const [timeString, setTimeString] = useState(time && formatTime(time))
```
mqp, line 188: I don't really think that the UI where you click the pencil and it becomes a writable input has any advantages at all over the UI where it's just an input in the first place, and it has disadvantages (like complexity, probable inaccessibility, and you have to click the pencil.)

mqp (approved):
Cool, makes sense to me. I didn't squint at the resolution logic very hard but I assume it basically just moved around rather than changed, so it should be OK.

functions/src/scripts/add-auto-resolution.ts
TrueMilli: I'll implement an eMail notification 1 day before the auto resolution to warn creators to change the auto resolution date in case they don't want their markets resolved yet in another PR.

We can think of other notifications for the creators as well since the feature is new.

functions/src/resolve-market.ts
TrueMilli: Fixed

functions/src/resolve-market.ts
TrueMilli: I thought about it but it should not be important when exactly the auto resolution happens - no one can trade anymore and +1h doesn't really make a difference after a week.

Is there anything I'm missing?

functions/src/scripts/add-auto-resolution.ts
```diff
@@ -0,0 +1,58 @@
...
+const firestore = admin.firestore()
+
+async function addAutoResolutionToContracts() {
+  console.log('Adding auto resolution to existing contracts')
+
+  const contracts = await getValues<Contract>(firestore.collection('contracts'))
+
+  console.log('Loaded', contracts.length, 'contracts')
+
+  for (const contract of contracts.filter((c) => !c.isResolved)) {
```
TrueMilli, line 21: Query for non resolved contracts above instead of this.

functions/src/resolve-market.ts
```diff
@@ -31,140 +35,207 @@ export const resolveMarket = functions
...
+const privateResolveMarket = async (
+  contract: Contract,
+  data: {
+    outcome: resolution
+    value?: number
+    probabilityInt?: number
+    resolutions?: { [outcome: string]: number }
+  },
+  contractDoc: admin.firestore.DocumentReference<admin.firestore.DocumentData>,
+  autoResolution = false
```
jahooma, line 118: There's no need to pass the contractDoc, this function can simply generate that reference.

jahooma, line 118: I'd prefer not to have this `autoResolution` param. If we can avoid having it, it makes the code simpler.

Instead, the `autoResolve` function should do any special tasks, e.g. updating the description to say it was auto resolved.

common/contract.ts
```diff
@@ -31,9 +31,10 @@ export type Contract<T extends AnyContractType = AnyContractType> = {
   closeTime?: number // When no more trading is allowed
 
   isResolved: boolean
-  resolutionTime?: number // When the contract creator resolved the market
+  resolutionTime?: number // When the market is resolved
   resolution?: string
-
+  autoResolution: resolution
```
jahooma, line 36: I don't see why we want this field if auto resolve  always uses one resolution.

If there's another usecase for it, we should add it at that time. I'm also skeptical you would want any specific resolution telegraphed in advanced, other than `N/A`. If we had some api that was auto-resolving it, that should be variable, not fixed in advanced.

jahooma (changes requested):
This is very cool, and a feature we may need.

I have a few concerns before this is merged.

- I'm not sure that the auto-resolution is adequately communicated to the market creator. They're not going to check the market info dialog. We probably want a label in the market creation page near the close date input that informs them the market will auto-resolve 1 week later by default. Secondly, we need an email and in-app notification that warns market creators that their contract is about to auto-resolve, probably generated 24 hours before.
- I'm very confident that auto-resolving to `MKT` is a bad idea. Leads to strong incentives for whales to camp out on these markets and bet them all the way to the opposite resolution, taking money from everyone who bet wisely. I would not be comfortable merging this unless we change it to `N/A` as the default, or you can convince me otherwise.
- I'm skeptical of other kinds of "auto" resolutions (wouldn't they just be normal resolutions but performed via api?). It doesn't make sense in general that the auto resolution is known in advance, except for `N/A`
- I'm not sure having both a close date and auto-resolution date is the right approach. I now think that we don't want markets to ever close. In that case, we might want one `resolvesInvalidTime` which is the time at which the market resolves `N/A` if the creator hasn't resolved by. In any case, it would be better to have more confidence in this approach before merging.

jahooma:
Hi @TrueMilli,

Thanks for creating this PR! Good news —we've discussed it a bit more in our standup, and we're on board with the overall change.

- Auto-resolving to MKT is ok for now
- When the change goes in, ping me and I'll update the email that gets sent when one of your markets closes, so that it warns you that the market is set to auto-resolve.
- FYI: The current plan is to eventually deprecate `closeTime` and just have the auto-resolution time, so markets never close.
- Specific comments raised in code review still need to be addressed fully before merge. We also want the change to be thoroughly tested since you are new to the codebase.

Cheers,
James

web/components/contract/contract-info-dialog.tsx
TrueMilli: I agree in general, but I don't like the way the datepicker doesn't fit in with the other fields:
image

It's probably good to think about it some more outside of this feature. E.g. there is no reason for me the close date shouldn't be editable there as well.

common/contract.ts
TrueMilli: You are right. I was worried I would need to re-add it if I removed it earlier, but it's worse if we have bloat in our codebase and db. I'll remove it tomorrow.

functions/src/scripts/add-auto-resolution.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const contracts = await getValues<Contract>(
+    firestore.collection('contracts').where('isResolved', '==', false)
+  )
+
+  console.log('Loaded', contracts.length, 'contracts')
+
+  await batchedWaitAll(
+    contracts.map((c) => async () => {
+      addAutoResolutionToContract(c)
```
jahooma, line 26: Need an await here. Or you could return. Best is probably `contracts.map((c) => () => addAutoResolutionToContract(c))` which returns for you.

web/lib/firebase/contracts.ts
```diff
@@ -42,17 +42,19 @@ export function contractUrl(contract: Contract) {
 }
 
 export function contractMetrics(contract: Contract) {
-  const { createdTime, resolutionTime, isResolved } = contract
+  const { createdTime, resolutionTime, isResolved, autoResolutionTime } =
+    contract
 
   const createdDate = dayjs(createdTime).format('MMM D')
+  const autoResolutionDate = dayjs(autoResolutionTime).format('MMM D')
```
jahooma, line 50: Ideally, this would display the year if the year is not the same as our current year.

I'd either add that logic, or remove, since nothing is using the autoResolutionDate at the moment.

TrueMilli:
> This looks pretty good — thanks for the changes you've made so far!
> 
> I left a few more comments. I think it's about ready for the next phase, where you run the script and/or and manually add a close date to contracts missing those.
> 
> Seems like it would make sense to manually add all the close dates first? Then you'd only need to run the script once.
> 
> What's the process for adding those close dates? I'd suggest
> 
> * a script to download those questions into a csv
> * use google sheets to manually set close dates for each in a column (have one of us double check the close dates)
> * download the updated csv
> * run another script to set those close dates

* All open markets on the dev db have close dates now and all unresolved markets have auto resolve dates now
* I've asked in #dev-and-api for contracts.json generated by get-json-dump.js to define close dates for the contracts lacking them

As soon as I get the file I'll define close dates and create a google doc with the suggestions. When they are double checked I create a script to set them.

TrueMilli:
> > This looks pretty good — thanks for the changes you've made so far!
> > I left a few more comments. I think it's about ready for the next phase, where you run the script and/or and manually add a close date to contracts missing those.
> > Seems like it would make sense to manually add all the close dates first? Then you'd only need to run the script once.
> > What's the process for adding those close dates? I'd suggest
> > 
> > * a script to download those questions into a csv
> > * use google sheets to manually set close dates for each in a column (have one of us double check the close dates)
> > * download the updated csv
> > * run another script to set those close dates
> 
> * All open markets on the dev db have close dates now and all unresolved markets have auto resolve dates now
> * I've asked in #dev-and-api for contracts.json generated by get-json-dump.js to define close dates for the contracts lacking them
> 
> As soon as I get the file I'll define close dates and create a google doc with the suggestions. When they are double checked I create a script to set them.

Scripts created and pushed to the PR. Direct links are in the discord as well.

jahooma:
Cool! I'll run the script later today

jahooma:
Ok, I've run the scripts!
- All markets now have a closeTime
- (I also modified the script to set a closeTime for markets that were already resolved to the resolutionTime)
- Every unresolved market now has an autoResolutionTime

A couple thoughts
- We should make the `closeTime` field non-optional
- But, we should make the `autoResolutionTime` field optional, because I did not fill it in for markets that were already resolved. It also makes sense that we might want to support a use case of not having an auto resolution time. This will require some changes to the PR.

One the whole, this is a nice cleanup for our contracts!

web/components/contract/contract-info-dialog.tsx
jahooma: Agree their's room for improvement here, in terms of how you edit the dates and also allowing the market close to be edited here, which is fairly important.
I'm ok for that to be a followup PR.

functions/src/resolve-market.ts
```diff
@@ -31,140 +40,199 @@ export const resolveMarket = functions
...
+
+  if (value && !isFinite(value))
+    return { status: 'error', message: 'Invalid value' }
+
+  if (
+    outcomeType === 'BINARY' &&
+    probabilityInt &&
+    (probabilityInt < 0 || probabilityInt > 100 || !isFinite(probabilityInt))
+  )
+    return { status: 'error', message: 'Invalid probability' }
```
jahooma, line 144: This is a little suspicious. Why isn't BINARY a case in the switch statement?

I think you've lost this check from before:
```
if (outcomeType === 'BINARY') {
        if (!RESOLUTIONS.includes(outcome))
          return { status: 'error', message: 'Invalid outcome' }
}
```
I would be very careful about refactoring code like this, especially if you were not the one that wrote it initially.

jahooma (commented):
A few comments from another pass on the code!

functions/src/resolve-market.ts
TrueMilli: I assumed `!RESOLUTIONS.includes(outcome)` was always false.

Outcome is of type resolution:
 ```
data: {
    outcome: resolution
    value?: number
    contractId: string
    probabilityInt?: number
    resolutions?: { [outcome: string]: number }
}
```

with `resolution` and `RESOLUTIONS` being defined as.

```
export type resolution = 'YES' | 'NO' | 'MKT' | 'CANCEL'
export const RESOLUTIONS = ['YES', 'NO', 'MKT', 'CANCEL'] as const
```

File: common/contract.ts [hunk 15]
```diff
@@ -31,9 +31,9 @@ export type Contract<T extends AnyContractType = AnyContractType> = {
   closeTime?: number // When no more trading is allowed
 
   isResolved: boolean
-  resolutionTime?: number // When the contract creator resolved the market
+  resolutionTime?: number // When the market is resolved
   resolution?: string
-
+  autoResolutionTime: number // When the market will be resolved automatically
```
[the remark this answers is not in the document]
TrueMilli: The field being optional is definitely the right decision even if we are not able to remove the auto resolution date on the website yet. I didn't know how to make the date optional without adding another button.

Removing the date is nicer than having to set it to the year 9999.

functions/src/resolve-market.ts
jahooma: That's true — you're normally right, but this needs to be robust to rouge client calls. They might not abide by the required type, so we have to check all the parameters we have been passed to make sure they are correct.

common/contract.ts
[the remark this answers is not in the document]
jahooma: I see, yes. So you need to update the type here to add the '?' and make it optional.

jahooma (changes requested):
Great. I'll approve and merge this once the comments are all addressed. Thanks!

functions/src/resolve-market.ts
```diff
@@ -31,140 +41,204 @@ export const resolveMarket = functions
...
+    return { status: 'error', message: 'Contract already resolved' }
 
-      const creator = await getUser(creatorId)
-      if (!creator) return { status: 'error', message: 'Creator not found' }
+  const creator = await getUser(creatorId)
+  if (!creator) return { status: 'error', message: 'Creator not found' }
 
-      const resolutionProbability =
-        probabilityInt !== undefined ? probabilityInt / 100 : undefined
+  const resolutionProbability = probabilityInt
```
jahooma, line 158: This is incorrect if `probabilityInt` is 0. Switch back to checking `probabilityInt !== undefined`

web/components/contract/contract-info-dialog.tsx
```diff
@@ -180,3 +203,69 @@ const getTweetText = (contract: Contract, isCreator: boolean) => {
...
+
+  const onSave = () => {
+    const newTime = dayjs(timeString).valueOf()
+    if (newTime === time) setIsEditing(false)
+    else if (contract.closeTime && newTime > (contract.closeTime ?? Date.now)) {
```
jahooma, line 220: I think you need to call `Date.now()`. Although, closeTime is always defined at the moment.

jahooma (approved):
Found a few more things to fix, but otherwise approved!

I'll rerun the script to update `autoResolutionTime` for all contracts, and merge this PR in — hopefully later today if you make the small fixes suggested.

Thanks!

functions/src/create-contract.ts
```diff
@@ -113,6 +116,7 @@ export const createmarket = newEndpoint(['POST'], async (req, auth) => {
     ante,
     closeTime.getTime(),
     tags ?? [],
+    autoResolutionTime,
```
jahooma, line 119: This is not in the right order, `autoResolutionTime` should be the last parameter. (To make this more robust, we could switch to taking all the parameters as an object instead of a param list. Not sure if that should be part of this change.)

functions/src/resolve-market.ts
```diff
@@ -31,140 +41,203 @@ export const resolveMarket = functions
...
+  )
+
+export const autoResolveMarkets = functions.pubsub
+  .schedule('every 1 minutes')
+  .onRun(async () => {
+    const contracts = await getValues<Contract>(
+      firestore
+        .collection('contracts')
+        .where('isResolved', '==', false)
+        .where('closeTime', '>', Date.now())
```
jahooma, line 63: closeTime should be '<' Date.now() if the market is closed!

functions/src/resolve-market.ts
```diff
@@ -31,140 +41,203 @@ export const resolveMarket = functions
...
+        .where('autoResolutionTime', '<', Date.now())
```
jahooma, line 64: Firebase does not allow two queries with not equal operators. Advice is to move the closeTime query into a filter operation after loading the contracts.

functions/src/resolve-market.ts
```diff
@@ -31,140 +41,203 @@ export const resolveMarket = functions
...
-        return { status: 'error', message: 'Invalid value' }
+      if (contract.creatorId !== userId)
+        return { status: 'error', message: 'User not creator of contract' }
 
+      return privateResolveMarket(contract, data)
+    }
+  )
+
+export const autoResolveMarkets = functions.pubsub
+  .schedule('every 1 minutes')
```
jahooma, line 57: Needs `.runWith({ secrets: ['MAILGUN_KEY'] })` since it sends emails.

jahooma (commented):
The PR was reverted because of a few bugs I found.

PR #410, opened 2022-06-03
Add Dockerfile

wasabipesto:
Add a Dockerfile for easy setup. 

Run in docker with:
```
git clone https://github.com/manifoldmarkets/manifold.git
docker build -t manifold-img .
docker run --name manifold-dev -p 3000:3000 -d manifold-img
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM node:18
+
+ADD package.json /tmp/package.json
+RUN cd /tmp && npm install
+RUN mkdir -p /opt/manifold && cp -a /tmp/node_modules /opt/manifold/
+
+WORKDIR /opt/manifold
+ADD . .
+RUN yarn
+
+EXPOSE 3000
+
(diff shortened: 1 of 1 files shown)


wasabipesto:
I guess I could also include a docker-compose to build directly from the repo, but that'd be secondary.

sipec:
Interesting!
I personally think our setup is already pretty simple and cross-platform.
I don't know anything about docker though so assigning someone else to review this 😅

wasabipesto:
Every app on my server runs in its own docker so that was the first thing I set up, if it's not something y'all are interested in feel free to nix this.

jahooma:
I'm not that familiar with docker, but the presence of another config file seems like a negative, because it takes more attention, and is another thing to keep up to date.

mqp:
I agree, none of us are going to use this or maintain it, so I would prefer not to have it unless we see further signals that other people would appreciate it.

PR #413, opened 2022-06-03
Add total market volume to API response

wasabipesto:
We already had `volume7Days` and `volume24Hours`, but adding total volume lets us do e.g. external leaderboards.

### DIFF
--- a/web/pages/api/v0/_types.ts
+++ b/web/pages/api/v0/_types.ts
@@ -28,6 +28,7 @@ export type LiteMarket = {
   p?: number
   totalLiquidity?: number
 
+  volume: number
   volume7Days: number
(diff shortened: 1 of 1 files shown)


wasabipesto:
I'm planning on adding more elements but we can discuss those in individual PRs (unless you'd prefer I do them in bulk).
Also I realize I'm adding elements to LiteMarket right as we hit the vercel limit, I can hold off until we have pagination or something.

mqp:
Seems totally reasonable, no idea why this wasn't already in there.

If we're over the limit, it doesn't make anything worse to be further over the limit :-)

mqp (approved):

PR #418, opened 2022-06-04
Switch to Google Secret Manager for function secrets

mqp:
***Deploy status: Secrets are moved over on dev and prod. Functions redeployed on dev but not prod.***

This is important because v2 Cloud Functions doesn't support our old config stuff. So we need to do this switch before we can make any of our functions be v2.

### DIFF
--- a/functions/src/send-email.ts
+++ b/functions/src/send-email.ts
@@ -1,10 +1,12 @@
 import * as mailgun from 'mailgun-js'
-import * as functions from 'firebase-functions'
 
 const DOMAIN = 'mg.manifold.markets'
-const mg = mailgun({ apiKey: functions.config().mailgun.key, domain: DOMAIN })
 
 export const sendTextEmail = (to: string, subject: string, text: string) => {
+  const mg = mailgun({
+    apiKey: process.env.MAILGUN_KEY as string,
+    domain: DOMAIN,
+  })
   const data: mailgun.messages.SendData = {
     from: 'Manifold Markets <info@manifold.markets>',
     to,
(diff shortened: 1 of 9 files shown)


functions/src/send-email.ts
```diff
@@ -1,10 +1,12 @@
 import * as mailgun from 'mailgun-js'
-import * as functions from 'firebase-functions'
 
 const DOMAIN = 'mg.manifold.markets'
-const mg = mailgun({ apiKey: functions.config().mailgun.key, domain: DOMAIN })
 
 export const sendTextEmail = (to: string, subject: string, text: string) => {
+  const mg = mailgun({
```
jahooma, line 6: Would it be better to move this to a function, e.g. `initMailgun()` and then reuse below?

functions/src/stripe.ts
```diff
@@ -42,8 +37,13 @@ const manticDollarStripePrice = isProd
     }
 
 export const createCheckoutSession = functions
-  .runWith({ minInstances: 1 })
+  .runWith({ minInstances: 1, secrets: ['STRIPE_APIKEY'] })
   .https.onRequest(async (req, res) => {
+    const stripe = new Stripe(process.env.STRIPE_APIKEY as string, {
+      apiVersion: '2020-08-27',
+      typescript: true,
+    })
```
jahooma, line 45: Ditto, could be better to move to function `initStripe()`

jahooma (approved):
Sweet! Could you update the README on the new process for adding / reading function secrets?

mqp:
> Sweet! Could you update the README on the new process for adding / reading function secrets?

Yes, I also need to verify that it works in emulators. According to the documentation the emulators should just be able to talk to the cloud secrets no problem.

functions/src/send-email.ts
mqp: OK, I buy that.

mqp:
Seems to work on the emulators, so everything is good to go. Deploying on prod.

PR #427, opened 2022-06-05
Fix free response answers with 0 bets disappearing

jackc271:
Before this PR, free response answers and their associated comments disappeared when all shares in the answer were sold. If it's just an answer, this is a surprising UX but not that bad. But this can also disappear an entire comment thread with great discussion, as I noticed on one of my markets recently.

I suppose the downside is that these answers take up space and are more likely to be undesired - but I think answers with M$1 of bets are much more common anyway.

### DIFF
--- a/web/components/feed/activity-items.ts
+++ b/web/components/feed/activity-items.ts
@@ -200,9 +200,7 @@ function getAnswerGroups(
 ) {
   const { sortByProb, abbreviated, reversed } = options
 
-  let outcomes = uniq(bets.map((bet) => bet.outcome)).filter(
-    (outcome) => getOutcomeProbability(contract, outcome) > 0.0001
-  )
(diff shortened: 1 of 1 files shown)


jackc271:
CC @IanPhilips since you authored the relevant code.

jackc271:
Real-world example: https://manifold-git-fork-jackc271-free-response-disappea-ee916c-mantic.vercel.app/jack/how-should-manifold-encourage-tradi#eqoYju0XFqkYQdynDClf#answer-11

jahooma (approved):
Thanks, Jack! This is originally my mistake, not Ian's.

If we want to hide low-probability options, there would really need to be a way to click to expand them. Simply not hiding these options is an improvement!

mqp:
Well, I agree with that.

PR #430, opened 2022-06-06
Add documentation for new API endpoints

mqp:
Just wrapping this up so we can publicize these if we wish.

The whole API documentation is pretty sloppy but that's probably fine for now.

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -400,14 +400,15 @@ Parameters:
   and an additional `value` parameter is required which is a number representing
   the target value. (Bet on numeric markets at your own peril.)
 
-- Example request
-  ```
-  $ curl https://manifold.markets/api/v0/bet -X POST -H 'Content-Type: application/json' \
-      -H 'Authorization: Key {...}' \
-      --data-raw '{"amount":1, \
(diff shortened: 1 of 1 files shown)


docs/docs/api.md
```diff
@@ -347,23 +374,76 @@ If you have questions, come chat with us on [Discord](https://discord.com/invite
...
+
+Places a new bet on behalf of the authorized user.
+
+Parameters:
+
+- `amount`: Required. The amount to bet, in M$, before fees.
+- `contractId`: Required. The ID of the contract (market) to bet on.
+- `outcome`: Required. The outcome to bet on. For binary markets, this is `YES`
+  or `NO`. For free response markets, this is the ID of the free response
+  answer. For numeric markets, this is a string representing the target bucket,
```
jahooma, line 399: Nice work for even including numeric markets! I think we will probably change them soon to be simple and no longer take a `value` param, but for now this is good.

jahooma (commented):
Thanks Marshall! Looks pretty reasonable on a first read-through. Wait for Sinclair's review though!

jahooma:
@sinclairity Did you want to review this PR?

PR #432, opened 2022-06-06
Migrate placeBet and createContract to v2 functions

mqp:
***Deploy status: See below.***

This should dramatically improve our cold start -- and warm -- performance issues with these two functions, and if everything is working good, we can migrate other functions to work this way.

Some differences between the old and new:

- Annoyingly, v2 Cloud Functions don't allow capital letters in names. They allow hyphens, but the Javascript + Firebase CLI toolchain makes it impossible to get hyphens in there. Whatever.
- The v2 Cloud Functions have beefier CPUs, so the idle 1 minimum instance costs more money, but we don't care.
- I'm renaming `createContract` to `createmarket` while I'm at it.

Here's how this is gonna go down:

1. I merge this.
2. I `firebase deploy --only functions` on prod, without deleting old functions, creating new `placebet` and `createmarket` v2 endpoints with new URLs. (Cloud Functions v2 URLs aren't predictable until you deploy them. They include some nonce.)
3. I change the client code to point at the new functions and test it. Nobody deletes the old functions in between step 2 and 3.
4. I deploy again and delete the old functions.

See #433 for client changes.

### DIFF
--- a/functions/src/api.ts
+++ b/functions/src/api.ts
@@ -1,5 +1,8 @@
 import * as admin from 'firebase-admin'
-import * as functions from 'firebase-functions'
+import { Response } from 'express'
+import { logger } from 'firebase-functions/v2'
+import { onRequest, Request } from 'firebase-functions/v2/https'
+
 import * as Cors from 'cors'
 import { z } from 'zod'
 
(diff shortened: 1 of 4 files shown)


jahooma (approved):
Nice, can't wait to see how this affects latency! Did you try this out in dev?

PR #433, opened 2022-06-06
Point client at new v2 versions of `createmarket` and `placebet`

mqp:
See #432 for more details.

### DIFF
--- a/common/envs/prod.ts
+++ b/common/envs/prod.ts
@@ -40,10 +40,9 @@ export const PROD_CONFIG: EnvConfig = {
     appId: '1:128925704902:web:f61f86944d8ffa2a642dc7',
     measurementId: 'G-SSFK1Q138D',
   },
-  // TODO: fill in real endpoints for prod
   functionEndpoints: {
-    placebet: 'https://placebet-w3txbmd3ba-uc.a.run.app',
-    createmarket: 'https://createmarket-w3txbmd3ba-uc.a.run.app',
+    placebet: 'https://placebet-nggbo3neva-uc.a.run.app',
+    createmarket: 'https://createmarket-nggbo3neva-uc.a.run.app',
   },
   adminEmails: [
     'akrolsmir@gmail.com', // Austin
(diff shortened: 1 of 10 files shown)


common/envs/theoremone.ts
```diff
@@ -12,6 +12,11 @@ export const THEOREMONE_CONFIG: EnvConfig = {
     appId: '1:698012149198:web:b342af75662831aa84b79f',
     measurementId: 'G-Y3EZ1WNT6E',
   },
+  // TODO: fill in real endpoints for T1
+  functionEndpoints: {
+    placebet: 'https://placebet-w3txbmd3ba-uc.a.run.app',
+    createmarket: 'https://createmarket-w3txbmd3ba-uc.a.run.app',
+  },
```
jahooma, line 19: I wonder if we should bother with keeping Theoremone up to date. It's a question for @akrolsmir if we can deprecate these.

common/envs/prod.ts
```diff
@@ -37,6 +40,11 @@ export const PROD_CONFIG: EnvConfig = {
     appId: '1:128925704902:web:f61f86944d8ffa2a642dc7',
     measurementId: 'G-SSFK1Q138D',
   },
+  // TODO: fill in real endpoints for prod
```
jahooma, line 43: I guess you would get the new endpoints for prod after deploying those v2 functions, but not before?

Probably the suffix of the domain name is common across our prod app so the endpoint going forward could be predicted by switching out the function name.

jahooma (approved):
LGTM!

common/envs/prod.ts
mqp: I suspect that's correct, but it's not documented anywhere I can see.

common/envs/theoremone.ts
mqp: Austin earlier expressed to me that he was basically interested in spending zero effort to keep them up, so zero effort is what I am doing at the moment.

PR #434, opened 2022-06-06
Switch to SF Compact as default font

mantikoros:


jahooma:
Cool!

You should post a screenshot with any PR with visual changes.

akrolsmir:
Yeah definitely not a fan, but I guess I could be wrong. As usual, would suggest setting up a FR prediction market to see what users think.

mqp:
I agree that I think I prefer the status quo, although I'm not married to the current font and I think someone could find a new one with new weights that I liked a lot better.

mantikoros:
I feel stronger that the current font is bad, than that my current choice is good.

sipec:
I like our current font. It's fun

PR #437, opened 2022-06-06
Rename /analytics to /stats to work around adblockers

akrolsmir:


web/pages/embed/analytics.tsx
```diff
@@ -5,7 +5,7 @@ import Analytics, {
   CustomAnalytics,
```
jahooma, line 5: Move this file to pages/embed/stats.tsx? Would break existing embeds I guess

jahooma (approved):
Nice!

web/pages/embed/analytics.tsx
akrolsmir: Yeah I didn't do this because I didn't want to break existing embeds. Not sure if a redirect would solve it, but I asusme it's not as much of a problem.

One reason to do this would be if existing embeds are getting flagged by adblock too, but I haven't heard any reports of this.

mqp: Interesting. I wonder what the relevant adblock rules look like.

PR #449, opened 2022-06-07
Fix secret access for some email sending functions

mqp:
Broken via #418. Omitting these meant that our "market closed" emails and our "thanks for paying us money" emails were not operational.


jahooma:
Np, thanks for fixing!

Seems it might also have prevented people from gaining mana they paid for, if `stripeWebhook` wasn't running? If so, we would need to transfer those people Manifold Dollars.

PR #451, opened 2022-06-07
Fr comment ux improvements

IanPhilips:
- Flattened responses to FR answer replies into 1 level.
- Only the most recent bet on a free response market is given a comment input (other open comment inputs are hidden)
- Comment inputs are only as large as the number of lines in a comment. 
- The submit button is always an arrow inside the comment input box now. 
- Abstracted the replies code into a component that comment threads and fr comment threads share. It could be more condensed but would take a lot more work for not too much gain yet.

Screen Shot 2022-06-07 at 3 30 52 PM

### DIFF
--- a/web/components/feed/activity-items.ts
+++ b/web/components/feed/activity-items.ts
@@ -29,7 +29,6 @@ export type CommentInputItem = BaseActivityItem & {
   type: 'commentInput'
   betsByCurrentUser: Bet[]
   commentsByCurrentUser: Comment[]
-  answerOutcome?: string
 }
 
 export type DescriptionItem = BaseActivityItem & {
@@ -74,10 +73,10 @@ export type BetGroupItem = BaseActivityItem & {
 
 export type AnswerGroupItem = BaseActivityItem & {
   type: 'answergroup'
+  user: User | undefined | null
   answer: Answer
-  items: ActivityItem[]
-  betsByCurrentUser?: Bet[]
-  commentsByCurrentUser?: Comment[]
+  comments: Comment[]
+  bets: Bet[]
 }
 
 export type CloseItem = BaseActivityItem & {
@@ -232,31 +231,19 @@ function getAnswerGroups(
 
   const answerGroups = outcomes
     .map((outcome) => {
-      const answerBets = bets.filter((bet) => bet.outcome === outcome)
-      const answerComments = comments.filter((comment) =>
-        answerBets.some((bet) => bet.id === comment.betId)
-      )
       const answer = contract.answers?.find(
         (answer) => answer.id === outcome
       ) as Answer
 
-      let items = groupBets(answerBets, answerComments, contract, user?.id, {
-        hideOutcome: true,
-        abbreviated,
-        smallAvatar: true,
-        reversed,
-      })
-
-      if (abbreviated)
-        items = items.slice(-ABBREVIATED_NUM_COMMENTS_OR_BETS_TO_SHOW)
-
+      // TODO: this doesn't abbreviate these groups for activity feed anymore
       return {
         id: outcome,
         type: 'answergroup' as const,
         contract,
-        answer,
-        items,
         user,
+        answer,
+        comments,
+        bets,
       }
     })
     .filter((group) => group.answer)
@@ -276,7 +263,6 @@ function getAnswerAndCommentInputGroups(
   outcomes = sortBy(outcomes, (outcome) =>
     getOutcomeProbability(contract, outcome)
   )
-  const betsByCurrentUser = bets.filter((bet) => bet.userId === user?.id)
 
   const answerGroups = outcomes
     .map((outcome) => {
@@ -284,25 +270,14 @@ function getAnswerAndCommentInputGroups(
         (answer) => answer.id === outcome
       ) as Answer
 
-      const answerBets = bets.filter((bet) => bet.outcome === outcome)
-      const answerComments = comments.filter(
-        (comment) =>
-          comment.answerOutcome === outcome ||
-          answerBets.some((bet) => bet.id === comment.betId)
-      )
-      const items = getCommentThreads(bets, answerComments, contract)
-
       return {
         id: outcome,
         type: 'answergroup' as const,
         contract,
-        answer,
-        items,
         user,
-        betsByCurrentUser,
-        commentsByCurrentUser: answerComments.filter(
-          (comment) => comment.userId === user?.id
-        ),
+        answer,
+        comments,
+        bets,
       }
     })
     .filter((group) => group.answer) as ActivityItem[]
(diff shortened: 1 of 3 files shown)


IanPhilips:
This breaks some things for the /activity feed but I'd rather not support that right now until we know we want it.

web/components/feed/activity-items.ts
```diff
@@ -362,7 +337,9 @@ function getCommentThreads(
   comments: Comment[],
   contract: Contract
 ) {
-  const parentComments = comments.filter((comment) => !comment.replyToCommentId)
+  let parentComments = comments.filter((comment) => !comment.replyToCommentId)
+  if (contract.outcomeType === 'FREE_RESPONSE')
+    parentComments = comments.filter((comment) => !comment.replyToCommentId)
```
jahooma, line 342: Not sure I understand — this is the same filter operation as above?

web/components/feed/activity-items.ts
```diff
@@ -425,13 +402,13 @@ export function getAllContractActivityItems(
...
-      betsByCurrentUser: [],
-      commentsByCurrentUser: [],
-    })
+    // items.push({
+    //   type: 'commentInput' as const,
+    //   id: 'commentInput',
+    //   contract,
+    //   betsByCurrentUser: [],
+    //   commentsByCurrentUser: [],
+    // })
```
jahooma, line 411: Better to delete than comment out

web/components/feed/feed-answer-comment-group.tsx
```diff
@@ -16,44 +14,74 @@ import { Linkify } from 'web/components/linkify'
...
   const isFreeResponseContractPage = !!commentsByCurrentUser
-  if (mostRecentCommentableBet && !showReply) setShowReplyAndFocus(true)
+  const mostRecentCommentableBet = getMostRecentCommentableBet(
+    betsByCurrentUser,
+    commentsByCurrentUser,
+    user,
+    answer.number.toString()
+  )
+  if (mostRecentCommentableBet && !showReply)
+    scrollAndOpenReplyInput(undefined, answer)
```
jahooma, line 65: Should put this in an effect. Even if you have to add a line to ignore the linter with empty deps list.

web/components/feed/feed-answer-comment-group.tsx
```diff
@@ -16,44 +14,74 @@ import { Linkify } from 'web/components/linkify'
...
+      comment.replyToCommentId &&
+      !comment.answerOutcome &&
+      answerComments.map((c) => c.id).includes(comment.replyToCommentId)
   )
+  const commentsList = answerComments.concat(commentReplies)
+
   const prob = getDpmOutcomeProbability(contract.totalShares, answer.id)
   const probPercent = formatPercent(prob)
   const [open, setOpen] = useState(false)
   const [showReply, setShowReply] = useState(false)
```
jahooma, line 54: I would add some new lines around these useState hooks for readability. Consider also grouping them with the other useState above, with `replyToUsername`.

web/components/feed/feed-answer-comment-group.tsx
```diff
@@ -16,44 +14,74 @@ import { Linkify } from 'web/components/linkify'
...
+  const betsByCurrentUser = (user && betsByUserId[user.id]) || []
```
jahooma, line 55: Prefer '??' to '||' if you can.

web/components/feed/feed-comments.tsx
```diff
@@ -266,20 +314,22 @@ export function CommentInput(props: {
...
   betsByCurrentUser: Bet[]
   commentsByCurrentUser: Comment[]
-  // Tie a comment to an free response answer outcome
-  answerOutcome?: string
-  // Tie a comment to another comment
-  parentComment?: Comment
   replyToUsername?: string
   setRef?: (ref: HTMLTextAreaElement) => void
+  // Reply to a free response answer
+  answerOutcome?: string
```
jahooma, line 320: Could be `parentAnswerOutcome`

web/components/feed/feed-comments.tsx
```diff
@@ -373,74 +424,45 @@ export function CommentInput(props: {
                 )}
             </div>
 
-            <Row className="grid grid-cols-8 gap-1.5 text-gray-700">
```
jahooma, line 376: Nice, no more css grid!

web/components/feed/feed-comments.tsx
```diff
@@ -465,6 +484,18 @@ export function CommentInput(props: {
                 )}
               </Col>
             </Row>
+            <Row>
+              {!user && (
+                <button
+                  className={
+                    'btn btn-outline btn-sm text-transform:capitalize mt-2'
```
jahooma, line 491: We should discuss this, but I believe it's common practice to not capitalize any words other than the first in UI. I think we should go with this style.

E.g. Sign in to comment

web/components/feed/feed-comments.tsx
```diff
@@ -238,7 +286,7 @@ export function getMostRecentCommentableBet(
         )
       ) {
         if (!answerOutcome) return true
-        // If we're in free response, don't allow commenting on ante bet
+        if (sortedBetsByCurrentUser[0].id !== bet.id) return false
```
jahooma, line 289: This is a little confusing. Might be clearer to add an if case at the top level of this function for `if (answerOutcome) { }`, and then consider just the first bet in `sortedBetsByCurrentUser`.

jahooma (approved):
LGTM! Thanks for the cleanup. It's fine to break the /activity feed.

web/components/feed/activity-items.ts
IanPhilips: Whoops! not sure what that was

web/components/feed/feed-answer-comment-group.tsx
IanPhilips: Thanks for the reminder! I tried moving this in a useEffect loop and was getting an endless refresh and couldn't figure out why. Just add betsByCurrentUser in the deps list and it's perfecto now.

web/components/feed/feed-comments.tsx
IanPhilips: Good idea!

web/components/feed/feed-comments.tsx
IanPhilips: See ya never!

web/components/feed/feed-comments.tsx
IanPhilips: Sounds good to me

web/components/feed/feed-comments.tsx
IanPhilips: Good idea!

PR #453, opened 2022-06-07
Speed up notification loading by prepopulating relevant info

IanPhilips:


jahooma (approved):
Nice! It's good to have the code to gracefully handle old notifications for this change.

But maybe in a followup, we should just delete all the notifications before this change, and then change all the code to depend on the source data you have filled in. (Alternatively, we could try to backfill those properties, but I don't think it's that important.)

mqp:
I agree with jahooma, I don't see any reason to clutter up this code when either deleting old notifications or backfilling the old data is easy.

IanPhilips:
Awesome, I'd love to do that! Maybe next week I'll just delete the old style? It might be good to let the new notifications fill in a bit before the old ones disappear.

PR #459, opened 2022-06-08
Simulate Uniswap v3

akrolsmir:
Not really intending to merge for a while, but hoping that this simulator:
- Gives us all a better intuition of how Uniswap/Maniswap v3 could work
- Tests out our implementation of Uniswap v3

TODOs:
- [x] Implement betting
- [ ] Implement limit orders
- [ ] Implement withdrawing (track what happens to different LP's liquidity)
- [x] Approximate the Maniswap curve in Uniswap v3
- [ ] Play with visualizations for injecting and understanding liquidity
- [ ] Implement resolutions?
- [ ] Figure out for n-dimensional case?
- [ ] Add fees? (this looks pretty hard/annoying ☹️)


common/calculate-swap3.ts
```diff
@@ -0,0 +1,260 @@
...
+  pool: { YES: number; NO: number }
+
+  // Uniswap uses 0.01, 0.003, 0.0005. Let's stick with 0.003 for now.
+  // fee: number
+
+  // Min/max is expressed as a odds ratio of cost of YES to cost of NO
+  // E.g. ratio of 1 = 1:1 = 50%; ratio of 3 = 3:1 = 75%
+  // minRatio: number
+  // maxRatio: number
+  minTick: number
```
mantikoros, line 21: Might want to create a design doc for the specific data structures you plan to use. 

I think it would probably be easier to just store the probability range here. It's easy to compute the tick from the probability, and this version is not easily human parseable. (It also isn't robust to changing the tick size in the future.)

PR #461, opened 2022-06-08
dpm fix multi resolution

TrueMilli:
With the current implementation the pool is split among all "winning" shares with the respective probability.
That means an answer with twice the shares gets assigned twice the winnings.

Correct would be to split the pool amongst the "winning" answers and then equally among shares for each answer.

Left old, right new
image

### DIFF
--- a/common/calculate-dpm.ts
+++ b/common/calculate-dpm.ts
@@ -363,13 +363,10 @@ function calculateMktDpmPayout(contract: DPMContract, bet: Bet) {
             )
           }
         )
-      : (probs[outcome] * shares) / weightedShareTotal
+      : (shares / totalShares[outcome]) * probs[outcome]
 
   const totalPool = sum(Object.values(pool))
-  const winnings =
-    outcomeType === 'NUMERIC'
-      ? poolFrac * totalPool
-      : (shares / totalShares[outcome]) * probs[outcome] * totalPool
+  const winnings = poolFrac * totalPool
 
   return deductDpmFees(amount, winnings)
 }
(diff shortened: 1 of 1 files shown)


mantikoros:
> That means an answer with twice the shares gets assigned twice the winnings.

As it should! 

The current behavior seems way more correct to me. Why should someone who invested $90 on a correct outcome, get the same amount as someone who only invested $10?

TrueMilli:
> > That means an answer with twice the shares gets assigned twice the winnings.
> 
> As it should!
> 
> The current behavior seems way more correct to me. Why should someone who invested $90 on a correct outcome, get the same amount as someone who only invested $10?

From the example: The market expected _a_ to be 5/75 = 6.7% probable and _b_ to be 70/75 = 93.3% probable. The market creator decided that the actual numbers were 50-50. People who bet on _a_ should receive much more winnings than people who bet on _b_. With the current model they get just 
People who bet on _b_ actually over estimated the probability and should (without the 100 free liquidity from the market creator) loose money.

Maybe this example makes it clearer: When a market is resolved as MKT with the current mechanism you lose M$ when you bet on choices;
image

mantikoros (approved):
Ok, I'm convinced. Thanks @TrueMilli and @Gurkenglas!

Gurkenglas:
@TrueMilli Yep, you should get money for trades that bring the probabilities closer to the resolve-probabilities. Your last example doesn't quite fit, though: You bought a from 0 to 17%, of course if it resolves as 7% you will lose some money.

TrueMilli:
> @TrueMilli Yep, you should get money for trades that bring the probabilities closer to the resolve-probabilities. Your last example doesn't quite fit, though: You bought a from 0 to 17%, of course if it resolves as 7% you will lose some money.

Good point. I should have bought in lockstep. I feel M$10 -> M$4 with subsidy is still to extreme. And we did the math™.

PR #467, opened 2022-06-08
Update FR comment prob and input after bet

IanPhilips:
Updates fr answers in the comments section in realtime after a bet is placed. Also only scrolls user to an input field for their bet if they made a bet while on the page, (not for previous commentable bets).

### DIFF
--- a/web/components/feed/feed-answer-comment-group.tsx
+++ b/web/components/feed/feed-answer-comment-group.tsx
@@ -54,7 +54,6 @@ export function FeedAnswerCommentGroup(props: {
   )
   const commentsList = answerComments.concat(commentReplies)
   const thisAnswerProbOnMount = useRef<number>()
-  // get most recent bet on this answer to find probability
   const thisAnswerProb = bets
     .filter((bet) => bet.outcome === answer.number.toString())
     .sort((a, b) => b.createdTime - a.createdTime)[0].probAfter
--- a/web/components/feed/feed-answer-comment-group.tsx
+++ b/web/components/feed/feed-answer-comment-group.tsx
@@ -64,6 +64,7 @@ export function FeedAnswerCommentGroup(props: {
 
   useEffect(() => {
     thisAnswerProbOnMount.current = thisAnswerProb.valueOf()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


web/components/feed/feed-answer-comment-group.tsx
```diff
@@ -54,13 +53,22 @@ export function FeedAnswerCommentGroup(props: {
       answerComments.map((c) => c.id).includes(comment.replyToCommentId)
   )
   const commentsList = answerComments.concat(commentReplies)
-
-  const prob = getDpmOutcomeProbability(contract.totalShares, answer.id)
-  const probPercent = formatPercent(prob)
+  const thisAnswerProbOnMount = useRef<number>()
```
jahooma, line 56: One simpler way to get this is using useState:
```
const [thisAnswerProbOnMount] = useState(thisAnswerProb.valueOf())
```

Since useState only initializes the value on first render.

web/components/feed/feed-answer-comment-group.tsx
```diff
@@ -54,13 +53,22 @@ export function FeedAnswerCommentGroup(props: {
...
   const commentsByCurrentUser = (user && commentsByUserId[user.id]) ?? []
   const isFreeResponseContractPage = !!commentsByCurrentUser
+
+  useEffect(() => {
+    thisAnswerProbOnMount.current = thisAnswerProb.valueOf()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
   useEffect(() => {
+    if (thisAnswerProbOnMount.current === thisAnswerProb) return
```
jahooma, line 71: It seems like the logic is you only want to scroll if the most recent bet is made after you loaded the page?

Couldn't you save the time at first render, and then if `mostRecentCommentableBet.createdTime` is greater, proceed with scrolling?

Seems like you would calculate the `mostRecentCommentableBet` outside the effect, and then add have deps of `[mostRecentCommentableBet, showReply, answer, scrollAndOpenReplyInput]`.

And you'd wrap `scrollAndOpenReplyInput` in a useEvent so the ref doesn't change.

web/components/feed/feed-answer-comment-group.tsx
IanPhilips: Good idear

web/components/feed/feed-answer-comment-group.tsx
IanPhilips: Great! that works and I suppose is better than prob as in the far future when there is as much liquidity as the oceans on earth, the probability might not change with one measly bet...

jahooma (approved):
Great! Very nice code~

web/components/feed/feed-answer-comment-group.tsx
IanPhilips: Finally got useEvent to work
https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md

PR #468, opened 2022-06-09
Add paging to `/markets` API endpoint

mqp:
This is ugly, but the status quo is that this endpoint is totally broken so I am doing the simplest thing that seems sane.

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -40,7 +40,7 @@ Lists all markets, ordered by creation date descending.
 
 Parameters:
 
-- `n`: Optional. How many markets to return. The maximum and the default is 1000.
+- `limit`: Optional. How many markets to return. The maximum and the default is 1000.
 - `before`: Optional. The ID of the market before which the list will start. For
   example, if you ask for the most recent 10 markets, and then perform a second
   query for 10 more markets with `before=[the id of the 10th market]`, you will
@@ -50,7 +50,7 @@ Requires no authorization.
 
 - Example request
   ```
-  http://manifold.markets/api/v0/markets?n=1
+  http://manifold.markets/api/v0/markets?limit=1
   ```
 - Example response
   ```json
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -24,8 +24,9 @@ APIs that require authentication accept an `Authorization` header in one of two 
   identity. This is what our web client uses. It will probably be annoying for
   you to generate and we will not document it further here.
 
-API requests that accept parameters should have a body with a JSON object with
-one property per parameter.
+API requests that accept parameters should either have the parameters in the
+query string if they are GET requests, or have a body with a JSON object with
+one property per parameter if they are POST requests.
 
 API responses should always either have a body with a JSON result object (if
 the response was a 200) or with a JSON object representing an error (if the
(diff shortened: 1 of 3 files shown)


docs/docs/api.md
```diff
@@ -35,13 +36,21 @@ response was a 4xx or 5xx.)
 
 ### `GET /v0/markets`
 
-Lists all markets.
+Lists all markets, ordered by creation date descending.
+
+Parameters:
+
+- `n`: Optional. How many markets to return. The maximum and the default is 1000.
```
akrolsmir, line 43: Slight preference for `limit` over `n`; a bit more clear. It's what Stripe uses: https://stripe.com/docs/api/pagination

web/pages/api/v0/markets.ts
```diff
@@ -4,15 +4,46 @@ import { listAllContracts } from 'web/lib/firebase/contracts'
...
+  }
+  if (n < 1 || n > 1000) {
+    res.status(400).json({ error: 'n must be between 1 and 1000.' })
+    return
+  }
+
+  try {
+    const contracts = await listAllContracts(n, before)
+    // Serve from Vercel cache, then update. see https://vercel.com/docs/concepts/functions/edge-caching
+    res.setHeader('Cache-Control', 's-maxage=1, stale-while-revalidate')
```
akrolsmir, line 40: Hm, do you know how the Vercel cache interacts with different query params? Does everything just magically work (eg `/markets?limit=8&before=blah` gets cached for 15 min)?

docs/docs/api.md
akrolsmir: Also not sure that `before` is optimal; Stripe has `ending_before`? But it's kind of weird since our result is creation date with newest first.

akrolsmir (approved):
Looks great. Thanks for making our /markets API usable again!

docs/docs/api.md
mqp: I'm definitely not going to argue that these are optimal, but few things about our API are optimal. We will just have to expect to do a big pass at some point in the future with breaking changes to make the "v1" API.

akrolsmir: Yeah I think "before" is fine, and I'd just do "limit" for clarity

web/pages/api/v0/markets.ts
mqp: I don't really know. Glancing at the documentation, it looks like probably each independent URL will be cached separately for 1 second (not sure where you got the 15 minute idea?)

In general, I think this whole endpoint is a bad idea. If people want a list of all markets, we should provide periodic database dumps. If people want to query markets programmatically, then an online API makes more sense. I don't think it makes sense to jam them both into the same code path and try to make the performance magically work out with clever caching headers.

akrolsmir: Oh, huh, I was mixing it up with the per-market endpoint which is updated every 2 minutes (not sure why I thought 15).

fwiw I'm on board with providing database dumps (maybe hourly?) for clients which want to go over everything and do analysis or something

mqp: (Although we should probably make sure that our API at least returns client caching headers that make sense and works with etags. I have no idea whether it does.)

PR #469, opened 2022-06-09
Comment tips

sipec:
Demo: https://www.loom.com/share/c26688dcef604f6a8dd22f05816dbec4

- [x] transaction logic


jahooma (commented):
This is really cool! I appreciate the loom video explainer, and I'm intrigued by the quadratically growing tip amounts!

I'm convinced enough with the design to proceed, though it's more complex than I would have attempted.

The code is pretty solid. Very clean and clear.

I've left a few comments with some suggestions so far. Happy to review again as the feature is completed or as you have questions for me.

web/components/feed/feed-comments.tsx
```diff
@@ -256,19 +273,116 @@ export function FeedComment(props: {
...
+  const me = useUser()
+  const myId = me?.id || ''
+  const savedTip = tips[myId] || 0
+
+  // optimistically increase the tip count, but debounce the update
+  const [localTip, setLocalTip] = useState(savedTip)
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const score = useMemo(
+    flow(
```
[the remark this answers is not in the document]
mqp: It's cool, and I didn't know lodash had this, but I think it would be clearer to write `const tipValues = Object.values(tips).concat(localTip); return sumBy(tipValues, invQuad)`.

IanPhilips:
General feedback:
The comment tipping looks really cool, I'm excited to tip the good commenters out there! 
We had some more thoughts at standup today:
- Side arrows look a bit unintuitive
    - [i] I’d rather have up/down arrows or a flying money emoji :money_with_wings:
- Shady to allow you to come back much later and take back the tip
    - [S] untipping is not core feature, only short undo
    - [A] Maybe after 1 hour, you cannot un-tip
    - [i] I would vote 1 min or 30 seconds
- [A] Money pump if people go negative
    - Possible fix: Cannot refund yourself if they have negative money
- [A] Downvoting could be good next feature for the down arrow, though
    - [i] Pay money to burn their money

akrolsmir:
@IanPhilips Thank you so much for extracting and summarizing the discussion we had! It's super valuable to help us all stay on the same page, in the context of an async & remote team~

web/components/feed/feed-comments.tsx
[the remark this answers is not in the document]
sipec: mqp I agree, with the caveat that we need the `{ ...tips, [me]: localTip } thing to override your stored tip otherwise you'll double count your own tipped contribution if any of it is stored already.

ofc this is moot if we just do it in the txns

sipec:
> [i] I’d rather have up/down arrows or a flying money emoji 💸

imo up / down looks worse inline, but putting it on the side of the comment is ugly.

web/components/feed/feed-context.tsx
```diff
@@ -0,0 +1,18 @@
+import { createContext, ReactNode, useContext } from 'react'
+import { CommentTipMap, useTipTxns } from 'web/hooks/use-comment-tips'
+
+const FeedContext = createContext<CommentTipMap | null | undefined>(null)
```
jahooma, line 4: I think your implementation using Context is good. Clean and reasonable.

However, we don't use context anywhere else in our codebase. I'm concerned about introducing a new pattern because it adds complexity.

Secondly, I'm not a fan of context in general. It's an implicit parameter and those are bad. Someone who tries to use the FeedComment elsewhere will be mysteriously missing tip data.

I think part of the problem here is that the feed items code is complex and not following normal react structure: we compute a list of activity items as a model first and then pass them dynamically into the relevant feed item components. This makes it hard to just pass props all the way through.

However, I think passing the tips data all the way from `ContractActivity` is still a better solution. Just follow how `comments` are passed through. It's probably less of a code change than setting up this context.

One alternate solution would be to let each comment fetch its own data, which is then cached with react query. It would have to query the txns for its own comment id. I think this would be an ok solution as well.

Something like:
```
export const useCommentTips = (commentId: string) => {
  const result = useDatabaseSnapshot(
    ['comment-tips', commentId],
    // insert query for 'TIP' txns involving commentId,
    { subscribe: true, includeMetadataChanges: true }
  )
  return result.isLoading ? undefined : result.data
}
```

Any thoughts?

web/components/tipper.tsx
```diff
@@ -0,0 +1,160 @@
...
+  useEffect(() => {
+    if (savedTip && !initialized.current) {
+      setLocalTip(savedTip)
+      initialized.current = true
+    }
+  }, [savedTip])
+
+  const score = useMemo(() => {
+    const tipVals = Object.values({ ...tips, [myId]: localTip })
+    return sumBy(tipVals, invQuad)
```
jahooma, line 44: I'm skeptical of showing users a square-root scaled sum. I think they will say it doesn't make sense, especially if they add some M$, and it goes up by less than they paid. We can try it on users, I guess. I give 80% odds we'll switch it to a plain sum though.

web/hooks/use-comment-tips.ts
```diff
@@ -0,0 +1,24 @@
...
+export type CommentTips = { [userId: string]: number }
+export type CommentTipMap = { [commentId: string]: CommentTips }
+
+export function useTipTxns(contractId: string): CommentTipMap {
+  const [txns, setTxns] = useState<TipTxn[]>([])
+
+  useEffect(() => {
+    return listenForTipTxns(contractId, setTxns)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [contractId])
```
jahooma, line 15: You should be able to add the deps here without changing behavior.

web/components/tipper.tsx
```diff
@@ -0,0 +1,160 @@
...
+        data: {
+          contractId: comment.contractId,
+          commentId: comment.id,
+        },
+        description: `${user.name} tipped M$ ${change} to ${comment.userName} for a comment`,
+      })
+    }, 1500)
+  )
+  // instant save on unrender
+  useEffect(() => () => void saveTip.flush(), [])
```
jahooma, line 71: Adding saveTip as a dep should be fine since it is only created once.

web/components/tipper.tsx
```diff
@@ -0,0 +1,160 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
jahooma, line 1: It's better not to use this if you can. It looks pretty straightforward to add the correct dependencies to your effects here.

common/txn.ts
```diff
@@ -13,9 +13,27 @@ export type Txn = {
...
+
+export type TipTxn = Txn & {
+  fromType: 'USER'
+  toType: 'USER'
+  category: 'TIP'
+  data: {
+    contractId: string
+    commentId: string
+  }
+}
```
jahooma, line 39: Nice!

jahooma (commented):
Great, this is good progress! The code looks good. The main comment I have is that I don't like React Context haha.

A couple other points:
- Include a screenshot of the latest visual changes
- The leaderboard and portfolio page calculate profit using your balance. We should subtract out the tips you earn (and add back the tips you sent). You could load your tips in these places to fix the profit calculation, or, I think it may be easier to have the `transact` cloud function update your `totalDeposit` field (in `private-users/{userId}` doc) on each tip appropriately (even if this is a temporary solution).

functions/src/transact.ts
```diff
@@ -25,7 +35,7 @@ export const transact = functions
         message: 'Must be authenticated with userId equal to specified fromId.',
       }
 
-    if (amount <= 0 || isNaN(amount) || !isFinite(amount))
```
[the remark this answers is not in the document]
jahooma: Probably it's ok to have negative transfers? I imagine we might want them at some point. You can still check that the other user doesn't go to a negative balance (but I'm not sure that we need to right now).

sipec:
current design:

https://user-images.githubusercontent.com/18368938/173958681-42dd0361-533c-4f38-a68d-d8ecd2b9c402.mov

web/components/feed/feed-context.tsx
sipec: Using react-query is interesting. But if we let each comment fetch it's own data, we may as well just have add a new firebase listener to each comment directly instead of using react-query. Each query would be unique anyways, so we wouldn't be deduping, and we aren't using the queries loading/error states. It would make more sense to use react-query if we want to grab all the tips in the contract and just filter it on a per-comment level.

We also currently don't use react-query much and it runs into similar code complexity issues as context (like, devs need to learn about query keys).

web/components/tipper.tsx
sipec: it makes more sense in the context of sorting by comment karma (which isn't in this PR)

sipec: I think the actual way people will be confused in practice is thinking that they can only upvote once

web/components/feed/feed-context.tsx
sipec: > passing the tips data all the way is probably less of a code change than setting up this context

Depends how you count. The prop gets added in 20+ different places

jahooma: Good point on just using a firebase listener for the comment tip. That could be the best solution. If you want to be able to sort over all the tips, you could just fetch all the txns again from the parent component.

sipec:
React context removed. I've also tested out that balances move correctly on tip / untip but profit metrics are unaffected.
One thing I want to figure out before pushing is deploying the cloud function to prod

PR #476, opened 2022-06-10
Optimize `updateContractMetrics`

mqp:
I don't know if this is moving the needle on cost, but it's easier to just fix stuff to be better and look at what happens rather than try to discern it by thinking.

I guess this does half as many database queries as the previous version and reads way less data, so that should be good.


jahooma (approved):
LGTM!

PR #477, opened 2022-06-10
Optimize `updateUserMetrics`

mqp:
Similar to #476, I don't know whether this is moving the needle, but it's common sense to not do a thousand database queries when we could do one.

It's possible this will blow the memory budget but we can bump the memory on the instances if so.

### DIFF
--- a/functions/src/update-user-metrics.ts
+++ b/functions/src/update-user-metrics.ts
@@ -1,6 +1,6 @@
 import * as functions from 'firebase-functions'
 import * as admin from 'firebase-admin'
-import { sum, sumBy } from 'lodash'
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Thanks. This does make sense!

Hopefully the memory options of V1 are enough, otherwise we can migrate to V2.

mqp:
> Thanks. This does make sense!
> 
> Hopefully the memory options of V1 are enough, otherwise we can migrate to V2.

I think we can actually up the memory on V1? It's increasing the max execution time that V2 opens up.

jahooma:
True, but I think I have maxed the memory before on V1 cloud functions (like 2GB)! Looks like V2 goes up to 16GB

PR #481, opened 2022-06-10
Fix busted comment permalink copying code

mqp:
This generated broken links, so it wasn't working.

### DIFF
--- a/web/components/feed/copy-link-date-time.tsx
+++ b/web/components/feed/copy-link-date-time.tsx
@@ -38,7 +38,7 @@ export function CopyLinkDateTimeComponent(props: {
     <div className={clsx('inline', className)}>
       <DateTimeTooltip time={createdTime}>
         <Link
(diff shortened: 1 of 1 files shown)


IanPhilips:
Oops I think the error is on line 42

mqp:
> Oops I think the error is on line 42

What do you mean? This version works and I see no obvious problem with line 42.

IanPhilips:
this is line 41-44
```
        <Link
          href={`/${contractCreatorUsername}/${contractCreatorUsername}#${elementId}`}
          passHref={true}
        >
```

IanPhilips:
This is the format of the buggy link: https://manifold.markets/EnopoletusHarding/EnopoletusHarding#HTfHPCk8Srv2eNWggt5m

mqp:
I understand, we just found two different bugs. Now they are both fixed.

PR #482, opened 2022-06-10
Must listen to contract updates for probs to update

IanPhilips:
It looks like listening to contract updates via `useContractWithPreload` is essential to getting the FR comments section answer probs and the top answer probs to be in sync. This makes sense to me, is there anything I'm missing?

### DIFF
--- a/web/components/feed/contract-activity.tsx
+++ b/web/components/feed/contract-activity.tsx
@@ -28,40 +28,38 @@ export function ContractActivity(props: {
   className?: string
   betRowClassName?: string
 }) {
-  const { user, mode, contractPath, className, betRowClassName, contract } =
-    props
+  const { user, mode, contractPath, className, betRowClassName } = props
 
-  const liveContract = useContractWithPreload(contract) ?? props.contract
+  const contract = useContractWithPreload(props.contract) ?? props.contract
 
   const updatedComments =
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    mode === 'only-recent' ? undefined : useComments(liveContract.id)
+    mode === 'only-recent' ? undefined : useComments(contract.id)
   const comments = updatedComments ?? props.comments
 
   const updatedBets =
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    mode === 'only-recent' ? undefined : useBets(liveContract.id)
+    mode === 'only-recent' ? undefined : useBets(contract.id)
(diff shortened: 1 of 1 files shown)


web/components/feed/contract-activity.tsx
```diff
@@ -28,38 +28,40 @@ export function ContractActivity(props: {
...
   const comments = updatedComments ?? props.comments
 
-  // eslint-disable-next-line react-hooks/rules-of-hooks
-  const updatedBets = mode === 'only-recent' ? undefined : useBets(contract.id)
+  const updatedBets =
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    mode === 'only-recent' ? undefined : useBets(liveContract.id)
   const bets = (updatedBets ?? props.bets).filter((bet) => !bet.isRedemption)
-
+  // if (!liveContract) return <div />
```
jahooma, line 45: Remove?

web/components/feed/contract-activity.tsx
```diff
@@ -28,38 +28,40 @@ export function ContractActivity(props: {
   className?: string
   betRowClassName?: string
 }) {
-  const { user, mode, contractPath, className, betRowClassName } = props
+  const { user, mode, contractPath, className, betRowClassName, contract } =
+    props
 
-  const contract = useContract(props.contract.id) ?? props.contract
+  const liveContract = useContractWithPreload(contract) ?? props.contract
```
jahooma, line 34: This only makes sense if useContract is not getting updates, but I thought it would be. Could be a problem with how I'm using the `react-query` library...

In any case, happy to proceed if you think this solves the issue.

jahooma, line 34: I would just leave the name as `contract`. Many of our hooks return live updating data, but we don't prefix the name.

jahooma (approved):
Thanks for investigating!

jahooma:
P.S. one reason this is needed is that we are reading answers from `contract.answers` instead of from the subcollection of answers. These days we mostly ignore the `answers` subcollection... Probably shouldn't have stored answers in two ways.

web/components/feed/contract-activity.tsx
IanPhilips: Yes I did a/b testing and this does fix the issue. I suppose ListenForValues and/or onSnapshot works slightly differently than useFirestoreDocumentData?

PR #484, opened 2022-06-11
Turn on no unused variables linting, kill dead code

mqp:
Continuing to boil the frogs with more linting.


sipec:
I personally like this a lot!

web/hooks/use-contracts.ts
```diff
@@ -50,21 +49,6 @@ export const useInactiveContracts = () => {
...
-
-  useEffect(() => {
-    if (!tags || tags.length === 0) return
-    return listenForTaggedContracts(tags, setContracts)
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [tagsKey])
-
-  return contracts
-}
-
```
jahooma, line 67: I'm not so sure about deleting this code as it seems likely to be useful in the future. Yes, it's possible we could dig through git history to revive it, but probably someone other than me will end up rewriting it.

Could we keep this and `listenForTaggedContracts`?

web/components/feed-create.tsx
```diff
@@ -67,90 +61,3 @@ export function FeedPromo(props: { hotContracts: Contract[] }) {
     </>
   )
 }
-
-export default function FeedCreate(props: {
```
jahooma, line 71: This also seems like a nice, polished component that we might want to revive at some point. It's not as bad to kill because I think we would actually dig up this code if we needed to. So, I'm am ok to delete and decrease our maintenance burden.

jahooma (approved):
I'm skeptical of unused vars causing errors that show the build as broken. I think it wastes more time than it saves. Good engineers would only leave unused variables for a reason. Plus, unused vars are very benign and often useful at signaling intention.

 But I'm willing to try it out. I reserve the right switch it to being just an eslint warning or turn it off completely again.

web/hooks/use-contracts.ts
mqp: Sure.

mqp:
> I'm skeptical of unused vars causing errors that show the build as broken. I think it wastes more time than it saves. Good engineers would only leave unused variables for a reason.

Manifold Markets engineers do not only leave unused variables for a reason! 90% of the examples in this PR are clearly just cases where nobody noticed it was unused anymore, or nobody cared.

As a good engineer, the *way* that I don't leave dead code around is by having my editor point it out. That's why I am turning on these lints -- I personally want these features to help me write better code, and unless I fix the existing warnings and stop people from producing more ones, I can't easily use them myself, because I will be swamped by a giant list of 500 warnings and every file I open will already have warnings everywhere. The reasonable equilibria are 0 and lots. I prefer 0.

> Plus, unused vars are very benign and often useful at signaling intention.

I am totally content to keep around benignly useful variables by prefixing them with an underscore, or benignly useful dead code by commenting it out, or by adding lint exceptions. Fundamentally, I just want to be able to see when I make a mistake without also seeing a giant list of other outstanding mistakes that nobody cares about.

web/components/feed-create.tsx
mqp: I agree with that assessment.

PR #485, opened 2022-06-11
Improve visibility of `updateFooMetrics` functions behavior

mqp:
Basically I wanted to add a bunch of logging so that I can see what's going on in these functions, and add a script to test it.

Based on looking at graphs so far, I think these will be our new cost culprits after `updateFeed`, and we can easily continue to make them more efficient with minimal effort. For example, if we put these two functions together it's basically going to do all the work in the same amount of time, and we aren't taking advantage of Firebase's batched update functionality either. Probably I am going to do this if it saves e.g. twenty bucks per day. (These guys basically scale linearly with the number of things in the DB, so they are only going to get more expensive.)

But I'm getting tired of feeling like I am in the dark while I am changing stuff so I am gonna start lighting things up.

### DIFF
--- a/functions/src/utils.ts
+++ b/functions/src/utils.ts
@@ -1,5 +1,6 @@
 import * as admin from 'firebase-admin'
 
+import { chunk } from 'lodash'
 import { Contract } from '../../common/contract'
 import { PrivateUser, User } from '../../common/user'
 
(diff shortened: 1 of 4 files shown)


functions/src/utils.ts
```diff
@@ -1,8 +1,34 @@
...
+  concurrency = 100
+) => {
+  const results = []
+  const chunks = chunk(xs, concurrency)
+  for (let i = 0; i < chunks.length; i++) {
+    log(`${i * concurrency}/${xs.length} processed...`)
+    results.push(...(await Promise.all(chunks[i].map(fn))))
+  }
+  return results
+}
```
mqp, line 30: Compared to `batchedWaitAll`, obviously I wanted the logging, but I also think this signature is more ergonomic.

mqp:
I'm gonna click this in, since it's pretty innocuous, and if you want to critique anything in it you should critique my next PR where I fix up the rest of the stuff in the metrics functions. I like to keep this stuff moving so I can keep accumulating experimental data about how backend metrics respond to different kinds of changes.

PR #486, opened 2022-06-11
Add quadratic matching to Manifold for Charity

akrolsmir:
So instead of using a fixed quadratic funding pool, Manifold uses a scaling pool - for every $1 donated, we increase the matching pool by $1. This has some nice properties:
- Your donation is always incentive-aligned; every $1 donated leads to at least $1 to your charity.
- Donations to other charities can only increase the total donated to your charity. Unlike with a fixed bounty pool, where other donations can reduce the match that you get!
- We still get the nice "many small donors > 1 large donor" effect of quadratic funding!

I think this achieves some of the nice properties in [this EA forum post](https://forum.effectivealtruism.org/posts/tXavWgk8Xp6Avg8No/quadratic-funding-without-a-matching-pool-1), but since we have DO have a matching pool generously offered by FTX FF, we can use it and make our system much easier to reason about. In our system, users just specify an amount to donate, instead of a payment limit, which should be more intuitive.

TODOs (can be in follow-up PRs)
- [x] Check to see if total match ever decreases. (A: it never does!)
- [ ] Show how much each individual donation produces in terms of matching amounts, on each individual charity page.
- [x] Improve visual design of the main Charity page
- [x] Ask to be added to https://wtfisqf.com/


common/quadratic-funding.ts
```diff
@@ -0,0 +1,30 @@
...
+  const weights: Record<string, number> = {}
+  for (const [charityId, byDonor] of Object.entries(donationsByDonors)) {
+    const sumByDonor = mapValues(byDonor, (txns) =>
+      sum(txns.map((txn) => txn.amount))
+    )
+    const sumOfRoots = Object.values(sumByDonor)
+      .map((s) => s ** 0.5)
+      .reduce((a, b) => a + b)
+    weights[charityId] = sumOfRoots ** 2
+  }
```
sipec, line 25:
```suggestion
  const weights = mapValues(donationsByDonors, (byDonor) => {
    const sumByDonor = Object.values(byDonor)
      .map((txns) => sumBy(txns, "amount"))
    const sumOfRoots = sumBy(sumByDonor, Math.sqrt)
    return sumOfRoots ** 2
  }
```

sipec (approved):
looks good!

filyp:
Hi @akrolsmir ! Sorry for responding so late, I was quite busy.

[This mechanism](https://forum.effectivealtruism.org/posts/tXavWgk8Xp6Avg8No/quadratic-funding-without-a-matching-pool-1#Parameter_choice) actually does have the property "when I send $10, the charity is guaranteed at least $10", because when you pay, the pot increases by your contribution, and also all the increases in others' payments triggered by your contribution. (There are some caveats which I described [here](https://forum.effectivealtruism.org/posts/tXavWgk8Xp6Avg8No/quadratic-funding-without-a-matching-pool-1#Strategic_thinking), but I think they are acceptable for now)

It's true that specifying "payment limit" instead of the actual payment is unintuitive, but the actual payment can be easily calculated based on the limit. So you could for example have a UI where users slide the payment limit, and see the actual payment immediately changing in response to that.


The easiest and most natural modification to make that method work with a matching pool, would be to make a very big contribution from that matching pool, but placed far to the left on the graph, so that it saturates really slowly. It would increase the leverage all the other contributors make, incentivising them to pay more.

The disadvantage is that you have to split your matching pool between all the projects right at the beginning. But thanks to that, the matching of funds is truly independent across projects, and when people fund one of them, it cannot decrease the funds for another.

PR #488, opened 2022-06-12
Dark mode

NcyRocks:
Added a dark mode that adapts to the OS dark mode setting.
Pretty simple conversion using the current colour scheme, I just made all the colours inverted in intensity when dark mode is active.
Things that need further work include the graphs (numbers are too dark), charity logos (made to go on a light background), market stats, and almost certainly other stuff.

### DIFF
--- a/web/components/amount-input.tsx
+++ b/web/components/amount-input.tsx
@@ -41,7 +41,7 @@ export function AmountInput(props: {
         <span className="bg-gray-200 dark:bg-gray-800 text-sm">{label}</span>
         <input
           className={clsx(
-            'input input-bordered max-w-[200px] text-lg dark:bg-black dark:border-gray-800',
+            'input input-bordered max-w-[200px] text-lg dark:bg-black dark:border-gray-700',
             error && 'input-error',
             inputClassName
           )}
--- a/web/components/amount-input.tsx
+++ b/web/components/amount-input.tsx
@@ -41,7 +41,7 @@ export function AmountInput(props: {
         <span className="bg-gray-200 dark:bg-gray-800 text-sm">{label}</span>
         <input
           className={clsx(
-            'input input-bordered max-w-[200px] text-lg',
+            'input input-bordered max-w-[200px] text-lg dark:bg-black dark:border-gray-800',
             error && 'input-error',
             inputClassName
           )}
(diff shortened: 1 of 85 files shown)


IanPhilips:
This is fun and will be way easier to use at night, thanks for making this! Just noticed this:
Screen Shot 2022-06-13 at 7 59 46 AM

akrolsmir:
I was prepared to be skeptical but this actually looks quite good -- thank you so much @NcyRocks!

Some misc things that could be improved:
image

image

image
(this entire pastel color scheme is quite bad and we should pick a new one anyways; but it's especially bad against a dark background. Also, the ticks are really hard to see)

image
The gray text is just a little bit hard to read (on the cards and main copy), I'd try making it one shade lighter

Some other questions:
1. How hard will this be to maintain moving forwards? I would endorse "dark mode is an available/experimental, but we promise no visual support, new features may not have dark mode enabled but we're always open to PRs to fix this)
2. Should there be an explicit toggle somewhere on the site so you can switch between them? (Where would it go? We don't really have "user settings" today haha)

IanPhilips:
Re: maintenance: the more maintainable way I've used dark mode is via a base level theme object that we reference like `theme.background` and `theme.primary` that is modulated by the current theme. This way we don't have to put `color1 dark:color1`  into all our components.

mqp:
> Re: maintenance: the more maintainable way I've used dark mode is via a base level theme object that we reference like theme.background and theme.primary that is modulated by the current theme. This way we don't have to put color1 dark:color2 into all our components.

Strongly support doing this. Looks to me like Tailwind makes it easy to do.

akrolsmir:
Set up https://manifold.markets/Austin/kickstart-will-the-manifold-communi to see if the community will want to pitch in to make this happen~

NcyRocks:
100% agree that using Tailwind to do the dark mode is the better way of doing it - I actually only realised it was possible after I'd already done this. And that swapping all explicit 'color1' mentions for 'theme.primary' or whatever is the right move, and that this commit as it stands will be burdensome to maintain. I didn't quite realise that when I made it, it sorta only became clear when I had to manually override some of the colours of the buttons and things.

If you open up any table (market info, betting history,  leaderboard) you'll see it's still using the same colours, because it's taking them from the theme and there's no way (that I found) to override that. Using a Tailwind/DaisyUI colour scheme was probably the only viable way to do this all along. So it's probably best to scrap this in its current form, swap all colours to a theme, and make the theme either togglable or adapt to device dark mode.

NcyRocks:
Closing for above reasons. I'll probably keep working on a dark mode but this is an inefficient and unsustainable way of implementing it.

PR #489, opened 2022-06-12
Finish optimizing `updateFooMetrics` functions

mqp:
This applies the last "obvious-seeming-to-me" changes to the metrics jobs. It merges `updateContractMetrics` and `updateUserMetrics` into the same `updateMetrics` job, which makes sense because they required basically all the exact same data to run. It also uses Firestore's write batching API to actually apply their updates -- I didn't see us ever using this before, so we will see if there is any surprise.

I think this change will presently probably only save us five or ten bucks a day, but I still like to do it because it's common sense, because the inefficiencies scale with the size of our data, because I am learning, and because I would really like to minimize "we are hammering the database unnecessarily" as a possible culprit for latency problems on our realtime functions.


functions/src/utils.ts
```diff
@@ -15,18 +15,25 @@ export const logMemory = () => {
   }
 }
 
-export const mapAsync = async <T, U>(
```
jahooma, line 18: I thought this was a cool function!

jahooma (approved):
Thanks Marshall! Very clean and efficient. I enjoy reading code like this :-)

functions/src/utils.ts
mqp: A function has to be useful to be cool! It makes sense to me but not if I'm not using it.

PR #492, opened 2022-06-13
Remove unused acitivity items-related code

IanPhilips:
As a precursor to adding liquidity injections into the betting history, remove lots of unused code from the activity items framework.


jahooma (approved):
Sweet! Glad to clean up this old code. Are there plans to abolish the whole feed items model? I think it made more sense when we were interleaving bets and comments.

PR #493, opened 2022-06-13
Don't use bets to filter answers

IanPhilips:
Totalbets wasn't updating like our answer listener was, so I removed it. This allows new answers to show up in the top part as well as the comments section of FR questions.


IanPhilips:
I also switched the useLayoutEffect to useEffect bc the command line is always yelling about how it doesn't do anything, and useEffect seems to work just fine.

IanPhilips:
Do we need that probability filter? I can figure out a diff way to do it if so.

PR #494, opened 2022-06-13
Listen for contract updates

jahooma:


jahooma:
@IanPhilips I think this should fix the problems with answer panel not updating!

PR #495, opened 2022-06-14
🧾 Limit orders! 

jahooma:
Notion design doc: https://manifoldmarkets.notion.site/Limit-orders-a48c86cdef724d3784b43ef574d2cf3a


jahooma:
First pass on bet UI (binary markets only for now).

Screen Shot 2022-06-13 at 8 11 49 PM
Screen Shot 2022-06-13 at 8 11 58 PM

ahalekelly:
Did some API performance testing, ran POST /bet in a loop on one market

| Clients  | Throughput (bets/s) | 90th percentile latency |
| ------------- | ------------- | ------------- | 
| 1  | 1.5  | 851ms| 
| 2  | 2.0  | 1988ms| 
| 4  | 1.6  | 5016ms|

Those aren't per-client throughput numbers, that's total.

mqp:
I have no idea how to make this work performance-wise given Firestore's constraints, but I am OK with taking an approach consistent with the rest of our software, which is roughly "experiment with features in a way that doesn't scale to more than a few users on the website at a time, and then somehow later if we determine the features are right, go rewrite them in a way that scales." As long as everyone is aware that this is what we are doing.

ahalekelly:
Hmm does this implementation write to the contract for every limit order? Austin pointed out that if you avoid that you could bypass the bottleneck.

common/bet.ts
```diff
@@ -25,8 +26,10 @@ export type Bet = {
   isAnte?: boolean
   isLiquidityProvision?: boolean
   isRedemption?: boolean
-
-  createdTime: number
+  // A record of each transaction that partially (or fully) fills the bet amount.
+  // I.e. A limit order could be filled by partially matching with several bets.
+  // Non-limit orders can also be filled by matching with multiple limit orders.
+  fills?: fill[]
```
mantikoros, line 32: The bets subcollection is going to get pretty big with all the limit order and fills being stored...

functions/src/place-bet.ts
```diff
@@ -97,11 +129,45 @@ export const placebet = newEndpoint({}, async (req, auth) => {
...
+
+      // Deduct balance of makers.
+      // TODO: Check if users would go negative from fills and cancel those bets.
+      const spentByUser = mapValues(
+        groupBy(makers, (maker) => maker.bet.userId),
+        (makers) => sumBy(makers, (maker) => maker.amount)
+      )
+      for (const [userId, spent] of Object.entries(spentByUser)) {
+        const userDoc = firestore.collection('users').doc(userId)
+        trans.update(userDoc, { balance: FieldValue.increment(-spent) })
```
mantikoros, line 166: Didn't know FieldValue.increment() was a thing.

common/new-bet.ts
```diff
@@ -30,38 +40,185 @@ export type BetInfo = {
...
+    shares: number
+    timestamp: number
+  }[] = []
+
+  let amount = betAmount
+  let cpmmState = { pool: contract.pool, p: contract.p }
+  let totalFees = noFees
+
+  let i = 0
+  while (true) {
```
mantikoros, line 164: elegant core loop

common/new-bet.ts
```diff
@@ -30,38 +40,185 @@ export type BetInfo = {
...
+    amount -= taker.amount
+
+    if (maker.matchedBetId === null) {
+      // Matched against pool.
+      cpmmState = maker.state
+      totalFees = addObjects(totalFees, maker.fees)
+      takers.push(taker)
+    } else {
+      // Matched against bet.
+      takers.push(taker)
```
mantikoros, line 180: there should also be a creator fee assessed on the taker which should be added to totalFees

mantikoros: I guess you can leave that for v 2.

web/components/limit-bets.tsx
```diff
@@ -0,0 +1,81 @@
...
+  contract: CPMMBinaryContract | PseudoNumericContract
+  bets: LimitBet[]
+  className?: string
+}) {
+  const { contract, bets, className } = props
+  const recentBets = sortBy(bets, (bet) => bet.createdTime).reverse()
+
+  return (
+    <Col className={clsx(className, 'gap-2 rounded bg-white')}>
+      <div className="px-6 py-3 text-xl">Limit bets</div>
```
mantikoros, line 23: I still like "limit order" more.

common/bet.ts
```diff
@@ -35,4 +38,21 @@ export type NumericBet = Bet & {
   allBetAmounts: { [outcome: string]: number }
 }
 
+// Binary market limit order.
+export type LimitBet = Bet & {
+  orderAmount: number // Amount of limit order.
+  limitProb: number // [0, 1]. Bet to this probability.
+  isFilled: boolean // Whether all of the bet amount has been filled.
+  isCancelled: boolean // Whether to prevent any further fills.
```
mantikoros, line 46: "canceled" is the preferred American spelling

common/calculate-cpmm.ts
```diff
@@ -115,6 +119,53 @@ export function calculateCpmmPurchase(
   return { shares, newPool, newP, fees }
 }
 
+export function calculateCpmmAmount(
```
mantikoros, line 122: I think there is probably a closed-form solution to find this amount...

mantikoros: at the very least, there definitely are ways to compute it more efficiently if this ever becomes the bottleneck (unlikely).

mantikoros (approved):
Very nice. A reasonably streamlined solution to adding limit orders based off the Bet object.

common/calculate-cpmm.ts
jahooma: Ok, added a comment to that effect.

common/bet.ts
jahooma: Oh, I guess I can fix that then. Maybe after my tests with discord people so as not to break things. Or, maybe I'll just leave it haha. Two L's makes more sense to me.

common/new-bet.ts
jahooma: I guess I'll leave it for v2. I actually was planning to do it before.

PR #503, opened 2022-06-14
Fix of pool in API

TrueMilli:
@wasabipesto noticed inconsistencies in the market data the api returns.

I've taken a look at LiteMarket, my suggestions are in this PR.
There is more data for non-binary markets now and hopefully it's clearer what the values mean.
The total number of shares in the liquidity pool which was written to "pool" before seems to have been a relic from dpm binary markets and is not in the API anymore.

### Details
* pool: set in the same way as in the "Market Overview" from contract-info-dialog.tsx now

Re-creation of https://github.com/manifoldmarkets/manifold/pull/479


TrueMilli:
@mantikoros: can you point out where the static type issues are/were on the PR?

mantikoros:
> [#479 (comment)](https://github.com/manifoldmarkets/manifold/pull/479#issuecomment-1155466606) @mantikoros: can you point out where the static type issues are/were on the PR?

./pages/api/v0/_types.ts
4:20Error: 'DPM' is defined but never used. Allowed unused vars must match /^_/u.  @typescript-eslint/no-unused-vars
6:10Error: 'sum' is defined but never used. Allowed unused vars must match /^_/u.  @typescript-eslint/no-unused-vars

TrueMilli:
> > [#479 (comment)](https://github.com/manifoldmarkets/manifold/pull/479#issuecomment-1155466606) @mantikoros: can you point out where the static type issues are/were on the PR?
> 
> ./pages/api/v0/_types.ts 4:20Error: 'DPM' is defined but never used. Allowed unused vars must match /^_/u. @typescript-eslint/no-unused-vars 6:10Error: 'sum' is defined but never used. Allowed unused vars must match /^_/u. @typescript-eslint/no-unused-vars

Already noticed and fixed that.

ahalekelly:
I agree the name is very confusing, but I'm relying on the current pool parameter to calculate the number of yes and no shares in the liquidity pool. I would only support removing this if we add yes and no liquidity pool share counts to the API.

Also, changing a value from one thing to another in an API is never a good idea. Better to create a new value and deprecate the old one. totalLiquidity displays the correct value of the pool right? So can we just remove pool, keep totalLiquidity, and add poolYes and poolNo?

And we need to have an API change policy. I would suggest ping the API Hacker role on discord when a change is suggested, again when a change is finalized, and give them a few days to update their scripts before merging.

ahalekelly:
Or if you think totalLiquidity is too confusing, come up with a name that's not currently in use

TrueMilli:
> I agree the name is very confusing, but I'm relying on the current pool parameter to calculate the number of yes and no shares in the liquidity pool. I would only support removing this if we add yes and no liquidity pool share counts to the API.
> 
> Also, changing a value from one thing to another in an API is never a good idea. Better to create a new value and deprecate the old one so you don't cause confusion for users. What if we remove pool, update totalLiquidity to give values for all market types, and add poolYes and poolNo?
> 
> And we need to have an API change policy. I would suggest ping the API Hacker role on discord when a change is suggested, again when a change is finalized, and give them a few days to update their scripts before merging.

An earlier version of this PR had the field TotalShares which contained the total number of shares. Nobody could tell me what the value is good for so I cut it. It's easy to reintroduce if there is demand.

There is a disclaimer that the API is still in alpha and things might break at any time.

When the API is live I agree it should be versioned so people can update on their own terms. But for the time being I would not slow down development to keep legacy code around.

ahalekelly:
Well I can tell you I made M$2,000 last night by knowing the total number of liquidity shares. So there's literally value in having it. Though separate yesShares and noShares would be preferable. https://discord.com/channels/915138780216823849/938171760237477998/987392719867486208

You don't have to slow down development, but at least let us know when changes are coming. Isn't this what the API Hacker role on discord is for?

mqp:
I don't think a few days of advance notice is going to fly. Right now the API is incredibly provisional, and we have a lot of velocity of people requesting and making reasonable changes, and stalling them all for a few days will result in a lot of conflicts and delays. In two or three months it may be appropriate to have a real policy about breaking changes. I agree that it's good to ping the role.

(I have no idea what makes sense with this particular change.)

TrueMilli:
> I don't think a few days of advance notice is going to fly. Right now the API is incredibly provisional, and we have a lot of velocity of people requesting and making reasonable changes, and stalling them all for a few days will result in a lot of conflicts and delays. In two or three months it may be appropriate to have a real policy about breaking changes. I agree that it's good to ping the role.
> 
> (I have no idea what makes sense with this particular change.)

That makes sense to me. Maybe a market could be created or vote held among "API Hackers" in the discord. I don't really use the API myself, I just looked at the code and noticed how little sense it made to me.

ahalekelly:
Alright, I understand that a few days would be too much of a delay. Let's just ping the role on discord when a new change is proposed then.

Milli would it be alright if I fork this and open a new PR?

TrueMilli:
> Alright, I understand that a few days would be too much of a delay. Let's just ping the role on discord when a new change is proposed then.
> 
> Milli would it be alright if I fork this and open a new PR?

Of course, go ahead.

PR #504, opened 2022-06-14
removal of totalLiquidity

TrueMilli:
The "totalLiquidity" field is confusing. It only contains information for binary markets and contains the info which is supposed to be under "Pool" (as in the market info page).

It can be removed after completion of #503.
This is a seperate PR since it needs to be discussed whether this is a breaking change.


TrueMilli:
Superseded by #553

PR #509, opened 2022-06-15
[Leaderboards][Work in Progress] Added period toggle for leaderboards

Pico2x:
@akrolsmir  @jahooma @jahooma Hey guys, thought I'd share the progress so far with a pull request. 

This is what it looks like let me if you'd like any changes to the design. 
[Click for Video](https://user-images.githubusercontent.com/6242616/173882715-f9280126-a2c4-42bb-984f-82db44ad88be.mov
)
1. Changed how firestore stores the relevant data. It used to be number fields in the 'user' collection, they are now maps that store the info for each period. They are still cached every 15mins by the update-user-metrics scheduled cloud function.

Assumption: My first idea here was to store the data in a last30daysVolume[30] array, and calculate the duration when querying. I decided against that as it seems that firebase is very limited when querying array fields. 

2. Updated the methods that compute the user's bettingProfits and creatorVolume in update-user-metrics. They can now take a startTime as a param. updateUserMetricsCore() now uses those methods to store the information for all the relevant time periods (daily,weekly,monthly,alltime). 

Note: Seems like Marshall deleted updateUserMetrics yesterday and moved all scheduled jobs into a new single file, so I'll need to update this. 

3. Updated the relevant queries to take a period parameter.

4. Updated the frontend to show the new UX. 
Note: I used getStaticProps to prebuild the default "weekly" view, but whenever the user selects something else in the toggle (i.e monthly, allTime, daily), I'm fetching the query again with the new period, dynamically. I haven't worked with Jamstack/serverless before, is this bad practice in this context? Should I just pre-fetch all the data in getStaticProps, and just swap what gets rendered when the user selects smth in the toggle?

**Important: There are some merge conflicts with @mqp's 4f96b9ef6342456bf7c9f30f7e90924f6abc8dfe, I'll fix those tomorrow, we can't land this until then.**

### DIFF
--- a/functions/src/update-user-metrics.ts
+++ b/functions/src/update-user-metrics.ts
@@ -9,10 +9,12 @@ import { User } from '../../common/user'
 import { calculatePayout } from '../../common/calculate'
 
 const firestore = admin.firestore()
+const DAY_MS = 24 * 60 * 60 * 1000
 
 const computeInvestmentValue = (
   bets: Bet[],
-  contractsDict: { [k: string]: Contract }
+  contractsDict: { [k: string]: Contract },
+  startTime: number = 0
 ) => {
   return sumBy(bets, (bet) => {
     const contract = contractsDict[bet.contractId]
@@ -20,16 +22,21 @@ const computeInvestmentValue = (
     if (bet.sale || bet.isSold) return 0
 
     const payout = calculatePayout(contract, bet, 'MKT')
+    const betTime = bet.createdTime ?? 0
+    if (betTime < startTime || betTime >= Date.now()) return 0
+
     return payout - (bet.loanAmount ?? 0)
   })
 }
 
 const computeTotalPool = (
   user: User,
-  contractsDict: { [k: string]: Contract }
+  contractsDict: { [k: string]: Contract },
+  startTime: number = 0
 ) => {
   const creatorContracts = Object.values(contractsDict).filter(
-    (contract) => contract.creatorId === user.id
+    (contract) =>
+      contract.creatorId === user.id && contract.createdTime >= startTime
   )
   const pools = creatorContracts.map((contract) =>
     sum(Object.values(contract.pool))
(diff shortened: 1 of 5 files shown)


functions/src/update-user-metrics.ts
```diff
@@ -58,21 +65,96 @@ export const updateUserMetricsCore = async () => {
...
+    return await firestore
+      .collection('users')
+      .doc(user.id)
+      .update({
+        totalPnLCached: {
+          daily: calculatedValues.dailyInvestmentValue,
+          weekly: calculatedValues.weeklyInvestmentValue,
+          monthly: calculatedValues.monthlyInvestmentValue,
+          allTime:
+            calculatedValues.allTimeInvestmentValue +
```
Pico2x, line 84: For the "Top Bettors" leaderboard, we currently add the current user cash balance, and subtract the mana deposits the user has done. 

For Mana deposits: we don't keep track of each time a user deposits mana, so I'm for now NOT subtracting this in the daily,weekly,monthly view - only on the AllTime view. The correct solution will be to start tracking deposits, which I can do as a followup to this. It'll be less trivial. 

For current cash balance: Why are we adding the cash balance when we really only care about profits? Do we want to add it to the weekly/monthly/daily views? I'm keeping it to AllTime only for now, let me know what your thinking is here.

functions/src/update-user-metrics.ts
```diff
@@ -9,27 +9,34 @@ import { User } from '../../common/user'
 import { calculatePayout } from '../../common/calculate'
 
 const firestore = admin.firestore()
+const DAY_MS = 24 * 60 * 60 * 1000
```
jahooma, line 12: This can be imported from `/common/util/time.ts`!

functions/src/update-user-metrics.ts
```diff
@@ -9,27 +9,34 @@ import { User } from '../../common/user'
...
+  contractsDict: { [k: string]: Contract },
+  startTime: number = 0
 ) => {
   return sumBy(bets, (bet) => {
     const contract = contractsDict[bet.contractId]
     if (!contract || contract.isResolved) return 0
     if (bet.sale || bet.isSold) return 0
 
     const payout = calculatePayout(contract, bet, 'MKT')
+    const betTime = bet.createdTime ?? 0
```
jahooma, line 25: `createdTime` should always be defined so coalescing with 0 is unnecessary.

functions/src/update-user-metrics.ts
```diff
@@ -58,21 +65,96 @@ export const updateUserMetricsCore = async () => {
...
+          weekly: calculatedValues.weeklyCreatorVolume,
+          monthly: calculatedValues.monthlyCreatorVolume,
+          allTime: calculatedValues.allTimeCreatorVolume,
+        },
+      })
   })
   log(`Updated metrics for ${users.length} users.`)
 }
 
+const calculateValuesForUser = (
```
jahooma, line 99: Nice consolidated function!

web/pages/leaderboards.tsx
```diff
@@ -30,31 +40,69 @@ export default function Leaderboards(props: {
...
+      setTopCreators(res.props.topCreators as User[])
+      setLoading(false)
+    })
+  }, [period])
 
   return (
     <Page margin>
+      <Title text={'Leaderboards'} className={'hidden md:block'} />
+      <Col className="gap-4 text-center">
+        <select
```
jahooma, line 62: Very nice. Consider using Tabs instead of a select (see tabs.tsx). But up to you, I haven't checked how it looks yet!

functions/src/update-user-metrics.ts
jahooma: Sweet, thanks for thinking about this!

There's two ways to compute profit currently. The leaderboard uses a robust approach by simply adding up your investment value + balance and subtracting your mana deposits.

The second way, used by the portfolio page, involves looking at every bet and counting how much profit (or loss) was made for each, and summing the total. Historically, this has been a little buggy because there are a lot of cases that can be tricky to get right.

I believe that for daily, weekly, and monthly leaderboards the first approach is a bit harder. Your current code computes the investment value of new bets within the respective period, but that's not equivalent to the profit — you would need to subtract the initial investment amount. I.e. Making large bets currently would lead to high investment, putting you on the leaderboard even if you had losses.

The other question is if daily, weekly, monthly leaderboards should count recent profits from old bets. I think they should. E.g. a year ago you bet on Pete Buttigieg for President, and in the last 24 hours the price doubled. That should count on the daily leaderboard.

I'll leave you to puzzle this out!

One more thing to mention: the `stripe-transactions` top-level collection does contain the data of mana deposits per user. Very occasionally we gift users mana, and increment the `totalDeposits`. So`totalDeposits` could still be higher than mana purchased. (We plan to move every transaction into the txns collection eventually, but haven't done this yet.)

jahooma (commented):
Awesome work so far, Fede! I'll be eager to see how you update this PR.

PR #513, opened 2022-06-15
Update Trending Markets

SirSaltyy:
Updated the fixed list of trending markets on the landing page. Chose ones which are slightly more friendly to new users and mainstream.


wasabipesto:
I didn't realize these were hardcoded - any interest in making them rotate through currently popular markets?

SirSaltyy:
> 

We use to do something like this but then people would boost their markets to the front page by spam buying and selling.
Another thing to consider is the people who are going to be interacting with the trending markets are likely to be signed-out new users. Thus, I want I don't want markets that are too complicated or meta to be appearing.

Issue #508, opened 2022-06-15
Market creation API should include new market URL in response


bcongdon:
Example binary market creation response:

```
{
    "id": "JNz71KomjIuJ79bjpCJD",
    "slug": "testing-binary-market-creation-thro",
    "mechanism": "cpmm-1",
    "outcomeType": "BINARY",
    "totalLiquidity": 100,
    "initialProbability": 0.99,
    "p": 0.99,
    "pool": {
        "YES": 100,
        "NO": 100
    },
    "creatorId": "HjUZah4YkjUImgPS1AR5tnfWjTg2",
    "creatorName": "Ben",
    "creatorUsername": "bcongdon",
    "creatorAvatarUrl": "https://lh3.googleusercontent.com/a-/AOh14GhNf21x93pt4uam3fXie3D3QYySjAUyPxE6TdOkwnA=s96-c",
    "question": "Testing\n        Binary Market creation through API",
    "description": "Going to resolves as N/A",
    "tags": [
        "fun"
    ],
    "lowercaseTags": [
        "fun"
    ],
    "visibility": "public",
    "isResolved": false,
    "createdTime": 1655304763532,
    "closeTime": 4102444800000,
    "volume": 0,
    "volume24Hours": 0,
    "volume7Days": 0,
    "collectedFees": {
        "creatorFee": 0,
        "liquidityFee": 0,
        "platformFee": 0
    }
}
```

For consistency, the response should include the `url` field, so it's closer to the `LiteMarket` response provided by other API routes

PR #517, opened 2022-06-16
Include market URL in createmarket response

bcongdon:
My intention with this is to resolve https://github.com/manifoldmarkets/manifold/issues/508.

I'm not sure how to test this, since the relevant change is to the backend Firebase function

### DIFF
--- a/common/contract.ts
+++ b/common/contract.ts
@@ -12,6 +12,7 @@ export type AnyContractType =
 export type Contract<T extends AnyContractType = AnyContractType> = {
   id: string
   slug: string // auto-generated; must be unique
+  url: string
 
   creatorId: string
   creatorName: string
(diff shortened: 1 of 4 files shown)


common/contract.ts
```diff
@@ -12,6 +12,7 @@ export type AnyContractType =
 export type Contract<T extends AnyContractType = AnyContractType> = {
   id: string
   slug: string // auto-generated; must be unique
+  url: string
```
mqp, line 15: The purpose of this type is to represent the data in our database, so it would be incorrect to add this here.

mqp (commented):
I am in the middle of figuring out what to do about this "nobody knows how to make URLs" business. I think what is going to happen is we are going to return slugs from the API and then make it clear that you can build a URL with `https://manifold.markets/m/{slug}`, with no username. So I will address this once we have consensus, which will be soon. Until then, let's not fiddle with it.

common/contract.ts
bcongdon: Thanks for the context! Agreed it makes sense to wait on this, given the discussion around URLs

Issue #507, opened 2022-06-15
Sending a POST to the /markets API route should return an error


bcongdon:
`curl -X POST https://manifold.markets/api/v0/markets` should not return the list of markets, as it currently does. Should likely return a "invalid method" error

PR #518, opened 2022-06-16
Add HTTP method checking to API routes

bcongdon:
Resolves https://github.com/manifoldmarkets/manifold/issues/507

Adds validation to the public API routes so they only respond to the HTTP methods they were intended for

### DIFF
--- a/web/pages/api/v0/bet/index.ts
+++ b/web/pages/api/v0/bet/index.ts
@@ -4,11 +4,16 @@ import {
   CORS_ORIGIN_LOCALHOST,
 } from 'common/envs/constants'
 import { applyCorsHeaders } from 'web/lib/api/cors'
+import { assertHTTPMethod } from 'web/lib/api/validation'
 import { fetchBackend, forwardResponse } from 'web/lib/api/proxy'
 
 export const config = { api: { bodyParser: false } }
 
 export default async function route(req: NextApiRequest, res: NextApiResponse) {
+  if (!assertHTTPMethod(req, res, 'POST')) {
+    return
+  }
+
   await applyCorsHeaders(req, res, {
     origin: [CORS_ORIGIN_MANIFOLD, CORS_ORIGIN_LOCALHOST],
     methods: 'POST',
--- a/web/pages/api/v0/market/[id].ts
+++ b/web/pages/api/v0/market/[id].ts
@@ -3,12 +3,17 @@ import { Bet, listAllBets } from 'web/lib/firebase/bets'
 import { listAllComments } from 'web/lib/firebase/comments'
 import { getContractFromId } from 'web/lib/firebase/contracts'
 import { applyCorsHeaders, CORS_UNRESTRICTED } from 'web/lib/api/cors'
+import { assertHTTPMethod } from 'web/lib/api/validation'
 import { FullMarket, ApiError, toLiteMarket } from '../_types'
 
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse<FullMarket | ApiError>
 ) {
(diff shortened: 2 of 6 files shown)


web/pages/api/v0/bet/index.ts
```diff
@@ -4,11 +4,16 @@ import {
...
 import { applyCorsHeaders } from 'web/lib/api/cors'
+import { assertHTTPMethod } from 'web/lib/api/validation'
 import { fetchBackend, forwardResponse } from 'web/lib/api/proxy'
 
 export const config = { api: { bodyParser: false } }
 
 export default async function route(req: NextApiRequest, res: NextApiResponse) {
+  if (!assertHTTPMethod(req, res, 'POST')) {
+    return
+  }
```
mqp, line 15: I don't believe we should need these checks on the routes that just proxy to Firebase, because the Firebase functions should be checking the method and forwarding back appropriate error messages and status codes. Have you found otherwise?

web/pages/api/v0/market/[id].ts
```diff
@@ -3,12 +3,17 @@ import { Bet, listAllBets } from 'web/lib/firebase/bets'
...
+import { assertHTTPMethod } from 'web/lib/api/validation'
 import { FullMarket, ApiError, toLiteMarket } from '../_types'
 
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse<FullMarket | ApiError>
 ) {
+  if (!assertHTTPMethod(req, res, 'GET')) {
+    return
+  }
```
Comment (mqp) [inline web/pages/api/v0/market/[id].ts:15]: I'm concerned about these `GET` methods that aren't proxying to Firebase because they should also support `HEAD` and `OPTIONS` requests.

mqp (changes requested):

web/pages/api/v0/bet/index.ts
bcongdon: Currently the error message for proxied requests is suboptimal:

```
curl -i -X GET https://manifold.markets/api/v0/market
HTTP/2 500
content-type: application/json; charset=utf-8
x-matched-path: /api/v0/market
cache-control: public, max-age=0, must-revalidate
content-length: 47
date: Thu, 16 Jun 2022 23:39:49 GMT
etag: "2f-NLzThRQiuTD7mRueNLeWmzSmcSA"
vary: Origin
x-vercel-cache: MISS
age: 0
server: Vercel
x-vercel-id: pdx1::cle1::4qt5v-1655422789083-3869a3577425
strict-transport-security: max-age=63072000

{"message":"Error communicating with backend."}
```

With the change, it looks like:
```
HTTP/1.1 405 Method Not Allowed
Allow: POST
Date: Thu, 16 Jun 2022 23:40:14 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Transfer-Encoding: chunked

Method GET Not Allowed
```

mqp: That's on me then, I will look and understand what is going on. The goal was for the Vercel routes to be a very transparent proxy -- having logic on both sides is a recipe for confusion.

web/pages/api/v0/market/[id].ts
Comment (mqp) [reply web/pages/api/v0/market/[id].ts:15]: I definitely care about correctly interpreting the method and producing appropriate errors, I just don't want to not support functionality we want.

web/pages/api/v0/bet/index.ts
mqp: Fixed in https://github.com/manifoldmarkets/manifold/pull/542.

bcongdon: Thanks!

Issue #515, opened 2022-06-15
Market API should expose text of free response outcomes


bcongdon:
Currently, it doesn't seem possible to get the text of free response options from the API. In the `bets[]` array, the API returns things like `"outcome": "4"`, which aren't descriptive enough.

Perhaps `LiteMarket` should contain a list of `Outcomes`, for free response markets.

I'd be happy to work on this

PR #519, opened 2022-06-16
Include free response answers in FullMarket API response

bcongdon:
Resolves https://github.com/manifoldmarkets/manifold/issues/515

Adds "free response" market answers to the FullMarket responses in API routes

### DIFF
--- a/web/pages/api/v0/market/[id].ts
+++ b/web/pages/api/v0/market/[id].ts
@@ -1,7 +1,10 @@
 import { NextApiRequest, NextApiResponse } from 'next'
 import { Bet, listAllBets } from 'web/lib/firebase/bets'
 import { listAllComments } from 'web/lib/firebase/comments'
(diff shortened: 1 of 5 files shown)


web/pages/api/v0/market/[id].ts
```diff
@@ -25,6 +28,8 @@ export default async function handler(
     'userId'
   >[]
 
+  const answers = (contract as FreeResponseContract).answers
```
Comment (mqp) [inline web/pages/api/v0/market/[id].ts:31]: I think it would be more idiomatic Typescript to sort of be more explicit rather than typecasting:
```typescript
const answers = contract.outcomeType === 'FREE_RESPONSE' ? contract.answers : undefined;
```
For example, that would result in `answers` having the correct type of `Answer[] | undefined`, instead of the incorrect type of `Answer[]`.

mqp (approved):

PR #520, opened 2022-06-16
Step charts

ahalekelly:
Step charts accurately represent the price at each point in time. And they look better too!
image
Additional photos: https://imgur.com/a/priJVgt

### DIFF
--- a/web/components/answers/answers-graph.tsx
+++ b/web/components/answers/answers-graph.tsx
@@ -108,6 +108,7 @@ export const AnswersGraph = memo(function AnswersGraph(props: {
         }}
         colors={{ scheme: 'pastel1' }}
         pointSize={0}
+        curve='stepAfter'
         enableSlices="x"
         enableGridX={!!width && width >= 800}
         enableArea
--- a/web/components/answers/answers-graph.tsx
+++ b/web/components/answers/answers-graph.tsx
@@ -108,7 +108,7 @@ export const AnswersGraph = memo(function AnswersGraph(props: {
         }}
         colors={{ scheme: 'pastel1' }}
         pointSize={0}
-        curve='stepAfter'
+        curve="stepAfter"
         enableSlices="x"
         enableGridX={!!width && width >= 800}
         enableArea
--- a/web/components/answers/answers-graph.tsx
+++ b/web/components/answers/answers-graph.tsx
@@ -113,7 +113,7 @@ export const AnswersGraph = memo(function AnswersGraph(props: {
         enableGridX={!!width && width >= 800}
         enableArea
         areaOpacity={1}
-        margin={{ top: 20, right: 28, bottom: 22, left: 40 }}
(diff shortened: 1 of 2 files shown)


jahooma (approved):
Awesome, let's do it! Thanks for thinking of this and implementing it!

web/components/answers/answers-graph.tsx
```diff
@@ -108,6 +108,7 @@ export const AnswersGraph = memo(function AnswersGraph(props: {
         }}
         colors={{ scheme: 'pastel1' }}
         pointSize={0}
+        curve='stepAfter'
```
jahooma, line 111: Sorry, could you change to double quotes, please! (Or, run prettier, our auto-formatter.)

web/components/contract/contract-prob-graph.tsx
```diff
@@ -84,7 +84,8 @@ export const ContractProbGraph = memo(function ContractProbGraph(props: {
           format: (time) => formatTime(+time, lessThanAWeek),
         }}
         colors={{ datum: 'color' }}
-        pointSize={bets.length > 100 ? 0 : 10}
+        curve='stepAfter'
```
jahooma, line 87: Ditto!

jahooma (approved):

web/components/answers/answers-graph.tsx
ahalekelly: Sure, done!

wasabipesto:
I like both the look and utility of this. The "old" line charts always had this weird cognitive dissonance where no point on the chart was technically accurate except the actual bet points. These new step charts now exactly match how I calculate "average historical probability" in my analytics.

web/components/contract/contract-prob-graph.tsx
```diff
@@ -84,13 +78,14 @@ export const ContractProbGraph = memo(function ContractProbGraph(props: {
...
-        pointSize={bets.length > 100 ? 0 : 10}
+        curve="stepAfter"
+        pointSize={0}
         pointBorderWidth={1}
         pointBorderColor="#fff"
         enableSlices="x"
         enableGridX={!!width && width >= 800}
         enableArea
-        margin={{ top: 20, right: 28, bottom: 22, left: 40 }}
+        margin={{ top: 20, right: 0, bottom: 25, left: 40 }}
```
jahooma, line 88: The reason for the right margin is the labels get cut off:
Screen Shot 2022-06-16 at 11 03 04 AM

jahooma (approved):
Thanks! I was thinking about not making the line extend so far to the right :-)

web/components/contract/contract-prob-graph.tsx
ahalekelly: Ah interesting. Well with the other changes in my last commit, we don't need to get rid of the right margin anymore, I can put it back.

mqp:
+1 for strongly preferring these to the status quo.

PR #521, opened 2022-06-16
Clean some API stuff up and instrument `placeBet` with a bunch of logging

mqp:
After this, we should be able to start getting some idea for what is taking all the time in `placeBet`.

### DIFF
--- a/functions/src/api.ts
+++ b/functions/src/api.ts
@@ -1,9 +1,7 @@
 import * as admin from 'firebase-admin'
-import { Response } from 'express'
 import { logger } from 'firebase-functions/v2'
-import { onRequest, Request } from 'firebase-functions/v2/https'
+import { HttpsOptions, onRequest, Request } from 'firebase-functions/v2/https'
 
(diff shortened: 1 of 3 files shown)


functions/src/api.ts
```diff
@@ -122,18 +107,21 @@ export const validate = <T extends z.ZodTypeAny>(schema: T, val: unknown) => {
   }
 }
 
+const DEFAULT_OPTS: HttpsOptions = {
+  minInstances: 1,
+  cors: [CORS_ORIGIN_MANIFOLD, CORS_ORIGIN_LOCALHOST],
```
jahooma, line 112: Cool! This looks simpler than before.

jahooma (approved):
Thanks Marshall! Let's see what the logs reveal.

functions/src/api.ts
mqp: Yes, including CORS control is new functionality in v2 Cloud Functions, so before we had to do it ourselves, but now we don't.

PR #525, opened 2022-06-16
Kill `updateRecommendations` job

mqp:
James mentioned we can kill this because it's also just serving old dead activity feed functionality. It costs us a couple dollars a day, so sure, let's kill it then.


jahooma (approved):
LGTM!

PR #528, opened 2022-06-17
Don't unnecessarily query contract in `onCreateBet`

mqp:
I just happened across this and it looks to me like there's no value in doing this check, so let's not do it.


jahooma (approved):
LGTM!

PR #529, opened 2022-06-17
New tooltip for binary market chart, with date and time

ahalekelly:
Always open to suggestions, I'm no designer
image

### DIFF
--- a/web/components/contract/contract-prob-graph.tsx
+++ b/web/components/contract/contract-prob-graph.tsx
@@ -1,5 +1,6 @@
 import { DatumValue } from '@nivo/core'
-import { ResponsiveLine } from '@nivo/line'
+import { ResponsiveLine, SliceTooltipProps } from '@nivo/line'
+import { BasicTooltip } from '@nivo/tooltip'
 import dayjs from 'dayjs'
 import { memo } from 'react'
 import { Bet } from 'common/bet'
@@ -72,7 +73,7 @@ export const ContractProbGraph = memo(function ContractProbGraph(props: {
           min: startDate,
           max: latestTime.toDate(),
         }}
-        xFormat={(d) => formatTime(+d.valueOf(), lessThanAWeek)}
+        xFormat={(d) => formatTooltipTime(+d.valueOf())}
         axisBottom={{
           tickValues: numXTickValues,
           format: (time) => formatTime(+time, lessThanAWeek),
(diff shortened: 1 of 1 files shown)


ahalekelly:
I would have liked to have a dot or something on the selected point because it can be kind of confusing which line you're looking at, but nivo doesn't have an option for that. We could switch from the "slice" tooltip to the "mesh" tooltip, but the mesh tooltip is fixed to the selected point and doesn't center itself on the screen like the slice one does and so hangs way over the edge of the chart. Just replace `enableSlices="x"` with `useMesh={true}` to compare.

akrolsmir (approved):
This looks great - thank you so much, @ahalekelly! I love that we've gotten rid of the extraneous "[]YES" info since it provided zero helpful information.

Some things on my design wishlist:
- It's kind of weird that the granularity changes on a single graph (it moves from just "Jan 3" to "Jun 6 7pm")
- Would love if the horizontal line didn't get "stuck" between two bets, but rather can just always show what the current time is
- Having it on one line might be visually better than having it on two; two lines takes up more visual weight than I'd like with a large white box. I think having the bold percent first, with possibly a bit of spacing, should do well to keep them visually distinct

You've worked with Nivo a lot more than I have at this point so I'm not sure which of these are feasible; I trust you to use your best judgement!

Also, protip from one non-designer to another: copy UI/UX designs from places with good design taste! For things like financial charts, I generally refer to Robinhood, Coinbase, and Monarch Money for inspiration~

ahalekelly:
Thanks! Good suggestions.

The cursor getting stuck is something I noticed too. I think the only way to fix that would be to generate fake data points that nivo can jump to. I'll look into how much work that would be.

The other two should be straightforward. I did have the text on one line at first and thought that took up too much horizontal space, I'll send a comparison.

ahalekelly:
Here's the tooltip on one line.

Thoughts on date formatting for markets less than a week long? Currently for the x axis ticks it displays just the time if the tick is today, otherwise displays the date and time. If you don't want the granularity on the tooltip to change I think we should say Today and Yesterday in place of the date, and I think that's more readable on the x axis ticks as well.
image
Current tick formatting for reference:
image

akrolsmir (approved):
Great, this looks amazing! Have a slight preference for showing the minute of the bet "Today, 6:34pm" instead of "Today, 6pm", because it makes the legend feel more dynamic as you move over the different times within the same hour.

No strong feelings about whether to use Today or the date in the axis below, I'll let you decide on that one~

akrolsmir:
This is beautiful; thank you so much!

Issue #527, opened 2022-06-17
Feature request: Top Followed users leaderboard


bcongdon:
It'd be nice to have a leaderboard with the top followed users. -- I'd be interested in working on this.

My understanding is that Firebase doesn't easily allow for aggregation queries (e.g. `SELECT user, COUNT(*) FROM follows GROUP BY user`), so this is nontrivial.

One approach would be to:
1. Add a `followersCached` property to the User document (similar to `creatorVolumeCached`)
2. Increment/decrement this cached value upon follow/unfollow.
3. Backfill this value for existing users.
4. Base the leaderboard off of `followersCached`.

Does this seem reasonable?

(See also https://manifold.markets/ScottLawrence/will-manifold-create-a-leaderboard)

jahooma:
Yes! That's a good approach. If you don't want to write a backfilling script, you could query for the followers on every follow / unfollow, count them, and set `followersCached` on the user.

I tend to like verbose names, but you might try `followerCountCached` instead, so it doesn't sound like a list of followers. Also, I'm not sure you really need the `Cached` postfix after all. So I'd go with just `followerCount`.

PR #531, opened 2022-06-17
Create Top Followed Users leaderboard

bcongdon:
Addresses https://github.com/manifoldmarkets/manifold/issues/527.

This takes the approach outlined by @jahooma, where instead of backfilling all follower counts for users, the backend queries/caches the follower count upon follow/unfollow. (This will mean that follower counts will be inaccurate until each user gets followed/unfollows at least once)

---

I don't have access to the Manifold dev db, so I wasn't able to test the backend function changes


sipec (approved):

sipec:
just deployed the cloud functions onto the dev db, so you should be able to test this now

bcongdon:
This logic is currently incorrect -- it calculates the number of users a users follows, rather than the number of followers for a user.

Given the way the Firestore documents are laid out, I think we'll have to do the backfill + increment/decrement on follow/unfollow approach.

I'll ping this PR once I've made the change

bcongdon:
OK, this is ready for another review. I added a backfill script, which I tested against the dev environment. With the backfill and the new function, the leaderboard seems to be working: (Note this is in the dev environment, which doesn't have many follower records)
Screen Shot 2022-06-17 at 6 32 07 PM

functions/src/scripts/backfill-followers.ts
```diff
@@ -0,0 +1,47 @@
...
+
+async function backfillFollowers() {
+  console.log('Backfilling user follower counts')
+  const followerCounts: { [userId: string]: number } = {}
+  const users = await getValues<User>(firestore.collection('users'))
+
+  console.log(`Loaded ${users.length} users. Calculating follower counts...`)
+  for (const [idx, user] of users.entries()) {
+    console.log(`Querying user ${user.id} (${idx + 1}/${users.length})`)
+    const followsSnaps = await getValues<Follow>(
```
jahooma, line 20: I wouldn't call these snaps, because they are not firestore snapshots, they are Follow objects. Probably just `follows`

jahooma (approved):
Hey Ben, this looks very clean and correct. Thanks for implementing!

Since I have access to the prod database, I can run the backfilling script, deploy the cloud functions, and merge your code in.

functions/src/scripts/backfill-followers.ts
bcongdon: Done, thanks for the review!

jahooma:
Ran the script, deployed the cloud functions and merged. Thanks!

PR #540, opened 2022-06-18
[Leaderboards] Added period toggle for leaderboards

Pico2x:
Had to create a separate PR - closed the original but you can still see it here: https://github.com/manifoldmarkets/manifold/pull/509

The new diff (3f4614f34850c9362ebae831903984be0b214c85) fixes the Top Bettors leaderboard so it correctly calculates the profit in a time period. 

In order to calculate the, for example, Weekly profit of a particular user, we need to do:
weeklyProfit = allTimeProfitNow - allTimeProfitAWeekAgo

We can easily calculate the allTimeProfit (its what was already in the leaderboard), but retroactively calculating the allTimeProfit-a-week-ago is a rabbithole. As far as I can tell, I'd have to figure out a way to reconstruct the old "investmentValue" based on the bets at the time, find a way to deduce the sharePrice a week ago for CPMM2 markets, find a different method for DPM markets, find a way to deduce the totalDeposits a week ago based on the stripe-transaction history, find a way to deduce the cash balance at the time, etc. 

I decided to take a caching approach instead. I added a "PortfolioHistory" field to the user collection, which looks like this: 

`[{
  investmentValue: number
  balance: number
  totalDeposits: number
  timestamp: number
}]`

Whenever the update-metrics scheduled function runs, it calculates new portfolio values, **checks whether there have been changes vs the last element of the portfolioHistory**, and if so, appends the new values. 

Each value is a snapshot of a users portfolio in one moment in time. The allTimeProfit for any given portfolio = investmentValue + balance - totalDeposits. To find the all time profit a week ago, I search the portfolioHistory for the most recent values that are over a week old. I'm doing this in descending order to make it performant (as the portfolioHistory will get longer and longer with time). 

Let me know if this approach seems reasonable. Obviously, we won't get retroactive information, so when shipping this we can start by showing only daily & allTime leaderboards, and show the weekly/monthly ones after that amount of time has passed. 

Another thing we should discuss is which cadence we'd like to store this at. Currently new values get appended to the array every 15mins (ONLY if there's been an update). That's an upper ceiling of 35k values stored per user per year. It's just very simple text so it's not terrible, but we can definitely get away with updating at much longer intervals. 

For the leaderboards we could do it every few hours without affecting the user experience much, imo. 
But you guys wanted to also show a 'portfolio-over-time' graph in the Portfolio tab, and in that case it might be nice to have some more granularity. We could update it every 15mins during the day, and then run a job once a day that deletes 95% of the older values. 

Anyway, let me know what you think! 

[DEMO VIDEO](https://user-images.githubusercontent.com/6242616/174415161-623d1b2d-98a4-4974-bfea-97c50fed391c.mp4
)


web/pages/leaderboards.tsx
```diff
@@ -30,31 +40,68 @@ export default function Leaderboards(props: {
...
+      setLoading(false)
+    })
+  }, [period])
 
   return (
     <Page margin>
+      <Title text={'Leaderboards'} className={'hidden md:block'} />
+      <Col className="gap-4 text-center">
+        <select
+          onChange={async (e) => {
```
jahooma, line 63: Is there a reason for this to be async?

yarn.lock
```diff
@@ -250,65 +250,11 @@
   resolved "https://registry.yarnpkg.com/@amplitude/analytics-types/-/analytics-types-0.2.1.tgz#ab4f9b4bbec8afa768a27af78f75bb69dd2a3fc7"
   integrity sha512-+FfXlCjHysYWliRBjD2wQ2gZ4V6jGKskdt9j8npv9Rmzdehj04OAHudZ/UrgH5++88As9ww1wyaS7y4sm4x8vA==
 
-"@amplitude/eslint-config-typescript@^1.10.0":
```
jahooma, line 253: I think you can safely discard these yarn.lock changes.

functions/src/create-user.ts
```diff
@@ -67,8 +68,9 @@ export const createUser = functions
       balance,
       totalDeposits: balance,
       createdTime: Date.now(),
-      totalPnLCached: 0,
-      creatorVolumeCached: 0,
+      profitCached: { daily: 0, weekly: 0, monthly: 0, allTime: 0 },
+      creatorVolumeCached: { daily: 0, weekly: 0, monthly: 0, allTime: 0 },
+      portfolioHistory: [] as PortfolioMetrics[],
```
jahooma, line 73: Love your solution! One thing I'd suggest though is to put the portfolio history into a user subcollection. e.g. `users/{userId}/portfolio`, with each doc containing the PortfolioMetrics.

This solves a lot of the scaling concerns because:
1. Firebase has a 1MB cap on the size of a doc.
2. User objects are public and downloaded by everyone frequently, so it's best not to store a lot in it anyway.
3. We can feel free to store as much data as we want, even on a 15 minute interval with little tradeoff.

web/pages/leaderboards.tsx
```diff
@@ -30,31 +40,68 @@ export default function Leaderboards(props: {
...
+            setPeriod(e.target.value as LeaderboardPeriod)
+          }}
+          value={period}
+          className="!select !select-bordered max-w-xs"
+        >
+          <option value="allTime">All Time</option>
+          <option value="monthly">Monthly</option>
+          <option value="weekly">Weekly</option>
```
jahooma, line 71: Great. Reminder to remove these options before merging. (Maybe add a TODO to add them back in at the relevant time?)

jahooma (changes requested):
Thanks Fede! Caching the relevant portfolio metrics is actually a great approach that will also set you up to do the portfolio graph quite nicely :-D

The code is very clean and clear. Thanks also for starting again with a new branch and new approach to get the leaderboards working! We knew that this task wasn't straightforward when we assigned it to you of course, haha, so kudos for navigating these complexities.

functions/src/create-user.ts
Pico2x: @jahooma Thanks for reviewing. Sounds good, I'll move that over to a subcollection. Great points!

web/pages/leaderboards.tsx
Pico2x: Definitely!

Pico2x:
-I moved the portfolioHistory to a subcollection
- Switched the layout to a Tab view as you suggested. 
-The top creator leaderboards are all shown from the get-go, while  I'm only showing the allTime top Bettors leaderboard for now, and will toggle the rest on after a day/week/month. 

Some pics:
Screenshot 2022-06-18 at 21 45 01
Screenshot 2022-06-18 at 21 46 18

functions/src/update-metrics.ts
```diff
@@ -26,15 +27,25 @@ const computeInvestmentValue = (
...
-  const [users, contracts, bets] = await Promise.all([
+  const [users, contracts, bets, allPortfolioHistories] = await Promise.all([
     getValues<User>(firestore.collection('users')),
     getValues<Contract>(firestore.collection('contracts')),
     getValues<Bet>(firestore.collectionGroup('bets')),
+    getValues<PortfolioMetrics>(
+      firestore
+        .collectionGroup('portfolioHistory')
+        .where('timestamp', '>', Date.now() - 31 * DAY_MS) // so it includes just over a month ago
+    ),
```
jahooma, line 48: A cheaper way to do this would be to do three queries for history entries a day ago, a week ago, a month ago. E.g. filter for before specific times, sort by descending timestamp, and use limit(1).

jahooma (commented):
Looks good, thanks Fede! Just a couple small ideas for you.

functions/src/update-metrics.ts
```diff
@@ -53,32 +64,74 @@ export const updateMetricsCore = async () => {
...
+      user,
+      contractsById,
+      currentBets
+    )
+    const lastPortfolio = last(portfolioHistory)
+    const didProfitChange =
+      lastPortfolio === undefined ||
+      lastPortfolio.balance !== newPortfolio.balance ||
+      lastPortfolio.totalDeposits !== newPortfolio.totalDeposits ||
+      lastPortfolio.investmentValue !== newPortfolio.investmentValue
```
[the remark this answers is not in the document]
Pico2x: It's not - the timestamp will be different.

functions/src/update-metrics.ts
Pico2x: I thought about doing that, but I think it runs into some of the limitations of firestore. 

If you wanted to query the monthly portfolio for example, you'd want to:
 -do a collectionGroup query for 'portfolioHistory'
- where one_week_ago > timestamp > one_month_ago
- orderBy desc
- groupBy userId
- limit(1) within each groupBy group

But firestore doesn't support groupBy at all. 

If you don't add limit(1) it defeats the purpose  of the optimization - you still need to iterate. 

I'd suggest leaving it as is - by doing the timestamp check I'm already doing I ensure that the max number of values you'd have to iterate through per user is around 3k (i.e num of 15min intervals in a month).

jahooma: Ah, yeah that's true! It wouldn't be that simple. You'd have to do those three queries per user. I suppose we can punt on that for now, though I do think loading the portfolio history will be the largest cost here. Firebase charges by document loaded.

Re: iterating in reverse. Reversing modifies the array, so it would be iterating through all the elements. Plus, `[...portfolioHistory]` creates another copy of the array. I'm not as concerned about the performance though.

functions/src/update-metrics.ts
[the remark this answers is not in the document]
jahooma: Ah right, this is good then.

jahooma (approved):
Awesome, I think this is ready to go then! Fix the merge conflicts, and I suppose I'll merge / deploy the cloud functions.

Pico2x:
@jahooma Fixed up the merge conflicts, should be good to go!

jahooma:
Hi Fede, there's still at least one static analysis error, and now merge conflicts after merging #531. Could you fix these? Thanks.

Pico2x:
> https://vercel.com/mantic/prod/96h9R8BRhZe8nvqQehWJQDq1Daca

Will James. Can you give my github account access to the Vercel - I can't see the errors. The errors aren't showing up in my linter config.

jahooma:
Ok cool, I approved dev Vercel access for this PR. Yeah, the previous static error I fixed, but then there was another one.

Pico2x:
Fixed the conflicts. Checked for build errors with yarn build, checked for lint errors with Eslint setup to typescript, and ran in localhost. Let me know if there's any other signal you folks also look for.

jahooma:
Thanks Fede!

jahooma:
Alright! Deployed the relevant functions and merged. It's live!

I had to add a few tweaks in my own fork here: #557 

Some notes
- Firestore rules should be edited in firestore.rules and then also copied over to the console to publish.
- I had to create an index in prod for the `portfolioHistory` collection. Normally this is the responsibility of the person creating the PR, but Fede did not have access to the prod database.
- The `updateMetrics` timed out with 60 seconds limit, so I had to increase the time limit.

Otherwise, it seems to be working!

One thing I noticed as a nice-to-have is if the `getStaticProps` computed all the leaderboards (for all time, monthly, weekly and daily) and passed the data to the client. Then there would be no need for the loading spinner which is a little intrusive.

PR #543, opened 2022-06-18
Clean up unclean user names

mqp:
Supersedes #211.

In addition to the sanitization in #211, I wrote a script to look for existing "unclean" names in the DB. There were two that exceeded our length limit by a couple characters. They were just reasonable but long names, so instead of just rudely truncating them, I decided to just extend the length limit slightly. Now no names in the database do or should in the future violate the sanitization rules.


functions/src/utils.ts
```diff
@@ -36,8 +36,9 @@ export const writeUpdatesAsync = async (
   }
 }
 
-export const isProd =
-  admin.instanceId().app.options.projectId === 'mantic-markets'
+export const isProd = () => {
+  return admin.instanceId().app.options.projectId === 'mantic-markets'
+}
```
mqp, line 41: I did this because, the way it was, if this file was imported and executed in a script before calling `initAdmin`, it would blow up.

PR #545, opened 2022-06-18
Add Statistics link to sidebar

wasabipesto:
There's currently no user-discoverable link to https://manifold.markets/stats (previously https://manifold.markets/analytics).


akrolsmir (approved):
Awesome. I've been wanting to link our stats more widely, thanks for going ahead and doing this!

I do worry that we're getting into "nav item inflation" - at some point we may want to restructure some of these to be findable from http://docs.manifold.markets/ vs the home app - but for now, this is perfect

Issue #524, opened 2022-06-16
API should expose resolutionProbability for binary markets resolved to MKT


wasabipesto:
Currently, binary markets have their `resolution` set to YES, NO, or MKT. If they resolve to MKT, there is no way to know via the API what value they resolved to. 

The value `resolutionProbability` is is in [contract.ts](https://github.com/manifoldmarkets/manifold/blob/ead13b4e08e039637863b5ae9340754433d6218f/common/contract.ts#L73) and I would suggest exposing that to the API, possibly alongside `initialProbability`.

I'll take a swing at this at some point if no one beats me to it.

PR #546, opened 2022-06-18
Add resolutionProbability to LiteMarket

wasabipesto:
Currently, binary markets have their resolution set to YES, NO, or MKT. If they resolve to MKT, there is no way to know via the API what value they resolved to (Issue #524).

This PR exposes `resolutionProbability` to the API via LiteMarket. This is a key step in being able to calculate payouts for markets.


jahooma:
Seems like a good thing to add. Not sure how we're doing versioning.

mqp:
> Seems like a good thing to add. Not sure how we're doing versioning.

We simply think to ourselves, "now it's a new version."

PR #547, opened 2022-06-18
Add Users API endpoint

wasabipesto:
This is a first swing at adding a `users` endpoint to the API. It's read-only, exposing basic account info and cached values.

Basic uses at this stage include:
- Getting current balance for intelligent betting via API
- Getting total market volume for leaderboard placement
- Getting username and avatar for rendering bets

Future plans may include:
- Exposing extended data in FullUser, akin to FullMarket
  - `Markets[]`, a list of all markets a user has created
  - `Comments[]`, a list of all comments a user has made
  - `Bets[]`, a list of all bets a user has made
- List of followers and followed accounts
- Amount received via tips
- Groups(?)

I would love feedback on styling, layout, etc. We may want to move some of this out of `_types` if we plan on adding much more. We may also want pagination here, right now the response on prod is a little over 1MB but it'll grow quickly.

Once we're set on how we want to implement this, I'll also do up some docs.

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -446,8 +446,7 @@ Requires no authorization.
     twitterHandle?: string
     discordHandle?: string
 
-    // Note: the following are fully rederivable from the transaction history.
-    // They are here for convenience only and may be removed in the future.
+    // Note: the following are here for convenience only and may be removed in the future.
     balance: number
     totalDeposits: number
     totalPnLCached: number
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -396,6 +396,66 @@ Requires no authorization.
   ```
 - Response type: A `FullMarket` ; same as above.
 
+### `GET /v0/users`
+
+Lists all users.
+
+Requires no authorization.
+
+- Example request
+  ```
+  https://manifold.markets/api/v0/users
+  ```
+- Example response
+  ```json
+  [
+    {
+      "id":"igi2zGXsfxYPgB0DJTXVJVmwCOr2",
+      "createdTime":1639011767273,
+      "name":"Austin",
+      "username":"Austin",
+      "url":"https://manifold.markets/Austin",
+      "avatarUrl":"https://lh3.googleusercontent.com/a-/AOh14GiZyl1lBehuBMGyJYJhZd-N-mstaUtgE4xdI22lLw=s96-c",
+      "bio":"I build Manifold! Always happy to chat; reach out on Discord or find a time on https://calendly.com/austinchen/manifold!",
+      "bannerUrl":"https://images.unsplash.com/photo-1501523460185-2aa5d2a0f981?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=1531&q=80",
+      "website":"https://blog.austn.io",
+      "twitterHandle":"akrolsmir",
+      "discordHandle":"akrolsmir#4125",
+      "balance":9122.607163564959,
+      "totalDeposits":10339.004780544328,
+      "totalPnLCached":9376.601262721899,
+      "creatorVolumeCached":76078.46984199001
+    }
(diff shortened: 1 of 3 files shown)


ahalekelly:
Great, thank you! Would be nice to also have an endpoint to get a LiteUser via username instead of id, since there's no way to get the user id from the website UI. Similar to the slug endpoint.

mqp:
This makes total sense to me and I would be happy to merge it as is and improve it later, if you want.

mqp:
By the way, someone on Discord mentioned they were working on a GraphQL adapter for reading data, so I'm personally kind of avoiding thinking very hard about how these reading-stuff-from-the-DB APIs should be designed until I can see how that pans out. I don't personally have any implementing GraphQL APIs so I have a hard time judging it from here.

ahalekelly:
Yeah that all makes sense to me.

wasabipesto:
GraphQL would be nice! 

If you're good with this as-is I'll submit the rest in future PRs.

web/pages/api/v0/_types.ts
```diff
@@ -140,3 +141,61 @@ function augmentAnswerWithProbability(
...
+
+  bio?: string
+  bannerUrl?: string
+  website?: string
+  twitterHandle?: string
+  discordHandle?: string
+
+  balance: number
+  totalDeposits: number
+  totalPnLCached: number
```
akrolsmir, line 162: I'm not especially sure how long we want to commit to supporting all these values; `creatorVolumeCached` and `totalPnLCached`, but also `totalDeposits` and maybe even `balance`.

The reason is that these are all denormalized values, basically cached here for convenience; they should be fully rederivable from txn history. If API clients start depending on these, we still want the flexibility to change this in the future.

But I think this is fine to merge as-is, we might just want to make a note in the doc

akrolsmir (approved):
This is great, thanks! Eventually it'd be cool to have all the info that's exposed on eg `https://manifold.markets/Austin` to be viewable on the API (comments, bets, markets, etc). But this is a great first step.

Either here or in a follow-up PR, would you mind documenting this new endpoint on the docs page? Thank you so much~

wasabipesto:
> Also, might want to have some example usages in mind, so we understand how best to structure the data and what to provide. 

I'll take whatever I can get for stats, if it tells a story or explains a concept I'll stick it in the page. 

But in addition to that page I'm kicking around the idea of making some sort of souped-up dashboard. Some data I'd like to show:

- For each user, what amount of their money came from:
  - Real-money deposits?
  - Market creation profits?
  - Liquidity profits?
  - Trade profits?
  - Comment tips?
  - In the future:
    - Market tips?
    - Manalinks?
    - Funds?
- What amount of their money went to:
  - Charity?
  - Fees?
  - Tipping others?
- What amount of their portfolio is currently invested in:
  - Creating markets?
  - Adding liquidity?
  - Trades?
- For each action (betting, creating markets, adding liquidity, investing in funds), what was their return on investment?
- I'd also like to be able to calculate all of these values over time.

I've not looked into the transaction table but if each user's balance can be re-derived from those numbers I'd like to get it to the API somehow.

docs/docs/api.md
```diff
@@ -396,6 +396,66 @@ Requires no authorization.
...
+    url: string // link to user's profile
+    avatarUrl?: string
+
+    bio?: string
+    bannerUrl?: string
+    website?: string
+    twitterHandle?: string
+    discordHandle?: string
+
+    // Note: the following are fully rederivable from the transaction history.
```
akrolsmir, line 449: Would drop this first sentence because it's not true (from an API consumer's perspective, anyways, because they can't read the txn log or stripe payment records yet)

akrolsmir (approved):
> I've not looked into the transaction table but if each user's balance can be re-derived from those numbers I'd like to get it to the API somehow.

Unfortunately, at the moment bets themselves, along with user deposits, are not stored on the txn log; instead, they're captured in various less-easy-to-parse places. I'd love to get all of these consolidated on a single txn system, for sure! Someday...

PR #549, opened 2022-06-19
Updating docs

wasabipesto:
- Added some notes to the API docs
- Copied in new API example responses
- Tweaked general docs & FAQ
- Added free response details page

Future PRs:
- Finish souping up the API docs
- Check all links (some notion pages are dead/require auth)

In the future I would love to find someone new and get their notes/what they have questions on and beef out those sections. May trawl through the FBB notes and pull from that too.


docs/docs/free-response-markets.md
```diff
@@ -0,0 +1,41 @@
...
+- The creator of each market is responsible for resolving each market. They can resolve to any single answer, or even multiple answers.
+  - Resolving to multiple answers allows the creator to choose a percentage for each selected answer (or distribute equally). The payout for any answer is taken from the amount of the total pool allocated to that answer.
+    - For example, let's take a free-response market with many answers. The pool for this market is $500, and you own 100 out of 500 total shares of answer #1.
+    - If the creator resolves to answer #1 only, you will receive `M$500 * (100 / 500) = M$100`.
+    - If the creator resolves 50% to answer #1 and 50% to answer #2, you will receive `(M$500 * 50%) * (100 / 500) = M$50`.
+    - Note that your payout is dependent on the total number of shares, and thus may decrease if more people buy shares in that answer.
+  - Creators can also resolve N/A to cancel all transactions and reverse all transactions made on the market - this includes profits from selling shares.
+
+# Market creation
```
mqp, line 22: It seems clear that rather than repeating this stuff between the "free response" and "binary" market documents, we should reorganize stuff somehow. Same thing for the fee structure below.

mqp:
Thank you, it's impressive how many wrong statements you found. Further work is welcome.

Maybe someone else should take a look at whether the free response market documentation is totally correct. It looks reasonable to me but I don't happen to understand how they work.

docs/docs/free-response-markets.md
wasabipesto: Good idea! Just merged those files and condensed it a tad.

docs/docs/faq.md
```diff
@@ -35,17 +37,17 @@ No. See [here](https://manifold.markets/hamnox/will-manifold-markets-add-nongoo
...
+### The payout probabilities I'm shown sometimes aren't right. For example if a market is at 15% and I bet M$1 on "no", it tells me that I'll make a 42% profit if I win, but the listed payout is just M$1. What's going on?
 
-Payout amounts are visually rounded to the nearest M$ 1, and only integer amounts can be put into markets. Behind the scenes however, your balance does track fractional amounts, so you're making a M$ 0.42 profit on that bet. Once you win another M$ 0.08, that fractional M$ 0.5 will display as an extra M$ 1 in your account. (There's no way to view your exact balance, you can only see the rounded value.)
+Payout amounts are visually rounded to the nearest M$1, and only integer amounts can be put into markets. Behind the scenes however, your balance does track fractional amounts, so you're making a M$0.42 profit on that bet. Once you win another M$0.08, that fractional M$0.5 will display as an extra M$1 in your account. (There's no way to view your exact balance, you can only see the rounded value.)
```
ahalekelly, line 42: I believe payment amounts are always rounded down in the UI

docs/docs/market-details.md
```diff
@@ -0,0 +1,105 @@
...
+    - Note that all current binary markets use this fee schedule.
+    - The post-bet probability is what the market probability would be after your bet if there were no fees.
+    - Example:
+      - If you bet M$100 on NO and the resulting probability without fees would be 10%, then you pay `M$100 * 10% * 10% = M$1.0`.
+      - If you bet M$100 on YES and the resulting probability without fees would be 50%, then you pay `M$100 * 10% * 50% = M$5.0`.
+    - 100% of this fee is used to provide a commission to the market creator, which is paid out after the market is resolved.
+  - Our DPM fee schedule is currently: `5% * (1 - post-bet probability) * bet amount`
+    - Note that all free-response markets use this fee schedule. The calculation for this is the same as above.
+    - 4% is used to provide a commission to the market creator, which is paid out after the market is resolved. 1% is "burnt" to prevent inflation.
+  - No fees are levied on sales.
```
ahalekelly, line 104: If you have existing shares in a binary market and buy shares on the opposite side, that is equivalent to selling your shares and you do not pay fees

docs/docs/market-details.md
```diff
@@ -55,10 +55,10 @@ for the pool to be sorted into.
...
+- Behind the scenes, when a bet is placed the CPMM mechanism does [a bunch of math](http://bit.ly/maniswap). The end result is that for each M$1 bet, 1 YES share and 1 NO share is created. Some amount of shares are then given to the user who made the bet, and the rest are stored in the liquidity pool.
+ Due to this mechansim, the number of YES shares in the whole market always equals the number of NO shares.
+- You can manually add liquidity to any market to increase the incentives for traders to participate. You can think of added liquidity as a subsidy for getting your question answered. You can do this by opening up the market info popup window located in the (...) section of the header on the market page.
   - Adding liquidity provides you with a number of YES and NO shares, which can be withdrawn from the same interface. 
   - These shares resolve to M$ like normal when the market resolves, which will return you some amount of your investment. It is often wise to inject liquidity if you believe the current probability is very close to the real probability.
```
ahalekelly, line 63: > It is often wise to inject liquidity if you believe the current probability is very close to the real probability.

I would remove this and add something like:

> If the market moves significantly in either direction, your liquidity will become significantly less valuable. You are currently very unlikely to make money by investing liquidity in a market, it is a way to subsidize a market and encourage more people to bet, to achieve a more accurate answer. Adding liquidity to a market also makes it require more capital to move the market, so if you want to subsidize a market, first make sure the market price is roughly where you think it should be.

ahalekelly: > Every market starts with M$100 of liquidity. Every day, each user gets 1 free market for which Manifold provides the initial liquidity, otherwise you must invest M$100 for the initial liquidity pool. The initial liquidity provider receives their liquidity back when the market resolves.

wasabipesto: Added the first bit. The free markets are covered elsewhere, I don't want to repeat that in case they get removed soon.

docs/docs/faq.md
wasabipesto: Interesting... this came from the old community FAQ, maybe it's changed since then? Or was it wrong the whole time? Maybe I should just say it rounds and leave it at that.

mqp:
Since this is a big chunk of good stuff I'm going to merge it, feel free to continue fixing more stuff in future PRs.

PR #550, opened 2022-06-19
VSCode: added recommended extensions

TrueMilli:
* prettier
* eslint
* firestore security rules


mqp:
I only want this file if other VS Code users on the team are going to keep it up to date, which by the fact that it doesn't already exist, I am slightly skeptical of. But if they say it's a good file that they like, then OK.

jahooma:
I guess we could try it, since it might be useful for open source contributors

sipec:
I'm in favor of this!

PR #551, opened 2022-06-19
Make type definitions for Txn more precise

mqp:
This is directly analogous to #355. The goal is to express that a `Txn` is either a `DonationTxn` or a `TipTxn` (or later one of the future kinds of transactions), so that type narrowing on transactions works. If we don't write it like this then code that handles generic transactions will become a big mess and won't be very typesafe.


akrolsmir:
Hm, I'm a bit skeptical about this proposal. I think especially the "fromType/toType" might be useful, because we're going to be routing mana (and down the line, YES/NO shares or other fungible (or NFTs)) between different accounts through the txn system.

Maybe the use case would become clearer to me as the number of transaction types increase? Not sure.

jahooma (commented):
Looks reasonable to me, LGTM, pending Austin / Sinclair's approval.

jahooma:
> Hm, I'm a bit skeptical about this proposal. I think especially the "fromType/toType" might be useful, because we're going to be routing mana (and down the line, YES/NO shares or other fungible (or NFTs)) between different accounts through the txn system.
> 
> Maybe the use case would become clearer to me as the number of transaction types increase? Not sure.

The `fromType` and `toType` fields are still required, just moved into Donation and Tip (one of which is required)!

common/txn.ts
```diff
@@ -1,34 +1,31 @@
...
+
+export type Txn<T extends AnyTxnType = AnyTxnType> = {
   id: string
   createdTime: number
 
   fromId: string
-  fromType: SourceType
-
   toId: string
-  toType: SourceType
```
sipec, line 11: Are we ever going to have txns without `fromType` or without `toType?` I think it should be required in every Txn. I suggest we don't delete these lines. Even with these lines, the types should still get narrowed correctly.

image
(In the above example, the last two lines complain that "BANK" is not assignable to "USER" and that c.z has no type overlap with "BANK")

common/txn.ts
```diff
@@ -1,34 +1,31 @@
...
+
   // Human-readable description
   description?: string
-}
-
-export type SourceType = 'USER' | 'CONTRACT' | 'CHARITY' | 'BANK'
+} & T
 
-export type DonationTxn = Omit<Txn, 'data'> & {
+export type Donation = {
```
sipec, line 22: I agree with this!

common/txn.ts
```diff
@@ -1,34 +1,31 @@
 // A txn (pronounced "texan") respresents a payment between two ids on Manifold
 // Shortened from "transaction" to distinguish from Firebase transactions (and save chars)
-export type Txn = {
+export type AnyTxnType = Donation | Tip
```
sipec, line 3: this is also not a useful type to export

common/txn.ts
sipec: An alternative approach: instead of `Txn<T extends AnyTxnType>` do `Txn<T extends { fromId: SourceType,  toId...`

sipec (approved):
I've picked some nits and I think the names of some of these new types is unintuitive, but I broadly agree with this even as is.

mqp:
> Hm, I'm a bit skeptical about this proposal. I think especially the "fromType/toType" might be useful, because we're going to be routing mana (and down the line, YES/NO shares or other fungible (or NFTs)) between different accounts through the txn system.
> 
> Maybe the use case would become clearer to me as the number of transaction types increase? Not sure.

This proposal doesn't change anything about the data. All txns still must have a from and to type.

common/txn.ts
mqp: I just deleted these declarations because they are redundant, but the facts that you are displeased and that Austin instantly misunderstood it seems like good evidence that it's good documentation, so I will put them back.

I guess nobody noticed when I analogously removed the redundant `outcomeType` declaration from `Contract`.

functions/src/transact.ts
```diff
@@ -69,7 +69,7 @@ export const transact = functions
 
       const newTxnDoc = firestore.collection(`txns/`).doc()
 
-      const txn: Txn = removeUndefinedProps({
+      const txn = removeUndefinedProps({
```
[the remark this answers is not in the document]
mqp: I took the type away because the type is wrong. This doc is not known to be of type `Txn` because we don't validate the input data for this cloud function to ensure that it is.

Issue #535, opened 2022-06-17
Feature Request: Sort markets by pool size


ahalekelly:

wasabipesto:
I think this is a great suggestion, and should pretty pretty simple a la https://github.com/manifoldmarkets/manifold/commit/688ebb797a9964fb6523f0ea2a1eaa90dfb9d860

Unfortunately I can't get it to work on my build, I think there are some firestore changes that may need to be added as part of this.

ahalekelly:
Sorry what I meant was sort by the size of the betting pool, not the size of the liquidity pool. ie how much money goes to the winners. Could be called Most Invested or something. Since selling has a negative amount, I think you can just sum the amount of all the bets. Compared to Most Traded which is the sum of the absolute value of all the bets.

Sort by liquidity pool would be cool too though. Except free response markets don't have a liquidity pool?

ahalekelly:
oh pool is an object not a number, so it makes sense you can't just sort by it
https://github.com/manifoldmarkets/manifold/blob/dca957708c03430948d20d67b2de321a390d591b/common/contract.ts#L65

sipec:
added and removed (because with limit orders, pool size is no longer a good measure of liquidity)

PR #553, opened 2022-06-20
API update pool

ahalekelly:
Inspired by @TrueMilli #503

pool: return the pool dict. In the cast of binary markets, this contains the number of Yes and No shares in the liquidity pool, and in the case of free response markets, this contains the total value invested in each of the answers. Previously the pool value returned the sum of the pool shares for binary markets and null for free response markets, so this is a breaking change but provides more info and I might be the only one using this value currently.

totalLiquidity: for binary markets this is unchanged, for free response markets it now returns the total money invested in the market. Used @TrueMilli's function

But it feels like CPMM and DPM markets are fundamentally different enough that I'm not sure it makes sense to combine them like this. DPM markets don't exactly have liquidity like CPMM markets do, I would rather keep totalLiquidity as CPMM only and make a new variable totalValue that is the total amount invested in the market. If we do this, we should make sure it would support #535 as well, which I'm not quite sure how to do yet


ahalekelly:
Reverted the totalLiquidity changes, #554 and #535 made me more convinced on the need for a totalValue variable, I think we should implement that instead and expose that via the API.

ahalekelly:
Also I think the Pool value on the homepage should show totalValue, the amount up for grabs for the winner.

docs/docs/api.md
```diff
@@ -96,7 +96,8 @@ Requires no authorization.
     tags: string[]
     url: string
 
-    pool: number
+    pool: { outcome: number } // For binary markets, the number of shares in the liquidity pool. For free response markets, the amount of mana invested in each answer
+    totalLiquidyt?: number // Binary markets only, the amount of mana deposited into the liquidity pool
```
mqp, line 100: Typo here.

mqp (approved):
Makes total sense to me.

docs/docs/api.md
ahalekelly: Oh thanks

ahalekelly:
Aw shit, I hate git

mqp:
I love git.

PR #555, opened 2022-06-21
Patch infinite money exploit from untipping

sipec:


sipec:
I've tested this and it works. Leaving visual indications for later.

PR #558, opened 2022-06-22
[M4T] Hanania instance

akrolsmir:


mqp:
Hey @akrolsmir I have a question, how do you deploy Firebase stuff to these private instances? Don't you need an entry in, like, `.firebaserc`, so you can do `firebase use hanania` or something?

mqp:
(Also note #562.)

Issue #559, opened 2022-06-22
Feature request: Add an API route to resolve markets


bcongdon:
This was discussed in the Discord, adding an issue here just for tracking

PR #560, opened 2022-06-22
Update resolve-market to be a v2 function

bcongdon:
This is a prerequisite for adding an API for market resolution (#559).

The other passthrough write APIs use V2 functions, so figured I'd update this one first.

The diff is a bit hard to read, but the only thing that really changes is that now this uses `zod` for the body schema

---

Also, I didn't have permissions to create a new Cloud Function in the dev project:

```
Error: Missing required permission on project dev-mantic-markets to deploy new HTTPS functions. The permission cloudfunctions.functions.setIamPolicy is required to deploy the following functions:

- resolvemarket

To address this error, please ask a project Owner to assign your account the "Cloud Functions Admin" role at the following URL:

https://console.cloud.google.com/iam-admin/iam?project=dev-mantic-markets
error Command failed with exit code 1.
```

(I don't necessarily want this permission, `setIamPolicy` is pretty powerful... Hopefully if someone else creates the function, others can update it without needing this permission, but I'm not sure)


mqp:
In general, the additional tasks that will need to be done on top of what's currently present to make the API public:

1. Change the existing error reporting to return real HTTP status codes instead of JSON blobs with`{ status: 'error' }`. (i.e. basically everywhere we `return` JSON except at the end, we should be `throw`ing an `APIError` with a status code.)
2. Make sure we endorse the output as a kind of reasonable output, e.g. it doesn't have meaningless wrappers, etc.
3. Deploy the new function.
4. Change the client code to point at the new function, handle the new error format, and parse the new output. Add docs.
5. Verify.
6. Delete the old function.

> (I don't necessarily want this permission, setIamPolicy is pretty powerful... Hopefully if someone else creates the function, others can update it without needing this permission, but I'm not sure)

Yes, I believe so.

bcongdon:
Thanks for the review, this was helpful.

I updated the PR to start throwing `APIError`s instead of `{ status: 'error' }`. The Cloud Function now returns the updated Contract object upon success. I also updated the web client to start calling the new function (and updated its error handling accordingly). 

I think the API inputs are "OK" as-is. I wrote a [draft doc](https://github.com/manifoldmarkets/manifold/commit/580d4db6f32c8c0cec228510d245c7959d0484da#diff-04b537d3dbb935a7356d57c21696de081f22011a8225eba32e8962364bf61851) for the API, and it felt logical enough. One weird thing is that the "MKT" resolution type in the API is overloaded to support "PROB" and "MULTI" resolutions... Not sure if it's worth promoting those to first-class API resolution types, lmk what you think

Planning on adding the API route and docs in a separate PR.

mqp:
> One weird thing is that the "MKT" resolution type in the API is overloaded to support "PROB" and "MULTI" resolutions... Not sure if it's worth promoting those to first-class API resolution types, lmk what you think

Weird, and I don't know why we designed it that way, but I am fine with punting it and making breaking changes later. I don't think we should block on improving that since we would have to go through the database and fix up all the old ones.

mqp:
Regarding client integration, note #562.

bcongdon:
> Weird, and I don't know why we designed it that way, but I am fine with punting it and making breaking changes later.

SGTM

> Regarding client integration, note https://github.com/manifoldmarkets/manifold/pull/562.

Thanks for the heads up. I'll rebase on top of that once it's merged

bcongdon:
Rebased on #562

functions/src/resolve-market.ts
```diff
@@ -15,156 +15,146 @@ import {
...
+  resolutions: z.map(z.string(), z.number()).optional(),
+})
+
+export const resolvemarket = newEndpoint(['POST'], async (req, auth) => {
+  const { outcome, value, contractId, probabilityInt, resolutions } = validate(
+    bodySchema,
+    req.body
+  )
+  const userId = auth.uid
+  if (!userId) throw new APIError(403, 'Not authorized')
```
mqp, line 34: This is now taken care of by the `newEndpoint` internals and should be deleted.

functions/src/resolve-market.ts
```diff
@@ -15,156 +15,146 @@ import {
 } from '../../common/payouts'
 import { removeUndefinedProps } from '../../common/util/object'
 import { LiquidityProvision } from '../../common/liquidity-provision'
+import { APIError, newEndpoint, validate } from './api'
+
+const bodySchema = z.object({
+  outcome: z.enum(RESOLUTIONS),
```
mqp, line 21: I think this field needs to accept not only things in `RESOLUTIONS` but also an integer index, for free response markets.

functions/src/resolve-market.ts
```diff
@@ -15,156 +15,146 @@ import {
...
+    if (
+      isNaN(+outcome) &&
+      !(outcome === 'MKT' && resolutions) &&
+      outcome !== 'CANCEL'
+    )
+      throw new APIError(400, 'Invalid free response outcome')
+  } else if (outcomeType === 'NUMERIC') {
+    if (isNaN(+outcome) && outcome !== 'CANCEL')
+      throw new APIError(400, 'Invalid numeric outcome')
+  }
```
mqp, line 53: I think it would be better to handle these validations with additional `zod` schemas applied conditional on the `outcomeType`, like in `placebet`.

functions/src/resolve-market.ts
```diff
@@ -15,156 +15,146 @@ import {
...
+
+  if (value !== undefined && !isFinite(value))
+    throw new APIError(400, 'Invalid value')
+
+  if (
+    outcomeType === 'BINARY' &&
+    probabilityInt !== undefined &&
+    (probabilityInt < 0 || probabilityInt > 100 || !isFinite(probabilityInt))
+  )
+    throw new APIError(400, 'Invalid probability')
```
mqp, line 63: The range checks here overlap with the schema validation and should be removed.

functions/src/resolve-market.ts
```diff
@@ -15,156 +15,146 @@ import {
...
+    throw new APIError(403, 'User is not creator of contract')
 
-      if (contract.resolution)
-        return { status: 'error', message: 'Contract already resolved' }
+  if (contract.resolution) throw new APIError(400, 'Contract already resolved')
 
-      const creator = await getUser(creatorId)
-      if (!creator) return { status: 'error', message: 'Creator not found' }
+  const creator = await getUser(creatorId)
+  if (!creator) throw new APIError(400, 'Creator not found')
```
mqp, line 71: This seems like it should be a 500.

functions/src/resolve-market.ts
```diff
@@ -15,156 +15,146 @@ import {
...
-          closeTime: newCloseTime,
-          resolutionProbability,
-          resolutions,
-          collectedFees,
-        })
-      )
+  const { payouts, creatorPayout, liquidityPayouts, collectedFees } =
+    getPayouts(
+      outcome,
+      Object.fromEntries(resolutions || []),
```
mqp, line 98: Kind of awkward to parse into a `Map` and turn it into an object here. Why not have the schema parse it into an object?

functions/src/resolve-market.ts
```diff
@@ -15,156 +15,146 @@ import {
...
+    contract,
+    outcome,
+    resolutionProbability,
+    Object.fromEntries(resolutions || [])
   )
 
+  const updatedContractSnap = await contractDoc.get()
+  if (!updatedContractSnap)
+    throw new APIError(500, 'Updated contract does not exist')
+  return updatedContractSnap.data() as Contract
```
mqp, line 155: Kind of ridiculous to fetch it again from the DB when we just put this data. I would prefer to either return the data as we have it in memory, or just return the ID/slug and let the user look it up if they want the new data.

mqp (changes requested):
If you fix any correctness bugs (e.g. make sure that resolving FR markets works) then I will merge this.

I am OK merging it without docs and documenting it at our leisure. Migrating it to v2 is in itself valuable.

functions/src/resolve-market.ts
bcongdon: I took another pass at this which removed most of these validations

functions/src/resolve-market.ts
```diff
@@ -221,4 +201,34 @@ const sendResolutionEmails = async (
...
+      resolutions: undefined,
+      probabilityInt: undefined,
+    }
+  } else if (outcomeType === 'FREE_RESPONSE') {
+    const { outcome, resolutions } = validate(freeResponseSchema, body)
+    if (!(outcome === 'MKT' && resolutions))
+      throw new APIError(
+        400,
+        'Resolutions cannot be specified with MKT outcome'
+      )
```
mqp, line 217: This isn't right. The corresponding old code was this:
```typescript
 if (isNaN(+outcome) && !(outcome === 'MKT' && resolutions) && outcome !== 'CANCEL')
  // .. error case
```
That is, it's trying to express that either the outcome must be a numeric response index, 'MKT' with multiple resolved numeric indices, or 'CANCEL'. It expressed that extremely badly, but your version expresses it even worse :-)

It would be nice to use some combination of `zod` and manual checks here to actually do better than the old code, e.g. check that the indices correspond to valid answers. But at least it has to accept all legal resolutions.

mqp:
There were a few remaining issues I found during testing, so I made #598 to supersede this.

PR #561, opened 2022-06-23
Port /api/v0/ to graphql on /api/graphql

joy-void-joy:
This PR is a backbone for a GraphQL API which I think Manifold would really benefit using.
For now only markets, slug and market are implemented as proof of concept. Changing what gets exposed and the API is very simple, most of the time only the graphql files will need to change.

To test: ``/api/graphql``

Example query:

```graphql
{
  slug(url: "can-i-please-have-a-free-answer-mar")
  {
    bets
    {
      amount
      createdTime
      id
      isAnte
      isSold
      market
      {
        id
      }
      outcome
      probAfter
      probBefore
      sale
      {
        amount
      }
      shares
      user
      {
        avatarUrl
        id
        name
        username
      }
    }
    closeTime
    comments
    {
      answerOutcome
      answers
      {
        id
        text
      }
      createdTime
      id
      market
      {
        id
      }
      replyTo
      {
        id
      }
      text
      user
      {
        avatarUrl
        id
        name
        username
      }
    }
    createdTime
    creator
    {
      avatarUrl
      id
      name
      username
    }
    description
    id
    isResolved
    mechanism
    outcome
    {
      __typename
      ...on MarketOutcomeBinary
      {
        probability
      }
      ...on MarketOutcomeFreeResponse
      {
        answers
        {
          createdTime
          creator
          {
            avatarUrl
            id
            name
            username
          }
          id
          number
          probability
          text
        }
      }
      ...on MarketOutcomeNumeric
      {
        max
        min
      }
    }
    p
    pool
    question
    resolution
    resolutionTime
    tags
    totalLiquidity
    volume
    volume7Days
    volume24Hours
  }
}
```

Caching is implemented, as well as lazy responses (only what is queried will be fetched from the database). Other benefits include being able to keep track of what field gets used (for deprecation rollout), ease of dynamic use for programmers, and higher versatility compared to a REST api.


mqp:
The Vercel logs say that `yarn build` failed like so:

image

You should try `yarn build` on your box and see if it works.

joy-void-joy:
Thanks, is there a way to run all workflow approval locally? I'm running yarn build and yarn verify before pushing. The typescript error on vercel is due to it not running ``yarn generate`` before typecheck, maybe I should add it to ``yarn verify``?

There is still an error on my end, which is due to database initialization I've missed. I'm trying to solve it now!

joy-void-joy:
Actually, it seems the problem I have is also present on the main branch (it says 
```Error [FirebaseError]: Firebase: No Firebase App '[DEFAULT]' has been created - call Firebase App.initializeApp() (app/no-app).``` ).

joy-void-joy:
Okay, I've added ``yarn generate`` to every scripts, which should fix the pipeline. I can rebase the branch and reopen a PR if you'd like

wasabipesto:
This is amazing! Being able to specify exactly what I need in a query would be so useful, rather than getting a dump of the entire database and parsing out what I want.

mqp:
Just to manage expectations -- I am spinning some other plates right now so it might be a few days before I provide substantial feedback on this.

sipec:
@joy-void-joy sorry for us ghosting you on this. I think it's very unlikely at this point that we'll do anything with this particular branch. There's been so much drift in the api code by now and will be more when we move the whole thing off vercel pretty soon. So unfortunately I'll be closing this.

PR #562, opened 2022-06-23
Clean up definition of v2 cloud function URLs

mqp:
It seems clear how these work now, so let's clean this up so that we don't have to keep adding individual v2 cloud functions here when we create them in the future.

(The `uc` region stands for `us-central1`. The ID as far as I can tell is some meaningless hash.)


jahooma (approved):
Sweet, should make it easier to use V2 functions! Thanks.

PR #564, opened 2022-06-23
Ameliorate homepage search spam

mqp:
The status quo ante: When you load the homepage it does at least three (probably more in some cases) search queries to Algolia before you even touch the UI, each pulling JSON results with thousands of contracts over the wire, and maybe rendering janky search results that disappear afterward.

Two fixes to try to stem the bleeding:

1. Don't treat `undefined` lists of `followedCategories` and `follows` different than empty lists. Previously, it would do a new search every time these came in, even if it was going to be the same search. Now it will only do a new search if you actually have follows or if you filtered the categories. That still kind of sucks but at least we are fixing it for many users.

2. Don't wait a long time to load the saved search sort from local storage. Previously, we would wait until the router is ready to look at both the URL sort/query parameters and also the local storage sort/query from the last usage. This meant that if anyone ever changed it from the default, it was guaranteed to do a search with the default first, and then have to do another search when the saved sort was loaded. Now it loads the saved sort immediately and uses it as the default, so they will only have to do another search if the URL query parameters disagree with the saved sort.

This code still seems kind of unfortunate (it's just incredibly confusing to me, and the design where we do one "default search" that we are going to throw away 500ms later as soon as we load the `follows` etc. from the DB seems not good) but this seems like an 80/20 fix.

### DIFF
--- a/web/components/contract-search.tsx
+++ b/web/components/contract-search.tsx
@@ -116,8 +116,8 @@ export function ContractSearch(props: {
     showCategorySelector,
     mode,
     Object.values(additionalFilter ?? {}).join(','),
-    followedCategories?.join(','),
-    follows?.join(','),
+    (followedCategories ?? []).join(','),
+    (follows ?? []).join(','),
   ])
 
   const indexName = `${indexPrefix}contracts-${sort}`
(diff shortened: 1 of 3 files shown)


jahooma (approved):
Awesome, thanks Marshall!

The Algolia query is paginated so it only loads 20 contracts at a time, but it's still definitely worth cutting the requests down.

And I'm most excited that for the case where I've previously sorted by Newest it might load the results faster because the request starts before the router is ready now!

PR #565, opened 2022-06-23
Show resolution time in market cards when appropriate

bcongdon:
Shows "resolve time" in the market card when the "resolve time" sort option is selected.

---

Before:
Screen Shot 2022-06-23 at 6 39 30 AM

After:

Screen Shot 2022-06-23 at 6 39 13 AM


jahooma (approved):
Awesome, looks good! Thanks Ben

PR #570, opened 2022-06-23
[PortfolioGraph] Shows a graph of the portfolio value over time

Pico2x:
Adds the portfolio graph to the portfolio page. 

[Here's a video of how it looks](
https://user-images.githubusercontent.com/6242616/174615819-af072c60-4fad-4e4b-94d5-5c34e5c5e1a0.mp4
)

Notes:
-We'll need to update the firestore security rules in production 
- I'm starting it off with the period toggle not showing, defaulting to 'allTime'. Once data gets cached we can turn it on.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -32,10 +32,6 @@ service cloud.firestore {
       allow read;
     }
 
-    match /{somePath=**}/portfolioHistory/{portfolioHistoryId} {
-      allow read;
-    }
-
     match /private-users/{userId} {
       allow read: if resource.data.id == request.auth.uid || isAdmin();
       allow update: if (resource.data.id == request.auth.uid || isAdmin())
--- a/firestore.rules
+++ b/firestore.rules
@@ -32,6 +32,10 @@ service cloud.firestore {
       allow read;
     }
 
+    match /{somePath=**}/portfolioHistory/{portfolioHistoryId} {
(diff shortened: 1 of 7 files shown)


Pico2x:
@jahooma Instead for data to merge this, can we ship it asap, and then add the component to user-page again when there's data?

firestore.rules
```diff
@@ -32,6 +32,10 @@ service cloud.firestore {
       allow read;
     }
 
+    match /{somePath=**}/portfolioHistory/{portfolioHistoryId} {
```
jahooma, line 35: I actually added this just above! I had thought it was necessary for the last change (b/c saw it in the dev rules), although now I see that it wasn't

jahooma (approved):
Right, I understand that it takes effort to maintain a branch in sync with main. Happy to merge this in and enable the graph after a couple days.

An alternative would be to just enable all time and daily now, and add the others in later. Up to you!

Pico2x:
@jahooma Thanks for reviewing. I think the graph is kind of lame until there's at least a few days of data in it, I'd merge it as is and enable Daily & AllTime in a few days. I don't have permissions to merge from my account (no 'write access'), so go for it when you have some time :)

jahooma:
Alright, let's merge.

PR #571, opened 2022-06-23
Remove deprecated useUserById implementation

sipec:
It's confusing to have two hooks that do the same thing.
I've tested - and subjectively doesn't seem to change load time of avatars.


jahooma (approved):
Thanks Sinclair!

PR #573, opened 2022-06-23
Comment tips v1.1: Remove undo. Show full tip amount. Scale linear

sipec:
No UI changes. Reverse (-5) arrow is simply disabled except in the 1.5 second window, and you can't reverse lower than your already saved tip.


jahooma (approved):
Great, let's ship it! Thanks, Sinclair!

PR #574, opened 2022-06-23
Add How to page

SirSaltyy:
Removed the about button. Added a "How to" button which links to a new on-site "How to Manifold"  page.

### DIFF
--- a/web/components/nav/sidebar.tsx
+++ b/web/components/nav/sidebar.tsx
@@ -81,7 +81,7 @@ function getMoreNavigation(user?: User | null) {
     { name: 'Discord', href: 'https://discord.gg/eHQBNBqXuh' },
     { name: 'Twitter', href: 'https://twitter.com/ManifoldMarkets' },
     { name: 'Statistics', href: '/stats' },
-    { name: 'About', href: 'https://docs.manifold.markets/$how-to' },
+    { name: 'How to', href: '/how-to' },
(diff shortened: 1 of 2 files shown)


wasabipesto:
If we remove the link to the docs, they are no longer user-discoverable. Additionally, we just duplicated an entire page of content. Are we migrating away from the docs system?

SirSaltyy:
> If we remove the link to the docs, they are no longer user-discoverable. Additionally, we just duplicated an entire page of content. Are we migrating away from the docs system?

Yes the intent is to move away from the docs page eventually. Long term there will be something similar to credits Help Center button. This current PR aims to be an easy, tentative solution to help new users who are trying to figure out what the heck is going on with the site.

Firstly, the docs page takes users to an external site with no way of navigating back easily which is horrible for new users.

Secondly, the docs page right now is pretty bad imo. So until I spend some time to improve it I'm okay with it being somewhat hidden at the bottom of a How to page.

SirSaltyy:
Also, to clarify, we can keep a docs page with stuff like how the API works, AMM's etc.

But things such as a "How to page" or a "Who Manifold is", shouldn't be contained there.

mqp:
OK, if that's your intent, let's merge this and then we can see how people react and argue more about it later if we want.

sipec:
Closing this now that how-to is implemented with notion

PR #578, opened 2022-06-24
Add market liquidity addition events to bets feed

bcongdon:
Example:
Screen Shot 2022-06-24 at 7 22 56 AM

Unfortunately, this only currently works when someone _adds_ liquidity to a market. The liquidity doc has it's "amount" zeroed out upon liquidity withdrawal, so liquidity withdrawals are not shown (the "add" event just disappears)

### DIFF
--- a/web/components/contract/contract-tabs.tsx
+++ b/web/components/contract/contract-tabs.tsx
@@ -8,15 +8,17 @@ import { Spacer } from '../layout/spacer'
 import { Tabs } from '../layout/tabs'
 import { Col } from '../layout/col'
 import { CommentTipMap } from 'web/hooks/use-tip-txns'
+import { LiquidityProvision } from 'common/liquidity-provision'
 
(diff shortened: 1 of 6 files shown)


akrolsmir (approved):
Great, thank you so much! Showing the liquidity is one of those things we've talked about forever but never had time to implement. Everything here looks great~

(Wondering if we want to rename the "bets" tab to "events" or something like that -- but we can figure that out later haha)

akrolsmir:
Oh actually - on the screenshot, it shows the liquidity provisions with a slightly smaller avatar icon than for bets, and also with a short grey line below the icon. Could we just use the same styling as we use for bets?

bcongdon:
Good catch, I didn't notice that. Fixed --
Screen Shot 2022-06-25 at 9 12 48 AM

akrolsmir:
Awesome, thanks!

PR #579, opened 2022-06-24
[Leaderboard] Show daily 'topBettor' leaderboard

Pico2x:
We've been caching values for over a day now, so we can enable the Top Bettors daily leaderboard.


Pico2x:
I can't read the prod failure - unsure if it's an actual error or just about permissions.

PR #580, opened 2022-06-24
[Badges] Shows Badges in User profile 

FRCassarino:
WIP PR for the badges functionality, so people can start pitching in. The code still needs to be cleaned up a bit - I'll be able to continue tomorrow. 

[Video of how it looks like](https://user-images.githubusercontent.com/6242616/175611917-f2bb5555-6940-4d05-abfa-5cc714ca4284.mov)

The design for each badge is a placeholder - does anybody have any opinions on how they should look?

This PR adds badges the user profile section, but the cooler feature will be to later add them anywhere that a user's name is mentioned. I'll add that bit tomorrow as well.

### DIFF
--- a/functions/src/update-metrics.ts
+++ b/functions/src/update-metrics.ts
@@ -133,6 +133,41 @@ export const updateMetricsCore = async () => {
     'set'
   )
   log(`Updated metrics for ${users.length} users.`)
+
+  const badgesUpdates = await userUpdatesWithBadges(users)
+  await writeAsync(firestore, badgesUpdates)
+}
+
+const userUpdatesWithBadges = async (users: User[]) => {
+  const [topTenTraders, topTenCreators] = await Promise.all([
+    getValues<User>(
+      firestore.collection('users').orderBy('profitCached', 'desc').limit(10)
+    ),
+    getValues<User>(
+      firestore
+        .collection('users')
+        .orderBy('creatorVolumeCached', 'desc')
+        .limit(10)
+    ),
+  ])
+
+  const userUpdatesWithBadges = users.map((user) => {
+    const newWeeklyTraderRank =
+      topTenTraders.findIndex((t) => t.id === user.id) + 1
+    const newWeeklyCreatorRank =
+      topTenCreators.findIndex((t) => t.id === user.id) + 1
+    return {
+      doc: firestore.collection('users').doc(user.id),
+      fields: {
+        badges: {
+          weeklyTraderRank: newWeeklyTraderRank,
+          weeklyCreatorRank: newWeeklyCreatorRank,
+        },
+      },
+    }
+  })
+
+  return userUpdatesWithBadges
 }
 
 const computeVolume = (contractBets: Bet[], since: number) => {
--- a/web/components/badges/badge.tsx
+++ b/web/components/badges/badge.tsx
@@ -1,0 +1,22 @@
+import React from 'react'
+import { Col } from '../layout/col'
+import { Row } from '../layout/row'
+
+export function Badge(props: { icon: React.ReactNode; label: string }) {
+  const { icon, label } = props
+  return (
+    <div className="margin-left-5 flex items-center">
+      <span
+        className={`inline-block whitespace-nowrap rounded-full bg-blue-600 py-2.5 px-2.5 text-center align-baseline text-xs font-bold leading-none text-white`}
+      >
+        <Row>
+          <Col>{icon}</Col>
+
+          <Col>
+            <span className="text-sm">{label}</span>
+          </Col>
+        </Row>
+      </span>
+    </div>
+  )
+}
--- a/web/components/badges/badges-dialog.tsx
+++ b/web/components/badges/badges-dialog.tsx
@@ -1,0 +1,85 @@
+import React, { useState } from 'react'
+import { Col } from '../layout/col'
+import { Row } from '../layout/row'
+import {
+  LightningBoltIcon,
+  PresentationChartLineIcon,
+} from '@heroicons/react/solid'
+import { Badge } from './badge'
+import { Modal } from '../layout/modal'
+import { Title } from '../title'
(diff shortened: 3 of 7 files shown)


jahooma:
Looks cool!

web/components/badges/badge.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { Col } from '../layout/col'
+import { Row } from '../layout/row'
+
+export function Badge(props: { icon: React.ReactNode; label: string }) {
+  const { icon, label } = props
+  return (
+    <div className="margin-left-5 flex items-center">
+      <span
+        className={`inline-block whitespace-nowrap rounded-full bg-blue-600 py-2.5 px-2.5 text-center align-baseline text-xs font-bold leading-none text-white`}
```
Pico2x, line 10: creates the placeholder badge shape in a solid blue color

web/components/badges/badges-dialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+      label={'#' + user.badges.weeklyCreatorRank}
+    />
+  )
+
+  const topTraderBadge = () => (
+    <Badge
+      icon={<LightningBoltIcon className="h-5 w-5" />}
+      label={'#' + user.badges.weeklyTraderRank}
+    />
+  )
```
jahooma, line 29: These don't need to be functions

web/components/badges/badges-dialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </Col>
+      </Row>
+      <Modal open={open} setOpen={setOpen}>
+        <Col className="gap-4 rounded bg-white p-6">
+          <Title className="!mt-0 !mb-0" text="Badges" />
+
+          <div>{user.name} has the following badges:</div>
+          {showTopCreatorBadge && (
+            <Row className="justify-start gap-4">
+              <Col>{topCreatorBadge()}</Col>
```
jahooma, line 62: `<Col>` probably unnecessary.

web/components/badges/badge.tsx
jahooma: Sweet! This is a decent first pass on the badge design. I wasn't immediately able to guess which badge was for traders / creators (I had it backwards).

I'm not a designer, but I would remove the `#` and make the rank number larger. Maybe add a border. The blue background is a little strong.

Would be nice to contract out some icons, e.g. https://dribbble.com/shots/16325934-Gamification-Badges but that would come later. For the first version, I wouldn't worry too much :-)

jahooma (commented):
This is neat, thanks for pushing the weekly badge idea forward!

web/components/badges/badge.tsx
Pico2x: I'll remove the #, thanks!

functions/src/update-metrics.ts
```diff
@@ -133,6 +133,41 @@ export const updateMetricsCore = async () => {
...
+    ),
+    getValues<User>(
+      firestore
+        .collection('users')
+        .orderBy('creatorVolumeCached', 'desc')
+        .limit(10)
+    ),
+  ])
+
+  const userUpdatesWithBadges = users.map((user) => {
```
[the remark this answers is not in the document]
Pico2x: This updates every user, so it does reset the ones who lost their rank. IndexOf returns -1 if the user isn't in topTenCreators. This means that it'll store 1-10 for the current top10 creators, and 0 for everybody else. 

I don't love updating every user when it could just be the current top 10 and the previous top 10, but these won't be the only badges, and all of them will involve updating separate sets of users each time the function runs (Ex: markets created streak in days, >100 markets created, top bettor in x community, 1th/10th/100th bet made, etc).

web/components/badges/badges-dialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+          </button>
+        </Col>
+        <Col>
+          <button
+            className="margin-left-5 flex items-center"
+            onClick={() => setOpen(true)}
+          >
+            <Row>{showTopTraderBadge && topTraderBadge()}</Row>
+          </button>
+        </Col>
```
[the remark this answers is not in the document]
Pico2x: Its just to help with the styling.

web/components/badges/badges-dialog.tsx
Pico2x: If we end up adding dozens of different types of badges, it's nicer to have them as functions rather than cluttering the main return statement. But can rm for now.

web/components/badges/badges-dialog.tsx
Pico2x: Its just for styling - gets them side by side, with the icon and text centered in the y-axis. I'm new to tailwinds, is there another way of doing this?

Pico2x:
@jahooma So I wanted to get the badges to show up in next to the username in every comment/market etc. But I'm running into smth that I'd like to understand better. 

In the comment or contract firebase documents, you've denormalized some of the user documents fields into them: userName, userUsername, userAvatarUrl. Whenever a user changes any of these, you update every one of their coments. 

I'm planning to break from this design pattern when it comes to badges. Badges can change every 15 minutes, so adding a denormalized "badges" field to each comment, contract, etc is not a good idea, as we'd be constantly updating them. Instead, I'd add a creatorUserId field, and query for the badges each time we display a users name. 

Does this sound reasonable, or do you have a better idea? We could instead denormalize the badge field, and have badges only update once a day for example. But it changes the feature a bit, and would make it out of sync with the Leaderboards.

jahooma:
@Pico2x Yeah, makes sense! No need to embed badges in all those objects. We can store them in the user object, and then fetch the user in these cases.

There should be no need to add a `creatorUserId` field, all these objects (comments, contracts, etc) should already have the relevant user id stored.

sipec:
something like this may be a good idea, and it's come up in discussion a lot, but at this point we should probably start over

Issue #533, opened 2022-06-17
Feature Request: Don't allow people to create duplicate answers in free response questions


ahalekelly:
Duplicate answers create a lot of headache, basically every market has to state "in case of duplicates i will resolve to whoever was first"
This change wouldn't stop people from duplicating with a slight change, but it would help

akrolsmir:
Yeah, this ought to be fairly trivial (could just be a client-side fix). Would be a good first-issue!

PR #581, opened 2022-06-24
Prevent duplicate Free Response answers

bcongdon:
Addresses #533

This PR prevents duplicate Free Response answers, and also adds a warning when creating an answer that is similar to an existing one. I'm not sure of Manifold's philosophy on adding dependencies, so please lmk if the added dependency on `string-similarity` is a dealbreaker

---

Example duplicate prevention:
Screen Shot 2022-06-24 at 10 22 34 AM

Example similar answer warning:
Screen Shot 2022-06-24 at 10 22 29 AM

### DIFF
--- a/web/components/answers/create-answer-panel.tsx
+++ b/web/components/answers/create-answer-panel.tsx
@@ -64,13 +64,15 @@ export function CreateAnswerPanel(props: { contract: FreeResponseContract }) {
 
   const changeAnswer = (text: string) => {
     setText(text)
-    const answerAlreadyExists = answers.find(
+    const existingAnswer = answers.find(
       (a) => lowerCase(a.text) === lowerCase(text)
     )
 
-    if (answerAlreadyExists) {
+    if (existingAnswer) {
       setAnswerError(
-        answerAlreadyExists ? `An answer for "${text}" already exists` : ''
+        existingAnswer
+          ? `"${existingAnswer.text}" already exists as an answer`
+          : ''
       )
       return
     } else {
@@ -127,7 +129,7 @@ export function CreateAnswerPanel(props: { contract: FreeResponseContract }) {
           <AnswerError
             key={2}
             level="warning"
-            text={`Did you mean "${possibleDuplicateAnswer}"?`}
+            text={`Did you mean to bet on "${possibleDuplicateAnswer}"?`}
           />
         ) : undefined}
         <div />
@@ -214,11 +216,10 @@ type answerErrorLevel = 'warning' | 'error'
 const AnswerError = (props: { text: string; level: answerErrorLevel }) => {
   const { text, level } = props
   const colorClass =
-    level === 'error'
-      ? 'text-red-500'
-      : level === 'warning'
-      ? 'text-orange-500'
-      : ''
+    {
+      error: 'text-red-500',
+      warning: 'text-orange-500',
+    }[level] ?? ''
   return (
     <div
       className={`${colorClass} mb-2 mr-auto self-center whitespace-nowrap text-xs font-medium tracking-wide`}
--- a/web/components/answers/create-answer-panel.tsx
+++ b/web/components/answers/create-answer-panel.tsx
@@ -1,6 +1,7 @@
 import clsx from 'clsx'
 import { useState } from 'react'
 import Textarea from 'react-expanding-textarea'
+import { findBestMatch } from 'string-similarity'
 
 import { FreeResponseContract } from 'common/contract'
 import { BuyAmountInput } from '../amount-input'
@@ -23,6 +24,7 @@ import { firebaseLogin } from 'web/lib/firebase/users'
 import { Bet } from 'common/bet'
 import { MAX_ANSWER_LENGTH } from 'common/answer'
 import { withTracking } from 'web/lib/service/analytics'
+import { lowerCase } from 'lodash'
 
 export function CreateAnswerPanel(props: { contract: FreeResponseContract }) {
   const { contract } = props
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -8026,11 +8031,6 @@ nanoid@^3.1.23, nanoid@^3.1.30, nanoid@^3.3.4:
   resolved "https://registry.yarnpkg.com/nanoid/-/nanoid-3.3.4.tgz#730b67e3cd09e2deacf03c027c81c9d9dbc5e8ab"
   integrity sha512-MqBkQh/OHTS2egovRtLk45wEyNXwF+cokD+1YPf9u5VfJiRdAiRwB2froX5Co9Rh20xs4siNPm8naNotSD6RBw==
 
-nanoid@^3.3.4:
-  version "3.3.4"
-  resolved "https://registry.yarnpkg.com/nanoid/-/nanoid-3.3.4.tgz#730b67e3cd09e2deacf03c027c81c9d9dbc5e8ab"
-  integrity sha512-MqBkQh/OHTS2egovRtLk45wEyNXwF+cokD+1YPf9u5VfJiRdAiRwB2froX5Co9Rh20xs4siNPm8naNotSD6RBw==
-
```
mqp, line 8033: ?

mqp:
My opinion on the dependency: I personally would not consider taking a dependency on this. `string-similarity` and `findBestMatch` are terrible names that obscure what it is doing. It looks like it's computing [this metric](https://en.wikipedia.org/wiki/S%C3%B8rensen%E2%80%93Dice_coefficient) (CTRL-F for "string similarity") which is trivial to compute ourselves in the same fashion. I would write that function and call it `bigramSimilarity` and link to the Wikipedia article, and then people reading the code would know what the hell it was. However, the social norm for this team is clearly to put in the dependency and never think about it again, so feel free to do that.

My opinion on the feature: I would not add this feature. It seems like you are observing a general UX problem (it's hard to see at a glance what answers already exist, so you don't know whether what you are thinking of has already been expressed, and if you create an overlapping answer then you can't undo or delete what you did) and applying a hack that solves like 25% of the problem in a very specific way. I would try to attack the general UX problem and only apply the hack if I ran out of ideas and it was still a problem. For example, I might change the UI so that the answers didn't take up so much space on the screen, and I might make it so that you could delete answers with no bets attached, or make it so that the question creator could merge two answers.

akrolsmir:
Hm, on one hand, I think Marshall's right that making FR answers more readable, and manageable by the author could be a good way of addressing the same problems that.

On the other... I do think this is a nice 80/20 solution. Those other solutions sound a lot more annoying, and it's nice to just ship small incremental improvements quickly; don't let perfection be the enemy of good enough!

String similarity checking is a cool idea -- in the original issue I imagined it would just be exact string match haha. I'm not yet sure whether the string similarity checking thing is useful, or if it's overkill - leaning towards trying the former, and taking it out if there are too many false positives.

Don't feel strongly about the addition of a library for this; additional deps are a bit scary, but this one could be useful; eg what would be cool is to also use this library to identify whether a new question you're about to ask has already been asked -- but def out of scope for this PR.

yarn.lock
akrolsmir: Look like it's a duplicate of the above nanoid reference, and thus maybe safe to remove?

web/components/answers/create-answer-panel.tsx
```diff
@@ -163,3 +208,22 @@ export function CreateAnswerPanel(props: { contract: FreeResponseContract }) {
     </Col>
   )
 }
+
+type answerErrorLevel = 'warning' | 'error'
+
+const AnswerError = (props: { text: string; level: answerErrorLevel }) => {
+  const { text, level } = props
+  const colorClass =
```
akrolsmir, line 216: There's a code structure I like to make this kind of thing more concise & readable:
```
const colorClass= {
  error: 'text-red-500',
  warning: 'text-orange-500',
}[level] ?? ''
```

web/components/answers/create-answer-panel.tsx
```diff
@@ -79,12 +115,21 @@ export function CreateAnswerPanel(props: { contract: FreeResponseContract }) {
...
           rows={1}
           maxLength={MAX_ANSWER_LENGTH}
         />
+        {answerError ? (
+          <AnswerError key={1} level="error" text={answerError} />
+        ) : possibleDuplicateAnswer ? (
+          <AnswerError
+            key={2}
+            level="warning"
+            text={`Did you mean "${possibleDuplicateAnswer}"?`}
```
akrolsmir, line 130: Maybe change to "Did you mean to bet on ..."

web/components/answers/create-answer-panel.tsx
```diff
@@ -54,6 +62,34 @@ export function CreateAnswerPanel(props: { contract: FreeResponseContract }) {
...
 
+  const changeAnswer = (text: string) => {
+    setText(text)
+    const answerAlreadyExists = answers.find(
+      (a) => lowerCase(a.text) === lowerCase(text)
+    )
+
+    if (answerAlreadyExists) {
+      setAnswerError(
+        answerAlreadyExists ? `An answer for "${text}" already exists` : ''
```
akrolsmir, line 73: Change text to "${text} already exists as an answer"

akrolsmir (approved):
LGTM, thanks again~

web/components/answers/create-answer-panel.tsx
bcongdon: Neat, I haven't seen that pattern before, but I like it

web/components/answers/create-answer-panel.tsx
bcongdon: Done

yarn.lock
bcongdon: Yeah, this was an automatic change by `yarn`. `nanoid@^3.3.4` is still present on line 8029 in the updated lockfile

bcongdon:
> For example, I might change the UI so that the answers didn't take up so much space on the screen, and I might make it so that you could delete answers that you made with no bets attached, or make it so that the question creator could merge two answers.

Yeah, I think that's a better solution long term. I think some other UX improvements like having the ability to lock FR answer submission to just the market creator would help too.

bcongdon (commented):
Updated screenshots:
Screen Shot 2022-06-25 at 9 21 35 AM
Screen Shot 2022-06-25 at 9 21 40 AM

akrolsmir:
Lgtm, pending ESLint!

bcongdon:
Looks like `main`'s ESLint is broken -- e.g. https://github.com/manifoldmarkets/manifold/runs/7051942840?check_suite_focus=true

PatMyron:
could use something similar for groups:

Screen Shot 2023-04-02 at 9 50 39 PM

PR #583, opened 2022-06-26
Add Firebase schema collection helpers (kind of an RFC)

mqp:
Once I learned that you could correctly type Firestore collections to represent the data inside their documents and it would work with all the rest of the Firestore querying machinery, it made me want to more visibly expose and use the main collections in our database. How does this look as a way to slightly clean up the status quo?

Probably there are other parts of the code that could benefit from referencing the collections explicitly, especially now that they are correctly typed (e.g. there isn't a big incentive to use the `getValue` and `getValues` helpers anymore) but if we like the general outline, we can get around to those whenever.


mqp:
@akrolsmir @jahooma any feedback on this? The tl;dr is to define the main database collections with appropriate document types in `web/lib/firebase/schema.ts`, and then use them to make all the code digging stuff up out of those collections slightly cleaner.

web/lib/firebase/schema.ts
```diff
@@ -0,0 +1,19 @@
+import { db } from './init'
```
jahooma, line 1: Mixed feelings on a schema file, which would be another file to update instead of just the one for the collection.

I kind of understand the benefit of being able to see all the top-level collections at once. But also, this currently doesn't include the sub collections which makes it a bit misleading.

jahooma (commented):
Cool! On the whole, I like this way of typing collections and then importing them. Less sure about he schema file.

web/lib/firebase/schema.ts
mqp: I think including the subcollections would be totally reasonable, but I am not 100% confident whether it's better to expose them as a Firebase `collectionGroup`, a function from `parentId` to `collection`, or something else, so I decided I would just get your guys' feedback on this version first.

mqp:
> Cool! On the whole, I like this way of typing collections and then importing them. Less sure about he schema file.

So you would prefer that we just ditch the schema file and put the collection definitions into the individual Firebase lib helper files? That seems totally reasonable to me, I like having the schema file but I will not say your way is worse. I will make that change in lieu of further feedback.

mqp:
By the way, I will just say the obvious, which is that it's too bad that the Firebase admin SDK and Firebase client SDK are two different things, so that it's hard to define Firebase helper code that both the frontend and backend can use.

jahooma (approved):
LGTM — Thanks for the collection type cleanup!

PR #584, opened 2022-06-26
Compute stats in Firebase instead of Vercel

mqp:
This is going to be substantially more efficient because of the proximity to the database, and it means our stats computation function won't get killed by Vercel for being too slow and crappy, so now our stats page will actually work on prod.


mqp:
@jahooma I'm just gonna deploy this so people stop complaining about the stats page not working, but if you have any feedback or disagreement that it's a good change I am happy to hear it whenever and we can go from there.

jahooma:
Wow, thanks Marshall!

PR #586, opened 2022-06-27
Typescript project references take 2

mqp:
OK, here are the changes from #575, except not the ones for the web project, so maybe if the problem was just with `tsc --watch` there will be no problem anymore. Unless @jahooma can tell me more information, I am just going to make PRs with less and less stuff until he doesn't revert them.


jahooma:
Ok great, this worked!

PR #589, opened 2022-06-28
Reorganize verify scripts

ForrestWeiswolf:
On #378, @mqp suggested
> I think it would be better if the verify scripts in each subdirectory did all the verify work for that subproject, and then in the root directory it recursed on all the subdirectories. To me that would be the ideal version. But I like this better than the status quo. Maybe someone can improve it later.

This is my attempt to have it work like that, while preventing the possibility of running `verify` in a subdirectory and thinking you'd run it for the whole project.

Also updates the one for `functions` to match the change to the github workflow made in #586.

### DIFF
--- a/common/package.json
+++ b/common/package.json
@@ -3,7 +3,8 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "verify": "(cd .. && yarn verify)"
+    "verify": "(cd .. && yarn verify)",
+    "verify:dir": "npx eslint . --max-warnings 0"
   },
   "sideEffects": false,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


mqp:
I definitely like this better than the status quo, so cool.

FYI I personally don't use these scripts at all because I expect my editor to do 100% of the "flagging down lint errors for me" work... I cared much more about the developer UX of this kind of thing 10 years ago, when my editor was less good at it. So to me, these scripts are basically just documentation. Which is to say, that's also why I "broke" the one script the other day. We should probably at least make the GitHub actions workflow actually use these scripts, so the "documentation" doesn't get out of touch with reality any more.

PR #590, opened 2022-06-28
Clean up emails.ts

ForrestWeiswolf:


mqp:
Sorry, I don't think this change is good. The original code is bad. It's checking things that should be impossible (e.g. "the user that just resolved a market doesn't exist; the user that just resolved a market doesn't have an email") and then if it finds them, it fails silently instead of producing an error that anyone can see.

By abstracting out these bad checks, you make it less likely for someone to notice that they are bad and fix them in the future. Bad code should look bad.

PR #593, opened 2022-06-29
Initial draft of Vercel Firebase auth

mqp:
I want to continue exploring options with this, but I wanted to put this out so that anyone who was curious could start getting an idea for what I was up to.

Basically, here is what this PR does:

1. When you log in to a Manifold instance serving a given Firebase project, it stores two cookies: one with your ephemeral Firebase ID token (containing a UID), and one with your long-lived Firebase refresh token (a token that you can exchange to Google for a new ID token.)

2. If you access a page that e.g. is only for authenticated users, `getServerSideProps` on that page can check those cookies. If you have a valid ID token cookie, it immediately knows your authenticated UID. If not, then if you have a valid refresh token cookie, it can do one round trip to Google to get a new ID token, which it will set as the new cookie, and it knows your authenticated UID. If you have neither cookie, you are not logged in.

3. Based on your UID, we can do whatever, e.g. serve a redirect.

This is already enough to be a solid improvement over the status quo, because the grossest problems are just when we are loading a page and on the server we literally don't know whether someone is logged in or not, so we have to guess. Now we know whether they are logged in, typically without a network round trip.

I am interested in doing better than this. If we create a cloud function that lets you exchange a valid ID token for a long-lived Firebase custom token, and we put that in a cookie, we can authenticate the Firebase client SDK on the Vercel server. Then, not only can we know whether you are logged in, we should be able to server-side render full pages that depend on your user content, like your user portfolio page, or whatever, without changing any of the rest of our code.

Another thing that should be investigated is whether we should put the cached user info that we are currently putting into local storage into a cookie. If we did that, we could use that cached information on the server for initial rendering, without taking the time to look it up from Firebase, even though it wouldn't be authenticated or up to date.


mqp:
@akrolsmir You mentioned you wanted to look at this -- just reminding you that it exists.

web/lib/util/cookie.ts
```diff
@@ -0,0 +1,26 @@
+type CookieOptions = string[][]
```
jahooma, line 1: Nice utility file!

web/lib/firebase/server-auth.ts
```diff
@@ -0,0 +1,63 @@
+import * as admin from 'firebase-admin'
+import fetch from 'node-fetch'
+import { IncomingMessage, ServerResponse } from 'http'
+import { FIREBASE_CONFIG, PROJECT_ID } from 'common/envs/constants'
+import { getAuthCookies, setAuthCookies } from './auth'
+
+const ensureApp = async () => {
+  // Note: firebase-admin can only be imported from a server context,
+  // because it relies on Node standard library dependencies.
```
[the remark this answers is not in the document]
jahooma: Oh, I guess it's only referenced within `/web` in `getServerSideProps`. Carry on.

web/pages/index.tsx
```diff
@@ -22,14 +34,10 @@ export async function getStaticProps() {
...
-    revalidate: 60, // regenerate after a minute
-  }
+  return { props: { hotContracts } }
 }
 
-const Home = (props: { hotContracts: Contract[] }) => {
+export default function Home(props: { hotContracts: Contract[] }) {
   const { hotContracts } = props
 
   const user = useUser()
```
jahooma, line 43: Probably we can remove the below `if (user) { Router.replace('/home') ... }`

jahooma (approved):
This looks great! Pretty clear how it works since you already explained `id` vs `refresh` tokens to me.

Shouldn't this already enable full server side rendering without creating a custom firebase token that is longer life? The custom token seems like an optimization.

Is the plan to continue to use different routes for authenticated server side rendered pages vs. non-authenticated pages?

Excited to see what code can be cleaned up with server side rendering, and what blinks and jumps can be fixed from the users' perspective!

mqp:
> This looks great! Pretty clear how it works since you already explained `id` vs `refresh` tokens to me.
> 
> Shouldn't this already enable full server side rendering without creating a custom firebase token that is longer life? The custom token seems like an optimization.
> 
> Is the plan to continue to use different routes for authenticated server side rendered pages vs. non-authenticated pages?
> 
> Excited to see what code can be cleaned up with server side rendering, and what blinks and jumps can be fixed from the users' perspective!

Well, we have some different options we can explore from here.

As context, here are some facts about the tokens, just to be super clear, because it's confusing to keep them straight:

- An "ID token" is a short-lived JWT signed by Firebase servers attesting to a specific UID.
- A "refresh token" is a blob associated in Firebase with a specific user that you can submit to Firebase servers to get an ID token. You get a refresh token as the output of the OAuth flow. The client SDK stores an ID token and refresh token for the logged in user in memory.
- A "custom token" is a custom JWT signed by Firebase servers attesting to a specific UID with arbitrary other claims. We can use the Firebase admin SDK to produce one. You can also submit a custom token back to the Firebase servers to get an ID token and a refresh token.

Now here are some facts about the software:

- Our cloud functions can just accept any kind of authentication we program them to. The `onCall` ones by default accept an ID token, but we can just do whatever we please.
- However, Firestore itself specifically needs to accept an ID token. We have no control over its auth code. So unless you have an ID token, you can't do any of the private stuff in the security rules.
- The client SDK is designed to log you in either with OAuth, or with a custom token. It then retrieves an ID token and refresh token, and uses the ID token to talk to Firestore. In principle, you could initialize it with your own refresh token and ID token, that you saved. In practice, this would require us to fork the SDK, so it would be a PITA to maintain.
- The admin SDK is designed to log you in either with a service account key. It might be possible to use it with a refresh token, I am not sure.

So let's consider the options available to use during Vercel SSR. With this pull request, it is now easy for SSR to have our ID token and refresh token. So then in order to interact with Firestore:

1. We can do direct HTTP REST calls using our ID token, using no library. Of course this is ugly code.
2. We can try to auth the admin SDK using the user's refresh token. Then we could easily write good looking server-side code to fetch data in `getServerSideProps`, but we couldn't run the React code using the client Firebase SDK.
3. We can make a cloud function that gives you a custom token, given a refresh token, and use that custom token to log into the client Firebase SDK, on the server. Then all the React code using the client Firebase SDK could run on the server too.

Number 3 would be the cleanest solution from the perspective of a guy writing React components -- the server-side rendering would Just Work and you would never have to think about it. So that might be what we want to aim at, unless there is some roadblock in our way.

mqp:
Some things I'm thinking about right now before merging this:

1. How comfortable am I with the robustness of "in some cases, you now can't access any authorized pages unless Vercel can talk to Google to get a new ID token for your refresh token?" Can we do something to improve this?
2. How comfortable am I with the fact that after merging, signed-in users will get a "redirect to logged out" once before they run the client code that gives them a new cookie? Can we do something to improve this?
3. Do I understand the client behavior enough to be sure that our auth code is correct now that it's running on `onIdTokenChanged` instead of `onAuthStateChanged`?

mqp:
OK, I'm gonna say I think this works and we like it and merge it.

Issue #473, opened 2022-06-10
Have 1.9 shares, can't sell 1 share


whosyourjay:
<img width="762" alt="Screenshot 2022-06-09 at 21 13 29" src="https://user-images.githubusercontent.com/4008952/172970942-6fa8efb1-550c-4649-99cc-bd2fd4660e66.png">

akrolsmir:
Another report by Jack: 
![image](https://user-images.githubusercontent.com/856034/176711134-2dd2eff0-4a2d-4992-a5bc-dd348f6b9a98.png)

PR #599, opened 2022-06-30
Allow people to sell all their shares

mqp:
This code has been here since selling shares was implemented, but it makes zero sense to me. Maybe @jahooma can explain how it would have gotten here.


akrolsmir:
Hm, I don't have context on why we added the epsilon=0.0000001; my guess is @mantikoros knows?

In any case, I don't think this is the root cause of the bug we're seeing in #473 , because there the share amounts seem to be much larger than the epsilon value, either 0.9+ or 0.4+ in the two examples found.

mqp:
> Hm, I don't have context on why we added the epsilon=0.0000001; my guess is @mantikoros knows?
> 
> In any case, I don't think this is the root cause of the bug we're seeing in #473 , because there the share amounts seem to be much larger than the epsilon value, either 0.9+ or 0.4+ in the two examples found.

What you didn't know is this code, which secretly rounds up the amount submitted to the API: https://github.com/manifoldmarkets/manifold/blob/a5a0a1370a23399446eeecf4d16233a94243e013/web/components/bet-panel.tsx#L416

These two bugs work together to make this outcome.

jahooma:
The extra `0.000000001` on max shares would only make this check less strict, not more strict.

I don't see what the bug is yet. It seems like the client and server disagree about how many shares you have.

Even if you enter `1` share while having 1.9xxx shares, the client should be sending a sell request for 1.9xxx shares, which you own.

Could be something about how your shares are stored to localstorage? Maybe epsilon should be bigger?? 😛

mqp:
> The extra `0.000000001` on max shares would only make this check less strict, not more strict.
> 
> I don't see what the bug is yet. It seems like the client and server disagree about how many shares you have.
> 
> Even if you enter `1` share while having 1.9xxx shares, the client should be sending a sell request for 1.9xxx shares, which you own.
> 
> Could be something about how your shares are stored to localstorage? Maybe epsilon should be bigger?? stuck_out_tongue

Oh, wow, I look dumb now. You're right, the check is not causing the problem, although I still want to delete the check. I don't know where the error is getting introduced.

Since in general, we make no effort to try to make the rest of our code avoid or quantify floating-point error, and if we fix wherever it's getting introduced it will just get introduced again, my suggestion would be to make the API specifically take "all" or "Infinity" as a possible amount of shares to sell.

jahooma:
I like the idea to pass a special value for all. Let's make it the string `'ALL'`. I think it's slight more clear than using `Infinity` which could be a programmer error.

mqp:
> P.S. I'm the one who designed it this way, but I like how if you type the floor of your number of shares, it sells them all. Seems like a small sin in exchange for a simpler UI.

To me I think that if we want users to care about their M$ balance, it's super important not to betray their trust in the UI and the market mechanism, or else they will just think "ok, the website just has bugs everywhere, so it's not serious." If they see a number, the number should be actually true.

jahooma:
Not showing all the decimal places would always be lying then! This UI doesn't inform you about your last < 1 share because it figures you don't care that much

mqp:
> Not showing all the decimal places would always be lying then! This UI doesn't inform you about your last < 1 share because it figures you don't care that much

Yep, I would definitely make sure all the decimal places were visible, e.g. by my lights a reasonable implementation would be if you can set desired number of sig figs in your profile and then maybe if you tap a displayed quantity you can see the exact quantity. Or alternatively, have them specify a number of shares to buy in the first place, rather than a number of M$, so that everyone isn't walking around with overly-specific numbers of fractional shares in everything all day as a matter of course.

PR #600, opened 2022-06-30
Market Resolution API

### LINKED ISSUE #559: Feature request: Add an API route to resolve markets

bcongdon:
Resolves #559

Notes:
- I added additional validation to `resolve-market` to prevent some invalid states, like resolving to a FR outcome that doesn't exist, or resolving `MULTI` with percentages that don't add up to 100%.
- Currently, we're using `contractId` for market resolution. Eventually, we should move to `slug`.

### DIFF
--- a/functions/src/resolve-market.ts
+++ b/functions/src/resolve-market.ts
@@ -1,14 +1,6 @@
 import * as admin from 'firebase-admin'
 import { z } from 'zod'
-import {
-  difference,
-  uniq,
-  mapValues,
-  groupBy,
-  sumBy,
-  sum,
-  isNumber,
-} from 'lodash'
+import { difference, uniq, mapValues, groupBy, sumBy } from 'lodash'
 
 import {
   Contract,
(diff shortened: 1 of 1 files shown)


functions/src/resolve-market.ts
```diff
@@ -74,6 +78,20 @@ export const resolvemarket = newEndpoint(opts, async (req, auth) => {
   const creator = await getUser(creatorId)
   if (!creator) throw new APIError(500, 'Creator not found')
 
+  if (contract.outcomeType === 'FREE_RESPONSE') {
+    if (resolutions) {
+      Object.keys(resolutions).forEach((outcome) =>
+        validateFreeResponseOutcome(contract, outcome)
+      )
+      const pctSum = sum(Object.values(resolutions))
+      if (pctSum !== 100) {
```
mqp, line 87: This is going to be incorrect due to floating point error.

functions/src/resolve-market.ts
```diff
@@ -74,6 +78,20 @@ export const resolvemarket = newEndpoint(opts, async (req, auth) => {
...
+        throw new APIError(400, 'Resolution percentages must sum to 100')
+      }
+    } else if (!isNaN(+outcome)) {
```
mqp, line 90: `z.number` already excludes `NaN`.

web/lib/api/proxy.ts
```diff
@@ -41,7 +41,12 @@ export const fetchBackend = (req: NextApiRequest, name: string) => {
     'Origin',
   ])
   const hasBody = req.method != 'HEAD' && req.method != 'GET'
-  const opts = { headers, method: req.method, body: hasBody ? req : undefined }
+  const body = req.body ? JSON.stringify(req.body) : req
```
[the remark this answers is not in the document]
mqp: I see, I hadn't thought about this.

mqp (commented):
Seems highly reasonable, I will test, make any necessary changes, and deploy.

functions/src/resolve-market.ts
```diff
@@ -74,6 +78,20 @@ export const resolvemarket = newEndpoint(opts, async (req, auth) => {
   const creator = await getUser(creatorId)
   if (!creator) throw new APIError(500, 'Creator not found')
 
+  if (contract.outcomeType === 'FREE_RESPONSE') {
```
mqp, line 81: Wouldn't it make more sense for this whole block to be in `getResolutionParams`?

functions/src/resolve-market.ts
bcongdon: Fixed, using a tolerance of 0.1 now.

functions/src/resolve-market.ts
bcongdon: This was intended to check "is outcome a number". Switched to using lodash's `isNumber`.

mqp:
I just went ahead and refactored it so that I liked it better. Basically, try to take advantage of Typescript more -- if you correctly use type narrowing then you shouldn't have to be doing stuff like checking `'resolutions` in freeResponseParams` all over the place.

Anyway this seems to work great so I am merging and deploying.

PR #601, opened 2022-06-30
Remove daily free market text from docs

bcongdon:


akrolsmir:
Thanks for catching this!

PR #603, opened 2022-06-30
[Fix] Clear suggested FR answer after submission

bcongdon:
Fixes a small bug with #581 where if you submit a FR answer that had a potential duplicate, the potential duplicate message is not cleared upon submission.


mqp:
Makes sense to me.

PR #605, opened 2022-07-01
Fix backup DB job to actually backup most things, refactor

mqp:
Pretty self-explanatory. I haven't tested yet whether there is a better way to do this export that saves us writing the collections here. I am skeptical of the comment that says the obvious thing that's supposed to work won't work.

### DIFF
--- a/functions/src/backup-db.ts
+++ b/functions/src/backup-db.ts
@@ -18,46 +18,52 @@
 
 import * as functions from 'firebase-functions'
 import * as firestore from '@google-cloud/firestore'
-const client = new firestore.v1.FirestoreAdminClient()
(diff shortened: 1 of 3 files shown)


functions/src/backup-db.ts
```diff
@@ -18,46 +18,62 @@
...
+  const collectionIds = [
+    'contracts',
+    'groups',
+    'private-users',
+    'stripe-transactions',
+    'transactions',
+    'users',
+    'bets',
+    'comments',
+    'followers',
```
jahooma, line 43: Add `follows` which is a subcollection of users. (Whereas `followers` is a subcollection of folds).

jahooma (approved):
Sweet, thanks for updating!

PR #606, opened 2022-07-01
Add a missing index

mqp:
It's missing, we need it, therefore, add it.

### DIFF
--- a/firestore.indexes.json
+++ b/firestore.indexes.json
@@ -337,6 +337,20 @@
           "order": "DESCENDING"
         }
       ]
+    },
+    {
+      "collectionGroup": "portfolioHistory",
+      "queryScope": "COLLECTION",
+      "fields": [
+        {
+          "fieldPath": "userId",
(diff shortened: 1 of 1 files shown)


jahooma (commented):
I didn't realize this file existed. How do we know it has all the current indexes?

Do we have a plan to deploy indexes exclusively from this file?

akrolsmir:
> I didn't realize this file existed. How do we know it has all the current indexes?
> 
> Do we have a plan to deploy indexes exclusively from this file?

I've been using this file as a way to repeatedly deploy indexes to custom instances.

Ideally, this file (and also firestore.rules) would be the source of truth for our Firestore config, and we wouldn't edit things via CLI or web GUI. It would be a bit of a pain to set up probably, but until then this represents a risk of random things failing via missing index or incorrect rules (plus no version control to represent changes in state)

mqp:
Haha I guess I know why the file keeps not having indexes now!

PR #607, opened 2022-07-01
Remove code for obsolete feed updater backend jobs

mqp:
We can also remove the machinery for batching up the work and calling sub-cloud functions now.


functions/src/fetch.ts
```diff
@@ -1,9 +0,0 @@
-let fetchRequest: typeof fetch
```
jahooma, line 1: This file seems useful to have? Is there a better alternative to node-fetch

functions/src/call-cloud-function.ts
```diff
@@ -1,17 +0,0 @@
-import * as admin from 'firebase-admin'
```
jahooma, line 1: I guess we can delete this, since there's probably a better alternative to calling cloud functions from cloud functions (i.e. factor our a common function and call it in both, or increase memory and timeout limits)

jahooma (approved):
Thanks, Marshall!

functions/src/fetch.ts
mqp: No, `node-fetch` is fine if we actually need it, but right now we don't, so we definitely don't need "a funny shim for using either the built in thing we don't need, or the replacement package that we don't need."

mqp:
Sorry, I guess it was too late at night, so I left out two of the things I meant to delete, the `update-feed` implementation and the old `keep-alive` warmup stuff that is already turned off, but now they are gone too.

jahooma (approved):

PR #609, opened 2022-07-01
Pseudo numeric market

mantikoros:


web/pages/create.tsx
```diff
@@ -129,6 +133,18 @@ export function NewContract(props: {
 
   const min = minString ? parseFloat(minString) : undefined
   const max = maxString ? parseFloat(maxString) : undefined
+  const initialValue = initialValueString
+    ? parseFloat(initialValueString)
+    : undefined
+
+  const adjustIsLog = () => {
+    if (min === undefined || max === undefined) return
+    const lengthDiff = Math.abs(Math.log10(max) - Math.log10(min))
```
jahooma, line 142: You should use `log10(max - min)` instead. It recommends everything as log scale if min = 0

jahooma (approved):
LGTM! Reasonable, clean implementation. Log scale is awesome. Our users will love this!

- Having both `resolutionValue` and `resolutionProbability` as fields is not ideal.
- Creators may want a warning when about to resolve to a number outside min and max.

- Might want to change font color from blue to green on numeric cards.

- Need to fix client-side errors, which are crashing the dev site.
Screen Shot 2022-07-02 at 1 55 00 PM
Screen Shot 2022-07-02 at 2 05 13 PM
- Graph not shaded all the way?
Screen Shot 2022-07-02 at 2 07 07 PM
- Should Higher and Lower buttons be swapped?
Screen Shot 2022-07-02 at 1 58 40 PM

PR #612, opened 2022-07-02
Use `getAll` Firestore technology to improve some code

mqp:
I just didn't know about this when I last touched this code, but now I do. My guess is we can expect this to be slightly quicker than parallel `Promise.all` requests mostly because it's going to be as slow as the latency of the one network request it does, rather than as slow as the max latency of any of the network requests it does.

Kind of weird that the client side Firebase SDK straight up doesn't expose this, but that's Firebase life for you.


jahooma (approved):
Cool stuff!

PR #613, opened 2022-07-02
Add `functions-framework` as explicit dependency

mqp:
[This is the runtime library](https://github.com/GoogleCloudPlatform/functions-framework-nodejs) that actually runs our functions and listens for requests. For some bizarre reason, if you don't include it in your dependencies, Google will download it for you in the build anyway. However, that's slower than having it in the dependencies, and it's also weird, because obviously we want to control what version it is, have it share dependencies with other stuff like every other Javascript library, etc. So let's not do that anymore.

We will also probably use it explicitly later for local development purposes.

### DIFF
--- a/functions/package.json
+++ b/functions/package.json
@@ -23,6 +23,7 @@
   "main": "functions/src/index.js",
   "dependencies": {
     "@amplitude/node": "1.10.0",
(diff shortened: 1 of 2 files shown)


jahooma (approved):
Sounds good!

PR #614, opened 2022-07-03
Redemption refactoring

mqp:
Making some changes here so that it's easier to understand this code and move things around further.

Right now, it still does exactly the same thing it used to do, just tidier looking and easier to inspect. But after this, I'm probably going to do another change where I merge this work into the "main transaction" for the various functions that use it, like `placebet`. That will probably save about four round trips that won't be necessary anymore (one for starting the transaction, two for looking up contract and user data, and one for committing the transaction), so 100-150ms per call. So that's where this is going.


mantikoros (approved):
LGTM. Nice refactoring.

PR #616, opened 2022-07-03
Toggle weekly leaderboard and daily/weekly/alltime portfolio graph

Pico2x:
Toggles on:
- All portfolio graph views except for the monthly one.
- The weekly leaderboard for "Top bettors".

Something to note, that I'm not sure if it's clear to you guys @jahooma: The portfolio graph will be public, i.e every user will be able to see any other user's portfolio over time. If that's *not* the intended functionality, please request a review and I'll hide it for all profiles except the one who's currently signed in.

### DIFF
--- a/web/components/portfolio/portfolio-value-graph.tsx
+++ b/web/components/portfolio/portfolio-value-graph.tsx
@@ -52,7 +52,7 @@ export const PortfolioValueGraph = memo(function PortfolioValueGraph(props: {
         margin={{ top: 20, right: 28, bottom: 22, left: 60 }}
         xScale={{
           type: 'time',
-          min: points[0].x,
+          min: points[0]?.x,
           max: endDate,
         }}
         yScale={{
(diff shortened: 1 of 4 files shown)


jahooma (approved):
Great! Yes, it is intended to be able to see others' portfolio over time.

I'll merge this in then!

jahooma:
Some post-merge thoughts!

- Weekly and Daily might be better named Week and Day, because they only show one week and day respectively.
- We definitely want an option to display your profit over time. Perhaps a toggle could switch between portfolio value and profit?
- It would be nice to show the current date in the hover tooltip (like was added to our other charts)

Pico2x:
> Some post-merge thoughts!
> 
> * Weekly and Daily might be better named Week and Day, because they only show one week and day respectively.
> * We definitely want an option to display your profit over time. Perhaps a toggle could switch between portfolio value and profit?
> * It would be nice to show the current date in the hover tooltip (like was added to our other charts)

Thanks for merging! I'll make those changes and send another PR

PR #620, opened 2022-07-05
Rich content

sipec:


mqp:
Just a general impression...out of the "increasingly complicated and featureful ways to edit descriptions":

1. Let people type plain text
2. Let people type some subset of HTML
3. Let people type something like Markdown, a very simple DSL that compiles to HTML
4. Let people edit a complicated DSL (ProseMirror) using a sophisticated UI, and render the AST for that into HTML
5. Number 4 except I guess there is another abstraction layer on top that I am not clear what it is for (Tiptap)

It's kind of weird to me that we would jump all the way from 1 to 5. Wouldn't it seem more practical to implement something more like number 2 or 3 and see if there was any demand for adding further complexity on top of that? What are ProseMirror and TipTap adding that we really wanted? If we just wanted a WYSIWYG UI, we could have made a much simpler one based on Markdown.

I guess you probably thought about this more than me but that's my opinion.

mqp:
(Another concern that's slightly related -- it's good for the API for creating markets and retrieving markets to be easy to use. Having a substantially complicated DSL to represent rich text descriptions is a big problem for that.)

jahooma:
I agree with Marshall. One improvement could be to try to save the plain text version as well in another field.

I tried out the editor, and I know it's in progress, but there's no floating format buttons, so it's hard for users to discover that you can do rich text. You have to just try adding the asterisks for bold and hyphens for lists.

I also got stuck in bold text mode without any idea of how to exit it.

Markdown does seem like a more obvious solution that solves some of these problems. I think Sinclair wanted to be able to eventually embed markets and images (that may also be possible with markdown?).

The other factor here is that I don't see very much value in rich text as a whole, so maybe @akrolsmir should chime in here. The task for rich text is currently of priority P3. Switching to a complex format for rich text that requires an npm dependency to parse doesn't seem like a great idea to me.

sipec:
@mqp regarding 4 vs 5 (TipTap) - I just added a custom paste handler and imo TipTap does a pretty good job of exposing prosemirror's APIs. I wouldn't consider it a significant cost.

Of course, what it gains us is that it works out of the box and the extensions support a lot of common use cases

sipec:
- I think going from plain text to side-by-side preview is a step backwards in terms of complexity to users, I insist on WYSIWYG being the best experience.
- I'm 80% sure TipTap/Prosemirror is the best WYSIWYG tool for our use case. Maybe Slate also would be fine? All the other editors I looked at were worse in terms of ease of extending
- I'm amenable to storing the data in a different format. If we keep using TipTap, some possibilities:
  - store pure html as a text string
  - convert to markdown on save (client-side?), and have md as the source of truth. GitLab does this.
  - store plaintext in separate field

akrolsmir:
Really quickly:

Agree with James that rich text and most other formatting options are something like P2/P3 in terms of immediate feature impact. But I'd think of this work to move to a rich text editor as somewhat foundational/infrastructure. Things this makes much easier:
- Embedding images
- @mention support (for users, AND for markets ala Notion)
- Manifold Stories

The better our rich text & connectivity, the less we can use Notion, and the more we can actually rely on Manifold dogfood for real use cases.

I agree with Sinclair in that WYSIWYG editing is a much better end-user experience, and the right choice here. Think about the number of people who compose in Word/Google Docs/Notion/Substack daily, vs.... I don't even know what the popular markdown composition tool is. Github _barely_ gets by because literally every user is a dev. We don't want to only target devs.

Do agree re: having a simple format for the API is good for devs (and for us, internally!). Compile-down-to-markdown is a promising idea IMO. Or just plaintext! My main question is how this scales up - once we start customizing with @mentions for users and for markets, can we still export that in a sensible way?

In any case, I wouldn't block on any of the above points. (I might also suggest making this easier to add and revert by having a separate `descriptionJSON` field rather than overwriting the old `description`, but I'll leave that for y'all to figure out)

jahooma:
Sweet! I guess I agree after all. WYSIWYG is a better experience than markdown (but we might want the toolbar of text format controls). And if this enables adding photos and other embeds, then that's great!

web/components/editor.tsx
```diff
@@ -0,0 +1,93 @@
...
+    Promise.all(files.map((file) => uploadImage('default', file)))
+  )
+
+  const editor = useEditor({
+    editorProps: {
+      attributes: {
+        class: clsx(proseClass, rowsClass, 'textarea textarea-bordered'),
+      },
+      handlePaste(view, event) {
+        const files = Array.from(event.clipboardData?.files ?? []).filter(
```
jahooma, line 37: Consider renaming `files` to `imageFiles`

web/package.json
```diff
@@ -24,6 +24,11 @@
     "@nivo/core": "0.74.0",
     "@nivo/line": "0.74.0",
     "@react-query-firebase/firestore": "0.4.2",
+    "@tiptap/extension-character-count": "^2.0.0-beta.31",
+    "@tiptap/extension-image": "^2.0.0-beta.30",
+    "@tiptap/extension-placeholder": "^2.0.0-beta.53",
+    "@tiptap/react": "^2.0.0-beta.114",
+    "@tiptap/starter-kit": "^2.0.0-beta.190",
```
jahooma, line 31: Ditto.

web/pages/create.tsx
```diff
@@ -185,6 +183,22 @@ export function NewContract(props: {
...
+
+  const { editor, upload } = useTextEditor({
+    rows: 3,
+    max: MAX_DESCRIPTION_LENGTH,
+    placeholder: descriptionPlaceholder,
+    disabled: isSubmitting,
+  })
+
+  useWarnUnsavedChanges(
+    !isSubmitting && (Boolean(question) || (editor != null && !editor.isEmpty))
```
jahooma, line 199: You could break out a local variable `const editorHasText = editor != null && !editor.isEmpty` or something

web/pages/create.tsx
```diff
@@ -225,15 +239,11 @@ export function NewContract(props: {
 
       await router.push(contractPath(result as Contract))
     } catch (e) {
-      console.log('error creating contract', e)
+      console.error('error creating contract', e, (e as any).details)
+      setIsSubmitting(false)
```
jahooma, line 243: Thanks for this!

jahooma (approved):
This is really cool! The code is pretty clean and minimal. Glad that all the text formatting is offloaded to a library.

This PR seems about ready to ship, but I also have some further thoughts, some of which could be addressed in a followup.

- I think I'm persuaded that storing just the rich text as json is fine. But, we might want to offer a plain text translation in the market API as well, if it's not that hard.
- The description input on the create page no longer has a placeholder label, which is not ideal.
- When I tried to drag an image file into the editor, it failed (this works in the github inputs). Is there a way to fix this for the drag and drop use case? (Pasting the image works.)
- In general, what's missing from this feature is any indication to the user that rich content is available. It's not obvious how to inform users, but I would favor something like a toolbar of buttons, including an upload image button. A less intrusive option could be something like the footer within Github:
Screen Shot 2022-07-07 at 5 36 55 PM
Maybe it could say rich text is supported and how to attach images.

mqp:
> I agree with Sinclair in that WYSIWYG editing is a much better end-user experience, and the right choice here. Think about the number of people who compose in Word/Google Docs/Notion/Substack daily, vs.... I don't even know what the popular markdown composition tool is.

Aren't the really obvious two Slack and Discord? Their entry isn't literally Markdown but it's 75% of Markdown, and it demonstrates that normal people understand the idea of including markup in text. 

I think an editor box with a simple markup language and, optionally, buttons that insert the markup language for you, is the obvious solution and it's much better than if you leave out the "simple markup language" part. If you make the simple markup language and you just want to spend more complexity budget than I think is reasonable on the WYSIWYG editor, then OK, I can understand that, maybe we have some understandable differing opinion about user behavior. But if you just make all descriptions the equivalent of a Word file, with WYSIWYG buttons the only way to insert formatting and with the descriptions compiled into a non-human-readable format, I think you are making a big objective mistake.

I'll happily bet you a hundred bucks that if you ask on our Discord whether people would prefer a Markdown + helper buttons editor or a WYSIWYG-only editor they will clearly tell you they prefer Markdown, if that convinces you.

akrolsmir:
@mqp no bet, because our Discord users aren't a particularly good representation of the audience we're going after. But I would bet $100 that after Manifold users use this tiptap WYSIWYG editor, they'll like it more than the equivalent Markdown editor that e.g. Github has.

Have you, like, actually tried Tiptap? Doing markdown-like things in Tiptap Just Works; hashtags turn into headings; dashes into bullet points; numbering, etc etc. 

Discord uses a markdown thing, but Slack actually turns its bullets and numbers and bold into rich text, WYSIWYG.

web/pages/create.tsx
```diff
@@ -194,9 +183,8 @@ export function NewContract(props: {
     disabled: isSubmitting,
   })
 
-  useWarnUnsavedChanges(
-    !isSubmitting && (Boolean(question) || (editor != null && !editor.isEmpty))
-  )
+  const isEditorFilled = editor != null && !editor.isEmpty
+  useWarnUnsavedChanges(!isSubmitting && (Boolean(question) || isEditorFilled))
```
sipec, line 187: oops, meant for this to be in the commit after

sipec:
I think I'll do the API stuff in the next PR. I have to update the docs too.

akrolsmir:
One thing that is missing from TipTap relative to Notion, though, is the Markdown formatting after copying a block. Copying the text in TipTap would ideally put a markdown-formatted string onto your clipboard, but Tiptap seems to strip that away. (This behavior is consistent with Slack, fwiw)

common/util/parse.ts
```diff
@@ -27,3 +47,31 @@ export function parseWordsAsTags(text: string) {
...
+  Italic,
+  ListItem,
+  OrderedList,
+  Paragraph,
+  Strike,
+  Text,
+
+  Image,
+]
+// export const exhibitExts = [StarterKit as unknown as Extension, Image]
```
jahooma, line 73: Remove this line?

common/util/parse.ts
```diff
@@ -27,3 +47,31 @@ export function parseWordsAsTags(text: string) {
     .join(' ')
   return parseTags(taggedText)
 }
+
+// can't just do [StarterKit, Image...] because it doesn't work with some imports
+// see
```
jahooma, line 52: Remove this line?

web/components/contract/contract-description.tsx
```diff
@@ -21,22 +22,33 @@ export function ContractDescription(props: {
   const descriptionTimestamp = () => `${dayjs().format('MMM D, h:mma')}: `
   const isAdmin = useAdmin()
 
+  const desc = contract.description ?? ''
+
   // Append the new description (after a newline)
   async function saveDescription(newText: string) {
-    const newDescription = `${contract.description}\n\n${newText}`.trim()
+    console.log(desc, exhibitExts)
```
jahooma, line 29: Probably should remove console.logs before merging

web/components/file-button.tsx
```diff
@@ -0,0 +1,26 @@
+import { useRef } from 'react'
+
+/** button that opens file upload window */
+export function FileButton(props: {
```
jahooma, line 4: nit: might be slightly clearer as `FileUploadButton`

web/components/editor.tsx
```diff
@@ -0,0 +1,146 @@
...
+  return (
+    <>
+      {/* hide placeholder when focused */}
+      <div className="[&:focus-within_p.is-empty]:before:content-none w-full">
+        {editor && (
+          <FloatingMenu
+            editor={editor}
+            className="w-full text-sm text-slate-300"
+          >
+            Type <em>*anything*</em> or even paste or{' '}
```
jahooma, line 96: What about:

`Type *emphasis* or **bold**. Or upload an image`

common/util/parse.ts
```diff
@@ -1,4 +1,24 @@
...
+import { Heading } from '@tiptap/extension-heading'
+import { History } from '@tiptap/extension-history'
+import { HorizontalRule } from '@tiptap/extension-horizontal-rule'
+import { Italic } from '@tiptap/extension-italic'
+import { ListItem } from '@tiptap/extension-list-item'
+import { OrderedList } from '@tiptap/extension-ordered-list'
+import { Paragraph } from '@tiptap/extension-paragraph'
+import { Strike } from '@tiptap/extension-strike'
+import { Text } from '@tiptap/extension-text'
+// import * as StarterKit from '@tiptap/starter-kit'
```
jahooma, line 20: Remove?

jahooma (approved):
Looks good! Thanks Sinclair!

I guess that last thing was to make sure Algolia search works over the rich text description?

You might want to store the plaintext description as well for the non-Algolia search. I.e. We need that case to work as well because it's used for the Atlas instances and for the tournament I have with Hanania.

PR #623, opened 2022-07-06
Show tip notifications

IanPhilips:
Screen Shot 2022-07-06 at 10 12 24 AM
Screen Shot 2022-07-06 at 10 12 16 AM

### DIFF
--- a/web/pages/notifications.tsx
+++ b/web/pages/notifications.tsx
@@ -357,6 +357,21 @@ function IncomeNotificationItem(props: {
     setNotificationsAsSeen([notification])
   }, [notification])
 
+  function getReasonForShowingIncomeNotification(simple: boolean) {
+    const { sourceText } = notification
+    let reasonText = ''
+    if (sourceType === 'bonus' && sourceText) {
+      reasonText = !simple
+        ? `for ${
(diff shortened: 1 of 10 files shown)


web/pages/notifications.tsx
```diff
@@ -1017,6 +1112,9 @@ function getReasonForShowingNotification(
           : ' for unique bettors'
       else reasonText = 'You earned your daily manna'
       break
+    case 'tip':
```
jahooma, line 1115: One day it would be nice to organize notification info so that all of one notification's params are specified together, rather than having functions like this that specify one aspect of all notifications.

jahooma (approved):
LGTM! Thanks, this will make tips way more useful and fun.

The notification code does seem a little unwieldy at the moment. One day we could try to refactor into a one file per notification type that specifies everything, plus some reusable notification components.

web/pages/notifications.tsx
IanPhilips: They all look so similar that this has worked well. I just moved the income reason notification code to the income notification though to cut down on growth in this section.

IanPhilips:
Removed extraneous question name and condensed income more
Screen Shot 2022-07-06 at 1 24 39 PM


Screen Shot 2022-07-06 at 1 24 49 PM

PR #625, opened 2022-07-07
Use react-query to cache notifications

IanPhilips:
Speeds up the notifications page quite a bit.


IanPhilips:
I gave the question in notifications two lines. Can't figure out how to stick the timestamp on the end such that the question truncates when it runs into the timestamp though so I removed the timestamp on smaller screens. 
Screen Shot 2022-07-07 at 11 21 45 AM

PR #627, opened 2022-07-07
Migrate addLiquidity and withdrawLiquidity functions to v2

mqp:
Straightforward, should be no visible changes.

(Background: I want to migrate our remaining v1 onCall functions to v2 because there is no reason not to and because it will be easier to then do further general backend changes, like deploying them in a single express app and making hot reloading of server code work.)


mantikoros (approved):
LGTM

PR #629, opened 2022-07-07
Enable tipping on group chats w/ notif

IanPhilips:
Screen Shot 2022-07-07 at 5 06 28 PM
Screen Shot 2022-07-07 at 5 06 46 PM


IanPhilips:
I also created a new function in `create-notification` specifically for tips to avoid adding more bulk to the current notification creation logic.

PR #641, opened 2022-07-12
Categories to groups

IanPhilips:
- Categories are now default groups that people are added to when they sign up
- These new default groups created by Manifold Markets don't have chat walls: those are for user-created groups only
- Current users are added to these default groups unless they unchecked any categories when those were available previously, in this case they are added only to the cateogires they were following
- Given we don't have free markets anymore I hypothesize we don't need to filter out personal and fun markets when a user signs up/or is signed out. 
  - Addendum: signed out users should see markets related to the market they landed on via similar bettors' betting history
- The personal filter filters for your followers' markets and markets in the groups you're in
Screen Shot 2022-07-12 at 11 42 50 AM
Screen Shot 2022-07-12 at 11 45 09 AM


IanPhilips:
By @mantikoros 's request I added a uniqueBettors attribute to `contract.ts` and taking @mqp 's advice, I now calculate this in the trigger `on-create-bet.ts`. Now new users will see the most popular markets by unique bettors. This isn't backfilled however so it would bias towards currently active, popular markets
Screen Shot 2022-07-12 at 4 40 41 PM

IanPhilips:
Perhaps I could rename the index to most popular?

jahooma:
+1 for Most popular

jahooma:
One problem with unique bettors is that it becomes static very quickly. Something like unique bettors in the last week might be better, as a suggestion.

IanPhilips:
Oh interesting idea! Hm, should we leave it and see how stale it gets or just change it now?

jahooma:
Oh, we could leave it for now I guess. Might have a different idea in a week or two

functions/src/on-create-bet.ts
```diff
@@ -26,8 +39,99 @@ export const onCreateBet = functions.firestore
...
+  if (!contract) {
+    log(`Could not find contract ${contractId}`)
+    return
+  }
+  const contractBets = (
+    await firestore
+      .collection(`contracts/${contractId}/bets`)
+      .where('userId', '!=', contract.creatorId)
+      .get()
+  ).docs.map((doc) => doc.data() as Bet)
```
jahooma, line 63: Seems like this won't scale well. Fetching all bets (except for the creator) on every bet is n^2.

Can't you just check if the new bet is a new unique bettor compared to the saved uniqueBettors field? If the uniqueBettors field doesn't exist then you could fetch all the bets.

IanPhilips:
Oh schniz that would actually be great so I could add an additional sort
for the number of followed user ids in the unique bettors array!

IanPhilips:
Okay so now the personal filter will show:
- contracts in groups that the user is a member of
- contracts created by users that the user follows
- contracts bet on by users that the user follows
- contracts bet on by the user

IanPhilips:
What should the personal filter be named?
- Recommended
- Interesting
- Personalized
- Something else?

jahooma:
Not sure. Maybe For you?

PR #646, opened 2022-07-13
Add typing, pasting links

sipec:


sipec:
fat fingered the close button lol

PR #647, opened 2022-07-14
minor tweaks of manalink form

ingawei:
- changing expire time to dropdown instead of calendar selection
- adding 'M$' in front of amount
Screen Shot 2022-07-14 at 2 27 22 AM
Screen Shot 2022-07-14 at 2 27 34 AM

### DIFF
--- a/web/components/manalinks/create-links-button.tsx
+++ b/web/components/manalinks/create-links-button.tsx
@@ -91,7 +91,7 @@ function CreateManalinkForm(props: {
     setNewManalink((m) => {
       return {
         ...m,
-        expiresTime: expiresTime,
+        expiresTime,
       }
     })
   }
--- a/web/components/manalinks/create-links-button.tsx
+++ b/web/components/manalinks/create-links-button.tsx
@@ -118,7 +118,7 @@ function CreateManalinkForm(props: {
                 <input
                   className="input input-bordered w-full pl-10"
                   type="number"
-                  min="0"
+                  min="1"
                   value={newManalink.amount}
                   onChange={(e) =>
                     setNewManalink((m) => {
@@ -134,6 +134,7 @@ function CreateManalinkForm(props: {
                 <input
                   className="input input-bordered"
                   type="number"
+                  min="1"
                   value={newManalink.maxUses ?? ''}
                   onChange={(e) =>
                     setNewManalink((m) => {
--- a/web/components/manalinks/create-links-button.tsx
+++ b/web/components/manalinks/create-links-button.tsx
@@ -63,6 +63,7 @@ function CreateManalinkForm(props: {
   const [finishedCreating, setFinishedCreating] = useState(false)
   const [copyPressed, setCopyPressed] = useState(false)
   setTimeout(() => setCopyPressed(false), 300)
+  const [expiresIn, setExpiresIn] = useState('Never')
 
   const [newManalink, setNewManalink] = useState<ManalinkInfo>({
     expiresTime: null,
@@ -71,6 +72,33 @@ function CreateManalinkForm(props: {
     uses: 0,
     message: '',
   })
+
+  const expireOptions = {
+    day: 'day',
+    week: 'week',
+    month: 'month',
+    never: 'never',
+  }
+
+  function getExpireTime(timeDelta: string) {
+    if (timeDelta in expireOptions && timeDelta != 'never') {
+      return dayjs().add(1, timeDelta)
+    } else {
+      return null
+    }
+  }
+
+  function setExpireTime(timeDelta: string) {
+    const expireTime = getExpireTime(timeDelta)
+    setNewManalink((m) => {
+      return {
+        ...m,
+        expiresTime: expireTime
+          ? dayjs(expireTime, 'YYYY-MM-DDTHH:mm').valueOf()
+          : null,
+      }
+    })
+  }
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


web/components/manalinks/create-links-button.tsx
```diff
@@ -72,6 +73,33 @@ function CreateManalinkForm(props: {
     message: '',
   })
 
+  const expireOptions = {
```
akrolsmir, line 76: Hm, it seems kind of weird to have expireOptions just mapping strings to themselves. Some designs I would prefer

1. Use a type for this: `type ExpiresIn = 'day' | 'week' | 'month' | 'never'`
2. Map
   ```
const EXPIRE_DURATIONS = {
  day: '1 day'
  ...
}
   ```
and then use the map to create the option values in the JSX in place of`<option value={expireOptions.never}>Never</option>`
3. Both 1 & 2!

web/components/manalinks/create-links-button.tsx
```diff
@@ -72,6 +73,33 @@ function CreateManalinkForm(props: {
...
+    day: 'day',
+    week: 'week',
+    month: 'month',
+    never: 'never',
+  }
+
+  function getExpireTime(timeDelta: string) {
```
akrolsmir, line 83: You could make this more concise with something like:
`return timeDelta === 'never' ? null : dayjs().add(1, timeDelta)`

akrolsmir: But actually - since it seems like setNewManalink is going to want a time in Ms anyways, we might as well express it in a time in ms; then we can simplify the below code to
```
const expiresTime = timeDelta === 'never' ? null : dayjs().add(1, timeDelta).valueOf()
setNewManalink((m) => { ...m, expiresTime})
```

web/components/manalinks/create-links-button.tsx
```diff
@@ -112,29 +145,22 @@ function CreateManalinkForm(props: {
...
-                        expiresTime: e.target.value
-                          ? dayjs(e.target.value, 'YYYY-MM-DDTHH:mm').valueOf()
-                          : null,
-                      }
-                    })
+                    setExpiresIn(e.target.value)
+                    setExpireTime(e.target.value)
                   }}
-                ></input>
+                  defaultValue={expireOptions.never}
```
akrolsmir, line 157: Let's actually make it expire in a week by default!

web/components/manalinks/create-links-button.tsx
```diff
@@ -87,22 +115,27 @@ function CreateManalinkForm(props: {
...
             <div className="flex flex-col gap-2 md:flex-row">
-              <div className="form-control">
+              <div className="form-control w-full md:w-1/2">
                 <label className="label">Uses</label>
                 <input
-                  className="input input-bordered w-full"
+                  className="input input-bordered"
                   type="number"
                   value={newManalink.maxUses ?? ''}
                   onChange={(e) =>
```
akrolsmir, line 141: Hm, we probably shouldn't be able to set negative maxUses: 
image

web/components/manalinks/create-links-button.tsx
```diff
@@ -87,22 +115,27 @@ function CreateManalinkForm(props: {
...
-                type="number"
-                value={newManalink.amount}
-                onChange={(e) =>
-                  setNewManalink((m) => {
-                    return { ...m, amount: parseInt(e.target.value) }
-                  })
-                }
-              ></input>
+              <div className="relative">
+                <span className="absolute mx-3 mt-3.5 text-sm text-gray-400">
```
akrolsmir, line 119: I _think_ I like the way this is all inline: 
image

But also it looks just a little bit odd to me being grayed out like that. Maybe because it's the only instance of gray-400 on the form; the placeholder text "From AustinChen" is a different shade too.

Also note that this is still inconsistent with the Amount field we show on our bet panel 
image

and we should maybe consolidate at some point?

akrolsmir (commented):
Looking great, really loving the simplicity of just picking time delta instead of a specific day!

One bug (which I'm pretty sure isn't related to this PR):
In the preview, it correctly shows "From AustinChen" 
image

But from the actual link page it just says "AustinChen" 
image

web/components/manalinks/create-links-button.tsx
jahooma: I was noticing yesterday that our inputs are not rounded, but I think they should be. Some consolidation would be good! I'm not sure whether I prefer the gray background or not.

ingawei: Will round other inputs, the poll I made shall chose which M$ style :)

web/components/manalinks/create-links-button.tsx
ingawei: Set both amount and uses to have a minimum of 1

web/components/manalinks/create-links-button.tsx
```diff
@@ -63,15 +63,39 @@ function CreateManalinkForm(props: {
...
+    return <option value={key}>{value}</option>
+  })
+
+  function setExpireTime(timeDelta: string) {
+    const expiresTime =
+      timeDelta === 'never' ? null : dayjs().add(1, timeDelta).valueOf()
+    setNewManalink((m) => {
+      return {
+        ...m,
+        expiresTime: expiresTime,
```
akrolsmir, line 94: not sure if you knew, but when the key and the value in a JS object have the same variable name like this, you can actually omit the value! aka `return {...m, expiresTime}` would just work, I think :)

akrolsmir (approved):
Looks great, thanks for making these improvements!

web/components/manalinks/create-links-button.tsx
ingawei: changed! Everyday I learn new things
image

PR #648, opened 2022-07-14
fix: let useUserById accept undefined userId

sipec:
As long as we never have a user with the id of "_" this should be fine 😬 
I can't think of a better way to fix this because of react's hook rules prevent us from simply returning early if the input is undefined


jahooma (approved):
Cool! Thanks for the quick fix. I'm not sure of a better way either at the moment.

jahooma:
Just realized that it might return `null` when it finds no user matching `-`. So we should add an if check to return `undefined`.

PR #649, opened 2022-07-14
refactor string matching for search

sipec:
cleans up some logic, and makes it easier to have fuzzy string matching later

### DIFF
--- a/common/util/parse.ts
+++ b/common/util/parse.ts
@@ -49,6 +49,16 @@ export function parseWordsAsTags(text: string) {
   return parseTags(taggedText)
 }
 
+// TODO: fuzzy matching
+export const wordIn = (word: string, corpus: string) =>
+  corpus.toLocaleLowerCase().includes(word.toLocaleLowerCase())
+
+const checkAgainstQuery = (query: string, corpus: string) =>
+  query.split(' ').every((word) => wordIn(word, corpus))
+
+export const searchInAny = (query: string, ...fields: string[]) =>
+  fields.some((field) => checkAgainstQuery(query, field))
+
 // can't just do [StarterKit, Image...] because it doesn't work with cjs imports
 export const exhibitExts = [
   Blockquote,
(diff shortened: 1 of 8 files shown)


common/util/parse.ts
```diff
@@ -49,6 +49,16 @@ export function parseWordsAsTags(text: string) {
   return parseTags(taggedText)
 }
 
+// TODO: fuzzy matching
+export const wordIn = (word: string, corpus: string) =>
+  corpus.toLocaleLowerCase().includes(word.toLocaleLowerCase())
+
+const checkAgainstQuery = (query: string, corpus: string) =>
+  query.split(' ').every((word) => wordIn(word, corpus))
```
jahooma, line 57: Concerned that for every word, corpus is being lowercased again. Probably worth special casing this by re-implementing wordIn again here so that corpus only gets lowercased once.

web/pages/charity/index.tsx
```diff
@@ -88,10 +89,12 @@ export default function Charity(props: {
...
-          charity.name.toLowerCase().includes(query.toLowerCase()) ||
-          charity.preview.toLowerCase().includes(query.toLowerCase()) ||
-          charity.description.toLowerCase().includes(query.toLowerCase()) ||
-          (charity.tags as string[])?.includes(query.toLowerCase())
+          searchInAny(
+            query,
+            charity.name,
+            charity.preview,
+            charity.description
+          ) || (charity.tags as string[])?.includes(query.toLowerCase())
```
jahooma, line 97: I guess charity tags are already lowercase?

common/util/parse.ts
```diff
@@ -49,6 +49,16 @@ export function parseWordsAsTags(text: string) {
...
+
+export const searchInAny = (query: string, ...fields: string[]) =>
```
jahooma, line 59: What about `matchInAny` or `findInAny` or `matchQueryInAny` or `queryMatchesAny`. Up to you.

jahooma (approved):
Nice simplification with an ergonomic utility function!

common/util/parse.ts
sipec: I'd be very surprised if people do searches with more than like 5 words

web/pages/charity/index.tsx
sipec: actually charities just don't have tags (except "Featured"). I never got around to it :/

IanPhilips (approved):
Thanks for consolidating and prettifying this logic! LGTM!

PR #650, opened 2022-07-14
Order groups by most recent chat activity

IanPhilips:


IanPhilips:
- Reduced the create-notification block
- Setting up for separate group/slug/questions notifications to show how many new unseen contracts there are on a group
- Order groups by most recent chat activity on the sidebar and 'My groups' tab on the groups homepage

PR #652, opened 2022-07-15
Add API endpoints for fetching user info by username and ID

mqp:
I know Phil wants this for his Twitch bot, and other people have asked for these before.

The URLs don't look so hot, but we kind of locked ourselves into this when we made `/user/[username]` paths, so here we are.


web/pages/api/v0/user/[username]/index.ts
```diff
@@ -0,0 +1,19 @@
+import { NextApiRequest, NextApiResponse } from 'next'
```
Comment (jahooma) [inline web/pages/api/v0/user/[username]/index.ts:1]: Is it better to rename the file `[username].ts` and move it up a directory?

jahooma (approved):
Thanks, Marshall. Looks good to me!

PR #653, opened 2022-07-16
Inga/manalink bug fixes

ingawei:
Fixed 2 manalink bugs
1. users are now unable to claim their own manalink
2. upon first login, user is first directed to login and then claim, instead of just logging upon claim and then not knowing if anything happened
Screen Shot 2022-07-15 at 5 17 21 PM
Screen Shot 2022-07-15 at 5 17 33 PM


web/components/manalinks/create-links-button.tsx
```diff
@@ -20,6 +20,7 @@ export function CreateLinksButton(props: {
 }) {
   const { user, highlightedSlug, setHighlightedSlug } = props
   const [open, setOpen] = useState(false)
+  console.log('open', open)
```
akrolsmir, line 23: Let's not commit console.log messages into the main codebase!

web/pages/link/[slug].tsx
```diff
@@ -33,14 +33,20 @@ export default function ClaimPage() {
       <div className="mx-auto max-w-xl">
         <Title text={`Claim M$${manalink.amount} mana`} />
         <ManalinkCard
-          defaultMessage={fromUser?.name || 'Enjoy this mana!'}
+          user={user}
+          defaultMessage={`from ${fromUser?.name}`}
```
Comment (akrolsmir) [inline web/pages/link/[slug].tsx:37]: As mentioned IRL: might be better to actually write in the `from Inga Wei` into the database, instead of having special behavior when the message isn't set.

akrolsmir (approved):
Thanks for fixing these two issues!

PR #654, opened 2022-07-16
Let users edit descriptions and questions

akrolsmir:
- Users can now edit their descriptions using rich editor
  image

- Users can now change their own question
  image
  Which then appends a change message in the description
  image

- Whitelisted admins can now do all of the above
  image


jahooma:
Nice work!

PR #657, opened 2022-07-18
Revamp backend code to support good local function development

mqp:
OK, so here's what happened:

- `functions/src/serve.ts` now runs an Express app that serves all our cloud functions in one web server.
- `yarn dev` from the functions directory runs that app under nodemon.
- `dev.sh [prod|dev|localdb]` from the root directory runs all of the web server and the backend concurrently in one invocation. This is the command I figure @akrolsmir will want to use all day.
- A bunch of refactoring to support this stuff.

This is still slightly provisional, mostly because I need to figure out how to get the Google Secret Manager secrets available to local functions (e.g. for email sending.) But it works. Hopefully it works on OS X too.


web/package.json
```diff
@@ -3,12 +3,14 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "dev": "concurrently -n NEXT,TS -c magenta,cyan \"next dev -p 3000\" \"yarn ts --watch\"",
-    "devdev": "cross-env NEXT_PUBLIC_FIREBASE_ENV=DEV concurrently -n NEXT,TS -c magenta,cyan \"cross-env FIREBASE_ENV=DEV next dev -p 3000\" \"cross-env FIREBASE_ENV=DEV yarn ts --watch\"",
```
mqp, line 7: @akrolsmir do you know why these old `FIREBASE_ENV=DEV` assignments were in here? They seem useless to me.

akrolsmir: no, I gave up on understanding that command entirely and just have been cargo-culting it ever since

mqp:
On reflection, it seems basically OK that the secrets aren't available for now. You could still get the secrets working locally by just pulling them out of the Firebase CLI and setting them as environment variables when you started the server, and that would work fine, if you really wanted them working locally. I am sure there's some polished workflow for this but we can put it aside for the moment.

mqp:
Merging this so I can see whether Austin and crew like using it or not and whether I like using it or not.

PR #660, opened 2022-07-18
Don't `setQuery` on group selector component during initial render

mqp:
This makes React mad because the rendering of the `ComboBox` component is setting state upstream in the `GroupSelector` component.

I don't see any reason to do it... why would it matter what the filter is when the box is closed and the filtered groups aren't showing? I can't observe any difference in behavior.


IanPhilips (approved):
I don't notice any difference in functionality either! It looks like the query gets cleared anyways. Thanks for fixing this!

PR #662, opened 2022-07-18
Cache notifs in local, gives instant load of old notifs

IanPhilips:


web/hooks/use-notifications.ts
```diff
@@ -19,31 +20,47 @@ export type NotificationGroup = {
...
+    cachedNotifications ?? []
+  )
 
-  const result = useFirestoreQuery([key], getNotificationsQuery(privateUser.id))
-  useEffect(() => {
+  const result = useFirestoreQueryData(
+    ['notifications-all', privateUser.id],
+    getNotificationsQuery(privateUser.id)
+  )
+  useMemo(() => {
```
jahooma, line 38: Shouldn't this still be an effect if it's not returning a memo-ed data? Ditto for other uses of useMemo in this PR.

jahooma (approved):
Thanks Ian! I tried it and it's now instant, which is awesome!

IanPhilips:
Oh yah I should return those notification arrays if I can, otherwise will
switch em back.

PR #669, opened 2022-07-19
Update Next.js 12.1.2 -> 12.2.0

mqp:
See https://nextjs.org/blog/next-12-2.

I'm interested in doing this upgrade because future Vercel server auth stuff may find the new middleware functionality extremely useful. It's also good to get closer to being able to upgrade to React 18.

The only issue I notice is a spurious warning when running the dev server -- see https://github.com/vercel/next.js/issues/38284. Unfortunately we should probably just put up with it until 12.2.1 is released with the fix. It doesn't seem to indicate any problem except being annoying.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -40,7 +40,7 @@
     "gridjs-react": "5.0.2",
     "lodash": "4.17.21",
     "nanoid": "^3.3.4",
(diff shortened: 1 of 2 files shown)


akrolsmir (approved):
LGTM~

PR #670, opened 2022-07-20
Add @ mentions to editor

sipec:
Not done yet:
- ~Notify people~ edit: in new PR
- ~render at mention using nextjs `<Link>` rather than `<a>`~


jahooma:
Tried to test it out in https://prod-git-editor-at-mention-mantic.vercel.app/
Got no results:
Screen Shot 2022-07-20 at 12 42 19 PM

web/components/editor.tsx
```diff
@@ -56,6 +63,71 @@ export function useTextEditor(props: {
           class: clsx('no-underline !text-indigo-700', linkClass),
         },
       }),
+      Mention.configure({
+        HTMLAttributes: {
+          class: clsx('not-prose text-indigo-700', linkClass),
+        },
+        // TODO: do a Next link instead of raw <a>
```
jahooma, line 70: Is there a reason to prefer the Next Link? I think it does some pre-caching, but also consumes cpu (according to Marshall's profiling).

If it's at all difficult to switch, I just wouldn't.

web/components/editor.tsx
```diff
@@ -56,6 +63,71 @@ export function useTextEditor(props: {
...
+        renderLabel: ({ options, node }) =>
+          [
+            'a',
+            { href: node.attrs.label },
+            `${options.suggestion.char}${node.attrs.label ?? node.attrs.id}`,
+          ] as any,
+        suggestion: {
+          items: ({ query }) =>
+            users
```
jahooma, line 79: Should these be sorted? E.g. alphabetically by name.

I think you'll also find that names with a prefix match should come before those with a mid-string match.

jahooma: This logic could be defined in a named function at the top level of this file (or possibly a util file).

web/components/editor.tsx
```diff
@@ -56,6 +63,71 @@ export function useTextEditor(props: {
...
+                })
+              },
+              onUpdate(props) {
+                component.updateProps(props)
+
+                if (!props.clientRect) {
+                  return
+                }
+
+                popup[0].setProps({
```
jahooma, line 112: It's a little mysterious to use popup[0]. Maybe save a reference directly to popup[0]?

web/components/editor.tsx
```diff
@@ -56,6 +63,71 @@ export function useTextEditor(props: {
           class: clsx('no-underline !text-indigo-700', linkClass),
         },
       }),
+      Mention.configure({
```
jahooma, line 66: Could we link to docs for tiptap somewhere in this file? Whatever would be helpful.

jahooma (commented):
Looks pretty good on a first pass! I'll have to see it in action to provide any further input on the design. You could also include a screenshot of the list of suggestions!

Thanks, Sinclair!

web/components/editor.tsx
sipec: > Should these be sorted?

The best UI would be sorting by your follows > your followers > number of followers.
The 80/20 UI is sorting by account balance I guess?

jahooma: Yup. Something pretty important is to show the prefix matches first. So, if I search for "Si" it should put Sinclair above names that merely contain "si", like Farsi (or whatever).

jahooma:
Let me know when you want me to do another review pass!

jahooma (approved):
LGTM! Thanks for making those fixes.

web/components/editor.tsx
sipec: Decided to render the mention using our existing react component (which uses Next Link - though we can change that later if we want).
This was the simplest way to fix certain bugs.

PR #674, opened 2022-07-20
migrate useUsers hook to react-query

sipec:
low priority.

This makes it faster when multiple components on a page call `useUsers` - but that's not super relevant yet.

I just did this to see if it would magically fix an unrelated bug (it didn't).


akrolsmir:
It reduces total lines of code in our codebase, I'm a fan~

PR #675, opened 2022-07-21
Make a React context to be the source of truth for authenticated user

mqp:
See here for an [intro to React contexts](https://reactjs.org/docs/context.html).

The main motivation for this: In our previous version, every time a component called `useUser` (which called `listenForLogin`), it would call `onIdTokenChanged` with a new event handler. That meant that when the ID token did in fact change (e.g. on initial render if you are logged in, or on refresh, or on login/logout), there would be a bunch of identical event handlers running -- it wouldn't be strange for there to be 10 or 20, since we have many components on a page that may call `useUser`.

This was pretty weird and counterintuitive, and besides being tricky to think about and obviously inefficient, it caused us to for example write a strange workaround so that we could create exactly one new user in the event handler. It's possible it was also incorrect if you made multiple user changes too fast, although I don't know if that ever happened in practice.

Now there's only one event handler at the top level, which stores the current user information in the context once, so it won't do extra work, everything will work how a reader would expect it to, and it will be easier to write code that runs "on login" or "on logout" in the future.

### DIFF
--- a/web/components/auth-context.tsx
+++ b/web/components/auth-context.tsx
@@ -13,6 +13,10 @@ import { randomString } from 'common/util/random'
 import { identifyUser, setUserProperty } from 'web/lib/service/analytics'
 import { useStateCheckEquality } from 'web/hooks/use-state-check-equality'
 
+// Either we haven't looked up the logged in user yet (undefined), or we know
+// the user is not logged in (null), or we know the user is logged in (User).
+type AuthUser = undefined | null | User
+
 const CACHED_USER_KEY = 'CACHED_USER_KEY'
 
 const ensureDeviceToken = () => {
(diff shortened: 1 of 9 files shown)


mqp:
Tagging @jahooma since I have the idea he is the "tries to write good React code" guy.

web/components/auth-context.tsx
```diff
@@ -0,0 +1,71 @@
...
+    deviceToken = randomString()
+    localStorage.setItem('device-token', deviceToken)
+  }
+  return deviceToken
+}
+
+export const AuthContext = createContext<User | null>(null)
+
+export function AuthProvider({ children }: any) {
+  const [currentUser, setCurrentUser] = useStateCheckEquality<User | null>(null)
```
jahooma, line 30: Concerned about the first render on the server and client. Previously, currentUser was undefined for this, then loaded from local storage on the first useEffect.

I think we should replicate that by setting the initial value to undefined. Some code handles undefined differently from null. Undefined means we're waiting to see whether you're logged in or not.

web/components/auth-context.tsx
```diff
@@ -0,0 +1,71 @@
...
+    })
+  }, [setCurrentUser])
+
+  useEffect(() => {
+    if (currentUser) {
+      identifyUser(currentUser.id)
+      setUserProperty('username', currentUser.username)
+      return listenForUser(currentUser.id, setCurrentUser)
+    }
+  }, [currentUser, setCurrentUser])
```
jahooma, line 66: Shouldn't this only update on e.g. `currentUser?.id` instead of `currentUser`? Looks like a pre-existing minor bug.

Could pull the id and username to local vars and pass just those into the effect.

jahooma (approved):
Thanks, Marshall! The overall approach looks fine. I knew this was happening, but it never seemed to cause problems. Good to fix though!

One thing I'll say is that I don't really like using React Context as a pattern, but I think it's a bit better when provided at the root level and for something ubiquitous like Auth.

web/components/auth-context.tsx
mqp: Agree that would be better.

web/components/auth-context.tsx
mqp: OK, I agree that it seems reasonable to draw the distinction between "not sure yet" and "not logged in" for now. (I'm not sure it will still be reasonable in the future if we're doing SSR everywhere, but we can revisit it at a later date.)

PR #677, opened 2022-07-21
Make main login/logout buttons reload server side props

mqp:
That means that if we create server-side redirects, they will trigger the redirects if appropriate (i.e. redirecting from `/` to `/home` if logging in, redirecting from random logged-in-users-only pages to `/` if logging out.) So now we can set up server side redirects when desired and not worry about also doing it on the client side in the React components anymore.

### DIFF
--- a/web/components/auth-context.tsx
+++ b/web/components/auth-context.tsx
@@ -41,6 +41,7 @@ export function AuthProvider({ children }: any) {
   useEffect(() => {
     return onIdTokenChanged(auth, async (fbUser) => {
       if (fbUser) {
+        setAuthCookies(await fbUser.getIdToken(), fbUser.refreshToken)
         let user = await getUser(fbUser.uid)
         if (!user) {
           const deviceToken = ensureDeviceToken()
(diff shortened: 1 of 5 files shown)


web/components/auth-context.tsx
```diff
@@ -41,6 +41,7 @@ export function AuthProvider({ children }: any) {
   useEffect(() => {
     return onIdTokenChanged(auth, async (fbUser) => {
       if (fbUser) {
+        setAuthCookies(await fbUser.getIdToken(), fbUser.refreshToken)
```
mqp, line 44: When you call `firebaseLogin`, it doesn't resolve until this event handler runs, but it doesn't wait for this event handler to also resolve. So now that we want to do stuff right way after `firebaseLogin` -- like reload and ask for new server side props based on our new cookies -- we ought to make sure we set the cookies right away.

mqp:
@jahooma here's my "actual suggestion" for what to do about the problem you noticed yesterday. This should resolve it everywhere.

web/components/create-question-button.tsx
```diff
@@ -26,7 +28,12 @@ export const CreateQuestionButton = (props: {
         </Link>
       ) : (
         <button
-          onClick={firebaseLogin}
+          onClick={async () => {
+            // login, and then reload the page, to hit any SSR redirect (e.g.
+            // redirecting from / to /home for logged in users)
+            await firebaseLogin()
+            router.replace(router.asPath)
```
mqp, line 35: This might look a little odd but it seems like it's the consensus for the best way to reload server-side props with Next right now. See https://github.com/vercel/next.js/discussions/12646.

jahooma (approved):
Awesome, thanks for researching and fixing this! Looks good.

PR #678, opened 2022-07-21
Inga/manalinks pagination bug

ingawei:
fixed pagination and link timing out bug
also changed UI a bit, I don't like that the inconsistency messages bring to the cards so I just put them in the bottom
Screen Shot 2022-07-21 at 12 13 11 PM

### DIFF
--- a/web/components/manalink-card.tsx
+++ b/web/components/manalink-card.tsx
@@ -27,49 +27,45 @@ export function ManalinkCard(props: {
   const { expiresTime, maxUses, uses, amount, message } = info
   return (
     <Col>
-      <Col>
-        <div
-          className={clsx(
-            className,
-            'min-h-20 group flex flex-col rounded-xl bg-gradient-to-br shadow-lg transition-all',
-            getManalinkGradient(info.amount)
-          )}
-        >
-          <Col className="mx-4 mt-2 -mb-4 text-right text-sm text-gray-100">
-            <div>
-              {maxUses != null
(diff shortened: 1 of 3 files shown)


web/pages/links.tsx
```diff
@@ -71,6 +68,35 @@ export default function LinkPage() {
...
+      <p className="text-gray-500">
+        You don't have any unclaimed manalinks. Send some more to spread the
+        wealth!
+      </p>
+    )
+  } else {
+    return (
+      <>
         <Col className="grid w-full gap-4 md:grid-cols-2">
           {displayedLinks.map((link) => {
```
jahooma, line 101: I think you need a `key` prop in the `<ManalinkCardFromView>` component below. 

https://reactjs.org/docs/lists-and-keys.html#keys

jahooma: You could also use the short form of an anonymous function without a return statement: 
`displayedLinks.map((link) => <ManalinkCardFromView ... />)

web/components/manalink-card.tsx
```diff
@@ -82,75 +84,79 @@ export function ManalinkCardFromView(props: {
   const [details, setDetails] = useState(false)
```
jahooma, line 84: `details` doesn't sound like a boolean variable. Maybe `showDetails` and `setShowDetails`?

jahooma (approved):
Thanks, Inga! The Manalinks look great!

I only have a few minor comments on the code.

PR #679, opened 2022-07-21
Challenge Bets

IanPhilips:
- Users can create quasi-limit bets at x probability with y amount on z outcome
- They get a link to share with anyone
  -  The format of the link is: `manifold.markets/challenges/username/market-slug/challenge-slug`
- Anyone with an account can use that link to take the other side of the trade.
  - Currently, the one who accepts the challenge puts up the just enough mana to make sure they have equal shares of each answer.
  - I could turn the probability input into a toggle that allows for 50:50 odds or current market probability odds. If set to the current market probability, the probability would be set in stone (would not change with the market even if it moves) and the acceptor will only put up enough mana such that they have equal shares of either side. 
  - If we allow the acceptor to get the current market price this may mean the offer may have gotten worse or better for the original creator, which I think is weird and perhaps unexpected.
- users get notifications for when their challenge was accepted and by whom

Screen Shot 2022-07-21 at 2 34 27 PM 

Screen Shot 2022-07-21 at 2 34 32 PM

Screen Shot 2022-07-21 at 3 07 03 PM

Screen Shot 2022-07-21 at 3 06 46 PM

Screen Shot 2022-07-21 at 3 21 45 PM

Screen Shot 2022-07-21 at 3 20 04 PM

Screen Shot 2022-07-21 at 2 34 03 PM
Screen Shot 2022-07-21 at 2 33 57 PM
Screen Shot 2022-07-21 at 2 43 08 PM


mantikoros (changes requested):
Great work! This is a fantastic, well-architected feature. And done very quickly. I'm impressed.

A few concerns:
- Challenge bets shouldn't change the liquidity pool in any way (as it is just a side bet between two parties)
- Numeric markets seem to be only partially supported. We should either add full support or remove them entirely. I think the latter is better.
- The challenge page and list of challenge bets looks pretty, but the UI on the market page could be better.

I think I will go ahead and make all of these changes myself, and I'll check with you before merging this branch in.

IanPhilips:
Screen Shot 2022-08-01 at 3 09 05 PM

IanPhilips:
Screen Shot 2022-08-02 at 9 38 47 AM

PR #680, opened 2022-07-21
Write script to fix old comments without IDs and user IDs

mqp:
In February we updated comments to have their ID in the document and to have the creator's user ID in the comment as part of #44, but many old comments were never fixed, so they still were missing IDs, producing a variety of creative bugs.


mqp:
@mantikoros for review since he's the one that was supposed to write this in the first place.

mantikoros (approved):
LGTM! Feel free to run.

PR #681, opened 2022-07-21
refactor createNotif - put optional args in one param

sipec:
This way you don't need a bunch of `undefined` and you don't have to remember the order of optional params.

I'm doing this because I want to let this function accept a list of recipients, but I found adding/changing params annoying


akrolsmir (approved):
I think this is much better! Perhaps notifications would be better (code-wise) if they were strongly typed, and I don't _love_ "miscData" as a name as it's very generic, maybe "optionalPayload" or sth?

But happy to ship as-is since it's a clear improvement!

PR #682, opened 2022-07-21
Render iframes inside the rich text editor

akrolsmir:
Thoughts:
- It's pretty hard to discover this feature atm; maybe we do a toolbar or something, later
- Not sure this is the most ideal path for getting more markets embedded - it might be possible to have TipTap hook directly into our React components https://tiptap.dev/guide/node-views/react
- This only works while editing a description, not on new market descriptions; my guess is this validation stuff breaks on iframes https://github.com/manifoldmarkets/manifold/blame/editor-embed/functions/src/create-contract.ts#L33


common/util/iframe.ts
```diff
@@ -0,0 +1,98 @@
+// Adopted from https://github.com/ueberdosis/tiptap/blob/main/demos/src/Experiments/Embeds/Vue/iframe.ts
```
akrolsmir, line 1: I'm not excited about having this file in `common/util/` instead of `web/components/editor/`, but it was the way I found to get parse.ts to be able to include this extension in `exhibitExts`

sipec: maybe just rename file to "extension-iframe.ts" or "tiptap-extension-iframe.ts"?

sipec (approved):

common/util/iframe.ts
akrolsmir: Hmm was wondering about more exotic options, like move exhibitExts into web/? Or somehow common/util/ just has the type definition, and the web implementation lives in web/? idk.

Renamed to tiptap-iframe for now

PR #687, opened 2022-07-22
500 mana email

SirSaltyy:
Creating market guide


mantikoros (changes requested):
Thoughts on 500 mana email:

- Needs a bit more exposition; should mention Manifold Markets upfront (many users will have forgotten they signed up and what our platform is), "Thanks for signing up for Manifold Markets!"
- Should use "M$" more. The terms "mana" or "Manifold Dollars" only appear in the docs—which users don't read.
- Don't capitalize mana
- I'd cut the section on other ways to earn mana. There should only be one clear message and call to action in this email: clicking on the link to get mana to bet on our site.
- Not a fan of the whimsical-looking button style. A plainer button with "Claim M$500" would be better, I think.

Thoughts on creating market email:

- Should be more explicit that this email is a guide to creating markets on Manifold. Mentioning what makes for a "good question" by itself is insufficient and may confuse more casual users.
- It would probably be helpful to discuss specific types of questions users can ask with examples
- Encourage creators to share their markets on social media using the link

PR #689, opened 2022-07-23
Image compression

TrueMilli:
## Summary
I added logic rejecting or compressing large images in the browser before uploading.
There is no new logic on the firebase.

## Parameters
* Images over a certain size are rejected because they take unreasonably long to compress. The target size might not even get reached (default max iterations: 10). I chose 20 MB as limit.
* I set 1 MB as target size.
* Height / Width is restricted to not waste space on resolutions which are not required. I chose 1920.

## Test files
Large images between 11 and 39 MB can be found under https://svs.gsfc.nasa.gov/vis/a030000/a030800/a030877/frames/5760x3240_16x9_01p/.

## Compression package
https://github.com/Donaldcwl/browser-image-compression

## Alternative
* I tried and decided against Compressor.js because it doesn't integrate as nicely in the current solution.
https://github.com/fengyuanchen/compressorjs 
* Compression in the firebase after image upload. Did not look into it: https://firebase.google.com/products/extensions/firebase-storage-resize-images


mqp:
I'm not really sure whether it's good to encourage people to upload insane avatars by automatically compressing them to be insane but smaller... avatars should be like 10kb 200x200, not 1MB 1920x1200. I would honestly rather it just be flatly broken so that they notice and upload something else instead.

I would be happy to accept a fix with UI that guided people to upload something smaller, instead.

sipec (approved):
Thanks for taking care of this!

sipec:
> I'm not really sure whether it's good to encourage people to upload insane avatars

Shrug. I don't think people will be encouraged by much - not like we're telling them. It might makes sense to have lower size limits (and perhaps different size limit for editor vs avatar) but that seems straightforward to change later if it actually is an issue

mqp:
> > I'm not really sure whether it's good to encourage people to upload insane avatars
> 
> Shrug. I don't think people will be encouraged by much - not like we're telling them. It might makes sense to have lower size limits (and perhaps different size limit for editor vs avatar) but that seems straightforward to change later if it actually is an issue

I mean, what I am saying is that when this pull request causes people to be able to upload an avatar, I would prefer them not to be able to upload it, so that they would instead choose a different avatar. I think compressing it to 1MB is worse than failing.

TrueMilli:
> > > I'm not really sure whether it's good to encourage people to upload insane avatars
> > 
> > 
> > Shrug. I don't think people will be encouraged by much - not like we're telling them. It might makes sense to have lower size limits (and perhaps different size limit for editor vs avatar) but that seems straightforward to change later if it actually is an issue
> 
> I mean, what I am saying is that when this pull request causes people to be able to upload an avatar, I would prefer them not to be able to upload it, so that they would instead choose a different avatar. I think compressing it to 1MB is worse than failing.

It's currently possible to upload files bigger than 20 MB so this actually reduces what's possible.
It's not only about profile pictures but also images in market descriptions. Files in higher resolution are reasonable there I feel.

mqp:
> > > > I'm not really sure whether it's good to encourage people to upload insane avatars
> > > 
> > > 
> > > Shrug. I don't think people will be encouraged by much - not like we're telling them. It might makes sense to have lower size limits (and perhaps different size limit for editor vs avatar) but that seems straightforward to change later if it actually is an issue
> > 
> > 
> > I mean, what I am saying is that when this pull request causes people to be able to upload an avatar, I would prefer them not to be able to upload it, so that they would instead choose a different avatar. I think compressing it to 1MB is worse than failing.
> 
> It's currently possible to upload files bigger than 20 MB so this actually reduces what's possible. It's not only about profile pictures but also images in market descriptions. Files in higher resolution are reasonable there I feel.
> 
> See e,g, https://manifold.markets/JoyVoid/what-dalle-prompt-will-i-choose-as-1f5e0795f091

That's a good point about the market descriptions. In that case I am basically neutral on this PR so if @sipec wants it they should merge it.

PR #691, opened 2022-07-24
Make tabs components better

mqp:
Previously, our tabs component that we used everywhere had only one mode, which would be normally called "uncontrolled" -- it kept its active index in local React state. This caused at least two problems:

- In some circumstances, a parent component might like to choose which tab is selected in some arbitrary way. But this becomes impossible to do in any principled way. This really bit me when I was making the manalinks page and I wanted to switch tabs to show the manalink you just created. (That particular case got redesigned, but obviously this is a very normal sounding desire.)
- In some circumstances, we are OK with having internal React-managed active tab state, but we want to represent it differently. For example, on the user page, we represent the active tab in the URL query string. This meant that when you changed tabs, it had to re-render everything twice -- once when the tabs component changed its intenal state, and once when outside code changed the query string on its behalf.

Now I split the tab component into three varieties:

- `ControlledTabs`, where you pass the active index in via props and the parent is responsible for setting it (e.g. via the `onClick` handler.
- `UncontrolledTabs`, which wraps `ControlledTabs` and manages an `activeIndex` which it sets in the `onClick` handler. This is synonymous with our existing `Tabs`.
- `QueryUncontrolledTabs`, which wraps `ControlledTabs` and sets the `activeIndex` based on the URL query parameter, which it then sets in the `onClick` handler.

Now the user page uses `QueryUncontrolledTabs`, so it fixes the problem with re-rendering.

Later I will go fix up other tab controls as appropriate.

### DIFF
--- a/web/components/layout/tabs.tsx
+++ b/web/components/layout/tabs.tsx
@@ -9,8 +9,6 @@ type Tab = {
   title: string
   tabIcon?: ReactNode
   content: ReactNode
-  // If set, change the url to this href when the tab is selected
-  href?: string
   // If set, show a badge with this content
   badge?: string
 }
@@ -33,25 +31,21 @@ export function UncontrolledTabs(props: TabProps & { activeIndex: number }) {
     currentPageForAnalytics,
   } = props
   const activeTab = tabs[activeIndex] as Tab | undefined // can be undefined in weird case
-
   return (
     <>
       <div className={clsx('mb-4 border-b border-gray-200', className)}>
         <nav className="-mb-px flex space-x-8" aria-label="Tabs">
           {tabs.map((tab, i) => (
-            <Link href={tab.href ?? '#'} key={tab.title} shallow={!!tab.href}>
+            <Link href="#" key={tab.title} shallow={false}>
               <a
                 id={`tab-${i}`}
                 key={tab.title}
                 onClick={(e) => {
+                  e.preventDefault()
                   track('Clicked Tab', {
                     title: tab.title,
-                    href: tab.href,
                     currentPage: currentPageForAnalytics,
                   })
-                  if (!tab.href) {
-                    e.preventDefault()
-                  }
                   onClick?.(tab.title, i)
                 }}
                 className={clsx(
--- a/web/components/layout/tabs.tsx
+++ b/web/components/layout/tabs.tsx
@@ -1,8 +1,6 @@
 import clsx from 'clsx'
-import Link from 'next/link'
 import { useRouter, NextRouter } from 'next/router'
 import { ReactNode, useState } from 'react'
-import { Row } from './row'
 import { track } from '@amplitude/analytics-browser'
 
 type Tab = {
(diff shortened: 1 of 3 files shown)


mqp:
Tagging @akrolsmir since he wrote the tabs component.

Also tagging @jahooma since he expressed interest in why the user page was so slow. (This makes it twice as fast, which is to say, still slow. You're right that rendering is the main bottleneck.)

web/components/layout/tabs.tsx
```diff
@@ -8,50 +9,42 @@ type Tab = {
...
-
   return (
     <>
       <div className={clsx('mb-4 border-b border-gray-200', className)}>
         <nav className="-mb-px flex space-x-8" aria-label="Tabs">
           {tabs.map((tab, i) => (
-            <Link href={tab.href ?? '#'} key={tab.title} shallow={!!tab.href}>
+            <Link href="#" key={tab.title} shallow={false}>
               <a
-                id={`tab-${i}`}
```
mqp, line 43: I happen to know that I added this for my hacky manalinks page workaround and nothing is using it anymore.

web/components/layout/tabs.tsx
```diff
@@ -1,82 +1,121 @@
...
   tabs: Tab[]
-  defaultIndex?: number
   labelClassName?: string
   onClick?: (tabTitle: string, index: number) => void
   className?: string
   currentPageForAnalytics?: string
-}) {
+}
+
+export function UncontrolledTabs(props: TabProps & { activeIndex: number }) {
```
jahooma, line 22: I think you have the terminology backward re: Controlled vs Uncontrolled components

Controlled components are "controlled" by a parent, and don't store any state.
https://stackoverflow.com/a/42522792/1592933

web/components/layout/tabs.tsx
```diff
@@ -1,82 +1,121 @@
...
+  const selectedIdx = tabs.findIndex((t) => isTabSelected(router, 'tab', t))
+  const activeIndex = selectedIdx !== -1 ? selectedIdx : defaultIndex ?? 0
+  return (
+    <UncontrolledTabs
+      {...rest}
+      tabs={tabs}
+      activeIndex={activeIndex}
+      onClick={(title, i) => {
+        router.replace(
+          { query: { ...router.query, tab: title.toLowerCase() } },
```
jahooma, line 110: A minor point might be that if the default tab is later selected, it should not add the query string

jahooma: Another slight change that is up to personal taste would be using `?tab='bets' vs `?bets`.

I slightly prefer the latter b/c it's shorter, but I am ok with how it works now.

jahooma (approved):
Thanks Marshall! This is a great change that makes it easier for us to use tabs correctly in the future.

web/components/layout/tabs.tsx
mqp: How embarrassing! In my mind it was like, controlled components control themselves, and uncontrolled components need something else to control them. I guess it's probably not the best jargon ever invented.

web/components/layout/tabs.tsx
mqp: I personally prefer the way it works now on these two points. If you feel strongly you can go change it later.

PR #692, opened 2022-07-24
Make `setNotificationsAsSeen` return a promise

mqp:
Not important, I just think it's easier to understand upstream that it's async when you're looking at a type signature if it returns a promise instead of returning void.


IanPhilips (approved):
Yup, makes sense, thanks for making this change!

PR #693, opened 2022-07-25
Clean up comments list rendering

mqp:
I really don't know how the markup ends up with so much literally useless HTML and CSS in it, but this list is a rendering bottleneck on the user profile page, so let's try to clean it up somewhat.

### DIFF
--- a/web/components/comments-list.tsx
+++ b/web/components/comments-list.tsx
@@ -17,58 +17,55 @@ export function UserCommentsList(props: {
   contractsById: { [id: string]: Contract }
 }) {
   const { comments, contractsById } = props
-  const commentsByContract = groupBy(comments, 'contractId')
 
-  const contractCommentPairs = Object.entries(commentsByContract)
-    .map(
-      ([contractId, comments]) => [contractsById[contractId], comments] as const
-    )
-    .filter(([contract]) => contract)
+  // we don't show comments in groups here atm, just comments on contracts
+  const contractComments = comments.filter((c) => c.contractId)
+  const commentsByContract = groupBy(contractComments, 'contractId')
 
   return (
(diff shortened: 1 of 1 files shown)


web/components/comments-list.tsx
```diff
@@ -17,58 +17,55 @@ export function UserCommentsList(props: {
...
+  const commentsByContract = groupBy(contractComments, 'contractId')
 
   return (
     <Col className={'bg-white'}>
-      {contractCommentPairs.map(([contract, comments]) => (
-        <div key={contract.id} className={'border-width-1 border-b p-5'}>
-          <div className={'mb-2 text-sm text-indigo-700'}>
-            <SiteLink href={contractPath(contract)}>
+      {Object.entries(commentsByContract).map(([contractId, comments]) => {
+        const contract = contractsById[contractId]
```
jahooma, line 28: I'm pretty sure there's some case where the contract doesn't exist. Like maybe we deleted some contracts without deleting the comments

jahooma (approved):
Thanks for cleaning up the excess divs!

web/components/comments-list.tsx
mqp: The purpose of the previous check was definitely to filter out group comments, so I think we don't know whether or not it's true that there might be ones where the comment doesn't exist.

I personally want it to explode in that case so that someone reports the bug and I can go fix the database, not to fail silently. It's unacceptable to have inconsistent data hanging around.

jahooma: Ok, cool. I thought I remembered it failing for me in the past, but using the vercel preview and navigating to my profile, there wasn't an error. I'm down to try it and see if it fails for anyone.

PR #694, opened 2022-07-25
Clean up rendering of user bets list

mqp:
This changes the bets list to not render all the individual bet details until you expand a contract. That means that there isn't an animation anymore, but the old animation was so fast that it looked worse anyway, in my opinion.

Anyway, the old way of rendering everything was taking 100-300ms on my laptop, so this is obviously the correct tradeoff direction.


jahooma (approved):
Thanks for the cleanup, and nice idea to reduce render time!

PR #696, opened 2022-07-26
Implement "sell all shares" functionality in `sellshares` and expose API

mqp:
This is partly for the Twitch bot, partly because it's a natural API to have, and partly so that our bet panel selling works reliably* to sell exactly 100% of your shares when desired.

*Well, I'm not really convinced the server code is correct with all magnitudes of shares, but it works better than if we didn't do this.


mqp:
Merging this so that it can be useful to Phil on the bot.

PR #701, opened 2022-07-28
Gift cards

IanPhilips:
I'm waiting on Tremendous' approval for depositing real money (aka getting a prod API key ) to fund the gift cards, but pending this and the secrets management, we'll be ready to go. 

Screen Shot 2022-07-28 at 10 49 49 AM

Screen Shot 2022-07-27 at 6 12 40 PM

Screen Shot 2022-07-27 at 6 12 53 PM

Screen Shot 2022-07-27 at 6 13 00 PM

### DIFF
--- a/common/txn.ts
+++ b/common/txn.ts
@@ -1,6 +1,6 @@
 // A txn (pronounced "texan") respresents a payment between two ids on Manifold
 // Shortened from "transaction" to distinguish from Firebase transactions (and save chars)
-type AnyTxnType = Donation | Tip | Manalink | Referral | Bonus
+type AnyTxnType = Donation | Tip | Manalink | Referral | Bonus | Purchase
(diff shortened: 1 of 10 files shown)


IanPhilips:
I have properly set the tremendous test key in google secrets: 

Screen Shot 2022-07-28 at 11 00 42 AM

But it won't work. @mqp would you have any ideas on how to make this work? 
I'm passing the secret to the functions like so: 
```
const opts = { secrets: ['TREMENDOUS_TEST_KEY'] }

export const convertmana = newEndpoint(opts, async (req, auth) => {
```
and reading it via `const apiKey = process.env.TREMENDOUS_TEST_KEY as string`

IanPhilips:
Tremendous denied the API request and haven't given a reason yet. Austin suggested we could just buy 10 amazon gift cards and offer them for mana to see if there's demand. Would be pretty easy to do actually!

PR #702, opened 2022-07-29
Add editor toolbar to choose and embed markets

akrolsmir:
Issues:
- [x] Clicking on the button brings up the annoying 'Changes you made may not be saved.' prompt for some reason
- [x] Can't embed multiple markets at a time, for some reason
- [x] Not sure about the UI. Maybe want the modal to have a sticky header, otherwise when you scroll down you might not know to go back up for the "Embed 1 question" button?

image
image


web/components/editor/market-modal.tsx
```diff
@@ -0,0 +1,93 @@
...
+            ) : (
+              <Row className={'justify-center'}>
+                <LoadingIndicator />
+              </Row>
+            )}
+          </Col>
+        </Col>
+
+        <div className={'overflow-y-scroll sm:px-8'}>
+          <ContractSearch
```
IanPhilips, line 76: I might prefer hiding the filter pills. You could add another option for `ContractSearch` or if you want to improve the ux you could keep track of which contracts the user has embedded already and not show them again them via: `         additionalFilter={{ excludeContractIds: [] }}` which will also hide the filter pills.

web/components/share-embed-button.tsx
```diff
@@ -29,7 +27,7 @@ export function ShareEmbedButton(props: {
       as="div"
       className="relative z-10 flex-shrink-0"
       onMouseUp={() => {
-        copyEmbedCode(contract)
+        copyToClipboard(embedCode(contract))
```
IanPhilips, line 30: Good idea to use the standard `copyToClipboard` method!

web/components/editor.tsx
```diff
@@ -269,7 +290,7 @@ const useUploadMutation = (editor: Editor | null) =>
     }
   )
 
-function RichContent(props: { content: JSONContent | string }) {
+export function RichContent(props: { content: JSONContent | string }) {
```
IanPhilips, line 293: Looks like this was exported but not used elsewhere?

web/components/editor.tsx
```diff
@@ -178,7 +183,23 @@ export function TextEditor(props: {
...
+                type="button"
+                onClick={() => setMarketOpen(true)}
+                className="-m-2.5 flex h-10 w-10 items-center justify-center rounded-full text-gray-400 hover:text-gray-500"
+              >
+                <MarketModal
+                  editor={editor}
+                  open={marketOpen}
+                  setOpen={setMarketOpen}
+                />
+                <PresentationChartLineIcon
```
IanPhilips, line 199: Cool! Or we could use a Foldy icon?

web/components/editor.tsx
```diff
@@ -178,7 +183,23 @@ export function TextEditor(props: {
...
-                <span className="sr-only">Embed an iframe</span>
+              </button>
+            </div>
+            <div className="tooltip flex items-center" data-tip="Add market">
+              <button
+                type="button"
+                onClick={() => setMarketOpen(true)}
+                className="-m-2.5 flex h-10 w-10 items-center justify-center rounded-full text-gray-400 hover:text-gray-500"
+              >
+                <MarketModal
```
IanPhilips, line 194: I'm not sure why it brings up the 'unsaved changes warning' but if we can't figure out how to officially let the window know it's no big deal we could toggle it off/on when interacting with this button. Check out `use-warn-unsaved-changes.ts` to see where it's toggled on.

sipec:
> Issues:
> 
> * [ ]  Clicking on the button brings up the annoying 'Changes you made may not be saved.' prompt for some reason
> * [ ]  Can't embed multiple markets at a time, for some reason

I can't reproduce either of those. There's no prompt and embedding multiple markets just works.

web/components/editor.tsx
sipec: I think an actual redirect happens. It should warn unsaved changes in that case, when adding stuff to the post

sipec:
I think this is good to be reviewed again, but assigning to Austin since Ian is on vacation

PR #703, opened 2022-07-29
Switch comments/chat to rich text editor

sipec:


web/components/feed/feed-comments.tsx
```diff
@@ -465,81 +464,43 @@ export function CommentInput(props: {
...
   return (
     <>
       <Row className="gap-1.5 text-gray-700">
-        <Textarea
-          ref={setRef}
-          value={commentText}
-          onChange={(e) => setComment(e.target.value)}
-          className={clsx('textarea textarea-bordered w-full resize-none')}
-          // Make room for floating submit button.
-          style={{ paddingRight: 48 }}
```
jahooma, line 507: Does the new editor look ok with the mobile submit button?

jahooma (commented):
Thanks, Sinclair — looks like good progress on this feature! Rich text comments will be very neat.

- Wasn't able to try it out since the Vercel preview didn't load.
- I think there are a few TODO's you've highlighted which should be done before this goes into production.

web/components/feed/feed-comments.tsx
```diff
@@ -568,28 +529,22 @@ export function CommentInputTextArea(props: {
...
-  comment: string
+  comment: JSONContent
   moreHref: string
   shouldTruncate?: boolean
 }) {
   const { comment, moreHref, shouldTruncate } = props
   let truncated = comment
 
-  // Keep descriptions to at most 400 characters
+  // TODO: Keep descriptions to at most 80 words (~400 characters)
```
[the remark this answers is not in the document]
sipec: It's as simple as using the CharCount extension I think.
An alternative to truncating would be to only display n lines of text and have overflow cut off.
In any case, there isn't anywhere we currently truncate comments so I'm just going to remove the option.

web/components/feed/feed-comments.tsx
sipec: it does

sipec:
@jahooma worth taking another look now

web/components/feed/feed-answer-comment-group.tsx
```diff
@@ -70,17 +70,23 @@ export function FeedAnswerCommentGroup(props: {
 
   const scrollAndOpenReplyInput = useEvent(
     (comment?: Comment, answer?: Answer) => {
-      setReplyToUsername(comment?.userUsername ?? answer?.username ?? '')
+      setReplyToUser(
```
jahooma, line 73: I think this could be refactored to look a little nicer. Maybe with if statements for comment vs answer?

web/components/feed/feed-comments.tsx
```diff
@@ -465,94 +442,89 @@ export function CommentInput(props: {
...
+    if (replyToUser) {
+      editor.commands.insertContentAt(0, {
+        type: 'mention',
+        attrs: { label: replyToUser.username, id: replyToUser.id },
+      })
+      editor.commands.focus()
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [user, replyToUsername, memoizedSetComment])
+  }, [editor])
```
jahooma, line 507: I think this means if you click reply on one comment, and then click reply to a different user's comment in the same thread, the @mention will not change as it should. That's not the worst behavior, but I wonder if there other problems.

web/components/groups/group-chat.tsx
```diff
@@ -292,16 +291,18 @@ function GroupChatNotificationsIcon(props: {
...
   tips: CommentTips
 }) {
-  const { comment, onReplyClick, group, setRef, highlight, user, tips } = props
-  const { text, userUsername, userName, userAvatarUrl, createdTime } = comment
-  const isCreatorsComment = user && comment.userId === user.id
+  const { comments, onReplyClick, group, setRef, highlight, user, tips } = props
+  const first = comments[0]
+  const { id, userUsername, userName, userAvatarUrl, createdTime } = first
+
+  const isCreatorsComment = user && comments[0].userId === user.id
```
jahooma, line 305: Could use `first`

jahooma (approved):
Thanks Sinclair! The code looks pretty good. Feel free to merge when you think it's ready!

PR #704, opened 2022-07-29
Api fixes

sipec:


web/.eslintrc.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:react-hooks/recommended',
     'plugin:@next/next/recommended',
+    'prettier',
```
jahooma, line 8: Nice miscellaneous change! Could just be a separate commit to main.

web/pages/api/v0/_types.ts
```diff
@@ -109,6 +119,10 @@ export function toLiteMarket(contract: Contract): LiteMarket {
         : closeTime,
     question,
     description,
+    textDescription:
+      typeof description === 'string'
+        ? description
+        : richTextToString(description),
```
jahooma, line 125: Nice!

jahooma (approved):
Seems like all good changes for the API. Thanks!

web/.eslintrc.js
sipec: actually this doesn't work :(

sipec:
trying to test this before I merge but deploys are failing

PR #706, opened 2022-08-01
Allow unspecfied outcome as input to `sellshares`

mqp:
This is particularly useful in the case of bots, where users may already know how they have bet, regardless of whether the bot knows, and want a very succinct command to sell all their shares.

### DIFF
--- a/functions/src/sell-shares.ts
+++ b/functions/src/sell-shares.ts
@@ -9,7 +9,7 @@ import { getCpmmSellBetInfo } from '../../common/sell-bet'
 import { addObjects, removeUndefinedProps } from '../../common/util/object'
 import { getValues, log } from './utils'
 import { Bet } from '../../common/bet'
-import { floatingLesserEqual } from '../../common/util/math'
+import { floatingEqual, floatingLesserEqual } from '../../common/util/math'
 import { getUnfilledBetsQuery, updateMakers } from './place-bet'
 import { FieldValue } from 'firebase-admin/firestore'
 import { redeemShares } from './redeem-shares'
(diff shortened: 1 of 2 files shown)


functions/src/sell-shares.ts
```diff
@@ -46,9 +46,28 @@ export const sellshares = newEndpoint({}, async (req, auth) => {
...
+    const sharesByOutcome = mapValues(betsByOutcome, (bets) =>
+      sumBy(bets, (b) => b.shares)
+    )
+
+    let chosenOutcome: 'YES' | 'NO'
+    if (outcome != null) {
+      chosenOutcome = outcome
+    } else {
+      const nonzeroShares = Object.entries(sharesByOutcome).filter(
+        ([_k, v]) => v
```
jahooma, line 59: Should we test this as `floatingEquals(v, 0)`?

Otherwise we'd have to make sure there were no floating rounding errors when you sell. And then you still could have a problem if you bought NO shares on purpose to cancel your YES shares, it might not cancel exactly even if you tried.

jahooma (approved):
Thanks for the API improvements!

functions/src/sell-shares.ts
mqp: You're probably right about the test here.

If you are trying to exactly cancel your shares you should be calling the API with no `amount`, and then it should use whatever amount it thinks is the exact sum of all your outstanding orders... although then if it added them up in a different order the next time you called this API, perhaps it would not add up to zero again... so yeah, anyway, you're right about this check.

PR #708, opened 2022-08-01
Backfill missing group IDs

mqp:
There are missing group IDs on numerous groups in the dev database. Not sure how they got there, but now they will be fixed.

It doesn't seem like the problem exists on prod.


IanPhilips:
Just ran into this as well! Going to merge it bc I need it.

PR #710, opened 2022-08-01
Add endpoint for reading currently authenticated user

keriwarr:
Use case: an app that acts on behalf of a user may want to check certain values about the current user such as their id and balance.

Note: I didn't test this because I can't run the gcloud CLI on this computer


mqp:
I think this looks OK, I will test and merge soon.

mqp:
I changed this after the fact to be simpler because the approach in this PR doesn't really handle errors and stuff. The whole `LiteUser` stuff doesn't seem well designed for us right now anyway so I have no regrets.

PR #711, opened 2022-08-02
Allow admins to resolve any market

mqp:
This is very convenient for admins since the backend has access to stuff like the Mailgun key, so it's easier to call this API than to run resolution code locally.


akrolsmir:
LGTM!

PR #712, opened 2022-08-03
Random contract page fixups

mqp:
I was looking at this code to try to check out what the deal with performance is, and these are just some random not big deal things I found to improve along the way. (The big deal things will be harder.)

The deal with `useRef` is that `useState` is going to force a re-render when the ref is set, but `useRef` doesn't. So it's better to use `useRef` when possible (i.e. when the ref is only going to be used after rendering, which I am almost sure is the case here...)

### DIFF
--- a/web/components/avatar.tsx
+++ b/web/components/avatar.tsx
@@ -47,14 +47,21 @@ export function Avatar(props: {
   )
 }
 
-export function EmptyAvatar(props: { size?: number; multi?: boolean }) {
-  const { size = 8, multi } = props
+export function EmptyAvatar(props: {
+  className?: string
+  size?: number
+  multi?: boolean
+}) {
+  const { className, size = 8, multi } = props
   const insize = size - 3
   const Icon = multi ? UsersIcon : UserIcon
 
   return (
     <div
-      className={`flex flex-shrink-0 h-${size} w-${size} items-center justify-center rounded-full bg-gray-200`}
+      className={clsx(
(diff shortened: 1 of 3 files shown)


web/components/feed/feed-comments.tsx
```diff
@@ -60,14 +60,14 @@ export function FeedCommentThread(props: {
...
   function scrollAndOpenReplyInput(comment: Comment) {
     setReplyToUsername(comment.userUsername)
     setShowReply(true)
-    inputRef?.focus()
+    inputRef.current?.focus()
   }
   useEffect(() => {
-    if (showReply && inputRef) inputRef.focus()
+    if (showReply) inputRef.current?.focus()
   }, [inputRef, showReply])
```
jahooma, line 71: `inputRef` not necessary as a dep. Also, the semantics are a little different. This effect might not be re-run like it was before, since the setState would trigger a re-render when the ref was set.

I would just double check that this is working. Probably it is because it waits for `showReply` to be true, and the effect would run after the first render of the input that also assigns the ref.

jahooma (approved):
LGTM. Thanks Marshall!

web/components/feed/feed-comments.tsx
mqp: I agree that the semantics changed here. I think your reasoning is correct. I did test it and it seems to be working great. I'll probably triple check before merging that I exercised all the possible code paths.

mqp:
I removed the ref stuff because it became totally obsolete with a bunch of recent changes to comment inputs.

PR #716, opened 2022-08-05
Add Basic Lands to MTG Guesser

marsteralex:


akrolsmir (approved):
Hrm so with the basics:
- On reveal, it's much harder to see why I was wrong, because it's not clear what the expansion symbols actually are. Unlike name (which is text to text), the player is left guessing what the expansion symbols are, or has to go look them up. 

Solving this is probably kinda hard though -- I guess you could hypothetically put the expansion icon next to the set name?

- Would change the game title, maybe "Visual Basics" if you're okay with a different pun, or "How Basic" or something if you want to keep the same one. "Basic B****" looks/feels aesthetically off

- Might be nice to introduce some difficulty levels on the front page. For now, could do something like just a subjective star rating (e.g. Basic lands are 5*, Counterspells are 3*, Burn is 2*, Beasts are 1*). This helps people calibrate against how frustrated to feel when they get things wrong.

(Even cooler would be to track the number of right/wrong answers in some database somewhere, and aggregate them all and show them to the users e.g. "1230 correct / 2100 guesses". Would be pretty straightforward to do on a Firestore database! And then we could set up prediction markets on "What will the % of right answers be on Finding Fantastic Beasts after 1 month?"

But would suggest this as a follow up, for sure)

sipec:
vercel preview doesn't work when you press start game 😞 
(I assume it actually does work when merged)

sipec:
Does Alex need us to merge this for him? If so, that would feel weird to me since ultimately its his fun personal project.

akrolsmir:
LGTM, nice!

PR #718, opened 2022-08-05
Create challenge without previous market

IanPhilips:
Will allow for any user to create a contract & challenge at once from the /challenges page using the defaults:
- M$100 ante, so they're charged the challenge amount and the create market ante
- 30 day trading close (easy to update later if they need to)

Screen Shot 2022-08-05 at 8 35 36 AM


IanPhilips:
Creating the contract once the challenge is made is a bit more difficult as challenges are a subcollection of contracts. So this would mean we either want draft contracts (with zero liquidity) that are not displayed on the home page - (maybe just displayed on a user's market tab?) or a draft-challenges collection. I'm more interested in trying draft markets than creating a new collection personally.

PR #719, opened 2022-08-05
Rich text in comments, fixed

sipec:
Continuation of #703

Issue1: on android chrome, deleting an at mention creates spooky behavior (text being duplicated, keyboard dissappearing)
- discovered by Matt P
- caused by [bug in android chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=612446). See also https://github.com/ProseMirror/prosemirror/issues/565, https://github.com/ueberdosis/tiptap/issues/677
- I fixed text duplicate issues - now delete behaves as it should. However, the keyboard still disappears on Android
- tested on Gboard and Swiftkey

### DIFF
--- a/common/comment.ts
+++ b/common/comment.ts
@@ -1,3 +1,5 @@
+import type { JSONContent } from '@tiptap/core'
+
 // Currently, comments are created after the bet, not atomically with the bet.
 // They're uniquely identified by the pair contractId/betId.
 export type Comment = {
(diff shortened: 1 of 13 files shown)


jahooma:
Cool! One more thing that would be nice is to make sure that comments created within this change don't error on old clients.

It could be common for most people to have our site open already. They would probably get an error after navigating to a market with a new style comment and have to refresh. Could be a reason to populate the `comment.text` field as well, temporarily.

sipec:
just tested - email notifs work as expected

PR #722, opened 2022-08-06
Fix up several pages to load user data on the server

mqp:
These are pages where previously we were doing a bunch of `if (user == null)` stuff to render something ridiculous if the user wasn't loaded yet, so this should be a solid improvement to the UX and also clean up the code. (The profile page was especially egregious.)

### DIFF
--- a/web/lib/firebase/users.ts
+++ b/web/lib/firebase/users.ts
@@ -52,6 +52,11 @@ export async function getUser(userId: string) {
   return (await getDoc(doc(users, userId))).data()!
 }
 
+export async function getPrivateUser(userId: string) {
+  /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+  return (await getDoc(doc(users, userId))).data()!
+}
+
 export async function getUserByUsername(username: string) {
   // Find a user whose username matches the given username, or null if no such user exists.
   const q = query(users, where('username', '==', username), limit(1))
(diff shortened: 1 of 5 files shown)


mqp:
@akrolsmir Here's the fruits of the custom-token-Firebase-SDK-server-auth I talked your ear off about that one day; now we can just go ahead and write normal Firebase SDK code to fetch stuff in `getServerSideProps`.

jahooma (commented):
This is amazing! In addition to the code simplification, somehow it's unbelievably fast to load the first part of the user page.

In fact, I don't understand why it's so fast. It's like it has cached `user` as a serverside prop. Especially when navigating between pages. https://prod-git-ssr-users-mantic.vercel.app/JamesGrugett?tab=bets

jahooma (approved):

mqp:
> This is amazing! In addition to the code simplification, somehow it's unbelievably fast to load the first part of the user page.

> In fact, I don't understand why it's so fast. It's like it has cached user as a serverside prop. Especially when navigating between pages. https://prod-git-ssr-users-mantic.vercel.app/JamesGrugett?tab=bets

I think you're just remarking on the status quo? This PR didn't touch the user page yet, because, as you speculated, the user page is using `getStaticProps` to cache the information about the guy you are looking at. You're probably confusing it with the `/profile` route, which is the "edit user info" page. That's the one this PR made much simpler (and sure, faster too.)

PR #724, opened 2022-08-06
Pass page props user to auth provider if present

mqp:
This establishes a special magic prop, `user`, which if it's present in the page props (e.g. because we wrote a `getServerSideProps` that returns it), will be the initial value in the page `AuthContext`. This will save a re-render on the client when we load the user later from either local storage and/or from the actual Firebase SDK (assuming those things don't disagree with the version that came down from the server.) It will also kind of make the server side rendering more "useful" because it will actually be able to render using the authed user.

### DIFF
--- a/web/components/auth-context.tsx
+++ b/web/components/auth-context.tsx
@@ -1,4 +1,4 @@
-import { createContext, useEffect } from 'react'
+import { ReactNode, createContext, useEffect } from 'react'
 import { User } from 'common/user'
 import { onIdTokenChanged } from 'firebase/auth'
 import {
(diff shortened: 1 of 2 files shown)


web/components/auth-context.tsx
```diff
@@ -28,11 +28,11 @@ const ensureDeviceToken = () => {
...
-export function AuthProvider({ children }: any) {
-  const [authUser, setAuthUser] = useStateCheckEquality<AuthUser>(undefined)
+export const AuthContext = createContext<AuthUser>(undefined)
 
+export function AuthProvider(props: { children: ReactNode; user?: AuthUser }) {
+  const { children, user } = props
+  const [authUser, setAuthUser] = useStateCheckEquality<AuthUser>(user)
   useEffect(() => {
     const cachedUser = localStorage.getItem(CACHED_USER_KEY)
     setAuthUser(cachedUser && JSON.parse(cachedUser))
```
jahooma, line 38: If `user` was passed as a prop, shouldn't we prefer it to the one in localStorage?

jahooma (approved):
Cool, seems reasonable!

web/components/auth-context.tsx
mqp: Yeah, I guess so? It seems like this shouldn't be relevant to the normal case, but if the two disagree, the server one is certainly more up to date, so there's no reason to use the local storage one.

I should think about what to actually do with this local storage user cache.

PR #726, opened 2022-08-08
Inga/manifund

ingawei:
This is very half baked, a couple of notes:

- scraping ea is in `/functions/src/scripts/contest/scrape-ea.ts`
- contest metadata is in `/common/contest.ts`
- scraped submissions are in `/web/lib/util/contests/[contestFileName].json`

- front end for contests is here `http://localhost:3000/contest/cause-exploration-prize`
- I'm have not figured out how to take submission data from the json and turn it into a market automatically, the section where I try to do this and fail is marked in a comment prefaced by `INGA/TODO`

- this is a big mess at the moment and I'm very sorry lol, it might just be worth copying directly over the scraping script and contest metadata and then looking at everything else as reference


sipec:
having trouble getting the contest pages to run

gappleto97:
Here's a market for this: https://manifold.markets/MarketManagerBot/how-many-days-after-september-1st-w

PR #727, opened 2022-08-09
add bets and user tables per market

Gurkenglas:
So it turns out you can add a bets table that lets the user sort and filter using like one line. Then I added some more lines for formating and styling and another table of each user's position. Feel free to redo any or all of it or give feedback.
Screenshot from 2022-08-09 02-50-05
Screenshot from 2022-08-09 02-50-13

### DIFF
--- a/web/components/contract/contract-tabs.tsx
+++ b/web/components/contract/contract-tabs.tsx
@@ -16,7 +16,7 @@ import { Avatar } from 'web/components/avatar'
 import { Grid, _ } from 'gridjs-react'
 import 'gridjs/dist/theme/mermaid.css'
 import { useState, useEffect, useRef } from 'react'
-import { maxBy, uniq } from 'lodash'
+import { maxBy } from 'lodash'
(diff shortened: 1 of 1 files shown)


Gurkenglas:
Note that this, uh, pays no attention to whether a bet is too old to have its owner named.

sipec:
I love the concept / implementation. I think a table display of bets and users that have bet is very useful.

I think it's worth it for me to take over the design work on this and finish it up, but probably not this week :/

Gurkenglas:
Eh, I can just use vercel's deployment, it's everyone else's loss if they don't have such tables. Except that annoyingly I can't log in through it.

jahooma:
Added the vercel deployment to the authorized domain whitelist, so you should be able to login now

gappleto97:
Here's a market for this:  https://manifold.markets/MarketManagerBot/how-many-days-after-september-1st-w-77e92d9620d1

PR #728, opened 2022-08-09
Backend robustness to email sending or analytics tracking failures

mqp:
It makes way more sense to log if these things fail rather than to blow up whatever they were in the middle of.


jahooma (approved):
Nice idea!

PR #729, opened 2022-08-09
Make homepage load user via SSR, pass it to contract stuff

mqp:
Besides avoiding re-rendering performance hits, one reason this will be nice is that SSR will now render a correct logged-in sidebar on this page and a correct customized contract search header, eliminating any loading jank. Since the homepage is what people are going to be landing on the most, it's nice to make it load as cleanly as possible.


jahooma (approved):
LGTM!

PR #730, opened 2022-08-09
Fix efficiency problems with visibility checking code

mqp:
This code was responsible for:

1. Loading more stuff on the homepage when you scroll to the bottom (very important!)
2. Tracking meaningless analytics in the database and local storage when you scroll to the part of the contract page that says "comments" and "bets" (pointless???)

There were basically two problems with it:

1. It stored the latest results from the `IntersectionObserver` in React state, but there was no reason to do this. That means it would trigger re-renders whenever the stuff that was being monitored came in or out of view.
2. The interface to it was a React hook that stored this state where it was being used; that place was in the root of giant components like the `ContractsGrid` and `FeedItems` which are slow to render, so re-renders on them were ultra painful.

Now both of these problems are fixed. For the first problem, we simply do not do that. For the second problem, I killed the old hook and instead packaged this stuff up in a separate `VisibilityObserver` component, which monitors itself. So now when it re-renders due to a change in internal state (which it still does once, to set up the `IntersectionObserver` object when the element ref is available) it doesn't re-render some giant parent, it just re-renders itself.


mqp:
@jahooma since he implemented the stuff I am rewriting here.

web/components/visibility-observer.tsx
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from 'react'
+
+export function VisibilityObserver(props: {
+  className?: string
+  onVisibilityUpdated: (visible: boolean) => void
+}) {
+  const { className, onVisibilityUpdated } = props
```
jahooma, line 7: One way to make this slightly more robust is to do
```const onVisibilityUpdated = useEvent(props.onVisibilityUpdated)```

Then it doesn't matter if the caller uses useCallback, the reference to onVisibilityUpdated won't change.

File: web/lib/firebase/seen-contracts.ts [hunk 2]
```diff
@@ -0,0 +1,22 @@
+import { mapValues } from 'lodash'
```
jahooma, line 1: This is good, although I would have just as soon removed it all, b/c I'm not sure we're using the seen contracts in any way today.  (It used to track contracts seen in the old feed.)

Actually, maybe we should still axe this and the visibility observer within the feed items, just to simplify.

jahooma (approved):
Thanks Marshall! Good to cut down on unnecessary re-renders.

web/lib/firebase/seen-contracts.ts
mqp: If you're cool with it of course I would love to axe it!

web/components/visibility-observer.tsx
mqp: OK, I'll defer to you on what you think good style is with regard to this stuff.

PR #731, opened 2022-08-09
Fix broken `useBets` filter implementation

mqp:
This was basically causing an infinite database-spam-and-render loop on the contract page (the only place where we use the filter option is the `ContractActivity` component on that page) because the filter option was an object literal defined in the parent component. So it would call this, this would run and set the bets, it would re-render, creating a new options object, so this would run and set the bets again, and so on.


jahooma (approved):
Thanks for discovering this! I think we had some reports of sluggishness and crashing.

PR #735, opened 2022-08-09
Editor improvements

sipec:


jahooma (approved):
Awesome! This is a great improvement, and fixes all the issues we thought up!

Makes it much more delightful to use. Nice work, Sinclair!

PR #736, opened 2022-08-10
Simplify and generalize Button component

sipec:
this allows `<Button>` to have any attributes that `<button>` accepts like autofocus, role=link, ref, ...

### DIFF
--- a/web/components/button.tsx
+++ b/web/components/button.tsx
@@ -1,31 +1,25 @@
-import { ReactNode } from 'react'
 import clsx from 'clsx'
 
-export function Button(props: {
-  className?: string
-  onClick?: () => void
-  children?: ReactNode
-  size?: '2xs' | 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2xl'
-  color?:
-    | 'green'
-    | 'red'
-    | 'blue'
-    | 'indigo'
-    | 'yellow'
-    | 'gray'
-    | 'gradient'
-    | 'gray-white'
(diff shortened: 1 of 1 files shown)


jahooma (commented):
Not sure if I like this change. It makes the main props you want to use less obvious, and adds a whole lot of cruft from the html spec.

Even if we need to rewrite the props we actually use, I think that's better.

Plus, we should think of Button as higher level. It should take config that is then translated into html attributes and css and such.

mqp:
I'm like the opposite of James, if the purpose of this work is to establish a common design language of colors and sizes, I would delete the `Button` component and instead expose mappings like
```typescript
const COLORS = {
  'red': { bg: 'bg-red-400 hover:bg-red-500', fg: 'text-white' }
  ...
} as const
const SIZES = {
  '2xs': { padding: 'px-2 py-1', text: 'text-xs' }
  ...
} as const
export const size(s: keyof typeof SIZES) {
  return clsx(SIZES[s].values())
}
export const color(c: keyof typeof COLORS) {
  return clsx(COLORS[c].values())
}
```
and then later do stuff like
```typescript
<button className={clsx(size('2xs'), color('red'))} ... />
```
It's easier to be able to use normal HTML elements in JSX instead of having to look up how the Button wrapper component works. And it's nice to be able to use the design language stuff outside of the context of a `<button>` element.

(If the Button wrapper component did more stuff, it would make sense to me. But not if it's just applying some CSS classes that anyone could apply.)

PR #737, opened 2022-08-10
Add cache headers to avatars

mqp:
We forgot to add cache headers to any uploaded avatars, presumably because we also didn't used to give them a unique filename. But now we do, so we can cache them.


sipec (approved):
🎉

PR #738, opened 2022-08-10
Teach `AuthContext` to manage the private user doc

mqp:
Previously, `AuthContext` tracked the user doc, and if you wanted the private user doc, well, you could go get it yourself.

Nowadays, literally every logged-in user page on our website needs the private user doc eventually, because it's used to serve information about notifications, which show up in the sidebar. There's no reason for the auth context to not just look it up too as soon as it knows who you are logged in as, instead of waiting until someone asks for it later, since doing two requests (for the user and private user) in parallel should be almost as fast as doing one anyway.

So now, the auth context fetches the private user doc when you log in, and SSR also will pass it through to the auth context if fetched, and the `usePrivateUser` hook gets it from the auth context.

### DIFF
--- a/web/components/auth-context.tsx
+++ b/web/components/auth-context.tsx
@@ -53,7 +53,7 @@ export function AuthProvider(props: {
         let user = await getUser(fbUser.uid)
         if (!user) {
           const deviceToken = ensureDeviceToken()
-          user = (await createUser({ deviceToken })) as User
+          user = (await createUser({ deviceToken })).user as User
         }
         setAuthUser(user)
         // Persist to local storage, to reduce login blink next time.
--- a/web/components/auth-context.tsx
+++ b/web/components/auth-context.tsx
@@ -1,7 +1,7 @@
 import { ReactNode, createContext, useEffect } from 'react'
-import { User, PrivateUser } from 'common/user'
 import { onIdTokenChanged } from 'firebase/auth'
 import {
+  UserAndPrivateUser,
   auth,
   listenForUser,
   listenForPrivateUser,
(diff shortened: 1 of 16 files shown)


web/components/auth-context.tsx
```diff
@@ -14,10 +15,10 @@ import { identifyUser, setUserProperty } from 'web/lib/service/analytics'
 import { useStateCheckEquality } from 'web/hooks/use-state-check-equality'
 
 // Either we haven't looked up the logged in user yet (undefined), or we know
-// the user is not logged in (null), or we know the user is logged in (User).
-type AuthUser = undefined | null | User
+// the user is not logged in (null), or we know the user is logged in.
+type AuthUser = undefined | null | UserAndPrivateUser
 
-const CACHED_USER_KEY = 'CACHED_USER_KEY'
+const CACHED_USER_KEY = 'CACHED_USER_KEY_V2'
```
mqp, line 21: The format of the data in `localStorage` changed (it now stores the private user info too) so I bumped this.

web/hooks/use-state-check-equality.ts
```diff
@@ -8,8 +8,9 @@ export const useStateCheckEquality = <T>(initialState: T) => {
   stateRef.current = state
 
   const checkSetState = useMemo(
-    () => (newState: T) => {
+    () => (next: SetStateAction<T>) => {
       const state = stateRef.current
+      const newState = next instanceof Function ? next(state) : next
```
mqp, line 13: Please never use this hook to try to set state which is a function, but not an updater function, or I will cry.

jahooma: 🙃

jahooma (approved):
This change was a good idea! Thanks Marshall!

PR #740, opened 2022-08-10
Componentize confetti to isolate re-renders due to window size

mqp:
Previously, the user and contract page would do extra re-renders of the whole content area on load (and on resize) due to the `useWindowSize` hook. Now that hook serves only the `FullscreenConfetti` component.

### DIFF
--- a/web/components/fullscreen-confetti.tsx
+++ b/web/components/fullscreen-confetti.tsx
@@ -1,0 +1,8 @@
+import Confetti, { Props as ConfettiProps } from 'react-confetti'
+import { useWindowSize } from 'web/hooks/use-window-size'
+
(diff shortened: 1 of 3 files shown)


web/components/fullscreen-confetti.tsx
```diff
@@ -0,0 +1,8 @@
+import Confetti, { Props as ConfettiProps } from 'react-confetti'
+import { useWindowSize } from 'web/hooks/use-window-size'
+
+export function FullscreenConfetti(props: ConfettiProps) {
+  const { width, height } = useWindowSize()
+  console.log('Width: ', width, 'Height: ', height)
```
jahooma, line 6: Remove console.log?

jahooma (approved):
Makes sense, LGTM!

PR #741, opened 2022-08-10
Move liquidity provision fetch down into `ContractTabs`

mqp:
That's the only thing that needs it, so no point re-rendering the rest (e.g. re-rendering the chart is very expensive.)

Actually, only the bets tab needs it, but our tab design is kind of fucked, so further improvements can come later.


jahooma (approved):
LGTM!

PR #742, opened 2022-08-10
Use `UserFollowButton` instead of `FollowButton` in `UserPage`

mqp:
Both cleaner and more efficient (since the `yourFollows` state is no longer at the top level of the user page.)


jahooma (approved):
LGTM! Thanks for the cleanup.

PR #745, opened 2022-08-11
Bail out earlier if `createmarket` is called with invalid group ID

mqp:
Previously, if you passed an invalid group ID, it would create the contract, and charge you money, but then error out and not do any of the other things you need to do for a new contract! Very bad. Now it validates the group up front.

### DIFF
--- a/functions/src/create-contract.ts
+++ b/functions/src/create-contract.ts
@@ -133,6 +133,26 @@ export const createmarket = newEndpoint({}, async (req, auth) => {
   if (ante > user.balance)
     throw new APIError(400, `Balance must be at least ${ante}.`)
 
+  let group = null
+  if (groupId) {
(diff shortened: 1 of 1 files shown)


functions/src/create-contract.ts
```diff
@@ -133,6 +133,26 @@ export const createmarket = newEndpoint({}, async (req, auth) => {
   if (ante > user.balance)
     throw new APIError(400, `Balance must be at least ${ante}.`)
 
+  let group = null
```
jahooma, line 136: Should this be typed better? `let group: Group | null`

jahooma (approved):
Thanks for fixing this, Marshall!

functions/src/create-contract.ts
mqp: Yes, I guess it should.

PR #747, opened 2022-08-11
Only load portfolio history inside user page bets tab

mqp:
Now that the tab state is persistent, we can easily move the portfolio history loading in here, so it doesn't have to re-render the main page when it loads.


jahooma (approved):
Looks good, thanks!

PR #750, opened 2022-08-12
Tweak Firestore user rules to be more robust

mqp:
The reason this is smart, is because if the Firestore rules refer to the document data, it will blow up requests that try to get or update a document that either doesn't exist or doesn't have that data (e.g. due to a bug.) So it's just better to not do that if the same data is in both the path and the document.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -20,17 +20,17 @@ service cloud.firestore {
 
     match /users/{userId} {
       allow read;
-      allow update: if resource.data.id == request.auth.uid
+      allow update: if userId == request.auth.uid
                        && request.resource.data.diff(resource.data).affectedKeys()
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Makes sense, I think. Would we want the db change to go through though if the underlying doc was deleted?

mqp:
> Makes sense, I think. Would we want the db change to go through though if the underlying doc was deleted?

I'm not sure what will happen if you call `update` on a nonexistent doc, but I would rather have "whatever usually happens" instead of "the security rules throw an exception." I got this "security rules throw an error" business on a read and it took me ten minutes to even figure out what was the deal, because the error message is very opaque.

jahooma:
Good point, the security rule violations have the most unhelpful error messages

PR #752, opened 2022-08-12
fix dependency peer-dep warnings, mostly

sipec:
There's two remaining.

For the firebase/data-compat warning we'd need to just upgrade `firebase`. See: https://github.com/firebase/firebase-js-sdk/pull/6319
For the other, we'd need to wait for nivo to stop using `react-motion` https://github.com/plouc/nivo/issues/667


mqp:
> [](https://github.com/firebase/firebase-js-sdk/pull/6319)
> For the other, we'd need to wait for nivo to stop using `react-motion` [plouc/nivo#667](https://github.com/plouc/nivo/issues/667)

I think the `react-motion` dep is already removed in the latest Nivo, but when I tried to upgrade us a month or two ago a bunch of mysterious stuff blew up. Probably if you spend half an hour figuring out what changed you could upgrade it. The `react-motion` dependency isn't small so getting rid of it would be a nice win.

PR #753, opened 2022-08-13
Clean up a bunch of markup and CSS on contract cards

mqp:
Rendering the contract cards is surprisingly expensive, so just cleaning up obviously unnecessary layers of cruft that has accumulated on them. Should be no visible changes.

### DIFF
--- a/web/components/contract/contract-card.tsx
+++ b/web/components/contract/contract-card.tsx
@@ -68,48 +68,41 @@ export function ContractCard(props: {
   return (
     <Row
       className={clsx(
-        'relative gap-3 self-start rounded-lg bg-white py-4 pl-6 pr-5 shadow-md hover:cursor-pointer hover:bg-gray-100',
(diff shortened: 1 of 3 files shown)


sipec (approved):

sipec:
> Rendering the contract cards is surprisingly expensive

I think we should start virtualizing lists - I've moved that up in priority in the tracker

mqp:
> > Rendering the contract cards is surprisingly expensive
> 
> I think we should start virtualizing lists - I've moved that up in priority in the tracker

I'm optimistic that it won't be necessary for lists of size e.g. ~50 if I just continue to profile, fix up the slow parts, and kill unnecessary re-renders, but it might be.

jahooma:
If we only render the next 20 cards at a time, I don't really see the benefit of virtualizing. In my experience, that would just make scrolling less responsive. Modern browsers have the memory to handle large lists.

PR #754, opened 2022-08-13
Make `RelativeTimestamp` a little more efficient

mqp:
These timestamps are most of the cost of rendering a big feed of comments or bets. I think basically the real culprit here is the tooltip code, which is doing a ton of work to compute where the tooltip should be, but might as well fix the easy stuff.

`dayjs(...)` isn't that expensive -- it's constructing a few objects and calling a few different things -- but it's expensive enough that we shouldn't call it thousands of extra times per render just for fun.


sipec:
oh I accidentally merged this. had too many tabs open.
this pr does actually look good to me though

PR #755, opened 2022-08-13
Improve performance of tooltip

sipec:
Currently chats are super slow to scroll, because every single tooltip has to recalculate its position.
To fix, we only render the tooltip when it is visible, rather than using `:hover` and `:focus` css.

See also the warning in: https://floating-ui.com/docs/react-dom#updating

### DIFF
--- a/web/components/tooltip.tsx
+++ b/web/components/tooltip.tsx
@@ -6,9 +6,14 @@ import {
   Placement,
   shift,
   useFloating,
-} from '@floating-ui/react-dom'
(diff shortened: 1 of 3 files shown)


jahooma (approved):
Thanks for looking into this! I tried it out, and it scrolls group chat much smoother than on the live site. LGTM!

sipec:
i suspected it was the issue but I didn't update my posterior all the way until seeing mqp suspect the same thing in his pr

PR #756, opened 2022-08-13
Simplify and fix inefficiencies in contract search component

mqp:
Previously, when you (for example) scrolled to the bottom of the contract search grid, this would happen:

1. The `loadMore` callback would be called.
2. It would call `setPage`, triggering a re-render of the grid. (This is obviously fishy since visually, absolutely nothing changed.)
3. The change in the `page` prop would trigger an effect that is looking at many dependencies and performs a new search query.
4. Later, your query would produce results.
5. `setHitsByPage` and `setNumPages` would be called. This would actually trigger two re-renders, because in React <18, state changes that aren't directly inside the React render code aren't batched.

Now:

1. The `loadMore` callback is called.
2. It directly invokes `queryNextPage`, instead of changing state, so no re-render happens.
3. Later, your query would produce results.
4. `setPages` and `setNumPages` are called, but we use a special React API to batch them, so this causes one re-render.

That's the useful change. Other than that, it just got a little simpler all around. `Contract[][]` is an easier data structure to work with than `{ [page: string]: Contract[] }`, and I changed the "superseding in-flight requests with new queries" logic slightly so that it doesn't have to rely on `useEffect` as the only interface to invoking the query machinery.

### DIFF
--- a/web/components/contract-search.tsx
+++ b/web/components/contract-search.tsx
@@ -13,7 +13,8 @@ import {
   ContractsGrid,
 } from './contract/contracts-grid'
 import { Row } from './layout/row'
-import { useEffect, useMemo, useState } from 'react'
+import { useEffect, useRef, useMemo, useState } from 'react'
+import { unstable_batchedUpdates } from 'react-dom'
 import { ENV, IS_PRIVATE_MANIFOLD } from 'common/envs/constants'
 import { useFollows } from 'web/hooks/use-follows'
 import { track, trackCallback } from 'web/lib/service/analytics'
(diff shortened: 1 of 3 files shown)


mqp:
You know, maybe I should feel a little bad about this change, because although it works fine today, it currently wouldn't automatically refresh the search results if you did a `setState` of one of the search parameters manually (outside of the UI event handlers that update that.)

I think I could fix that without too much hassle so I should probably do so before merging this.

sipec (approved):

web/components/contract-search.tsx
```diff
@@ -167,80 +169,67 @@ export function ContractSearch(props: {
...
+    setNumPages(1)
     track('select search filter', { filter: newFilter })
   }
 
   const selectSort = (newSort: Sort) => {
     if (newSort === sort) return
-
-    setPage(0)
     setSort(newSort)
+    setPages([])
```
jahooma, line 231: Setting pages to empty here would mean that while the next search is happening, we'd flash empty results (or a loading indicator), right?

The previous behavior is to keep showing the old search until the next search is loaded. It's less jumpy for the user.

Ditto for `updateQuery` and `selectFilter`

jahooma (changes requested):
I admire the goal to reduce unnecessary re-renders! The nested array of pages is also a good idea.

Probably would be a good idea to react to props changes with new search results.

web/components/contract-search.tsx
mqp: Oh, I actually didn't realize the old behavior was different. I personally like the new behavior, but I don't have a strong preference, so if you like the old behavior, I will change it back.

mqp:
OK, all concerns are fixed up.

PR #757, opened 2022-08-13
Move search controls into separate component

mqp:
I think this change is good because typing in the search bar needs to be responsive for the search to feel good, so it sucks if that typing triggers a re-render on the whole grid of loaded contracts. Anyway, it's nice to put things in components.

The diff looks scary, but I didn't change much -- basically I just uprooted all the "figure out what search to do" code into a separate component that calls an `onSearchParametersChanged` callback to do the search with the new parameters.

(P.S. The search still isn't perfectly snappy yet. The main thing I need to fix next to make it snappy is the `useQueryAndSort` machinery, which A) literally throws out characters you type B) basically triggers re-renders on the whole page when the query parameter is updated.)

(P.P.S. I might make some additional components here later, like breaking out the pill stuff into a component. But probably not because it's important for efficiency, just because I think it would be a little tidier.)

### DIFF
--- a/web/components/contract-search.tsx
+++ b/web/components/contract-search.tsx
@@ -22,7 +22,7 @@ import { useFollows } from 'web/hooks/use-follows'
 import { track, trackCallback } from 'web/lib/service/analytics'
 import ContractSearchFirestore from 'web/pages/contract-search-firestore'
 import { useMemberGroups } from 'web/hooks/use-group'
-import { Group, NEW_USER_GROUP_SLUGS } from 'common/group'
+import { NEW_USER_GROUP_SLUGS } from 'common/group'
 import { PillButton } from './buttons/pill-button'
 import { sortBy } from 'lodash'
 import { DEFAULT_CATEGORY_GROUPS } from 'common/categories'
@@ -204,10 +204,8 @@ function ContractSearchControls(props: {
     (group) => group.contractIds.length
   ).reverse()
 
-  const defaultPillGroups = DEFAULT_CATEGORY_GROUPS as Group[]
-
-  const pillGroups =
-    memberPillGroups.length > 0 ? memberPillGroups : defaultPillGroups
+  const pillGroups: { name: string; slug: string }[] =
+    memberPillGroups.length > 0 ? memberPillGroups : DEFAULT_CATEGORY_GROUPS
 
   const [filter, setFilter] = useState<filter>(
     querySortOptions?.defaultFilter ?? 'open'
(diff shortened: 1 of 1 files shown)


web/components/contract-search.tsx
```diff
@@ -227,115 +282,103 @@ export function ContractSearch(props: {
...
-              )
-            })}
-          </Row>
-        )}
-      </Col>
-
-      {filter === 'personal' &&
-      (follows ?? []).length === 0 &&
-      memberGroupSlugs.length === 0 ? (
-        <>You're not following anyone, nor in any of your own groups yet.</>
```
mqp, line 328: I believe this was dead code, since if someone has no groups, we treat them as if they are subscribed to the default groups for the purposes of this querying code. So `memberGroupSlugs.length` would never be 0.

web/components/contract-search.tsx
```diff
@@ -80,6 +92,105 @@ export function ContractSearch(props: {
...
 
+  const [numPages, setNumPages] = useState(1)
+  const [pages, setPages] = useState<Contract[][]>([])
+  const [showTime, setShowTime] = useState<ShowTime | undefined>()
+
+  const searchParameters = useRef<SearchParameters | undefined>()
+  const requestId = useRef(0)
+
+  const performQuery = async (freshQuery?: boolean) => {
+    if (searchParameters.current == null) {
```
jahooma, line 103: I would use `undefined` instead of `null` if that's what it is, even if `==` matches both.

web/components/contract-search.tsx
```diff
@@ -80,6 +92,105 @@ export function ContractSearch(props: {
...
   } = props
 
+  const [numPages, setNumPages] = useState(1)
+  const [pages, setPages] = useState<Contract[][]>([])
+  const [showTime, setShowTime] = useState<ShowTime | undefined>()
+
+  const searchParameters = useRef<SearchParameters | undefined>()
+  const requestId = useRef(0)
+
+  const performQuery = async (freshQuery?: boolean) => {
```
jahooma, line 102: Wouldn't it be nice to make this take an argument of `searchParameters`, which could include the boolean `freshQuery` as a prop?

Then you wouldn't need the Ref.

web/components/contract-search.tsx
jahooma: Cool.

jahooma (approved):
LGTM! Will be nice to see search operating in perfect condition!

web/components/contract-search.tsx
mqp: I believe I would need the ref, because we need to know "the last search parameters we used to load the grid" for when you scroll to the bottom and want to load a new page. Do you have a different plan?

jahooma: Oh, I see. I don't have another plan!

PR #758, opened 2022-08-14
Rewrite `useQueryAndSortParams` machinery to be faster/simpler/better

mqp:
OK, so the previous state of affairs looked like this:

1. There existed a hook called `useQueryAndSortParams` which served the query box and "sort" dropdown on contract grid search areas.
2. It's hard to explain exactly what it did because it had a bunch of flags, but basically, it would maintain some combination of React state, URL query string state, and local storage state corresponding to the most recent query and sort you did, and handed that back to the caller along with setters.
3. One problem with this code is that it was confusing.
4. Another problem was that it used the Next.js `Router.replace` API, which does a [ton of work](https://github.com/vercel/next.js/blob/canary/packages/next/shared/lib/router/router.ts#L952), including but not limited to re-rendering the page root. This is fine in many cases, but it's really not fine when you are typing into a search box at 100 WPM.
5. Another problem is that someone noticed problem number 4, and their solution was to debounce the query string updating, but this is literally debouncing the updates to the letters you type in the box, which is really bad. (Try mashing the keyboard and you will see that it doesn't catch all the stuff you typed.)
6. Another problem was that the way our application used this code didn't make sense to me. For example, the 'adding markets to a group' box and the 'user profile page market list' would fiddle with the URL parameters and use the saved sort from the homepage, which seems foolish.

So I rewrote this into something that seems better:

1. There exist hooks called `useQuery` and `useSort` which are a `useState` that optionally load their state from the URL parameters and keep those URL parameters updated with the state.
2. We do local storage management from outside.
3. It's fast because we just use the browser history API instead of the Next.js machinery.
4. Instead of debouncing the state update, we just debounce the actual queries to Algolia.
5. I changed all the different places that render a contract search widget so that none of them except the home page dick around with the URL or with local storage at all.

Now the typing responsiveness when you use the search box is very good. The only problem is the noticeable hiccup when it actually re-renders a new list of contracts on the screen, since rendering 20 contracts takes a decent chunk of time. Probably I can make future improvements to make that less bad.

### DIFF
--- a/web/components/contract-search.tsx
+++ b/web/components/contract-search.tsx
@@ -20,7 +20,7 @@ import ContractSearchFirestore from 'web/pages/contract-search-firestore'
 import { useMemberGroups } from 'web/hooks/use-group'
 import { NEW_USER_GROUP_SLUGS } from 'common/group'
 import { PillButton } from './buttons/pill-button'
-import { sortBy } from 'lodash'
+import { debounce, sortBy } from 'lodash'
 import { DEFAULT_CATEGORY_GROUPS } from 'common/categories'
 import { Col } from './layout/col'
 import { safeLocalStorage } from 'web/lib/util/local'
@@ -147,6 +147,12 @@ export function ContractSearch(props: {
     }
   }
 
+  const onSearchParametersChanged = useRef(
+    debounce((params) => {
+      searchParameters.current = params
+      performQuery(true)
+    }, 100)
+  )
   const contracts = pages
     .flat()
     .filter((c) => !additionalFilter?.excludeContractIds?.includes(c.id))
@@ -166,10 +172,7 @@ export function ContractSearch(props: {
         useQuerySortLocalStorage={useQuerySortLocalStorage}
         useQuerySortUrlParams={useQuerySortUrlParams}
         user={user}
-        onSearchParametersChanged={(params) => {
-          searchParameters.current = params
-          performQuery(true)
-        }}
+        onSearchParametersChanged={onSearchParametersChanged.current}
       />
       <ContractsGrid
         contracts={pages.length === 0 ? undefined : contracts}
--- a/web/components/contract-search.tsx
+++ b/web/components/contract-search.tsx
@@ -152,7 +152,8 @@ export function ContractSearch(props: {
       searchParameters.current = params
       performQuery(true)
     }, 100)
-  )
+  ).current
+
   const contracts = pages
     .flat()
     .filter((c) => !additionalFilter?.excludeContractIds?.includes(c.id))
@@ -172,7 +173,7 @@ export function ContractSearch(props: {
         useQuerySortLocalStorage={useQuerySortLocalStorage}
         useQuerySortUrlParams={useQuerySortUrlParams}
         user={user}
-        onSearchParametersChanged={onSearchParametersChanged.current}
+        onSearchParametersChanged={onSearchParametersChanged}
       />
       <ContractsGrid
         contracts={pages.length === 0 ? undefined : contracts}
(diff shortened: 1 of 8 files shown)


web/components/contract-search.tsx
```diff
@@ -132,31 +147,32 @@ export function ContractSearch(props: {
     }
   }
 
+  const onSearchParametersChanged = useRef(
```
jahooma, line 150: An alternative way to do this would be to use `useState`:

```
const [onSearchParametersChanged] = useState(() => debounce(...))
```

Initializes it once in a way where it is obvious it cannot change (whereas with refs you expect mutation).

web/hooks/use-sort-and-query-params.tsx
```diff
@@ -15,92 +11,52 @@ export type Sort =
...
+  return newUrl
+}
 
-  const [queryState, setQueryState] = useState(query)
-
-  useEffect(() => {
-    setQueryState(query)
-  }, [query])
+function updateURL(params: UpdatedQueryParams) {
+  // see relevant discussion here https://github.com/vercel/next.js/discussions/18072
```
jahooma, line 31: Good to know!

web/hooks/use-sort-and-query-params.tsx
```diff
@@ -15,92 +11,52 @@ export type Sort =
...
+}
 
-  const setQuery = (query: string | undefined) => {
-    setQueryState(query)
-    if (!disableQueryString) {
-      pushQuery(query)
-    }
+export function useQuery(defaultQuery: string, useUrl: boolean) {
+  const router = useRouter()
+  const initialQuery = useUrl ? getStringURLParam(router, 'q') : null
```
jahooma, line 44: Seems like this wouldn't work because on the first render the router is usually "not ready" and doesn't have the query properties defined. I could be wrong.

web/pages/contract-search-firestore.tsx
```diff
@@ -21,10 +16,10 @@ export default function ContractSearchFirestore(props: {
   }
 }) {
   const contracts = useContracts()
-  const { querySortOptions, additionalFilter } = props
+  const { additionalFilter } = props
 
-  const { query, setQuery, sort, setSort } =
-    useQueryAndSortParams(querySortOptions)
+  const [query, setQuery] = useQuery('', true)
```
jahooma, line 21: Not a big fan of boolean arguments because as a reader you have to check the definition to see what it's doing.

An alternative is an `options` object, e.g.

```
useQuery(defaultQuery: string, options: { useUrl: boolean })
```

jahooma (approved):
Thanks Marshall, this does look a like a big step forward in making the code cleaner and more correct.

web/components/contract-search.tsx
mqp: Hmm, I don't share your intuition at all, but I will change it if you like.

I note that in other parts of our codebase we use `useMemo` or `useCallback` for this purpose, but React explicitly calls that out as being incorrect.

web/hooks/use-sort-and-query-params.tsx
mqp: That doesn't really make any conceptual sense to me, since obviously the URL is available from time 0 on the client, and the code seems to work fine, but I will double check.

web/pages/contract-search-firestore.tsx
mqp: OK, I am kind of indifferent but I understand your POV so I will change it.

web/components/contract-search.tsx
jahooma: Another pattern is like what you have, except you call .current immediately: `useRef().current`. That also works because there's no way to modify it.

mqp: That makes sense to me, so if you like it I will do it.

web/hooks/use-sort-and-query-params.tsx
mqp: FYI, I can verify that when I test, on the first render both on client and server, `router.isReady` is true and `router.query` is populated.

PR #760, opened 2022-08-15
Denormalize some contract comment fields

mqp:
Now we have a `contractSlug` and `contractQuestion`.

This is a big deal for the user profile, because otherwise we have to load the contract doc for every comment. Now we don't.

### DIFF
--- a/web/components/comments-list.tsx
+++ b/web/components/comments-list.tsx
@@ -1,12 +1,8 @@
 import { useEffect, useState } from 'react'
-import { Dictionary, keyBy, uniq } from 'lodash'
 
 import { Comment } from 'common/comment'
-import { Contract } from 'common/contract'
-import { filterDefined, groupConsecutive } from 'common/util/array'
-import { contractPath } from 'web/lib/firebase/contracts'
+import { groupConsecutive } from 'common/util/array'
 import { getUsersComments } from 'web/lib/firebase/comments'
-import { getContractFromId } from 'web/lib/firebase/contracts'
 import { SiteLink } from './site-link'
 import { Row } from './layout/row'
 import { Avatar } from './avatar'
@@ -20,12 +16,20 @@ import { LoadingIndicator } from './loading-indicator'
 
 const COMMENTS_PER_PAGE = 50
 
-type ContractComment = Comment & { contractId: string }
+type ContractComment = Comment & {
+  contractId: string
+  contractSlug: string
+  contractQuestion: string
+}
+
+function contractPath(slug: string) {
+  // in honor of austin, who insists that contract URLs are prefixed with a username
+  return `/Austin/${slug}`
+}
 
 export function UserCommentsList(props: { user: User }) {
   const { user } = props
   const [comments, setComments] = useState<ContractComment[] | undefined>()
-  const [contracts, setContracts] = useState<Dictionary<Contract> | undefined>()
   const [page, setPage] = useState(0)
   const start = page * COMMENTS_PER_PAGE
   const end = start + COMMENTS_PER_PAGE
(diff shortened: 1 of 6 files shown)


web/components/comments-list.tsx
```diff
@@ -20,12 +16,20 @@ import { LoadingIndicator } from './loading-indicator'
 
 const COMMENTS_PER_PAGE = 50
 
-type ContractComment = Comment & { contractId: string }
+type ContractComment = Comment & {
```
jahooma, line 19: Would it be worth moving this type into `comment.ts` and exporting it for others to use?

web/components/comments-list.tsx
```diff
@@ -37,34 +41,23 @@ export function UserCommentsList(props: { user: User }) {
...
   }, [user.id])
 
-  useEffect(() => {
-    if (comments) {
-      const contractIds = uniq(comments.map((c) => c.contractId))
-      Promise.all(contractIds.map(getContractFromId)).then((contracts) => {
-        setContracts(keyBy(filterDefined(contracts), 'id'))
-      })
-    }
-  }, [comments])
```
jahooma, line 47: Nice to be able to remove!

web/components/comments-list.tsx
```diff
@@ -20,12 +16,20 @@ import { LoadingIndicator } from './loading-indicator'
...
+type ContractComment = Comment & {
+  contractId: string
+  contractSlug: string
+  contractQuestion: string
+}
+
+function contractPath(slug: string) {
+  // in honor of austin, who insists that contract URLs are prefixed with a username
+  return `/Austin/${slug}`
+}
```
jahooma, line 28: Hmm, we know the comment's `userUsername`, can't we just pass that in? This is just a joke, right?

jahooma (approved):
Nice change! This'll make the user page a lot snappier, and hopefully fix any loading problems.

Please do fill in their actual username in the contract path instead of Austin.

web/components/comments-list.tsx
mqp: No, this is for the URL to the contract. It wouldn't make any more sense for it to be the commenter's name. It's "supposed" to be the contract creator's name, but it has no effect.

web/components/comments-list.tsx
mqp: This is basically analogous to our contracts and the typing "should" be designed in the same way, but I don't know whether any other code really would care about having more precise typing on this right now. I'll fix it up if you like.

web/components/comments-list.tsx
mqp: If you are a robot who doesn't think it's funny to put "Austin" the most dignified thing would be to put "market" or something.

jahooma: I see, but users might share the URL after and then it would not have the correct info. Referral bonuses are also based on the creator of the contract. 

I don't think our users will get this joke. And why put "market" if everywhere else we use the creator username? I don't get it.

mqp: > I see, but users might share the URL after and then it would not have the correct info. Referral bonuses are also based on the creator of the contract.

Referral bonuses are based on the creator of the contract. They are not based on this string. This string has no effect.

Are you seriously suggesting I go add this extra database field to put here? If you really want me to, I will, but I think it's a total waste of time.

jahooma: `creatorUsername` is already a comment field! Plus we know it anyway from which user page this is.

jahooma: Oh I see `userUsername` is a field, not `creatorUsername`. But we could still pass in the creator username if we know it.

jahooma: One more idea is that after a contract loads, we could replace the url with the correct username if it wasn't correct.

web/components/comments-list.tsx
mqp: I'm going to leave this for a later PR since it's kind of reaching into stuff that is out of scope for this one, but I will do it.

PR #761, opened 2022-08-15
Tile contract cards in masonry layout

sipec:
Screen Shot 2022-08-15 at 4 24 09 PM


jahooma:
Whoa, nice! I wanted to do this, but couldn't figure out the magic css that was necessary. Thanks for figuring it out!

jahooma (approved):

PR #762, opened 2022-08-16
WIP persistence work

mqp:
This is a structure designed to make it so that we can persist React components like the homepage search in local storage and restore them again later, so that we can do stuff like showing you the same search results screen when you hit "back" without a ton of hassle.

Current state: working prototype, but open tasks:

- I have to deal with the user state like "member groups" and "follows" in the contract search controls, as well as how the close time filter uses `Date.now`.
- I pretty much want it to restore the state only when you hit `back`, and not in other cases.
- `usePersistentState` should offer a cookie store.


mqp:
@jahooma What do you think? Any complaints about this general approach?

web/hooks/use-persistent-state.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const usePersistentState = <T>(
+  defaultValue: T,
+  persist?: PersistenceOptions
+) => {
+  const store = persist?.store
+  const key = persist ? getKey(persist.prefix, persist.name) : null
+  let initialValue
+  if (key != null && store != null) {
+    const saved = loadState(key, store) as T
```
jahooma, line 40: If the initialValue is loaded immediately from storage, the first render will be different on the client vs what is server-side rendered. I think NextJS will complain and weird glitches would happen.

You might have to start with a default or loading value (maybe just `undefined`), and have a useEffect() that sets the saved value.

I guess I would try what you have first and see that it does cause problems before switching.

jahooma: I guess the cookie store would solve this problem

web/components/contract-search.tsx
```diff
@@ -207,24 +253,34 @@ function ContractSearchControls(props: {
...
+    persist: persistAs('query', sessionStore),
+  })
+  const [filter, setFilter] = usePersistentState<filter>(
+    defaultFilter ?? 'open',
+    persistAs('filter', sessionStore)
+  )
+  const [pillFilter, setPillFilter] = usePersistentState<string | null>(
+    null,
+    persistAs('pillFilter', sessionStore)
+  )
```
jahooma, line 283: These could actually just be url params as well. I just never got around to it.

web/components/contract-search.tsx
```diff
@@ -207,24 +253,34 @@ function ContractSearchControls(props: {
...
-  }, [sort])
+  const initialSort = defaultSort ?? 'score'
+  const [sort, setSort] = useSort(initialSort, {
+    useUrl: !!useQuerySortUrlParams,
+    persist: persistAs('sort', localStore),
+  })
+  const [query, setQuery] = useQuery('', {
+    useUrl: !!useQuerySortUrlParams,
+    persist: persistAs('query', sessionStore),
+  })
```
jahooma, line 275: Why do we need to persist the `query` in session storage? Wouldn't it be in the url?

web/components/contract-search.tsx
```diff
@@ -103,26 +91,73 @@ export function ContractSearch(props: {
...
+  }, [])
+
+  useEffect(() => {
+    if (persistPrefix && store) {
+      const handleScroll = (e: Event) => {
+        const scrollY = (e.currentTarget as Window).scrollY
+        console.log('Saving scroll position: ', scrollY)
+        saveState(getKey(persistPrefix, 'scrollY'), scrollY, store)
+      }
+      window.addEventListener('scroll', handleScroll)
```
jahooma, line 138: This kind of listener is pretty slow. Can't we measure the scroll on navigating away?

web/components/contract-search.tsx
```diff
@@ -103,26 +91,73 @@ export function ContractSearch(props: {
...
+    if (persistPrefix && store) {
+      const parameters = loadState(getKey(persistPrefix, 'parameters'), store)
+      const scrollY = loadState(getKey(persistPrefix, 'scrollY'), store)
+      if (parameters !== undefined) {
+        console.log('Restoring search parameters: ', parameters)
+        searchParameters.current = parameters as SearchParameters
+      }
+      if (scrollY !== undefined) {
+        console.log('Restoring scroll position: ', scrollY)
+        window.scrollTo(0, scrollY as number)
```
jahooma, line 126: So, we actually have another hook that is saving and restoring scroll position at the app level.

Check out `hooks/use-preserve-scroll.ts`. It would be great if this handled it for us and we didn't need to do anything special here. Otherwise, you might need to disable it somehow for this case.

jahooma (commented):
Thanks for investigating and coming up with a new approach! It's definitely one that could solve all our problems.

However:

- I'm concerned about the complexity involved. Makes the search code harder to understand.
- We would still need to load all the previous pages in Algolia again, which is not implemented yet, and which could add a lot of latency (query + render time). Really, we will have to switch to virtual lists and load just the page or two you were on. More complexity.
- It'll never match the speed we have now.

My current opinion is that we should keep this idea as a long term plan (unless we think of something better), but punt on it, and either keep what we have with the back bug, or switch to using the home + query param for the market to not break routing while still loading the contract page fast.

web/hooks/use-persistent-state.ts
mqp: Good point, not sure why I didn't see warnings? My implementation appeared to be working.

mqp:
> We would still need to load all the previous pages in Algolia again, which is not implemented yet, and which could add a lot of latency (query + render time). Really, we will have to switch to virtual lists and load just the page or two you were on. More complexity.

What do you mean by this?

web/components/contract-search.tsx
mqp: I agree, it was the simplest thing but that would be a much better implementation.

mqp:
> It'll never match the speed we have now.

Are you sure? It seems like the page transitions are basically doing a conceptually similar amount of work. What about the current approach do you think is necessarily faster?

mqp:
> My current opinion is that we should keep this idea as a long term plan (unless we think of something better), but punt on it, and either keep what we have with the back bug, or switch to using the home + query param for the market to not break routing while still loading the contract page fast.

I'm OK with keeping what we have, assuming we can easily fix the problem with it, but I think we should do it based on which makes more sense, because I don't agree that this version seems likely to be slower.

The added burden of having to configure serialization for every piece of state is concerning. It would be a lot more appealing if there were an easier way to manage "all the state owned by this component", but there's not. I don't like the homepage hack, but I also don't like this.

jahooma:
> > We would still need to load all the previous pages in Algolia again, which is not implemented yet, and which could add a lot of latency (query + render time). Really, we will have to switch to virtual lists and load just the page or two you were on. More complexity.
> 
> What do you mean by this? I think this work adds zero total Algolia queries.

Maybe I don't understand. Would the Algolia response be cached when you go back and make the same request again? We're only storing the search params, not the results of the search, so we need to query again.

Also, we need to load multiple pages worth of queries — how are we addressing that?

mqp:
> > > We would still need to load all the previous pages in Algolia again, which is not implemented yet, and which could add a lot of latency (query + render time). Really, we will have to switch to virtual lists and load just the page or two you were on. More complexity.
> > 
> > 
> > What do you mean by this? I think this work adds zero total Algolia queries.
> 
> Maybe I don't understand. Would the Algolia response be cached when you go back and make the same request again? We're only storing the search params, not the results of the search, so we need to query again.
> 
> Also, we need to load multiple pages worth of queries — how are we addressing that?

Yes, the whole point is to persist the results of the search, which are stored in `pages` and `numPages`, so that we can display them instantly when you hit "back". That's all implemented and working in this PR.

If we didn't persist the results of the search, it definitely wouldn't make any sense, the user experience would be ridiculous.

web/components/contract-search.tsx
mqp: That looks promising, I didn't realize it existed. I don't know whether it will just work perfectly out of the box (probably it will?) but if we move forward on this approach I will definitely unify these in a way that makes sense.

jahooma:
Oh ok I see that now. In that case, it should be pretty fast, but it will have to re-render all the market cards. Whereas with my hack it never destroys those DOM nodes, they are just visually hidden by css.

I like this approach because it seems sustainable for the long term. Maybe configuring it with a wrapping context is a good idea. I.e. the usePersistantState is configured by the context. But I think it should also work without a provided context and do some reasonable default thing. (e.g. store in session storage, use a random prefix).

Actually, do we really need the option for local storage? That could be hand done separately. Maybe the usePersistentState hook should be optimized for an ergonomic API. You can eliminate the second parameter by always doing session storage and a random prefix.

As you say, it could have other uses like remember the comment you were typing. Or caching notifications so that if you click one and go back, it loads immediately.

jahooma:
Actually, I'm not sure how you'd get the random prefix to work. You could use the context to save the order the hooks are called, but that's brittle.

mqp:
> Oh ok I see that now. In that case, it should be pretty fast, but it will have to re-render all the market cards. Whereas with my hack it never destroys those DOM nodes, they are just visually hidden by css.

That could be true, but I don't think it's true right now, right? The code in `home` conditionally renders the contract search based on whether there happens to be a contract right now.

jahooma:
> > Oh ok I see that now. In that case, it should be pretty fast, but it will have to re-render all the market cards. Whereas with my hack it never destroys those DOM nodes, they are just visually hidden by css.
> 
> That could be true, but I don't think it's true right now, right? The code in `home` conditionally renders the contract search based on whether there happens to be a contract right now.

I think it conditionally renders the contract page, but always renders the search

mqp:
> > > Oh ok I see that now. In that case, it should be pretty fast, but it will have to re-render all the market cards. Whereas with my hack it never destroys those DOM nodes, they are just visually hidden by css.
> > 
> > 
> > That could be true, but I don't think it's true right now, right? The code in `home` conditionally renders the contract search based on whether there happens to be a contract right now.
> 
> I think it conditionally renders the contract page, but always renders the search

Oh yeah, I just looked again and you are right. It's possible that would be annoying if there were enough contracts, but right now I think rendering performance on the contract grid with a few dozen contracts is pretty manageable. (The big costs according to the profiler, like most of our components, are links and tooltips, and we could really do a lot to optimize those if we had the motivation to do so.)

web/components/contract-search.tsx
```diff
@@ -103,26 +91,73 @@ export function ContractSearch(props: {
...
 
-  const [numPages, setNumPages] = useState(1)
-  const [pages, setPages] = useState<Contract[][]>([])
-  const [showTime, setShowTime] = useState<ShowTime | undefined>()
+  const store = safeSessionStorage()
+  const persistAs = (name: string) => {
+    return persistPrefix ? { prefix: persistPrefix, name, store } : undefined
+  }
+
+  const [numPages, setNumPages] = usePersistentState(1, persistAs('numPages'))
```
jahooma, line 103: Alternate idea: store persistent state keyed by the url in session storage. This hook would not take any prefix or name params, and simply load and save data based on the url. (Writing from multiple of the same hook at the same url would overwrite the same data.)

If you navigate back to the exact url as before, then it loads the saved data, otherwise, if it loads the same pathname but the query params are different, then it invalidates the cached data. (If it loads a different pathname, the cache is unaffected.)

To use it you'd basically make one object with all the data you want to save for the page. Could make this PR a lot simpler!

The scrolling bit can be independent, and might already work with the top level hook `usePreserveScroll` already discussed.

mqp: Interesting. I notice a few knobs on this idea:
- It sounds like you are suggesting we manage all the cached state with one hook that has a big blob, instead of calling many hooks with individual values. That alone would clean it up a lot.
- It sounds like you are saying the cache has exactly zero or one entry per page, but it could have more than one if that were useful.
- It sounds like you are saying the cache would invalidate itself if any part of the URL changed, but we could limit it to the path and the query parameters that were actually affecting the relevant state.

I think I agree that this would be simpler and shorter. The main drawback is that we actually would have to use the URL whenever we used this, or else we would have no cache key sitting around. So it might be inappropriate for a lot of the things we currently cache in local storage (whereas as it stands this PR would simplify all those quite nicely.) It's possible, though, that we can naturally make a `useURLState` hook that uses a slightly lower level `usePersistentState` hook and get the best of both worlds.

I'll experiment with this suggestion and these knobs for an hour or two and see what comes out. I like the idea.

mqp:
Poked around at this a little more...

1. The `usePreserveScroll` hook isn't going to work well in this case, I don't think, because the event it's listening to (`routerChangeComplete`) fires too late.
2. However, Next has a built-in option to store page scroll positions in session storage and restore them and that one works great.
3. There's another problem I realized with this design, which is that I really don't want to restore search results if someone, like, refreshes the page. Maybe storing the search results in an in-memory cache is what I actually want.

jahooma:
Cool, using the nextjs scroll restoration seems better. I have noticed the scroll restoration being delayed by a second or so.

That's a good point about refreshing the page actually refreshing the data. I think that when clicking a Nextjs Link it tries to do only clientside navigation, and probably if you go back, it's also client side?

Ah, I think I'm finally understanding how to make things go really fast. If we use their Link component for all navigation, we can cache a lot of data in-memory, like a single page app. That was what I always wanted.

So stuff like navigating to your user page and seeing your bets could be fast because we loaded all that data in the background, no matter which page you landed on, and then when you go there (with client-side navigation), you'll have all your bets data in memory already. I think that's achievable?

jahooma:
We could have some central in-memory store, accessible by hook. For the search results case, we could store results there too, but when we render ContractSearch, only use those stored results if you were navigating back (I think we can determine this).

jahooma:
Actually, maybe react-query is that in-memory store. Would work for notifications and bets and comments. But maybe not for storing these algolia search results.

jahooma:
Using react-query for caching notifications works (see #812)!

I think the right approach for in-memory caching these search results might also be to just use react query! Give that a shot! 

It might be a bit complex because of all the search params to pass in, but those can also be passed as query keys: https://react-query-v3.tanstack.com/guides/query-keys#array-keys

mqp:
I implemented a pretty decent looking version storing the search results state in history, and unified all the different kinds of "persistent state" (in URL params, in history, in local storage, in session storage) with a pretty clean abstraction. This looks pretty good so I'm probably going to tidy it up and go with that.

I agree that "in memory in the react-query cache" is also an interesting kind of persistent state, but I don't think it makes sense to apply it here? The search results shouldn't be in the react-query cache, because they want to act like your browser history. The search parameters shouldn't be in the react-query cache, because they are either part of the URL or want to be saved in local storage for your next session (until and unless we put them in the DB) or both.

mqp:
> That's a good point about refreshing the page actually refreshing the data. I think that when clicking a Nextjs Link it tries to do only clientside navigation, and probably if you go back, it's also client side?

Perhaps you already know this, but the deal is that `Link` does client-side navigation, but if it's navigating client-side to a page that has SSR or ISR, it will do a special request to pull down the SSR/ISR props as JSON.

mqp:
Funny that this PR is -50 lines even though I abstracted out all this useful generic stuff.

jahooma:
Whoa! So excited this got merged. Seems to be decently fast, and the navigation bugs seem fixed! Plus the scroll restoration is much better. The abstraction you came up with is pretty interesting as well and seems like it has many uses.

Nice work, Marshall!

PR #763, opened 2022-08-16
Kill weird group rendering height hack

mqp:
I replaced this height calculation with normal CSS. I don't really understand the purpose of the hack so maybe I am tearing down Chesterton's fence here, but it seems like it didn't achieve anything.

The only user-visible differences I know of are:
- In the old version, scrolling is more or less broken on the mobile layout if your phone is too short to show the list of groups, but now it works.
- In the new version, one tiny vertical margin that collapsed before doesn't collapse, so the vertical spacing near the bottom is slightly different, but I don't really think it's better or worse.

### DIFF
--- a/web/components/nav/sidebar.tsx
+++ b/web/components/nav/sidebar.tsx
@@ -18,7 +18,7 @@ import { ManifoldLogo } from './manifold-logo'
 import { MenuButton } from './menu'
 import { ProfileSummary } from './profile-menu'
 import NotificationsIcon from 'web/components/notifications-icon'
-import React, { useMemo, useState } from 'react'
+import React, { useMemo } from 'react'
 import { IS_PRIVATE_MANIFOLD } from 'common/envs/constants'
 import { CreateQuestionButton } from 'web/components/create-question-button'
 import { useMemberGroups } from 'web/hooks/use-group'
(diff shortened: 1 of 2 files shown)


sipec:
Looks good to me as well, although I am also curious if Ian has anything to say

jahooma:
This was my hack. Here's the case this fails:
Screen Shot 2022-08-16 at 4 38 49 PM
 
When you have lots of groups, it should make a scrolling list that extends to the bottom of your screen, and not further.

mqp:
Thanks for explaining, I will consider this matter this evening.

mqp:
What you said doesn't seem to work at all right now? In the production version, it cuts off with no way to see the groups whatsoever. In this screenshot I am a member of a group I can't see a link to:

image

I don't think an independently scrolling groups area or sidebar is a very good idea, because a scrollbar would be really ugly, but without a scrollbar it would be hard to know it can scroll.

My first suggestion of a design:
- Like the version in this PR, except:
- The bug where the background of the main content area is white when the sidebar extends past it is fixed.
- The sidebar only displays the "most interesting" N groups (i.e. you should be able to pick which ones, or the N with recent activity, or the N you chatted in most recently) so it doesn't get arbitrarily long.

But take that with a grain of salt because I have zero clue what the expected user flows for "people interacting with groups" is.

If anyone can tell me a real visual design they think is good for this, I guarantee I can produce it with minimal effort and no hacks.

jahooma:
It shows my groups on that screen. Maybe they didn't load?

Truncating the groups and making it not a list seems ok. Letting the groups keep extending down ruins the layout of many pages I think so I like that option the least.

Overall, this doesn't seem like a pressing problem. We should focus on the user experience, especially in ways that affect a lot of users.

mqp:
> Letting the groups keep extending down ruins the layout of many pages I think so I like that option the least.

Ruins it how? In what way would it have any effect on the layout of anything else?

> It shows my groups on that screen. Maybe they didn't load?

That's really weird, I tried it with both Firefox and Chrome.

I guess we don't really know what to do with this so I will just pitch it.

PR #764, opened 2022-08-17
Markets emails

IanPhilips:
I have to merge this with main to see how it looks with real markets (only sending the email to myself) but if that looks good will just have to change the interval and sub out the query for me for the query for all users.

The function gets all private users and filters for those who haven't unsubscribed from weekly trending emails (unsubscribe link at bottom), and gets the top 50 trending markets by unique individual bettor ids. For each user it filters the top 50 markets for ones they haven't bet in already and sends them 3 random ones. I was thinking about tracking the ones we've sent them (just a string array of slugs on the PrivateUser model) but Stephen thinks this isn't necessary so I removed it. 

I had to move a few different bits of code from web/lib to common so that I could build the og-image on the backend and send it with the email.

Screen Shot 2022-08-17 at 11 38 18 AM


functions/src/emails.ts
```diff
@@ -390,3 +391,53 @@ export const sendNewAnswerEmail = async (
...
+export const sendThreeContractsEmail = async (
+  privateUser: PrivateUser,
+  contractsToSend: Contract[]
+) => {
+  const emailType = 'weekly-trending'
+  const unsubscribeUrl = `${UNSUBSCRIBE_ENDPOINT}?id=${privateUser.id}&type=${emailType}`
+  if (!privateUser || !privateUser.email) return
+  await sendTemplateEmail(
+    privateUser.email,
+    contractsToSend[0].question + ' and 2 more questions for you.',
```
mantikoros, line 404: Not sure about this as an email subject. Something like "Interesting markets this week" would be better, I think.

functions/src/emails.ts
```diff
@@ -390,3 +391,53 @@ export const sendNewAnswerEmail = async (
...
   )
 }
+
+export const sendThreeContractsEmail = async (
+  privateUser: PrivateUser,
+  contractsToSend: Contract[]
+) => {
+  const emailType = 'weekly-trending'
+  const unsubscribeUrl = `${UNSUBSCRIBE_ENDPOINT}?id=${privateUser.id}&type=${emailType}`
+  if (!privateUser || !privateUser.email) return
```
mantikoros, line 401: You should check if the privateUser is unsubscribed here.

functions/src/weekly-markets-emails.ts
```diff
@@ -0,0 +1,69 @@
...
+import { Contract } from '../../common/contract'
+import { getPrivateUser, getValues, isProd, log } from './utils'
+import { filterDefined } from '../../common/util/array'
+import { sendThreeContractsEmail } from './emails'
+import { createRNG, shuffle } from '../../common/util/random'
+import { DAY_MS } from '../../common/util/time'
+
+export const weeklyMarketsEmails = functions
+  .runWith({ secrets: ['MAILGUN_KEY'] })
+  .pubsub.schedule('every 1 minutes')
```
mantikoros, line 13: This is just debugging code, I assume. When do you plan to send this email? I vote at a consistent time like Monday at 12 pm ET.

functions/src/weekly-markets-emails.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const firestore = admin.firestore()
+
+async function getTrendingContracts() {
+  return await getValues<Contract>(
+    firestore
+      .collection('contracts')
+      .where('isResolved', '==', false)
+      .where('closeTime', '>', Date.now() + DAY_MS)
+      .where('visibility', '==', 'public')
```
mantikoros, line 26: Do we even use the visibility flag?

mantikoros (commented):
Very nice! Looks like this basically ready to go.

A few additional comments:
- The "see more" text in the email template is awkwardly positioned. I would just remove it entirely.
- Please create a file for the email template at functions/src/email-templates

IanPhilips:
The see more text is only weird for markets without descriptions. Most have descriptions and will look like the last market. Do you think the last market's see more button is awkward?

functions/src/emails.ts
IanPhilips: If they're truly interesting then one of the titles should get their attention, right?

functions/src/emails.ts
IanPhilips: Well I do check that in the `weekly-markets-email` function which is what that flag is for. You can use this `sendThreeContractsEmail` function and check for your own flag.

IanPhilips: I suppose you'll use a different template though so I'll add it again here

functions/src/weekly-markets-emails.ts
IanPhilips: No time like the present! It is set on markets currently and we've been talking about using it.

functions/src/weekly-markets-emails.ts
IanPhilips: yup, just for debugging. I'll set it to that time after I merge it into main and uncomment the all users bit.

functions/src/emails.ts
jahooma: I like Ian's idea to include the question.

What do you think about sending more markets? Like 6?

IanPhilips: hmmm!! yeah we might as well, for all we know these could be the only markets they see all week

jahooma:
> The see more text is only weird for markets without descriptions. Most have descriptions and will look like the last market. Do you think the last market's see more button is awkward?

Maybe the description / see more paragraph should be aligned directly under the image (same width)

IanPhilips:
Hm I can't figure out how to download the html from the preview. I used the template builder and when I preview the content It's in an iframe but can't figure out how to download the iframe contents. Not going to block on that as they're just backups it seems like - they're stored on mailgun already.

IanPhilips:
I moved it to be centered after the description text but it sort of feels like it might take me to the next question so unsure of how much better it is.
Screen Shot 2022-08-17 at 4 58 22 PM

mantikoros:
Yeah, I think the market description is not adding anything. I'd remove it.

I'll figure out how to extract the html from the email. Feel free to merge this in and start running now.

mantikoros:
I created an html email with my design:
Screen Shot 2022-08-19 at 11 14 18 AM

PR #766, opened 2022-08-17
Replace `/markets` with `/home`

mqp:
Previously, the `/markets` page was kind of like a worse version of `/home`. It had no benefit over `/home` except that logged-out users could see it. So instead let's let logged-out users see `/home` and use that instead where we were using `/markets`.


jahooma (approved):
LGTM! This change makes a lot of sense, logically.

PR #772, opened 2022-08-18
Add comment type field to comments

mqp:
This is going to be helpful both for Typescript purposes (type narrowing on the comment will work well, like it does on contracts) and for database querying purposes (e.g. in the user profile page, where we want all contract comments.)

You might think that it's equally good to do a query like `query(comments, where('contractId', '!=', null))` instead of `query(comments, where('commentType', '==', 'contract'))`, but it's not, because `!= null` qualifies as a "range comparison" in [these docs](https://firebase.google.com/docs/firestore/query-data/order-limit-data#limitations) -- they don't say so, but I tried it -- and it prevents you from doing arbitrary sorts on other query fields.

### DIFF
--- a/common/comment.ts
+++ b/common/comment.ts
@@ -4,6 +4,8 @@ import type { JSONContent } from '@tiptap/core'
 // They're uniquely identified by the pair contractId/betId.
 export type Comment = {
   id: string
+  commentType: 'contract' | 'group'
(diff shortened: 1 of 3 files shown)


jahooma (approved):
Makes sense, LGTM!

PR #773, opened 2022-08-18
Use masonry on contract cards, sorted correctly

akrolsmir:
This is actually kind of a huge pain to get exactly right...
- The Tailwind `columns-2` solution doesn't order them in the ideal left-to-right ordering. But [`order-[x]`](https://tailwindcss.com/docs/order#class-reference) doesn't work either
- Firefox has support for Masonry https://stackoverflow.com/a/45200955/1222351
- MaterialUI has this implemented https://github.com/mui/material-ui/pull/27439

So here's this somewhat hacky solution, which seems to work okay.

TODO: Have to fix the 1-column version - unfortunately, detecting whether we're in md: breakpoint in React/JS seems to be annoying too

### DIFF
--- a/functions/package.json
+++ b/functions/package.json
@@ -39,6 +39,7 @@
     "lodash": "4.17.21",
     "mailgun-js": "0.22.0",
     "module-alias": "2.2.2",
+    "react-masonry-css": "1.0.16",
     "stripe": "8.194.0",
     "zod": "3.17.2"
   },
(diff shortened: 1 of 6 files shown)


akrolsmir:
Hrm, so this is now like an 80% solution. However, when you resize the screen, sometime you get an off-by-one type error where one of the contract cards slips to the bottom.

One possible solution would be to just use a library for this, https://github.com/paulcollett/react-masonry-css looks okay

sipec:
I think using a library is probably better than this hack.
But I think the best solution would be to just make all the cards the same height by clipping the titles, and mayybe enforcing a shorter character limit in titles.

IanPhilips:
Looks good on iPhone Safari!

PR #774, opened 2022-08-18
List manifold-sdk on the Awesome Manifold page

keriwarr:


jahooma:
Yeah! Let's add that!

PR #775, opened 2022-08-18
Simple bet interface in embeds

sipec:
No more cut off modals!

### DIFF
--- a/web/components/bet-button.tsx
+++ b/web/components/bet-button.tsx
@@ -9,8 +9,8 @@ import { useUserContractBets } from 'web/hooks/use-user-bets'
 import { useSaveBinaryShares } from './use-save-binary-shares'
 import { Col } from './layout/col'
 
-// Inline version of a bet panel. Opens BetPanel in a new modal.
-export default function BetRow(props: {
+/** Button that opens BetPanel in a new modal */
+export default function BetButton(props: {
   contract: CPMMBinaryContract | PseudoNumericContract
   className?: string
   btnClassName?: string
--- a/web/pages/embed/[username]/[contractSlug].tsx
+++ b/web/pages/embed/[username]/[contractSlug].tsx
@@ -94,6 +94,8 @@ export function ContractEmbed(props: { contract: Contract; bets: Bet[] }) {
 
   const [betPanelOpen, setBetPanelOpen] = useState(false)
 
+  const [probAfter, setProbAfter] = useState<number>()
+
   return (
     <Col className="h-[100vh] w-full bg-white">
       <div className="relative flex flex-col pt-2">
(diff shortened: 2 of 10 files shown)


sipec:
self review:
- it would be more reactish to lift the `amount` and `outcome` state out into the parent, and do the probability calculations in the parent, rather than pass a setState function
- missing outcome change for numeric

web/pages/embed/[username]/[contractSlug].tsx
```diff
@@ -88,18 +89,15 @@ export function ContractEmbed(props: { contract: Contract; bets: Bet[] }) {
...
-      ? windowHeight - topSectionHeight - paddingBottom
-      : 0
+  const [betPanelOpen, setBetPanelOpen] = useState(false)
+
+  const [probAfter, setProbAfter] = useState<number>()
 
   return (
-    <Col className="w-full flex-1 bg-white">
-      <div className="relative flex flex-col pt-2" ref={setElem}>
+    <Col className="h-[100vh] w-full bg-white">
```
Comment (jahooma) [inline web/pages/embed/[username]/[contractSlug].tsx:99]: Nice simplification of the embed sizing!

web/pages/embed/[username]/[contractSlug].tsx
```diff
@@ -114,25 +112,24 @@ export function ContractEmbed(props: { contract: Contract; bets: Bet[] }) {
...
-                <BetRow
-                  contract={contract as CPMMBinaryContract}
-                  betPanelClassName="scale-75"
-                />
-              )}
-              <BinaryResolutionOrChance contract={contract} />
-            </Row>
+            <BinaryResolutionOrChance
+              contract={contract}
+              probAfter={probAfter}
```
Comment (jahooma) [inline web/pages/embed/[username]/[contractSlug].tsx:126]: Nice touch

web/components/bet-button.tsx
```diff
@@ -9,8 +9,8 @@ import { useUserContractBets } from 'web/hooks/use-user-bets'
 import { useSaveBinaryShares } from './use-save-binary-shares'
 import { Col } from './layout/col'
 
-// Inline version of a bet panel. Opens BetPanel in a new modal.
-export default function BetRow(props: {
+/** Button that opens BetPanel in a new modal */
+export default function BetButton(props: {
```
jahooma, line 13: I would switch this to not be a default export, since you're editing this. Thanks~

Also, thanks for renaming it to something more appropriate!

jahooma (approved):
Cool change! This is pretty well done. It's nice to have a compact way to bet.

I think it works pretty well as is. One suggestion that would be nice is to put a small version of your avatar in somewhere in the bet panel, so you know that it is betting under your account. If you did, you could also consider putting the balance amount directly underneath your avatar.

One problem it seems is that it might not fit within smaller widths:
Screen Shot 2022-08-18 at 6 56 27 PM
Maybe it could switch into a 2-line layout in this case?
If the window size is too small, you could also consider hiding the bet panel / button.

PR #776, opened 2022-08-18
Make typing for comments more fancy

mqp:
Now it works like typing for contracts; the type narrows on the `commentType` field and then it knows whether it's supposed to have contract-related stuff or group-related stuff.


mqp:
@jahooma because you suggested doing this.

jahooma (approved):
LGTM!

PR #777, opened 2022-08-18
betting streaks

IanPhilips:
- every time a user places a bet the on-create-bet function checks if it's a new one since the past 9am UTC. If it's new, add one to their streak and send them a bonus + notification. 
- Every day at 9am UTC, the reset-betting-streaks function checks if the user has made a bet since a day ago. If not, their streak is set to 0.
- I'll probably add a fire emoji on someone's profile picture in the navbar/sidebar too

Screen Shot 2022-08-18 at 4 46 57 PM
Screen Shot 2022-08-18 at 4 46 47 PM

### DIFF
--- a/common/numeric-constants.ts
+++ b/common/numeric-constants.ts
@@ -5,3 +5,4 @@ export const NUMERIC_GRAPH_COLOR = '#5fa5f9'
 export const NUMERIC_TEXT_COLOR = 'text-blue-500'
 export const UNIQUE_BETTOR_BONUS_AMOUNT = 10
 export const BETTING_STREAK_BONUS_AMOUNT = 1
+export const BETTING_STREAK_RESET_HOUR = 9
--- a/common/numeric-constants.ts
+++ b/common/numeric-constants.ts
@@ -4,3 +4,4 @@ export const NUMERIC_FIXED_VAR = 0.005
 export const NUMERIC_GRAPH_COLOR = '#5fa5f9'
 export const NUMERIC_TEXT_COLOR = 'text-blue-500'
 export const UNIQUE_BETTOR_BONUS_AMOUNT = 10
+export const BETTING_STREAK_BONUS_AMOUNT = 1
--- a/common/numeric-constants.ts
+++ b/common/numeric-constants.ts
@@ -4,5 +4,5 @@ export const NUMERIC_FIXED_VAR = 0.005
 export const NUMERIC_GRAPH_COLOR = '#5fa5f9'
 export const NUMERIC_TEXT_COLOR = 'text-blue-500'
 export const UNIQUE_BETTOR_BONUS_AMOUNT = 10
-export const BETTING_STREAK_BONUS_AMOUNT = 1
+export const BETTING_STREAK_BONUS_AMOUNT = 5
 export const BETTING_STREAK_RESET_HOUR = 9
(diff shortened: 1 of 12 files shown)


functions/src/on-create-bet.ts
```diff
@@ -38,37 +44,96 @@ export const onCreateBet = functions.firestore
...
+  user: User,
+  bet: Bet,
+  contract: Contract,
+  eventId: string
+) => {
+  const betStreakResetTime = getPreviousBettingStreakResetTime()
+  // If they've already bet after the reset time, they've already gotten their streak bonus
+  if (user.lastBetTime ?? 0 > betStreakResetTime) return
+
+  const newBettingStreak = (user.currentBettingStreak ?? 0) + 1
```
mantikoros, line 78: You could also actually include the reset logic here if you wanted: If the current bet time is over 24 hrs past the last betStreakResetTime, the current bet streak is just 1. Your approach with the cloud functions also works though.

web/pages/notifications.tsx
```diff
@@ -362,27 +365,83 @@ function IncomeNotificationItem(props: {
...
+        {reasonText}
+        {sourceType === 'betting_streak_bonus' ? (
+          simple ? (
+            <span className={'ml-1 font-bold'}>Betting Streak</span>
+          ) : (
+            <SiteLink
+              className={'ml-1 font-bold'}
+              href={'/betting-streak-bonus'}
+            >
+              Betting Streak
```
mantikoros, line 404: Ideally this would be "10-day betting streak" but this is fine for now.

mantikoros (approved):
LGTM! I think this is going to be great!

I think the next step is creating a bet streak email.

common/numeric-constants.ts
```diff
@@ -4,3 +4,5 @@ export const NUMERIC_FIXED_VAR = 0.005
 export const NUMERIC_GRAPH_COLOR = '#5fa5f9'
 export const NUMERIC_TEXT_COLOR = 'text-blue-500'
 export const UNIQUE_BETTOR_BONUS_AMOUNT = 10
+export const BETTING_STREAK_BONUS_AMOUNT = 1
```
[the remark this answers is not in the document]
mantikoros: I'd raise this to M$5 or M$10 probably...

web/components/user-page.tsx
```diff
@@ -114,22 +123,34 @@ export function UserPage(props: { user: User }) {
...
+                  {formatMoney(profit)}
+                </span>
+                <span>profit</span>
+              </Col>
+              <Col
+                className={'w-20 cursor-pointer items-center text-gray-500'}
+                onClick={() => setShowBettingStreakModal(true)}
+              >
+                <span>🔥{user.currentBettingStreak ?? 0}</span>
+                <span>streak</span>
```
mantikoros, line 149: maybe "day streak"? Otherwise it's confusing

functions/src/on-create-bet.ts
IanPhilips: yeah that isn't a perfect solution because they need their streak to reset every day. Otherwise, if they don't make a bet for a few days their last streak will persist.

web/pages/notifications.tsx
IanPhilips: Sure, changed it

common/numeric-constants.ts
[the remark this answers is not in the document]
IanPhilips: 🤣

web/components/user-page.tsx
IanPhilips: Eh, they'll get a notification about it and they can click on the icon to learn more. I think just streak looks better. Open to changing it if people say they're confused.

PR #778, opened 2022-08-18
Fix user avatar in mention list not updating

sipec:
~I think this happened after Marshall's optimizations to Avatars, though I don't know exactly why.~
This happened due to https://github.com/manifoldmarkets/manifold/commit/34e8138e5096e1ad8a4944566328703bd080c645

Solution is to simply telling react to rerender the component by providing a key


jahooma (commented):
Thanks for investigating, Sinclair!

Now that you mention it, I might have noticed other avatar weirdness elsewhere. Like when I logged in as someone else, the old avatar was still showing.

Do you think there could be a broader fix for the Avatar component itself?

Otherwise, LGTM!

jahooma (approved):

PR #779, opened 2022-08-18
Fix tooltip popping up when modal is opened

sipec:
Headless UI's Modal component autofocuses the first focusable item inside it when opened. This is [by design for accessibility reasons](https://headlessui.com/react/dialog#managing-initial-focus). 

Ironically this means we'll have to remove keyboard focus for tooltips because this causes the tooltips to pop up unnecessarily for all users whenever the dialog is opened.
Doing this in an accessible way would involve is managing focus manually for several dialogs, which may not be possible yet as some of our modals lack a sensible element to focus by default. Or maybe disabling the autofocus unless the modal is opened via the keyboard (maybe complicated?)

I'm somewhat unhappy with this but I think doing it the unaccessible way is the right thing to do, in a utilitarian sense.

### DIFF
--- a/web/components/tooltip.tsx
+++ b/web/components/tooltip.tsx
@@ -6,7 +6,6 @@ import {
   Placement,
   shift,
   useFloating,
-  useFocus,
   useHover,
   useInteractions,
   useRole,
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Thanks for the quick fix! LGTM

PR #780, opened 2022-08-19
Make Avatar component update when avatarUrl updates

sipec:
> Now that you mention it, I might have noticed other avatar weirdness elsewhere. Like when I logged in as someone else, the old avatar was still showing. - @jahooma 

I can't reproduce, so I can't test your issue, but I think the ultimate cause is https://github.com/manifoldmarkets/manifold/commit/34e8138e5096e1ad8a4944566328703bd080c645
The moral of the story is don't replicate props as state


jahooma (approved):
Oh yeah, that's the bug! Thanks for tracking this down and fixing it!

PR #782, opened 2022-08-19
Fix notification for updated questions and descriptions

sipec:
Since descriptions are now rich text and not append-only, the previous way of calculating the changes will not work, and there isn't a simple way to fix it. For now, we just say that the market description changed without saying what in it changed.

### DIFF
--- a/functions/src/on-update-contract.ts
+++ b/functions/src/on-update-contract.ts
@@ -2,6 +2,7 @@ import * as functions from 'firebase-functions'
 import { getUser } from './utils'
 import { createNotification } from './create-notification'
 import { Contract } from '../../common/contract'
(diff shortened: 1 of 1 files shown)


IanPhilips (approved):
Cool, I *think* this will be better than not receiving a notification about the market description update! 

Maybe add parentheses around 'the market description' or even have it say '(see the question for updated market description)'?

jahooma:
I don't know if it's worth a notification for updating the description. That's most likely a typo fix or other small change?

I'd favor not getting a notification about it, especially if we can't say what the change was.

IanPhilips:
Yeah I'm also down to just skip it until people complain.

PR #784, opened 2022-08-20
Convert tags to groups on demand

mqp:
There are some old tags that people liked, so it's good to make groups out of them so that people can enjoy the fruits of their effort categorizing the old contracts.

P.S. there was some stuff busted on dev by trying to use the wrong Manifold user ID, like notification generation. Whatever.

### DIFF
--- a/functions/src/scripts/convert-tag-to-group.ts
+++ b/functions/src/scripts/convert-tag-to-group.ts
@@ -1,0 +1,66 @@
+// Takes a tag and makes a new group with all the contracts in it.
+
+import * as admin from 'firebase-admin'
+import { initAdmin } from './script-init'
+import { isProd, log } from '../utils'
+import { getSlug } from '../create-group'
+import { Group } from '../../../common/group'
+
+const getTaggedContractIds = async (tag: string) => {
+  const firestore = admin.firestore()
+  const results = await firestore
+    .collection('contracts')
+    .where('lowercaseTags', 'array-contains', tag.toLowerCase())
+    .get()
+  return results.docs.map((d) => d.id)
+}
+
+const createGroup = async (
+  name: string,
+  about: string,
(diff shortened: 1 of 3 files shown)


mqp:
Wait, I'm confused about this. Does the `ENV_CONFIG` stuff even work properly with Cloud Functions? How does `NEXT_PUBLIC_FIREBASE_ENV` ever become `==` to `DEV`?

functions/src/scripts/convert-tag-to-group.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const createGroup = async (
+  name: string,
+  about: string,
+  contractIds: string[]
+) => {
+  const firestore = admin.firestore()
+  const creatorId = isProd()
+    ? 'IPTOzEqrpkWmEzh6hwvAyY9PqFb2'
+    : '94YYTk1AFWfbWMpfYcvnnwI1veP2'
```
jahooma, line 26: This could probably be a constant for the environment. I created a function `isManifoldId` in `constants.ts` which could also be updated based on the environment. Probably something for a followup PR.

jahooma (approved):
Looks good to me! This should be handy for recreating the groups people want.

jahooma:
> Wait, I'm confused about this. Does the `ENV_CONFIG` stuff even work properly with Cloud Functions? How does `NEXT_PUBLIC_FIREBASE_ENV` ever become `==` to `DEV`?
> 
> I'm just going to keep that change for later and make this PR only the script.

I'm actually not sure!

functions/src/scripts/convert-tag-to-group.ts
mqp: Yes, I did all that but I reverted it because I got too confused and didn't feel like dealing with it right now.

PR #788, opened 2022-08-23
Fix link classes appending on paste

sipec:
I think with this, we can fix one of the football markets by admin-edit, ctrl a, ctrl x, ctrl v, save, and then it should be able to duplicate

### DIFF
--- a/web/components/editor.tsx
+++ b/web/components/editor.tsx
@@ -6,6 +6,7 @@ import {
   JSONContent,
   Content,
   Editor,
+  mergeAttributes,
 } from '@tiptap/react'
 import StarterKit from '@tiptap/starter-kit'
 import { Image } from '@tiptap/extension-image'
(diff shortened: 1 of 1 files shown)


akrolsmir (approved):
LGTM, though I can't shake the feeling that a paste handler is less good than a generic sanitization step before the JSON gets uploaded to Firestore.

Sanitization would work for other (?) cases and provide a guarantee of no weird stuff getting stored in Firebase. If we wanted to be super systematic, it could be a step on a cloud function (this would mean API markets would also not be inserting random junk)

But yeah I guess that's all follow-up work

akrolsmir:
(Sinclair says I am wrong and actually this makes it so that links never have classes)

I guess it's annoying to only be for links, but Sinclair says we don't add classes elsewhere so it's probably fine

PR #790, opened 2022-08-23
Toggle monthly leaderboards

FRCassarino:
I didn't get to enabling monthly leaderboards after my work trial was over (I enabled daily/weekly/alltime). The cache has been filled out for a while now, this toggles it on.


jahooma (changes requested):
Cool! I would update the tabs defaultIndex to be 2 so that weekly leaderboards are still the default. Otherwise, LGTM!

jahooma:
I wonder if maybe the tabs should be in the opposite order? Daily, Weekly, Monthly, All time?

PR #791, opened 2022-08-23
Dashboards

FRCassarino:
The first pass at dashboards. Based on what we talked about in the standup yesterday (cc @akrolsmir ) **we probably want to rethink a few things before shipping this.** I'm still sending this draft PR so you can follow along and discuss. 

This version does the following:
- Adds a dashboard tab to each Group.
- The tab is only visible by the group creator until they actually create a dashboard. 
- If a dashboard is created, the tab becomes visible to everyone and the default tab when clicking on a group. 
- The dashboard field is the same rich text field that's used for comments and groups - you can embed images, markets and youtube videos in them.

Based on yesterday's chat, I'll continue by uncoupling dashboards from groups in the backend. I'll refactor this so Dashboards becomes it's own firebase document  and each dashboard will have its own individual, shareable link. I'll still make it possible to add a dashboard to a group, but they'll also be able to exist completely independently from each group. 

Any suggestions on what to rename the feature to? Pages? ManiPages? Essays? Forecasts?


common/group.ts
```diff
@@ -1,8 +1,11 @@
+import { JSONContent } from '@tiptap/core'
+
 export type Group = {
   id: string
   slug: string
   name: string
   about: string
+  dashboard: JSONContent | null
```
jahooma, line 8: Just a note that you'd also have to run a script to update existing groups to have `dashboard: null` because otherwise the type would be inaccurate since they are undefined.

An alternate way to do this would be to give dashboard an optional type:
```dashboard?: JSONContent```
That seems a bit simpler to me and more consistent with how we typically add fields. To delete the dashboard field, you'd have to use the special `FieldValue.delete()` value from firestore.

A third option is to have both optional and null in the type, but that seems less clean than the second option.

**Review by jahooma — COMMENTED** [review]
Cool! Looks like a pretty clean implementation of a Dashboard tab that the creator of the group can edit as rich text.

As discussed, putting this within a group as a tab makes it a little hidden or, if linked to directly, cluttered. 

Feel free to pursue one specific use case at a time, like the Stories/Essay version of dashboards, which I think you're doing. We can also see about other use cases, like curated homepages of content.

One technical note is that you should delete the `package-lock.json` file because we are using yarn instead of npm. Get yarn here if you haven't already (https://classic.yarnpkg.com/lang/en/docs/install) and use the command `yarn` instead of `npm install`.

File: web/components/file-upload-button.tsx [hunk 2]
```diff
@@ -10,7 +10,11 @@ export function FileUploadButton(props: {
   const ref = useRef<HTMLInputElement>(null)
   return (
     <>
-      <button className={className} onClick={() => ref.current?.click()}>
+      <button
+        type={'button'}
```
FRCassarino, line 14: It was reloading the page when on the create dashboard form [https://stackoverflow.com/questions/1878264/how-do-i-make-an-html-button-not-reload-the-page]

FRCassarino:
The new changes undo for now all the Group related stuff, and instead:

- make dashboards their own independent firestore document
- Add a "Create Dashboard" page (as of yet not linked to from anywhere), and a "View Dashboard" page. 

I'll send further changes (adding dashboards to groups, showing dashboards in a user's profile, etc) as stacked PRs.

web/lib/firebase/api.ts
```diff
@@ -88,3 +88,7 @@ export function acceptChallenge(params: any) {
 export function getCurrentUser(params: any) {
   return call(getFunctionUrl('getcurrentuser'), 'GET', params)
 }
+
+export function createDashboard(params: any) {
```
jahooma, line 92: I really think we should type these params, even though that's not the pattern so far.
We have to manually keep them in sync with the cloud function, but I think it's worth it.

web/pages/create-dashboard.tsx
```diff
@@ -0,0 +1,94 @@
+import { useState } from 'react'
+import { Spacer } from 'web/components/layout/spacer'
+import { Page } from 'web/components/page'
+import { Title } from 'web/components/title'
+import Textarea from 'react-expanding-textarea'
+
+import { TextEditor, useTextEditor } from 'web/components/editor'
+import { createDashboard } from 'web/lib/firebase/api'
+import clsx from 'clsx'
+import { useRouter } from 'next/router'
```
jahooma, line 10: You can actually use the static router here and it's a bit simpler: 
`import Router from 'next/router'`
`Router.push(...)`

web/pages/create-dashboard.tsx
```diff
@@ -0,0 +1,94 @@
...
+      content: editor.getJSON(),
+    }
+
+    const result = await createDashboard(newDashboard).catch((e) => {
+      console.log(e)
+      setError('There was an error creating the dashboard, please try again')
+      return e
+    })
+    if (result.dashboard) {
+      await router.push(dashboardPath((result.dashboard as Dashboard).slug))
```
jahooma, line 39: Can remove cast once `createDashboard` is typed.

web/pages/dashboard/[...slugs]/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { ENV_CONFIG } from 'common/envs/constants'
+import Custom404 from 'web/pages/404'
+
+export const getStaticProps = fromPropz(getStaticPropz)
+export async function getStaticPropz(props: { params: { slugs: string[] } }) {
+  const { slugs } = props.params
+
+  const dashboard = await getDashboardBySlug(slugs[0])
+  const creatorPromise = dashboard ? getUser(dashboard.creatorId) : null
+  const creator = await creatorPromise
```
Comment (jahooma) [inline web/pages/dashboard/[...slugs]/index.tsx:27]: Alternately:
`const creator = dashboard ? await getUser(...) : null`

web/pages/dashboard/[...slugs]/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export async function getStaticPaths() {
+  return { paths: [], fallback: 'blocking' }
+}
+
+export default function DashboardPage(props: {
+  dashboard: Dashboard
+  creator: User
+}) {
+  props = usePropz(props, getStaticPropz) ?? {
```
Comment (jahooma) [inline web/pages/dashboard/[...slugs]/index.tsx:47]: I would unpack props here:
`const { dashboard, creator } = ...`

web/pages/create-dashboard.tsx
```diff
@@ -0,0 +1,94 @@
...
+import Textarea from 'react-expanding-textarea'
+
+import { TextEditor, useTextEditor } from 'web/components/editor'
+import { createDashboard } from 'web/lib/firebase/api'
+import clsx from 'clsx'
+import { useRouter } from 'next/router'
+import { Dashboard, MAX_DASHBOARD_NAME_LENGTH } from 'common/dashboard'
+import { dashboardPath } from 'web/lib/firebase/dashboards'
+
+export default function CreateDashboard() {
```
jahooma, line 14: You might consider moving this file to `pages/dashboard/create`. That url structure makes a little more sense to me and the file would be colocated with the dashboard index.

web/pages/dashboard/[...slugs]/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    dashboard: null,
```
Comment (jahooma) [inline web/pages/dashboard/[...slugs]/index.tsx:48]: Might want to start as `dashboard: undefined, creator: undefined`.

We often use the pattern of `undefined` meaning that it hasn't loaded yet, while `null` means that it's loaded and was not found.

web/pages/dashboard/[...slugs]/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { Button } from 'web/components/button'
+import { useState } from 'react'
+import { ShareDashboardModal } from 'web/components/share-dashboard-modal'
+import { Row } from 'web/components/layout/row'
+import { Col } from 'web/components/layout/col'
+import { ENV_CONFIG } from 'common/envs/constants'
+import Custom404 from 'web/pages/404'
+
+export const getStaticProps = fromPropz(getStaticPropz)
+export async function getStaticPropz(props: { params: { slugs: string[] } }) {
```
Comment (jahooma) [inline web/pages/dashboard/[...slugs]/index.tsx:22]: On reflection, I think we don't need the `Propz` pattern for dashboards.

It exists so that we can accommodate private instances where only authenticated users can load the data client-side.

But dashboards are an inherently public feature, so it's probably unnecessary and mucks up the code. If we really needed it private, we could always add it back. Plus, I think nowadays we would just use server side rendering to check your auth credentials first instead of using this pattern.

So, you can just turn this into `function getStaticProps` and don't call `usePropz` in the page component.

web/pages/create-dashboard.tsx
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    <Page>
+      <div className="mx-auto w-full max-w-2xl">
+        <div className="rounded-lg px-6 py-4 sm:py-0">
+          <Title className="!mt-0" text="Create a dashboard" />
+          <form>
+            <div className="form-control w-full">
+              <label className="label">
+                <span className="mb-1">
+                  Name<span className={'text-red-700'}>*</span>
```
jahooma, line 52: Perhaps "Title" is clearer for users? Name is like a person's name.

web/pages/create-dashboard.tsx
```diff
@@ -0,0 +1,94 @@
...
+                className="input input-bordered resize-none"
+                autoFocus
+                maxLength={MAX_DASHBOARD_NAME_LENGTH}
+                value={name}
+                onChange={(e) => setName(e.target.value || '')}
+              />
+              <Spacer h={6} />
+              <label className="label">
+                <span className="mb-1">
+                  Content<span className={'text-red-700'}>*</span>
```
jahooma, line 66: I would put a space between the label ("Content") and the "*". Same for above.

web/pages/create-dashboard.tsx
```diff
@@ -0,0 +1,94 @@
...
+              </label>
+              <Textarea
+                placeholder="e.g. Elon Mania Dashboard"
+                className="input input-bordered resize-none"
+                autoFocus
+                maxLength={MAX_DASHBOARD_NAME_LENGTH}
+                value={name}
+                onChange={(e) => setName(e.target.value || '')}
+              />
+              <Spacer h={6} />
```
jahooma, line 63: Might want to add a sentence or two that describes the purpose of creating a dashboard. Like `Write a long form post you can share online. Embed prediction markets and images to make your point.`

We could also add this in a follow up instead as we understand the purpose ourselves.

jahooma (approved):
This is awesome work, Fede! Very clean code.

Congrats on your first PR and big feature as an employee!

I've added a few comments. Once addressed, you can feel free to merge this PR in.

I also have an additional concern:
- Dashboards is not a very good user-facing name for this feature. It should definitely be changed in the UI and the URL to more closely match what it is. Possibly "Stories" or "Posts" make more sense.
- Once you've changed the name for users, I actually think it would make sense to change all the references in the code as well (filenames, variables names, collection name, etc.). It's a bit of work, but this is the very first PR on this feature, and I think it would save a lot of confusion in the future.

Thanks for joining our team!

jahooma:
One more thing: it's a good habit to include screenshots of whatever UI changes are included.

Makes it easier for everyone to stop by and scrutinize changes. (And for the author to notice small inconsistencies as well!)

FRCassarino:
Fixed all nits pointed out by James and renamed to 'Posts'

FRCassarino:
@jahooma I merged the branch, and followed the instructions to deploy the backend found here:  https://github.com/manifoldmarkets/manifold/blob/main/functions/README.md

image

I got this "With these options, your minimum bill will be $388.44 in a 30-day month" message. Can you confirm that whether this is normal/expected? I can finish deploying it afterwards. I believe I'll also need to manually update the firestore.rules in the firebase web console, is that right?

jahooma:
Awesome! Yes that's expected, do proceed with deploying, and you are right that you will also have to update the firestore rules in the web console.

Congrats on merging this!

FRCassarino:
Thanks, just deployed and verified it's working in production :)

PR #794, opened 2022-08-23
Allow to follow/unfollow markets, backfill as well

IanPhilips:
I'm not done with the UI, and actually I'm thinking of putting a heart up next to the bet button and making the unfollow easier to do in the notifications page.

- creating, betting, adding liquidity, and commenting all auto-follow you to a market
- you can unfollow to stop receiving notifications about the market
- if you interact after unfollowing you will re-follow the market
- added followerCount to markets, could be a possible sort in the future
- following a market is a much easier way of engaging with the site bc users may not: 
    - want to change this number, the number is already good
    - want to win money
    - believe they can win money 

Screen Shot 2022-08-23 at 4 29 50 PM

### DIFF
--- a/common/contract.ts
+++ b/common/contract.ts
@@ -57,6 +57,7 @@ export type Contract<T extends AnyContractType = AnyContractType> = {
   uniqueBettorIds?: string[]
   uniqueBettorCount?: number
   popularityScore?: number
+  followerCount?: number
 } & T
 
 export type BinaryContract = Contract & Binary
--- a/common/notification.ts
+++ b/common/notification.ts
@@ -70,3 +70,4 @@ export type notification_reason_types =
   | 'challenge_accepted'
   | 'betting_streak_incremented'
   | 'loan_income'
+  | 'you_follow_contract'
--- a/common/user.ts
+++ b/common/user.ts
@@ -42,6 +42,7 @@ export type User = {
   shouldShowWelcome?: boolean
   lastBetTime?: number
   currentBettingStreak?: number
+  hasSeenContractFollowModal?: boolean
 }
 
 export type PrivateUser = {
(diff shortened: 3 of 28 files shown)


jahooma (commented):
This is great, Ian! Definitely a top user request.

I do have the strong impression that we are sending not-useful emails/notifications of comments from every market you have bet on. Being able to turn those off individually is great.

But I wonder if we could get a strong signal on which markets are really important by not automatically following every market you bet on. I think that could be the most valuable part of this feature. Which markets do people really want to follow into the future and get updates on? Could help us recommend markets.

Plus, then we could send them more info on these followed markets, like a notification when the probability changes, or include probability updates on them in a weekly digest email.

jahooma:
I also don't like that if you unfollow, then you automatically follow again if you bet or comment.

jahooma:
To be clear my suggested change is to make betting completely independent of following markets. For commenting / liquidity providing, I think it's fine to auto-follow.

IanPhilips:
Screen Shot 2022-08-24 at 8 36 02 AM
Screen Shot 2022-08-24 at 8 35 54 AM

IanPhilips:
Right now the follow/unfollow is completed on the db triggers so there is a delay between placing a bet/comment and the heart being filled. Might just trigger it from the front end manually.

IanPhilips:
Hm that's kind of a good amount more work - ideally it'd be in the place-bet function but I don't want to slow down that esp bc it's in a fb transaction.

IanPhilips:
another unimplemented option: if you completely sell your shares should you unfollow a market? this is how it currently works - if you don't have shares in a market you don't get notifs from it

IanPhilips:
Will add it to place-bet otherwise the delay is too great

IanPhilips:
Selling all shares now auto unfollows

IanPhilips:
The meaning of less is a bit difficult to convey:
Screen Shot 2022-08-24 at 10 00 13 AM

IanPhilips:
Added an onboarding modal for first time follow-unfollows
Screen Shot 2022-08-24 at 10 13 49 AM

PR #797, opened 2022-08-25
Add tournament home page

sipec:
- [x] actually calculate number of bettors
- [ ] smash in CSPI data somehow
- [x] load markets server-side


sipec:
so far, just the UI layout is done

PR #798, opened 2022-08-25
[WIP] liking markets with tip/heart

IanPhilips:
Screen Shot 2022-08-25 at 11 43 50 AM

### DIFF
--- a/common/like.ts
+++ b/common/like.ts
@@ -1,0 +1,7 @@
+export type Like = {
+  id: string
+  userId: string
+  contractId: string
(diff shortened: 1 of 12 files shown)


wasabipesto:
I don't think it's obvious that "liking" will cost mana. I would highly suggest to make the tip reversible for at least 30 seconds like comment tips, or find a way to communicate it more clearly.

IanPhilips:
Screen Shot 2022-08-30 at 9 34 16 AM

PR #800, opened 2022-08-26
Replace style props with tailwind classes

sipec:
except where we need to use dynamic values for styles

also except daisy buttons


akrolsmir (approved):
LGTM, thanks for cleaning up the codebase!

akrolsmir:
(pending fixing the breakages)

PR #803, opened 2022-08-26
Post in a group about page.

FRCassarino:
This PR adds a Dashboard to a group's about page.

### DIFF
--- a/web/components/groups/group-about-post.tsx
+++ b/web/components/groups/group-about-post.tsx
@@ -11,15 +11,16 @@ import { DocumentRemoveIcon } from '@heroicons/react/solid'
 import { createPost } from 'web/lib/firebase/api'
 import { Post } from 'common/post'
 import { deletePost, updatePost } from 'web/lib/firebase/posts'
-import { useRouter } from 'next/router'
 import { useState } from 'react'
+import { usePost } from 'web/hooks/use-post'
 
 export function GroupAboutPost(props: {
   group: Group
   isCreator: boolean
   post: Post
 }) {
-  const { group, isCreator, post } = props
+  const { group, isCreator } = props
+  const post = usePost(group.aboutPostId) ?? props.post
   const isAdmin = useAdmin()
 
   if (group.aboutPostId == null && !isCreator) {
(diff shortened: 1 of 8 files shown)


jahooma:
I think this looks pretty good, but I'll review it again in more detail once the first Dashboards PR is merged. Thanks Fede!

FRCassarino:
I think the static analysis issues are unrelated to this PR.

web/pages/group/[...slugs]/index.tsx
```diff
@@ -139,6 +147,7 @@ export default function GroupPage(props: {
     topTraders,
     creatorScores,
     topCreators,
+    aboutPost,
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:150]: It's good to get `aboutPost` from `getStaticProps`.

However, it might also be nice to listen for changes to `aboutPost` client-side.

One reason to do this is `getStaticProps` is often cached for a period of time so if you saved changes, they might not appear for minutes, even after refreshing.

The solution is to create a hook called `usePost`, which fetches and listens for updates to a post. You can basically copy the implementation of `useGroup`. 

Then in this component, you could use `const aboutPost = usePost(group.aboutPostId) ?? props.aboutPost`.

web/pages/group/[...slugs]/index.tsx
```diff
@@ -121,6 +128,7 @@ export default function GroupPage(props: {
   creatorScores: { [userId: string]: number }
   topCreators: User[]
   messages: GroupComment[]
+  aboutPost: Post
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:131]: I think this should be typed `Post | undefined`.

web/components/groups/group-about-post.tsx
```diff
@@ -0,0 +1,143 @@
...
+      })
+      await updateGroup(group, {
+        aboutPostId: result.post.id,
+      })
+    } else {
+      await updatePost(post, {
+        content: newPost.content,
+      })
+    }
+    await router.replace(router.asPath)
```
jahooma, line 71: It's not the best solution to refresh the page. If you implement the `usePost` hook mentioned above, this might not be necessary. Then you could just set `setEditing(false)`.

jahooma (approved):
Awesome, Fede! I think it's a nice solution to show the post within the about page.

Left a few comments about the implementation which you should address for merging.

FRCassarino:
Fixed all nits, merging now.

PR #805, opened 2022-08-27
Make portfolio graph loading more efficient

mqp:
Since the default view period is a week, it sucked to pull down the whole history.

Note that the old code sort of implied that there might be history for prior to June, but in the actual database it seems that history was removed. There was no mechanism for filtering stuff only past June, and at a glance, it seems like the history collections I can find don't contain pre-June data points. So e.g. the code that was showing the "since June" label was never doing anything.

I also re-enabled the "daily" option. Austin claimed in [this commit](https://github.com/manifoldmarkets/manifold/commit/258b2a318fa626df536ca31ce178ec6ea6af4919) that it's "broken" but I see zero evidence of that. I'm guessing he got confused by the thing where the metrics update job was busted for a few days due to OOM.

### DIFF
--- a/web/components/portfolio/portfolio-value-section.tsx
+++ b/web/components/portfolio/portfolio-value-section.tsx
@@ -23,11 +23,8 @@ const periodToCutoff = (now: number, period: Period) => {
 }
 
 export const PortfolioValueSection = memo(
-  function PortfolioValueSection(props: {
-    userId: string
-    disableSelector?: boolean
-  }) {
-    const { disableSelector, userId } = props
+  function PortfolioValueSection(props: { userId: string }) {
+    const { userId } = props
 
     const [portfolioPeriod, setPortfolioPeriod] = useState<Period>('weekly')
     const [portfolioHistory, setUsersPortfolioHistory] = useState<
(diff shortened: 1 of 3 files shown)


web/lib/firebase/users.ts
```diff
@@ -252,11 +252,12 @@ export async function unfollow(userId: string, unfollowedUserId: string) {
   await deleteDoc(followDoc)
 }
 
-export async function getPortfolioHistory(userId: string) {
+export async function getPortfolioHistory(userId: string, since: number) {
   return getValues<PortfolioMetrics>(
     query(
       collectionGroup(db, 'portfolioHistory'),
       where('userId', '==', userId),
+      where('timestamp', '>=', since),
```
jahooma, line 260: Nice.

web/components/portfolio/portfolio-value-section.tsx
```diff
@@ -50,24 +55,22 @@ export const PortfolioValueSection = memo(
...
+            className="select select-bordered self-start"
+            value={portfolioPeriod}
+            onChange={(e) => {
+              setPortfolioPeriod(e.target.value as Period)
+            }}
+          >
+            <option value="allTime">All time</option>
+            <option value="weekly">Last 7d</option>
+            {/* Note: 'daily' seems to be broken? */}
+            {/* <option value="daily">Last 24h</option> */}
```
jahooma, line 68: Did you mean to enable this option? I think it's a good idea.

jahooma (approved):
Looks good. Thanks Marshall!

web/components/portfolio/portfolio-value-section.tsx
mqp: Oops, just forgot to push that commit.

PR #806, opened 2022-08-27
Fix some efficiency problems with `ContractProbGraph`

mqp:
1. The component was memoized, but the bets coming in were changing, so it was re-rendering unnecessarily.
2. The stuff it was doing was so inefficient it was taking O(tenths of a second) during render on large markets to just prepare the data set. (`dayjs` is really slow, don't use it more than you have to.)
3. It was generating interpolated graph points even before the bets loaded, which was both useless work and also ugly. This is what caused the graph to look like it was at 50% until the page loaded. Now it looks empty which I think is way better.

This substantially improves the loading jankiness of e.g. the Elon Musk market.

### DIFF
--- a/web/pages/[username]/[contractSlug].tsx
+++ b/web/pages/[username]/[contractSlug].tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import { ArrowLeftIcon } from '@heroicons/react/outline'
 
 import { useContractWithPreload } from 'web/hooks/use-contract'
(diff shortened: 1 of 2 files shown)


web/pages/[username]/[contractSlug].tsx
```diff
@@ -165,6 +165,10 @@ export function ContractPageContent(
   })
 
   const bets = useBets(contract.id) ?? props.bets
+  const nonChallengeBets = useMemo(
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:168]: Good

jahooma (approved):
Sweet! Looks like a bunch of good changes to make graphs render faster, thanks!

PR #807, opened 2022-08-27
Make tooltip rendering more efficient

mqp:
It looks sort of ridiculous, but a huge part of the cost of rendering our lists is these tooltips, so let's do basic things to try to make them more efficient. It won't matter as much once we're paginating these lists, but it won't not matter, either.

In particular, the `dayjs` formatting is really inefficient and you probably shouldn't use it unless you have some reason to. (For example, according to my profiler -- which I believe because I read the `dayjs` code -- formatting all the timestamps for the bet tooltips in the Elon Musk market took a large part of a second on my laptop.) The normal browser API does an equally good job efficiently.

### DIFF
--- a/web/components/datetime-tooltip.tsx
+++ b/web/components/datetime-tooltip.tsx
@@ -1,15 +1,12 @@
-import dayjs, { Dayjs } from 'dayjs'
-import utc from 'dayjs/plugin/utc'
-import timezone from 'dayjs/plugin/timezone'
-import advanced from 'dayjs/plugin/advancedFormat'
 import { Tooltip } from './tooltip'
 
-dayjs.extend(utc)
-dayjs.extend(timezone)
-dayjs.extend(advanced)
+const FORMATTER = new Intl.DateTimeFormat('default', {
+  dateStyle: 'medium',
+  timeStyle: 'long',
+})
 
 export function DateTimeTooltip(props: {
-  time: Dayjs
+  time: number
   text?: string
   className?: string
   children?: React.ReactNode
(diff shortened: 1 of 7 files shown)


web/components/datetime-tooltip.tsx
```diff
@@ -1,23 +1,20 @@
...
-import advanced from 'dayjs/plugin/advancedFormat'
 import { Tooltip } from './tooltip'
 
-dayjs.extend(utc)
-dayjs.extend(timezone)
-dayjs.extend(advanced)
+const FORMATTER = new Intl.DateTimeFormat('default', {
+  dateStyle: 'medium',
+  timeStyle: 'long',
+})
```
jahooma, line 6: Does this include the time zone?

jahooma (approved):
LGTM for improving performance!

We should aim to provide consistent formatter functions under `common`. It's good to have consistent formatting style, and it makes it easier to swap out the date-time implementation.

sipec (approved):

web/components/datetime-tooltip.tsx
mqp: Yes, at least on my en-US computer.

mqp:
> We should aim to provide consistent formatter functions under common. It's good to have consistent formatting style, and it makes it easier to swap out the date-time implementation.

I mostly disagree -- I don't think there is any value to "having consistent formatting style" of dates and times (I think you should pick a format that is the most user-friendly in the context it's being displayed) and I don't think that we will ever want to use a date-time formatter implementation other than `DateTimeFormat`, since it's feature-complete, built-in, and not going anywhere.

PR #808, opened 2022-08-27
De-feedify contract tab contents

mqp:
**WIP -- basically working but I need to test to make sure I didn't break any details.**

So basically my understanding of the situation is this:

1. We used to have a big need for a kind of "activity feed that shows all the different heterogeneous things that can happen on the website in chronological order, all mixed together." That's what we showed on the homepage.
2. Once we had code for the "heterogenous activity feed" it became useful to kind of use it for "any list of anything", because it knew how to render all stuff, so why not use it. For example, we used it for "the list of answers on the free response contract" even though that isn't really much like the old homepage. It didn't exactly make sense to use it for more homogeneous lists, but I guess it was the quickest way to get something on the screen.
3. At some point we also started using it for the bets and comment views in the contract page.
4. The original use case for this evaporated because we decided this wasn't a good UX for the homepage.
5. However, we still are using it for those bets and comment views.

This toasts the activity feed stuff and renders bets and comments in a more straightforward way. It is doing a lot less work, so it's probably a little more efficient, but my main motivation is cleaning stuff up for future changes. Now that these are not wrapped in the "activity feed" abstraction we can easily change how they are loading data to be more cromulent (e.g. denormalize, paginate, load on demand, etc.)


web/components/contract/contract-tabs.tsx
```diff
@@ -18,14 +27,18 @@ export function ContractTabs(props: {
   comments: ContractComment[]
   tips: CommentTipMap
 }) {
-  const { contract, user, bets, tips } = props
+  const { contract, user, tips } = props
   const { outcomeType } = contract
 
+  const updatedBets = useBets(contract.id, {
+    filterChallenges: false,
+    filterRedemptions: true,
```
jahooma, line 35: Redemptions are needed within the `userBets` to compute accurate invested/profit numbers, even if they don't appear in the list. Redemptions are not needed for the public bets list.

web/components/feed/feed-comments.tsx
```diff
@@ -231,31 +192,31 @@ export function FeedComment(props: {
...
-                  outcome={betOutcome ? betOutcome : ''}
-                  value={(matchedBet as any).value}
-                  contract={contract}
-                  truncate="short"
-                />
-              </>
-            )}
-          </>
+          {bought} {money}
+          {contract.outcomeType !== 'FREE_RESPONSE' && betOutcome && (
```
jahooma, line 196: Seems suspicious that this doesn't also exclude Multiple choice. I see that that's how it was though...

jahooma (approved):
Thanks Marshall! This is a big undertaking that improves code quality.

There's a lot of details here that I trust you'll get right haha. I did find one thing to fix about including redemption bets.

jahooma:
I also like that this refactor is -450 net lines of code 😎

mqp:
> I also like that this refactor is -450 net lines of code 😎

I believe there's more dead "feed stuff" code to clean up but I didn't bother hunting it all down right now.

web/components/contract/contract-tabs.tsx
mqp: OK, I'll make sure everything like this is getting computed correctly.

mqp:
OK, I tested more or less all the bets and comment feed functionality, so if I broke anything at least it's probably not something important. Merging this so that we can get on with other things.

PR #812, opened 2022-08-28
React-query-ify notifications

jahooma:
React query has a bug for subscribing to `useFirestoreQueryData`. I found a workaround online: https://github.com/invertase/react-query-firebase/issues/25

Secondly, I realized that if Nextjs Link components navigate client-side, we can always have the latest notifications fetched if we listen for them in the sidebar with the same query.

Now there is no fetch time at all to visit Notifications page! Except for the first load in which we show a loading indicator. Also removed the local storage for notifications since they are no longer necessary.


jahooma:
cc: @IanPhilips

PR #813, opened 2022-08-28
⚡ Cache user bets tab with react query!! 

jahooma:
From the insight that NextJS does client-side navigation when clicking Links, it means we can cache data in-memory.

This PR caches all the user bet data, including portfolio history, user bets, and the contracts corresponding to your bets using react query.

Further, we call these hooks on home and the market page to "precache" the data so that it will load instantly when navigating your user page bets tab.

To make this work, I backfilled the `uniqueBettorIds` in every contract so that we can query all the contracts you have bet on in one go.


jahooma:
cc: @mqp

PR #816, opened 2022-08-30
[WIP] Multiple choice line charts

ahalekelly:


ahalekelly:
People on discord seemed to prefer the current stacked area charts, maybe I'm in the minority here

akrolsmir:
> People on discord seemed to prefer the current stacked area charts, maybe I'm in the minority here

Yeah - I think in general, having the area filled in means you can get a sense of the total probability over time; I speculate that area is much easier for humans to interpret at a glance (compared to the relative positions of a thing on a 1d line). I think we'll keep the stacked area version for now, but thanks for proposing improvements to our design!

PR #822, opened 2022-08-31
Add embedded ContractGrid to Posts

FRCassarino:
Added embedded grids of contract cards to Posts:
https://user-images.githubusercontent.com/6242616/187686727-930a69ff-9196-484b-a408-b73b8c0c4560.mov

### DIFF
--- a/web/components/editor/market-modal.tsx
+++ b/web/components/editor/market-modal.tsx
@@ -29,7 +29,7 @@ export function MarketModal(props: {
   async function doneAddingContracts() {
     setLoading(true)
     if (contracts.length == 1) {
-      insertContent(editor, ...contracts.map(embedContractCode))
+      insertContent(editor, embedContractCode(contracts[0]))
     } else if (contracts.length > 1) {
       insertContent(editor, embedContractGridCode(contracts))
     }
--- a/web/components/editor/market-modal.tsx
+++ b/web/components/editor/market-modal.tsx
@@ -7,7 +7,7 @@ import { Col } from '../layout/col'
 import { Modal } from '../layout/modal'
 import { Row } from '../layout/row'
 import { LoadingIndicator } from '../loading-indicator'
-import { embedCode } from '../share-embed-button'
+import { embedContractCode, embedContractGridCode } from '../share-embed-button'
 import { insertContent } from './utils'
 
 export function MarketModal(props: {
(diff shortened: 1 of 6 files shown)


web/components/share-embed-button.tsx
```diff
@@ -9,13 +9,23 @@ import { DOMAIN } from 'common/envs/constants'
...
+export function embedContractGridCode(contracts: Contract[]) {
+  //TODO: Change before shipping
+  const src = `http://localhost:3000/embed/grid/${contracts
+    .map((c) => c.slug)
+    .join('/')}`
+  return embedCode(src, 'Grid of contracts')
 }
 
+const embedCode = (src: string, title: string) =>
+  `<iframe width="560" height="405" src="${src}" title="${title}" frameborder="0"></iframe>`
```
FRCassarino, line 27: @akrolsmir  @sinclairity  How can I make the iframe height cover the content? As far as I can tell those width and height attributes are not doing anything:

image

akrolsmir: I think it's actually controlled by the `padding-bottom: 20rem;` here. No idea why, but increasing it to 40rem does increase the iframe height, probably because the iframe itself is using `h-full` and ignoring the other params

image

akrolsmir: It's in https://github.com/manifoldmarkets/manifold/blob/26aba26da538fa978f1b0f205f5e09647709e362/common/util/tiptap-iframe.ts#L38

Simplest fix would be to add a param to dynamically control the iframe height, maybe?

Not really sure why the CSS works this way fwiw, I basically just copied the https://tiptap.dev/experiments/embeds example and hacked over it, very open to rearranging

web/components/editor/market-modal.tsx
```diff
@@ -28,7 +28,11 @@ export function MarketModal(props: {
 
   async function doneAddingContracts() {
     setLoading(true)
-    insertContent(editor, ...contracts.map(embedCode))
+    if (contracts.length == 1) {
+      insertContent(editor, ...contracts.map(embedContractCode))
```
sipec, line 32: nit: just do `insertContent(editor, embedContractCode(contracts[0]))` since there's exactly 1

sipec (approved):
looks good to me, pending height issue being fixed

PR #823, opened 2022-08-31
Clean up contract overview code

mqp:
This was hard for me to understand because it was a big rat's nest of "if it's contract type X, show Y", so I cleaned it up by factoring out the stuff that was the same between different kinds of contracts and making separate contract type overview components for the stuff that was different.

Basically no user-visible changes (I moved some of the pixels around a tiny bit.)

### DIFF
--- a/web/components/answers/answers-graph.tsx
+++ b/web/components/answers/answers-graph.tsx
@@ -18,19 +18,20 @@ export const AnswersGraph = memo(function AnswersGraph(props: {
 }) {
   const { contract, bets, height } = props
   const { createdTime, resolutionTime, closeTime, answers } = contract
+  const now = Date.now()
 
   const { probsByOutcome, sortedOutcomes } = computeProbsByOutcome(
     bets,
     contract
   )
 
-  const isClosed = !!closeTime && Date.now() > closeTime
+  const isClosed = !!closeTime && now > closeTime
   const latestTime = dayjs(
     resolutionTime && isClosed
       ? Math.min(resolutionTime, closeTime)
       : isClosed
       ? closeTime
-      : resolutionTime ?? Date.now()
+      : resolutionTime ?? now
   )
 
   const { width } = useWindowSize()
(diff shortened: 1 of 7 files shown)


web/pages/[username]/[contractSlug].tsx
```diff
@@ -232,6 +234,8 @@ export function ContractPageContent(
         )}
 
         <ContractOverview contract={contract} bets={nonChallengeBets} />
+        <ExtraContractActionsRow contract={contract} />
+        <ContractDescription className="mb-6 px-2" contract={contract} />
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:238]: Good idea to pull these out of `ContractOverview`

jahooma (commented):
LGTM, I think Ian was just modifying this so I added him as a reviewer!

web/components/answers/answers-graph.tsx
```diff
@@ -18,19 +18,20 @@ export const AnswersGraph = memo(function AnswersGraph(props: {
 }) {
   const { contract, bets, height } = props
   const { createdTime, resolutionTime, closeTime, answers } = contract
+  const now = Date.now()
```
sipec, line 21: wouldn't Date.now() still get recalculated every time the graph gets rerendered?

I would have done `const now = useMemo(Date.now)` or `const [now,] = useState(Date.now())` for this. Does React.memo() render that moot or am I missing something?

jahooma: Yes, but isn't it good to use the current time? Then the graph's x axis is more accurate.

mqp: Well, the `useMemo` on 14 means that this line won't run again unless the props change, so no, it won't get recalculated. But I agree with James -- that seems like a bug, not a feature. I would rather use the current time. Not enough to re-render the whole graph, though, which is quite slow.

IanPhilips (approved):
Looks cleaner, thanks for going through this! Is passing the user through props going out of style and we should default to using the `useUser()` hook everywhere it's needed now?

sipec:
> we should default to using the `useUser()` hook everywhere it's needed now?

Yes, since it'll just get the value cached by react query. Unless you do it in a component that has 100 instances like contact card - ime it might choke in those cases

mqp:
I agree that I don't think there's any important difference between passing down the user or using the `useUser` one so I just do whatever seems more convenient.

> Yes, since it'll just get the value cached by react query.

No, `useUser` doesn't have anything to do with react-query. The current user is stored in the `AuthContext` object at the root of the app.

> Unless you do it in a component that has 100 instances like contact card - ime it might choke in those cases

I would be interested to see this because it doesn't match my understanding of how it should work.

PR #826, opened 2022-08-31
Dramatically improve server auth stuff

mqp:
Old version (kind of embarrassing): We had multiple cookies, for the different kind of tokens we needed. Every time you hit the server on SSR paths, it would need to hit up Google to turn your custom token into an ID token and initialize the server-side SDK, which might take a quarter of a second. (Seems long, but that's what I measured in production.)

New version: One cookie stores the internal persistent SDK state, which we can reload instantly on the server. The only thing that will ever take time is if we need to refresh the ID token, which is basically unavoidable.

Note that we don't even need the admin SDK anymore.

The new version is really simple, not scary like how it used to be, so for more details just read the code. The only trick is the weird bit that reaches into the SDK to manipulate the persisted user.


mqp:
@akrolsmir @jahooma if you remember the old server auth stuff you might want to glance at this to see what's the deal now.

jahooma (approved):
Makes sense! Thanks for investigating this further to find a faster way and with only one cookie.

This makes SSR a better option for us, which will be useful.

PR #827, opened 2022-09-01
Fix prefetching to not populate useless state

mqp:
Since we (maybe?) have some reason to believe that this prefetching is hurting us due to increased memory pressure, let's at least not copy the prefetched data into state for no reason.


jahooma (approved):
Looks good! Thanks for improving this.

PR #828, opened 2022-09-01
emails: remove reference to creator commission

AnnikaCodes:


sipec (commented):
hmm I think we still want to remind them to close the market in this email. maybe like "Please resolve your market so the winners can get their mana payouts"

mantikoros:
Thank you for this PR. I just fixed this issue myself, but I appreciate the effort.

sipec:
ah, in e0ebdc644db

PR #830, opened 2022-09-01
Submit comments on ctrl/cmd-enter

akrolsmir:


web/components/feed/feed-comments.tsx
```diff
@@ -408,11 +409,9 @@ export function CommentInputTextArea(props: {
     submitComment,
     presetId,
     isSubmitting,
-    submitOnEnter,
+    submitOn,
     replyToUser,
   } = props
-  const isMobile = (useWindowSize().width ?? 0) < 768 // TODO: base off input device (keybord vs touch)
```
akrolsmir, line 414: @sipec wondering why we had this at all?

sipec: on touch devices, you should be able to submit on pressing enter, without needing to also press another key
or at least I think that's what it is - this existed before I refactored it way back when

IanPhilips: I think actually it's for the opposite: never allow submitting on enter on mobile, which it seems like noone wants.

IanPhilips: So I would add that back in!

PR #831, opened 2022-09-01
Fix editor rerendering when you load it

sipec:
also fixes bug where you are the only person worth mentioning


web/components/editor/mention-suggestion.ts
```diff
@@ -1,21 +1,24 @@
...
 import { searchInAny } from 'common/util/parse'
-import { orderBy } from 'lodash'
+import { memoize, orderBy } from 'lodash'
 import tippy from 'tippy.js'
+import { getUsers } from 'web/lib/firebase/users'
 import { MentionList } from './mention-list'
 
 type Suggestion = MentionOptions['suggestion']
 
+const users = memoize(getUsers)
```
jahooma, line 11: Seems like it would be better to use react query? The library kind of sucks if we can't cache the users and share that cache with other hooks, e.g. `useUserById`.

`useUserById` has the react query key ['users', userId]. The one you were using `useUsers` was just ['users']. So I think the cached results were not being shared between these hooks? Oh well.

I think you could replace this line with:
```
const queryClient = new QueryClient()
const users = () => queryClient.fetchQuery(['users'], getUsers)
```
Which would at least cache users for the useUsers hook.

jahooma (approved):
Thanks for fixing this! Kill two bugs with one PR.

web/components/editor/mention-suggestion.ts
mqp, line 11: Loading all the user documents from the database into memory to support a tiny @mention dropdown makes me extremely sad. This is degrading the user experience for everyone who ever looks at a page with a comment box, and it's going to get worse and worse forever.

I would consider these to be "generally reasonable, probably worth having" implementations:

- Querying on demand when you type characters in the box. It should be very possible to have a satisfying latency for this. I would probably do this.
- Providing a cloud function that returns all possible completions, and prefetching that, rather than the entire user docs. Although this doesn't scale indefinitely, it's 100x more practical than the current implementation.

I would suggest removing this feature until this is fixed because I don't believe it can possibly be net positive in its current form.

sipec: I agree that your proposed implementations would be better.
It looks like the `useUsers` hook is only used in cases where we search for particular users (like, adding to groups) so probably we could just replace all usages of that with some improved user search. I envision some combination of querying on demand and pre-filling with user data that is already there (like, the market creator and all the commenters)

> I would suggest removing this feature until this is fixed because I don't believe it can possibly be net positive in its current form.

disagree - we had already loaded in all the users via the useUsers hook. I favor doing the performant thing after fixing the editor rerendering bug.

mqp: I agree the new code doesn't look any worse than the old code in this regard, so this PR is OK with me, I am just saying I would advocate either fixing or deleting the thing instead.

PR #832, opened 2022-09-01
Make tournament page efficient

mqp:
The previous version loaded 200KB of compressed JSON, and since it's a `getStaticProps` page that's prefetched from the sidebar it was impacting the rest of the site. Now it loads the carousel contracts incrementally on the client.

I changed the carousel to sort by `popularityScore` of the contracts, both because it was convenient (we already have an index on that so I was able to do it in the query) and because it made way more sense to me than sorting by the current probability. If we really want it to sort by the current probability it seems to me we could add a new index and order by the `p` field on the contract to get that.


web/pages/tournaments/index.tsx
```diff
@@ -20,27 +14,33 @@ import { Col } from 'web/components/layout/col'
...
-import { getProbability } from 'common/calculate'
 import { Carousel } from 'web/components/carousel'
+import { usePagination } from 'web/hooks/use-pagination'
+import { LoadingIndicator } from 'web/components/loading-indicator'
 
 dayjs.extend(utc)
 dayjs.extend(timezone)
 dayjs.extend(customParseFormat)
-const toDate = (d: string) => dayjs(d, 'MMM D, YYYY').tz('America/Los_Angeles')
+const toDate = (d: string) =>
```
akrolsmir, line 31: Thanks, it's nice to standardize on numeric timestamps (even though I know you hate them haha)

web/pages/tournaments/index.tsx
```diff
@@ -107,130 +107,139 @@ const tourneys: Tourney[] = [
...
+        See more
+      </SiteLink>
+    </Carousel>
+  )
+}
+
+const MarketCarousel = (props: { slug: string }) => {
+  const { slug } = props
+  const q = contractsByGroupSlugQuery(slug)
+  const { allItems, getNext, isLoading } = usePagination({ q, pageSize: 12 })
```
akrolsmir, line 226: I doubled pagesize to 12 so that on first click, we don't require a load.

(The load looks really janky atm because the entire contract cards disappear on the next load; would like to get this fixed this later, but not high priority)

akrolsmir: Also really janky because the load resets your scroll position in the carousel....

Might not even be worth paginating, but rather have a final card that's "See all" that you click on and takes you to the entire page.

akrolsmir (approved):
Thanks a lot -- I totally did not realize that things in our sidebar would get loaded on every single page transition, so this will hopefully be much much better on our page load speeds!

web/pages/tournaments/index.tsx
mqp: It sounds like there's something going wrong here. It worked fine when I was using it.

PR #835, opened 2022-09-01
Make prefetching correctly use context cache

mqp:
We didn't understand how react-query worked, so we didn't realize that we needed to use the context's `QueryClient` to store the cached data. So our prefetches were having no effect.

Now the prefetching should work.


jahooma (approved):
Thanks for making the prefetch actually work! Code looks good.

PR #836, opened 2022-09-01
Disable group prefetching from contract links

mqp:
I wasn't able to convince James to disable most prefetching, but at least we compromised to not prefetch groups on contracts for as long as those groups are really expensive to prefetch (i.e. right now they load all their members, etc. etc.) since this is a particularly gross thing in the way of making the home/contract user flow better.

### DIFF
--- a/web/components/contract/contract-details.tsx
+++ b/web/components/contract/contract-details.tsx
@@ -116,25 +116,6 @@ export function AvatarDetails(props: {
   )
 }
 
-export function AbbrContractDetails(props: {
-  contract: Contract
-  showHotVolume?: boolean
-  showTime?: ShowTime
-}) {
-  const { contract, showHotVolume, showTime } = props
-  return (
-    <Row className="items-center justify-between">
-      <AvatarDetails contract={contract} />
-
-      <MiscDetails
-        contract={contract}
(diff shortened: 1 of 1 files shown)


web/components/contract/contract-details.tsx
```diff
@@ -162,12 +143,12 @@ export function ContractDetails(props: {
         isMobile ? 'max-w-[140px]' : 'max-w-[250px]'
       )}
     >
-      <SiteLink href={groupPath(groupToDisplay.slug)} className={'truncate'}>
-        <Row>
+      <Link prefetch={false} href={groupPath(groupToDisplay.slug)}>
```
mqp, line 146: I don't really think the `SiteLink` -> `MaybeLink` abstraction makes any sense to use when we know in advance whether or not the link is to something on Manifold.

jahooma: I think I created `SiteLink` and `ExternalLink` once. `SiteLink` made sure to use `Link` and `a`, whereas `ExternalLink` just used `a` and opened in a new tab.

If `SiteLink` were still internal we could add a prefetch prop, but oh well I guess this is fine.

jahooma (approved):
Looks good. Thanks

PR #838, opened 2022-09-02
Denormalize `betAmount` and `betOutcome` fields on comments

mqp:
This is making progress towards not forcing the comment tab to load all the bets on the contract.

### DIFF
--- a/web/components/feed/feed-comments.tsx
+++ b/web/components/feed/feed-comments.tsx
@@ -125,15 +125,12 @@ export function FeedComment(props: {
   } = props
   const { text, content, userUsername, userName, userAvatarUrl, createdTime } =
     comment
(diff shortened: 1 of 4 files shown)


web/components/feed/feed-comments.tsx
```diff
@@ -194,7 +191,6 @@ export function FeedComment(props: {
               of{' '}
               <OutcomeLabel
                 outcome={betOutcome ? betOutcome : ''}
-                value={(matchedBet as any).value}
```
mqp, line 197: I removed this since it only matters for old numeric contracts and that's not worth putting extra denormalized fields in our database for.

jahooma (approved):
LGTM!

PR #839, opened 2022-09-02
Disable SSR on /home

mqp:
Before, it was good for /home to have SSR because it was meant to kick you out if you were logged out, and SSR was the best method to accomplish that.

But now, /home basically looks the same whether you are logged in or logged out, so doing SSR auth has little benefit, and slightly slows down navigations to /home (e.g. it makes going back from contracts to the search slightly slower, and it's really good for that to be absolutely as fast as possible.)


jahooma (approved):
Great! Manifold will have the best performance of any prediction market home page.

PR #840, opened 2022-09-02
Remove custom token generation machinery

mqp:
This was for my old SSR scheme previous to #826, but there's no reason to have it around anymore. (And it's a little spooky to have around for no reason, due to security concerns.)


jahooma (approved):
LGTM — thanks for cleaning up!

PR #841, opened 2022-09-02
Clean up some old pre-Amplitude tracking code

mqp:
I don't believe anyone is using this anymore, so let's toss it.


jahooma (approved):
Thanks for the cleanup!

PR #842, opened 2022-09-02
Tidy up some dead code and markup in sidebar

mqp:
Not anything important, just noticing some stuff here I could clean up.

### DIFF
--- a/web/components/nav/sidebar.tsx
+++ b/web/components/nav/sidebar.tsx
@@ -234,11 +234,7 @@ export default function Sidebar(props: { className?: string }) {
 
       {!user && <SignInButton className="mb-4" />}
 
-      {user && (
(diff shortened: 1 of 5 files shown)


web/components/nav/sidebar.tsx
```diff
@@ -264,7 +260,7 @@ export default function Sidebar(props: { className?: string }) {
           buttonContent={<MoreButton />}
         />
 
-        {user && <CreateQuestionButton user={user} />}
+        {user && !user.isBannedFromPosting && <CreateQuestionButton />}
```
jahooma, line 263: One downside of moving the ban condition out is someone might forget to enforce it on another use of the button. But I guess the button isn't really reused much / not a big problem.

jahooma (approved):
Thanks for tidying these up!

PR #843, opened 2022-09-02
Remove some old code related to the old feed

mqp:
This is related to the old homepage activity feed and hasn't been relevant for a while.


jahooma (approved):
I remember this old code. LGTM

PR #845, opened 2022-09-02
Leaderboard calc: update profit even when portfolio didn't change 

jahooma:
If the portfolio didn't change since the cached one, that means all-time profit is the same, but not daily, weekly, or monthly.


jahooma:
cc: @FRCassarino

PR #846, opened 2022-09-02
Some changes to make auth better

mqp:
A variety of small things:

1. The `listenForUser` and `listenForPrivateUser` subscriptions that keep track of the current logged in user docs weren't handling the case where (surprisingly) they can't find the user by ID. But this case actually was coming up in the case where you (e.g. on your local dev server) have a cached user object for prod, and then log into dev, and now you have one with an invalid dev database ID. The consequence is that it would mysteriously explode on the first page load and make a scary message. Now we handle this much more gracefully.

2. Those subscriptions also weren't updating the local storage blob when your user info changes, which they should.

3. It was really weird that we called the `setCachedReferralInfoForUser` on every login even though it only matters when you create a user. Maybe someone can explain to me why this code was the way it was but I have no idea.

4. We no longer mix together the "listen for user stuff" and "update analytics username" in the same `useEffect`, so it makes more sense and doesn't do extra work when you change your username.

### DIFF
--- a/web/components/auth-context.tsx
+++ b/web/components/auth-context.tsx
@@ -103,11 +103,9 @@ export function AuthProvider(props: {
   }, [setAuthUser])
 
   const uid = authUser?.user.id
-  const username = authUser?.user.username
   useEffect(() => {
-    if (uid && username) {
+    if (uid) {
       identifyUser(uid)
-      setUserProperty('username', username)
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Cool, makes sense, and makes auth more robust. LGTM!

PR #848, opened 2022-09-02
Make `parse.richTextToString` more efficient

mqp:
This is necessary for rendering the OG meta description when loading the content page, so it's in the hotspot for the user browsing flow. Profiling suggests that recomputing the ProseMirror schema every time (the behavior of the previous version) may cost on the order of tens of milliseconds. That code is [here](https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/helpers/getSchemaByResolvedExtensions.ts#L23) -- I don't know if I believe the profiler but clearly we should just not do this.

I continue to update in the direction of "ProseMirror may or may not be good, but TipTap is often suspicious." It's kind of ridiculous to me that the `generateText` method they supplied took the extensions instead of the schema as an argument. Not a good design.

EDIT: Before merging this I want to look at some other TipTap code to understand what the deal is with extension resolution. It's also possible that in addition to this code I should just go and add a pre-rendered plaintext copy in the DB, which would be very useful, but it would also bloat our contract docs further, so it kind of sucks too.

### DIFF
--- a/common/package.json
+++ b/common/package.json
@@ -13,6 +13,7 @@
     "@tiptap/extension-link": "2.0.0-beta.43",
     "@tiptap/extension-mention": "2.0.0-beta.102",
     "@tiptap/starter-kit": "2.0.0-beta.191",
+    "prosemirror-model": "1.18.1",
     "lodash": "4.17.21"
   },
   "devDependencies": {
--- a/common/util/parse.ts
+++ b/common/util/parse.ts
@@ -1,5 +1,12 @@
 import { MAX_TAG_LENGTH } from '../contract'
-import { generateText, JSONContent } from '@tiptap/core'
+import {
+  getText,
+  getTextSerializersFromSchema,
+  getSchema,
+  JSONContent,
+} from '@tiptap/core'
+import { Node as ProseMirrorNode } from 'prosemirror-model'
+
 // Tiptap starter extensions
 import { Blockquote } from '@tiptap/extension-blockquote'
 import { Bold } from '@tiptap/extension-bold'
@@ -90,7 +97,6 @@ export const exhibitExts = [
   Paragraph,
   Strike,
   Text,
-
   Image,
   Link,
   Mention,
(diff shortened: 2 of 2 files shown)


sipec:
> Profiling suggests that recomputing the ProseMirror schema every time (the behavior of the previous version) may cost on the order of tens of milliseconds. 

the natural question I have is whether profiling the changed version shows an improvement on the order of tens of milliseconds

also, why are we trying to shave off tens of milliseconds? I think making code longer for small performance improvements is bad

mqp:
> > Profiling suggests that recomputing the ProseMirror schema every time (the behavior of the previous version) may cost on the order of tens of milliseconds.
> 
> the natural question I have is whether profiling the changed version shows an improvement on the order of tens of milliseconds

Yes, it was the schema generation step that cost the time, so now if we (for example) load a contract, go back to the homepage, and load another contract, the schema will already have been generated the first time, so the second time will skip that work.

I have to say a caveat which is that I don't happen to know now exactly which parts of the schema generation were slow, so I can't say whether in some condition they would be faster or slower. This change makes sense to me regardless so I just did it anyway without further research.

> also, why are we trying to shave off tens of milliseconds?

What? If I could shave off 50 milliseconds per half hour of work over and over, I would do nothing but that for two weeks, and then everything would work instantly and feel great, and it would be a great use of time.

> I think making code longer for small performance improvements is bad

Tens of milliseconds on page load is a large performance improvement, not a small performance improvement. If I could write 1000 lines of code to speed up the website by several seconds, I would obviously do that.

I strongly disagree that "length of the code in our file" is the right thing to optimize for. I want to optimize for the program being simple and straightforward, and then adjust as necessary to make it more efficient. This is doing the same thing as the previous code, except that it stores the schema instead of recomputing it every time, which I think makes common sense and is only slightly less simple. I really couldn't care less if the code is in our file or in TipTap's file.

The main thing I don't like is that this only helps after doing it once, so I am interested in approaches that make everything simpler and do less work to render rich text, rather than relying on storing intermediate results. But I don't currently understand ProseMirror and TipTap enough to suggest an approach like that, so this PR is what I can suggest in the meantime.

sipec (approved):
Well it certainly won't be 30 min per change like this if I take 6 days to approve 😅 

Sorry, I don't have a lot of context on prosemirror's internals either, but there's nothing obviously wrong about this code to me. Merge it if you want.

mqp:
> Well it certainly won't be 30 min per change like this if I take 6 days to approve 😅
> 
> Sorry, I don't have a lot of context on prosemirror's internals either, but there's nothing obviously wrong about this code to me. Merge it if you want.

It's just half an hour of work for me! I don't care if it sits for a week, that's fine.

PR #849, opened 2022-09-02
Don't re-create visibility observer for no reason

mqp:
I don't think this was causing any actual problem, but it was just obviously silly to re-create the `IntersectionObserver` object whenever the ref changes.

### DIFF
--- a/web/components/visibility-observer.tsx
+++ b/web/components/visibility-observer.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import { useEvent } from '../hooks/use-event'
 
 export function VisibilityObserver(props: {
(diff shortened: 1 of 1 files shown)


web/components/visibility-observer.tsx
```diff
@@ -8,17 +8,18 @@ export function VisibilityObserver(props: {
   const { className } = props
   const [elem, setElem] = useState<HTMLElement | null>(null)
   const onVisibilityUpdated = useEvent(props.onVisibilityUpdated)
-
-  useEffect(() => {
-    const hasIOSupport = !!window.IntersectionObserver
```
mqp, line 13: https://caniuse.com/intersectionobserver

jahooma: I guess we don't support IE11 or old versions of other browsers.

web/components/visibility-observer.tsx
```diff
@@ -8,17 +8,18 @@ export function VisibilityObserver(props: {
...
-    if (!hasIOSupport || !elem) return
-
-    const observer = new IntersectionObserver(([entry]) => {
+  const observer = useRef(
+    new IntersectionObserver(([entry]) => {
```
jahooma, line 12: This actually creates a new object on every render, and all but the first are discarded. Doesn't seem that much better. You could do the use state with init function to solve this. Otherwise, I don't think it really matters.

jahooma (approved):
LGTM

web/components/visibility-observer.tsx
mqp: Huh, TIL. Well, I won't bother changing it to something uglier unless it ever becomes an actual problem.

PR #850, opened 2022-09-03
Sort markets on tournament pages by %, using ContractSearchFirestore

akrolsmir:
This is super hacky and trades off a much longer initial load time, for a better browsing experience... Not sure if this is the right longterm solution but I _think_ it's better for the CEP contest?

### DIFF
--- a/web/pages/contract-search-firestore.tsx
+++ b/web/pages/contract-search-firestore.tsx
@@ -61,6 +61,10 @@ export default function ContractSearchFirestore(props: {
     matches = sortBy(matches, (c) =>
       getProbability(c as BinaryContract | PseudoNumericContract)
     ).reverse()
+  } else if (sort === 'lowest-percent') {
+    matches = sortBy(matches, (c) =>
+      getProbability(c as BinaryContract | PseudoNumericContract)
+    )
   }
 
   if (additionalFilter) {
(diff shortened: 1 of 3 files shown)


akrolsmir:
Just realized in ContractSearchFirestore, `const contracts = useContracts()` actually loads the ENTIRE set of Manifold contracts before doing an in-memory filter over them... quite suboptimal

mqp:
Why do you need to use `ContractSearchFirestore`? It seems to me this is just the homepage with a hardcoded group slug and a hardcoded sort so what's the problem with using Algolia?

akrolsmir:
> Why do you need to use `ContractSearchFirestore`? It seems to me this is just the homepage with a hardcoded group slug and a hardcoded sort so what's the problem with using Algolia?

Algolia didn't know about contract probabilities, so can't sort by them afaict. ContractSearchFirestore has the nice property of allowing us to calculate arbitrary sorts based on any computable/derivable property from the contract itself.

mqp:
> > Why do you need to use `ContractSearchFirestore`? It seems to me this is just the homepage with a hardcoded group slug and a hardcoded sort so what's the problem with using Algolia?
> 
> Algolia didn't know about contract probabilities, so can't sort by them afaict. ContractSearchFirestore has the nice property of allowing us to calculate arbitrary sorts based on any computable/derivable property from the contract itself.

It's in the database now (James added it [last week](https://github.com/manifoldmarkets/manifold/commit/9060abde8ec41a6cd37faf71c7b9328928b94558)) so presumably it should be easy to index.

gappleto97:
Here's a market for this: https://manifold.markets/MarketManagerBot/how-many-days-after-september-1st-w-bff738132665

akrolsmir:
Not needed now that this is supported in Algolia!

PR #851, opened 2022-09-03
Use the magic `auth` prop for groups SSR

mqp:
When loading a user via SSR, it's good to load the user and private user and put them in the `auth` prop, because [the `AuthContext` code will look at that prop to load up its initial state](https://github.com/manifoldmarkets/manifold/blob/main/web/pages/_app.tsx#L79). So passing it via the `auth` prop will make stuff using `useUser` outside your page component, like the sidebar, render correctly on the server, and save your component a re-render later on, because `useUser` will start out with the right thing in it, instead of starting out with e.g. the local storage cache and later loading the up-to-date user doc from the DB.

(Loading the user and private user in parallel is basically equally fast as loading just the user, and every page is eventually going to want the private user in order to render e.g. the notifications bubble. So if you're loading the user in SSR, there's basically no reason not to load the private user too.)

Sorry for not making it more obvious that the `auth` prop is "magic" but I don't know of any way that would have been better to express this. Maybe if I were really galaxy brained I would write an eslint plugin.

### DIFF
--- a/web/pages/groups.tsx
+++ b/web/pages/groups.tsx
@@ -9,7 +9,7 @@ import { Page } from 'web/components/page'
 import { Title } from 'web/components/title'
 import { useGroups, useMemberGroupIds } from 'web/hooks/use-group'
 import { groupPath, listAllGroups } from 'web/lib/firebase/groups'
-import { getUser, User } from 'web/lib/firebase/users'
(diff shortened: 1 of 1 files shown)


IanPhilips (approved):
Ohhh I see! thanks for the explainer, I'll stick to using auth next time.

PR #853, opened 2022-09-03
Denormalize user display fields onto bets

mqp:
Pretty self-explanatory. Previously we needed to get the user doc for every bet.

Changing this is more valuable than you might think, even in the world when the contract bets tab and order book (the two places where we show a lot of bets that aren't yours) are paginated, because not only is it relatively expensive to get all the whole user docs separately, it's actually relatively expensive to just create and destroy react-query queries (i.e. what we were currently doing with `useUser`.) It's quite bad to be creating dozens of them and then throwing them out when you leave the page.

### DIFF
--- a/functions/src/on-create-bet.ts
+++ b/functions/src/on-create-bet.ts
@@ -61,6 +61,12 @@ export const onCreateBet = functions
     const bettor = await getUser(bet.userId)
     if (!bettor) return
 
+    await change.ref.update({
+      userAvatarUrl: bettor.avatarUrl,
+      userName: bettor.name,
+      userUsername: bettor.username,
+    })
+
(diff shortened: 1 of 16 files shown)


functions/src/on-create-bet.ts
```diff
@@ -58,6 +58,12 @@ export const onCreateBet = functions.firestore
     const bettor = await getUser(bet.userId)
     if (!bettor) return
 
+    await change.ref.update({
+      userAvatarUrl: bettor.avatarUrl,
+      userName: bettor.name,
+      userUsername: bettor.username,
+    })
```
mqp, line 65: Right now, when we create for example a comment, we send some of the denormalized fields in the Firestore create request, like the denormalized user data. But I think it would be better to change those places so that they just sent the normalized version of the object, and Firestore triggers did the denormalization.

PR #855, opened 2022-09-06
Adds investmentValue to group leaderboard calculation

FRCassarino:
Relevant task: https://www.notion.so/manifoldmarkets/Change-group-leaderboard-to-update-based-on-bet-value-in-unresolved-markets-1a55b41650a04d468225212bb49a11e7

@jahooma Please help me validate whether the math here is correct. 

Sanity-checking on Localhost seems to work. I printed out the values of investmentValues for each user, and it's what I expect. For any given contract, they match the sum of these two values, which I believe is what's expected:

image

Looking at some example group leaderboards, it seems to be correct as well. For the group that contains the contract from the last image (the only one I bet on):

image

### DIFF
--- a/common/scoring.ts
+++ b/common/scoring.ts
@@ -34,17 +34,11 @@ export function scoreUsersByContract(contract: Contract, bets: Bet[]) {
   const profits = bets.map((bet) => {
     const { userId } = bet
     const payout = getContractBetMetrics(contract, [bet]).profit
-    console.log({
-      userId: userId,
-      metrics: getContractBetMetrics(contract, [bet]),
-    })
-    return { userId: userId, payout: payout }
+    return { userId, payout }
   })
 
-  const netPayouts = [...profits]
-
   const userScore = mapValues(
-    groupBy(netPayouts, (payout) => payout.userId),
+    groupBy(profits, (payout) => payout.userId),
     (payouts) => sumBy(payouts, ({ payout }) => payout)
   )
 
--- a/common/scoring.ts
+++ b/common/scoring.ts
@@ -31,18 +31,8 @@ export function scoreTraders(contracts: Contract[], bets: Bet[][]) {
 }
 
 export function scoreUsersByContract(contract: Contract, bets: Bet[]) {
-  const profits = bets.map((bet) => {
-    const { userId } = bet
-    const payout = getContractBetMetrics(contract, [bet]).profit
-    return { userId, payout }
-  })
-
-  const userScore = mapValues(
-    groupBy(profits, (payout) => payout.userId),
-    (payouts) => sumBy(payouts, ({ payout }) => payout)
-  )
-
-  return userScore
+  const betsByUser = groupBy(bets, bet => bet.userId)
+  return mapValues(betsByUser, bets => getContractBetMetrics(contract, bets).profit)
 }
 
 export function addUserScores(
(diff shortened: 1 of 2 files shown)


common/scoring.ts
```diff
@@ -62,7 +63,18 @@ export function scoreUsersByContract(contract: Contract, bets: Bet[]) {
       return { userId, payout }
     })
 
-  const netPayouts = [...resolvePayouts, ...salePayouts, ...investments]
+  const investmentValues = openBets.map((bet) => {
```
FRCassarino, line 66: investmentValues calculates the investmentValue for each open bet in the contract (only if the contract is not yet resolved).

common/scoring.ts
```diff
@@ -62,7 +63,18 @@ export function scoreUsersByContract(contract: Contract, bets: Bet[]) {
...
+    const { userId } = bet
+    const investmentValue = computeInvestmentValueForBet(bet, contract)
+    return { userId, payout: investmentValue }
+  })
+
+  const netPayouts = [
+    ...resolvePayouts,
+    ...salePayouts,
+    ...investments,
+    ...investmentValues,
```
FRCassarino, line 76: As far as I understand it, this is what we want to add to the current leaderboard. 

Before this PR:
The group leaderboard score for each user would only be based on profits on resolved markets. 
After this PR:
InvestmentValues is added, i.e the current value of the shares each user has on each **unresolved** contract in the group.

FRCassarino:
The static analysis failure seems unrelated to this PR.

common/scoring.ts
jahooma: I'll have more thoughts after all my meetings, but very quickly:

Investment value is not exactly right. If someone bets M$1000 that shouldn't automatically put them on the leaderboard. What we care about is current profit on their investment.

Profit is the current investment value minus the initial investment. You can also check how we calculate profit in `calculate.ts`.

FRCassarino: @jahooma Right, yeah makes sense. Will update after the standup.

FRCassarino:
@jahooma I am now subtracting the initial investment like you said, and using only the profit. Of course this also means that not only profits will affect the leaderboard, but losses as well. I'm assuming this is intended.

common/scoring.ts
```diff
@@ -62,7 +63,25 @@ export function scoreUsersByContract(contract: Contract, bets: Bet[]) {
...
+      metrics: getContractBetMetrics(contract, [bet]),
+    })
+    return { userId: userId, payout: payout }
+  })
+
+  const netPayouts = [
+    ...resolvePayouts,
+    ...salePayouts,
+    ...investments,
+    ...profits,
```
jahooma, line 83: I think the profit payouts include profits from resolution and sales and investment. So I think you can replace all this code with just the profit payouts. (The previous scoring code looks pretty old.)

For groups, I think we're going to want to not fetch all the contracts and bets to compute this in getStaticProps, so we should aim to do it in the `updateMetrics` cloud function.

FRCassarino: I early return in the profits method if the bets are resolved, so it is working as expected.

But fair, if we can replace all of it that's better, I'll look into it.

jahooma:
I've gone ahead and merged this as it seemed we needed the leaderboard for CEP tournament

PR #856, opened 2022-09-07
Members to memberIds

FRCassarino:


web/pages/group/[...slugs]/index.tsx
```diff
@@ -108,18 +109,27 @@ export async function getStaticPropz(props: { params: { slugs: string[] } }) {
...
+  userScores: { [userId: string]: number },
+  userIds: string[]
+) {
+  // sort userScores by score and trim to top 50
+  const sortedUserScores = Object.fromEntries(
+    Object.entries(userScores)
+      .sort(([, scoreA], [, scoreB]) => scoreB - scoreA)
+      .slice(0, 50)
+  ) as { [userId: string]: number }
+
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:122]: Why do we need to sort these, when we sort their entries below? Seems like you could remove this code.

web/pages/group/[...slugs]/index.tsx
```diff
@@ -108,18 +109,27 @@ export async function getStaticPropz(props: { params: { slugs: string[] } }) {
...
     10
   ).filter(([_, score]) => score >= 0.5)
 
-  const topUsers = topUserPairs.map(
-    ([userId]) => users.filter((user) => user.id === userId)[0]
-  )
-  return topUsers.filter((user) => user)
+  const topUserIds = topUserPairs
+    .map(([userId]) => userIds.filter((uid) => uid === userId)[0])
+    .filter((userId) => userId != null) as string[]
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:130]: Could be:
`filterDefined(topUserPairs.map(([userId]) => userIds.find(userId)))`

web/pages/group/[...slugs]/index.tsx
```diff
@@ -108,18 +109,27 @@ export async function getStaticPropz(props: { params: { slugs: string[] } }) {
...
+  return Promise.all(topUserIds.map((userId) => getUser(userId)))
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:131]: Could be `Promise.all(topUserIds.map(getUser))`

web/pages/group/[...slugs]/index.tsx
```diff
@@ -398,66 +408,11 @@ function GroupOverview(props: {
             />
           </Col>
         )}
-
-        <Col className={'mt-2'}>
-          <div className="mb-2 text-lg">Members</div>
-          <GroupMemberSearch members={members} group={group} />
-        </Col>
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:405]: So we are removing the list of members? I think that's ok for now.

jahooma (approved):
Look reasonable, thanks Fede!

Will be great to fix this error. The direction groups are going makes it less important to know all the other members in the group.

It looks like we can almost remove the `memberIds` prop as well, we just need to calculate `isMember` for yourself (probably with a query), and then not use `memberIds` in the leaderboard either.

web/pages/group/[...slugs]/index.tsx
Comment (FRCassarino) [reply web/pages/group/[...slugs]/index.tsx:405]: Yep, Stephen and Ian said that was fine.

PR #861, opened 2022-09-08
add commander category

marsteralex:
commanders


akrolsmir (approved):
LGTM! Commanders are one of the easiest categories I think, but still fun to learn about new/random ones

PR #866, opened 2022-09-10
WIP: add artist category

marsteralex:


akrolsmir (approved):
LGTM, though I will note that this category is impossible to get right

.vscode/settings.json
```diff
@@ -9,5 +9,8 @@
   },
   "editor.formatOnSave": true,
   "editor.formatOnPaste": true,
-  "editor.defaultFormatter": "esbenp.prettier-vscode"
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "[python]": {
```
akrolsmir, line 13: hrm, do we need to check this in? feels kind of weird since these vscode settings are used by the rest of the team, but python is only used for some light scripting for the mtg thing

akrolsmir (approved):
LGTM, as discussed the large mb doesn't seem to be too bad of an issue since it's getting compressed by gzip I think. It is a slight annoyance to have large files checked into git, but not _that_ bad. (if it becomes a real problem later, we can rehost the jsons on a firestore bucket or s3 or something)

I like the new text thing showing what the actual answer was! Suggestion: would be nice to show the correct answer when the user didn't guess. 

image

Not sure how to make this look good though, maybe using emoji ❌ and ✅  next to the answers instead of the crossouts?

PR #867, opened 2022-09-10
Add floating menu (bold, italic, link)

sipec:
Add a medium-style highlight menu for formatting text and adding links

### DIFF
--- a/web/lib/icons/bold-icon.tsx
+++ b/web/lib/icons/bold-icon.tsx
@@ -1,0 +1,20 @@
+// from Feather: https://feathericons.com/
+export default function BoldIcon(props: React.SVGProps<SVGSVGElement>) {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      width="24"
+      height="24"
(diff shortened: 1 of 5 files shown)


web/lib/icons/bold-icon.tsx
```diff
@@ -0,0 +1,20 @@
+// from Feather: https://feathericons.com/
+export default function BoldIcon(props: React.SVGProps<SVGSVGElement>) {
```
jahooma, line 2: Why use a default export? Seems better to be consistent with the rest of the codebase.

jahooma (approved):
This is great! Solves a lot of the remaining issues with the editor. Now it's easy to add links, and if the italics are acting up, it's easy to turn them off.

Code looks good!

web/lib/icons/bold-icon.tsx
sipec: it's how all the other svgs work, and it matches how the svgs would be imported if we saved these as .svg rather than .tsx and used an svg loader like [svgr](https://react-svgr.com/docs/webpack/)

PR #869, opened 2022-09-11
Use %mention to embed a contract card in rich text editor

akrolsmir:


akrolsmir:
(right now we just drop the @mention user functionality. I'm leaning towards copying over the code wholesale rather than trying some kind of code reuse, between %mention and @mention, but open to reuse if people have ideas on how to do it cleanly)

PR #872, opened 2022-09-12
"Fix "500 internal error" in large groups

FRCassarino:
This is a new PR for the 500 internal error. I had to revert the previous merge, as the updateMetrics function is throwing an error. 

@jahooma How can I test this in the dev instance? I've tested on localhost of course, by using yarn start and calling updateMetrics() manually. But there seems to be some difference between the behaviour there and in production, as I get no errors locally. 

Should I deploy this to dev instead and test there? 

Logs for updateMetrics are here: 
https://console.cloud.google.com/functions/details/us-central1/updateMetrics?project=mantic-markets&authuser=1&hl=es&tab=logs

### DIFF
--- a/common/group.ts
+++ b/common/group.ts
@@ -12,7 +12,18 @@ export type Group = {
   aboutPostId?: string
   chatDisabled?: boolean
   mostRecentContractAddedTime?: number
+  cachedLeaderboard?: {
+    topTraders: {
+      userId: string
+      score: number
+    }[]
+    topCreators: {
+      userId: string
+      score: number
+    }[]
+  }
 }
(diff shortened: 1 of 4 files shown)


jahooma (approved):
Thanks Fede! I decided to do a post-merge review to give some feedback.

Overall, I think this design for computing group leaderboards is good!

One thing I was mentioning in standup is that there's a script `update-metrics.ts` which calls `updateMetricsCore`. You can run this yourself to try to debug it e.g. add some console.logs and run the command `ts-node functions/src/scripts/update-metrics.ts`

PR #873, opened 2022-09-13
Move tabs to sidebar

FRCassarino:
Moves group tabs to the sidebar. 

image
image
image
image
image
image


FRCassarino:
Still has a couple of visual bugs that I'll need to fix before shipping, but it should be shippable after that if we agree on it.

@akrolsmir @jahooma Is there anything else you'd like added before we ship this?

web/pages/group/[...slugs]/index.tsx
```diff
@@ -196,85 +198,118 @@ export default function GroupPage(props: {
...
-            <div
-              className={'line-clamp-1 my-2 text-2xl text-indigo-700 sm:my-3'}
-            >
-              {group.name}
-            </div>
-            <div className={'hidden sm:block'}>
-              <Linkify text={group.about} />
-            </div>
+    <>
+      <header className="sticky top-0 z-50 w-full pb-2 md:hidden lg:col-span-12">
```
Comment (FRCassarino) [inline web/pages/group/[...slugs]/index.tsx:257]: https://user-images.githubusercontent.com/6242616/190210177-4cb11d60-8792-4af9-be01-35af7c234f69.mov

Does anybody have a good idea for how to fix this? Tailwinds 'sticky' position uses the browser window as its position reference, so ContractsSearchFilters overlaps the header. What I want is the same behaviour, but with the component stopping at the header instead.

gappleto97:
Here's a market for this: https://manifold.markets/MarketManagerBot/how-many-days-after-september-1st-w-7a49d7fb4b06

web/components/nav/group-nav-bar.tsx
```diff
@@ -0,0 +1,94 @@
...
+        className={clsx(
+          'block w-full py-1 px-3 text-center hover:bg-indigo-200 hover:text-indigo-700',
+          currentPage === item.key && 'bg-gray-200 text-indigo-700'
+        )}
+        onClick={track}
+      >
+        {item.icon && <item.icon className="my-1 mx-auto h-6 w-6" />}
+        {item.name}
+      </a>
+    </button>
```
jahooma, line 92: I'm not sure if this kind of link works in all cases. For example, if someone middle-clicks the button, would it open the link in a new tab? Or, if you tab to focus-select it and hit enter, does it navigate (and also trigger the track function)?

You might have to use the nextjs `<Link>` wrapper and set `href`, or use `SiteLink`.

web/components/nav/group-sidebar.tsx
```diff
@@ -0,0 +1,87 @@
...
+  buildArray(
+    user && {
+      name: 'Notifications',
+      href: `/notifications`,
+      key: 'notifications',
+      icon: NotificationsIcon,
+    }
+  )
+
+export default function GroupSidebar(props: {
```
jahooma, line 31: We don't typically use default exports, except for pages. It's good to be consistent so you always know how to import anything.

Two reasons for this convention: 1. can easily export more things from a file without updating all references 2. Renaming an export will automatically rename it everywhere it is referenced, which is not true for default exports.

web/pages/group/[...slugs]/index.tsx
Comment (jahooma) [reply web/pages/group/[...slugs]/index.tsx:257]: Hmm, good question. I'm not sure. There might be a way to add an offset to the search bar's sticky y-offset.

Alternatively, you could just pass an option to disable the sticky search bar for groups on mobile.

Comment (jahooma) [reply web/pages/group/[...slugs]/index.tsx:257]: One small suggestion might be to extract this header to a new component in this file.

web/pages/group/[...slugs]/index.tsx
```diff
@@ -196,85 +198,118 @@ export default function GroupPage(props: {
...
+            'mx-auto w-full pb-[58px] lg:grid lg:grid-cols-12 lg:gap-x-2 lg:pb-0 xl:max-w-7xl xl:gap-x-8'
+          }
+        >
+          <Toaster />
+          <GroupSidebar
+            groupName={group.name}
+            className="sticky top-0 hidden divide-gray-300 self-start pl-2 lg:col-span-2 lg:flex"
+            onClick={onSidebarClick}
+            joinOrAddQuestionsButton={joinOrAddQuestionsButton}
+            currentKey={sidebarPages[sidebarIndex].key}
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:285]: This should probably be represented in the URL, I think. Also could allow you to go back and forth.

A long time ago, you could go to `/group/JamesGroup/leaderboard` and it would show the leaderboard, etc.

Actually, now that these are not tabs, I think it's even more expected that the distinct pages of markets, about, leaderboard are different URLs. You can either change this PR, or follow up with the change.

Comment (jahooma) [reply web/pages/group/[...slugs]/index.tsx:285]: You might need to use nextjs catchall routes: https://nextjs.org/docs/routing/dynamic-routes#optional-catch-all-routes

Comment (jahooma) [reply web/pages/group/[...slugs]/index.tsx:285]: Actually, it might be better to make each a separate page as a new file under `pages/group/`.

E.g. `pages/group/my-group/markets`. Would be much better organization

jahooma (approved):
Sweet! This is really cool, Fede! Glad to see this groups overhaul come together quickly.

I tend to favor the approach of shipping it and seeing how it goes, with follow ups as needed. With that in mind, I approve this PR.

**Thoughts**
- It seems to me that on desktop the name of the group is not that prominent, so that someone wouldn't necessarily realize this is a sub-Manifold site.
One idea I had was to move the group name below the logo, which makes it more prominent. See image below.
- I don't like different indentation levels for sidebar items.
- Would be great if clicking notifications kept you within the group. Might be encoded as `/notifications?g=group-slug`. Not sure though.
- I'm not sure about the mobile navigation, seems like something to iterate on.
- I do like the sticky group name and back arrow on mobile. Great solution.

mockup groups

FRCassarino:
I addressed all the feedback items, except for the following ones which are all related:

> This should probably be represented in the URL, I think. Also could allow you to go back and forth.
> 
> A long time ago, you could go to `/group/JamesGroup/leaderboard` and it would show the leaderboard, etc.
> 
> Actually, now that these are not tabs, I think it's even more expected that the distinct pages of markets, about, leaderboard are different URLs. You can either change this PR, or follow up with the change.

> Actually, it might be better to make each a separate page as a new file under `pages/group/`.
> 
> E.g. `pages/group/my-group/markets`. Would be much better organization


> I'm not sure if this kind of link works in all cases. For example, if someone middle-clicks the button, would it open the link in a new tab? Or, if you tab to focus-select it and hit enter, does it navigate (and also trigger the track function)?
> 
> You might have to use the nextjs `<Link>` wrapper and set `href`, or use `SiteLink`.

The same changes should address all of those, but I'll do it in a separate PR. In the meantime I'm merging this version.

PR #875, opened 2022-09-13
Adjust market modal styles

sipec:
- Refactor so that add markets on groups page and add market embeds have similar style
- Fix bug where search bar doesn't stick


sipec:
tested and seems to work fine

PR #878, opened 2022-09-14
Precalculate and store positions for commentors

mqp:
Now we don't have to load all bets on a contract to show comments on non-FR markets. At the moment, we still have to load the current user's bets to figure out which bet a comment might be associated with, which is basically OK for now.

We are still in fact loading the bets for other reasons, but this is one step closer to a post-loading-all-bets world.


gappleto97:
Here's a market for this: https://manifold.markets/MarketManagerBot/how-many-days-after-september-1st-w-9968a4d243f2

PR #879, opened 2022-09-14
Inga/bettingfix

ingawei:
Cleaning up betting UI

### DIFF
--- a/web/components/contract/contract-details.tsx
+++ b/web/components/contract/contract-details.tsx
@@ -106,6 +106,11 @@ export function AvatarDetails(props: {
   )
 }
 
+export function getIsMobile() {
+  const { width } = useWindowSize()
+  return (width ?? 0) < 600
+}
+
 export function ContractDetails(props: {
   contract: Contract
   disabled?: boolean
@@ -123,7 +128,7 @@ export function ContractDetails(props: {
   const user = useUser()
   const isCreator = user?.id === creatorId
   const { width } = useWindowSize()
-  const isMobile = (width ?? 0) < 600
+  const isMobile = getIsMobile()
 
   return (
     <Col>
@@ -136,7 +141,7 @@ export function ContractDetails(props: {
           className="mr-1.5"
         />
         {!disabled && (
-          <div className="absolute mt-2 ml-[11px]">
+          <div className="absolute mt-3 ml-[11px]">
             <MiniUserFollowButton userId={creatorId} />
           </div>
         )}
@@ -234,7 +239,7 @@ export function MarketGroups(props: {
           isMobile ? 'max-w-[140px]' : 'max-w-[250px]'
         )}
       >
-        <div className="bg-greyscale-6 hover:bg-greyscale-4 text-2xs items-center truncate rounded-full px-2 text-white sm:text-xs">
+        <div className="bg-greyscale-4 hover:bg-greyscale-3 text-2xs items-center truncate rounded-full px-2 text-white sm:text-xs">
           {groupToDisplay.name}
         </div>
       </a>
@@ -248,7 +253,7 @@ export function MarketGroups(props: {
     >
       <div
         className={clsx(
-          'bg-greyscale-6 text-2xs items-center truncate rounded-full px-2 text-white sm:text-xs'
+          'bg-greyscale-4 text-2xs items-center truncate rounded-full px-2 text-white sm:text-xs'
         )}
       >
         No Group
@@ -268,7 +273,7 @@ export function MarketGroups(props: {
             {groupInfo}
             {user && (
               <button
-                className="text-greyscale-6 hover:text-greyscale-4"
+                className="text-greyscale-4 hover:text-greyscale-3"
                 onClick={() => setOpen(!open)}
               >
                 <PlusCircleIcon className="mb-0.5 mr-0.5 inline h-4 w-4 shrink-0" />
--- a/web/components/contract/contract-details.tsx
+++ b/web/components/contract/contract-details.tsx
@@ -122,139 +122,171 @@ export function ContractDetails(props: {
   const { volumeLabel, resolvedDate } = contractMetrics(contract)
   const user = useUser()
   const isCreator = user?.id === creatorId
-  const [open, setOpen] = useState(false)
   const { width } = useWindowSize()
   const isMobile = (width ?? 0) < 600
+
+  return (
+    <Col>
+      <Row>
+        <Avatar
+          username={creatorUsername}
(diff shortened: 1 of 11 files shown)


web/components/contract/contract-details.tsx
```diff
@@ -120,74 +127,157 @@ export function ContractDetails(props: {
...
+        </div>
+      )}
+      {/* {user && (
+        <>
+          <Row className="hidden items-center gap-1 md:inline-flex">
+            <DatabaseIcon className="h-5 w-5" />
+            <div className="whitespace-nowrap">{volumeLabel}</div>
+          </Row>
+        </>
+      )} */}
```
jahooma, line 220: nit: it's almost always not a great idea to commit commented out code.
Better to just remove. Someone could always dig it up in the git history.

web/components/contract/contract-details.tsx
```diff
@@ -120,74 +127,157 @@ export function ContractDetails(props: {
...
+                {!resolvedDate && closeTime && user && (
+                  <Row>
+                    <div>Closes&nbsp;</div>
+                    <EditableCloseDate
+                      closeTime={closeTime}
+                      contract={contract}
+                      isCreator={isCreator ?? false}
+                    />
+                  </Row>
+                )}
```
jahooma, line 187: Decent amount of nesting here, and this is logically related. A good candidate for extracting out into another component! E.g. `CloseOrResolveTime`

web/components/contract/contract-details.tsx
```diff
@@ -120,74 +127,157 @@ export function ContractDetails(props: {
...
-                size={'xs'}
-                color={'gray-white'}
+            {user && (
+              <button
+                className="text-greyscale-4 hover:text-greyscale-3"
                 onClick={() => setOpen(!open)}
               >
-                <PencilIcon className="mb-0.5 mr-0.5 inline h-4 w-4 shrink-0" />
-              </Button>
+                <PlusCircleIcon className="mb-0.5 mr-0.5 inline h-4 w-4 shrink-0" />
```
jahooma, line 279: Thanks, this icon is wayyy better than the pencil here.

web/components/contract/contract-details.tsx
```diff
@@ -280,12 +332,12 @@ export function ExtraMobileContractDetails(props: {
         !resolvedDate &&
         closeTime && (
           <Col className={'items-center text-sm text-gray-500'}>
+            <Row className={'text-gray-400'}>Closes&nbsp;</Row>
```
jahooma, line 335: Ditto, substitute `{' '}` for the non-breaking space.

web/components/follow-button.tsx
```diff
@@ -69,3 +73,36 @@ export function UserFollowButton(props: { userId: string; small?: boolean }) {
     />
   )
 }
+
+export function MiniUserFollowButton(props: { userId: string }) {
+  const { userId } = props
+  const currentUser = useUser()
+  const following = useFollows(currentUser?.id)
+  const isFollowing = following?.includes(userId)
+  const user = useUser()
```
jahooma, line 82: You're using `useUser` twice. I think the second one should be `useUserById(userId)`

jahooma: Or, you might not need it at all.

jahooma (approved):
Nice change, Inga! Getting the market page layout right is really important, and I like the push to toward simplicity, explaining the close date, and the more-compact follow button.

Some notes on the PR/code:
- It's a good idea to include screenshots in the PR description (even if slightly out of date haha). Makes it easier for us all to scrutinize the changes.
- Commenting out code is fine for drafts, but not for code that is pushed. Make sure to remove those before merging.
- Code looks really nice in general!

On the design:
- Like Stephen was saying in standup, I don't think we want to hide the like and watch buttons within the '...' menu. Seems like they would fit fine right next to the share button in the header?
- Did you mean for there to be a second share button below the graph?
- What do you think about showing more groups, say, up to 3?
- The follow button for creators is cool, but I'm sure how clear it is. What if you get one of those toast notifications (popping down from the top) that says you have followed them when clicked? For an example, see `like-market-button.tsx`.
- Is there no way to unfollow the creator from this page? Maybe that's fine?

web/components/contract/contract-details.tsx
```diff
@@ -120,74 +127,157 @@ export function ContractDetails(props: {
...
+                        <div>resolved&nbsp;</div>
+                        <b>{resolvedDate}</b>
+                      </Row>
+                    </DateTimeTooltip>
+                  </>
+                ) : null}
+
+                {!resolvedDate && closeTime && user && (
+                  <Row>
+                    <div>Closes&nbsp;</div>
```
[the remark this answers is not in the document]
ingawei: For some reason it just replaces it with a regular space and then doesn't actually translate? Also sorry this code is such a mess atm, I assumed no one would look at it until I assigned it to someone, lmk if I just shouldn't create the request in the future until it's review ready

ingawei:
> * Like Stephen was saying in standup, I don't think we want to hide the like and watch buttons within the '...' menu. Seems like they would fit fine right next to the share button in the header?

Yeah maybe I can put it in web view, but it's definitely too cluttered in mobile

web/components/contract/contract-details.tsx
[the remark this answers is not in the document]
jahooma: Np. Sorry for barging in with a review! You could always mark it as a draft PR. Though I tend to think it's good to get feedback as soon as possible.

Hmm, maybe just add some margin right then instead haha.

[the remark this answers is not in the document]
ingawei: Oh no I super appreciate the feedback! Your comments are always really helpful I'm just a little embarrassed at the state of this PR at the moment loll. Ehh is that less bad? The text size changes from web to mobile, I feel like conditional margins on this is kind of gross

web/components/contract/contract-details.tsx
```diff
@@ -104,6 +106,11 @@ export function AvatarDetails(props: {
   )
 }
 
+export function getIsMobile() {
```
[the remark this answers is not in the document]
ingawei: Done

web/components/contract/contract-details.tsx
ingawei: Done

web/components/contract/contract-details.tsx
ingawei: Done

web/components/contract/contract-details.tsx
ingawei: Thank you! :)

PR #881, opened 2022-09-15
Paginate contract bets tab

mqp:
This is just "placebo pagination" -- for now it's still loading all the bets. However, rendering bets is pretty slow, relatively speaking, because of the links and tooltips. So this is actually a nice speedup to loading and unloading on contract pages (i.e. when you press the back button and React has to clean up all the old crap) and we might as well just do it instead of waiting for "real pagination" to land.


jahooma (approved):
LGTM! Should help reduce render latency a bit for big contracts. Thanks!

PR #890, opened 2022-09-16
Adjust market header style

sipec:
behold, alignment:
image


ingawei (approved):
Thanks for getting rid of the redundant mobile stuff!

PR #898, opened 2022-09-20
Don't always require tips to render comments

mqp:
By not rendering a tip widget in the "resolved contract top trades" box, we only need to fetch the tips for rendering the tab area, not at the top level of the contract, so we can render more sooner and re-render less later.


jahooma (approved):
LGTM! Seems like a good idea

PR #899, opened 2022-09-20
Move comment-bet association code into comment creation trigger

mqp:
This is attractive because, as you can see by looking at the diff, it means that we don't have to thread down a ton of context about recent bets and recent comments into the comment posting UI, so it substantially simplifies the comment posting UI code and opens the door to further efficiency improvements.

The only functional difference between this code and the previous code is that the previous code would actually create comment documents with the same Firestore ID as their associated bet document (i.e. a comment associated with `contracts/ABC/bets/DEF` would go to document `contracts/ABC/comments/DEF`). This is cute but as far as I know we never made use of this property, and I am not sure why we would use it in the future instead of just querying by ID.

### DIFF
--- a/firestore.indexes.json
+++ b/firestore.indexes.json
@@ -102,6 +102,20 @@
     },
     {
       "collectionGroup": "comments",
+      "queryScope": "COLLECTION",
+      "fields": [
+        {
+          "fieldPath": "userId",
+          "order": "ASCENDING"
+        },
+        {
+          "fieldPath": "createdTime",
+          "order": "ASCENDING"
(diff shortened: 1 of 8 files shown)


mqp:
@mantikoros because recently he changed the UI to not show the bets here, so he may be interested in this.

mqp:
This is kind of a blocker on doing further contract page improvements, so merging it.

jahooma:
Cool, LGTM!

Sorry, lots of meetings today, will take a look at your other PRs.

PR #900, opened 2022-09-20
Kill counts of comments and trades on contract page

mqp:
It's pretty bad to have to load all comments and bets to get these counts; it means we have to load everything without pagination, and means we can't render the tab control until the counts are in. (As opposed to if we leave the counts out, we can render the comments tab as soon as comments are in, and the bets tab as soon as the bets are in.) I plan on making changes shortly to contract page loading that are incompatible with having these counts.

If we really wanted these numbers, we could make some complicated triggers that track them, but I don't think they are important, so that seems like more work than it's worth.

### DIFF
--- a/web/components/contract/contract-tabs.tsx
+++ b/web/components/contract/contract-tabs.tsx
@@ -47,9 +47,7 @@ export function ContractTabs(props: {
       l.amount > 0
   )
 
-  // Load comments here, so the badge count will be correct
-  const updatedComments = useComments(contract.id)
-  const comments = updatedComments ?? props.comments
+  const comments = useComments(contract.id) ?? props.comments
 
   const betActivity = lps != null && (
     <ContractBetsActivity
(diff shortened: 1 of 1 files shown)


jahooma (approved):
I actually think these bet and comment counts are fairly important. We should cache them on the contract and show them.

We were even thinking of putting them on the market card so users can judge activity, like how Reddit and Twitter show these numbers.

mqp:
> I actually think these bet and comment counts are fairly important. We should cache them on the contract and show them.
> 
> We were even thinking of putting them on the market card so users can judge activity, like how Reddit and Twitter show these numbers.

If we want these numbers, then in order to scale to ~hundreds of bets or comments without things sucking, we are pretty much either going to have to do it in some kind of job like `updateMetrics` and be OK with it being out of date, or we are going to have to implement a kind of fancy Firestore idempotent counter guy. Which might not be that hard, so probably we should do it.

jahooma:
> > I actually think these bet and comment counts are fairly important. We should cache them on the contract and show them.
> > We were even thinking of putting them on the market card so users can judge activity, like how Reddit and Twitter show these numbers.
> 
> If we want these numbers, then in order to scale to ~hundreds of bets or comments without things sucking, we are pretty much either going to have to do it in some kind of job like `updateMetrics` and be OK with it being out of date, or we are going to have to implement a kind of fancy Firestore idempotent counter guy. Which might not be that hard, so probably we should do it.
> 
> (Or we could bite the bullet and migrate away from Firestore, or implement a cache in between Firestore and the app which has a real data structure that we can query for things in.)

Haha, yep there's a lot of ways to do it. I think we should keep it simple for now. Here's how the `followerCount` is updated in `on-follow-user.ts`
```
    await firestore.doc(`users/${followedUserId}`).update({
      followerCountCached: FieldValue.increment(1),
    })
```

It might not be idempotent, but I think it works.

PR #902, opened 2022-09-20
Clean up some stuff in `AnswersPanel`

mqp:
Just fixing things I noticed when I was reading this code.

### DIFF
--- a/web/components/answers/answers-panel.tsx
+++ b/web/components/answers/answers-panel.tsx
@@ -57,7 +57,7 @@ export function AnswersPanel(props: {
 
   const answerItems = sortBy(
     losingAnswers.length > 0 ? losingAnswers : sortedAnswers,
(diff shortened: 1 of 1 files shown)


web/components/answers/answers-panel.tsx
```diff
@@ -56,6 +55,11 @@ export function AnswersPanel(props: {
     ),
   ]
 
+  const answerItems = sortBy(
+    losingAnswers.length > 0 ? losingAnswers : sortedAnswers,
+    (answer) => -getOutcomeProbability(contract, answer.number.toString())
```
jahooma, line 60: I think you can do `answer.id` instead of `answer.number.toString()`

jahooma (approved):
Thanks for the cleanup! Much better than it was.

PR #908, opened 2022-09-20
Make comment position calculation more robust

mqp:
This fixes three issues with this code:

1. If someone had both YES and NO shares, then it now nets them out instead of only 'seeing' the larger of the two. (This probably shouldn't happen, but there's no reason for this code to break if it does.)
2. If someone has some number of YES shares and `0 - epsilon` NO shares due to floating point error, which is totally possible, the code now works correctly. Previously it would end up always looking at the NO shares, which would get floored to -1, and end up not displaying a position in the UI.
3. The binary branch is now consistent with the free response branch in that it correctly records in the DB the exact number of shares and leaves it up to the client to do rounding or hiding as it chooses, which makes more sense.

### DIFF
--- a/common/calculate.ts
+++ b/common/calculate.ts
@@ -257,11 +257,6 @@ export function getTopAnswer(
 }
 
 export function getLargestPosition(contract: Contract, userBets: Bet[]) {
-  let yesFloorShares = 0,
-    yesShares = 0,
-    noShares = 0,
-    noFloorShares = 0
-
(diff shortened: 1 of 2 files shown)


jahooma (approved):
Makes sense! LGTM

mqp:
Stuff changed here, IDK if the problems I said are still at large (probably) but I am not going to merge this right now.

PR #911, opened 2022-09-21
Inga/mobilebetting

ingawei:
Functionality for mobile betting:
KNOWN BUG: embed betting M$ is off
Screen Shot 2022-09-20 at 9 54 42 PM
Screen Shot 2022-09-20 at 9 54 56 PM
Screen Shot 2022-09-20 at 9 55 06 PM
Screen Shot 2022-09-20 at 9 55 14 PM
Screen Shot 2022-09-20 at 9 55 35 PM


jahooma:
This looks really good! I'll review the code changes tomorrow. Thanks, Inga!

web/components/amount-input.tsx
```diff
@@ -34,46 +35,57 @@ export function AmountInput(props: {
...
-          onChange={(e) => onAmountChange(e.target.value)}
-        />
-      </label>
+    <>
+      <Col className={className}>
+        <label>
+          <span
+            className={clsx(
+              'text-greyscale-4 absolute ml-2 mt-[9px]',
+              isMobile ? 'font-sm' : 'font-lg'
```
jahooma, line 48: In general, it's better to use css media queries to adjust based on screen size, e.g.  replace this line with`font-sm sm:font-lg`

It's faster, and it can be correct when server-side rendered even when we don't know the screen size yet, which could prevent a flicker.

web/components/amount-input.tsx
```diff
@@ -34,46 +35,57 @@ export function AmountInput(props: {
...
+            )}
+          >
+            {label}
+          </span>
+          <input
+            className={clsx(
+              'placeholder:text-greyscale-4 border-greyscale-2 rounded-md pl-9',
+              error && 'input-error',
+              isMobile ? 'font-sm w-24' : 'font-lg',
```
jahooma, line 57: Ditto!

web/components/bet-button.tsx
```diff
@@ -64,11 +71,59 @@ export default function BetButton(props: {
...
+    return (
+      <>
+        <SignedInBinaryMobileBetting contract={contract} user={user} />
+      </>
+    )
+  } else {
+    return (
+      <>
+        <BetSignUpPrompt className="w-full" />
+      </>
```
jahooma, line 95: No need to wrap these returns in `<></>`

web/components/bet-button.tsx
```diff
@@ -64,11 +71,59 @@ export default function BetButton(props: {
...
+  const { contract, user } = props
+  const unfilledBets = useUnfilledBets(contract.id) ?? []
+
+  return (
+    <>
+      <Col className="w-full gap-2 px-1">
+        <Col>
+          <BuyPanel
+            hidden={false}
+            contract={contract as CPMMBinaryContract}
```
jahooma, line 113: You might want to type the contract field in props with `CPMMBinaryContract` so that no casting is required (and the props are accurate).

web/components/bet-panel.tsx
```diff
@@ -162,38 +165,53 @@ export function SimpleBetPanel(props: {
...
 
   const initialProb = getProbability(contract)
   const isPseudoNumeric = contract.outcomeType === 'PSEUDO_NUMERIC'
 
-  const [outcome, setOutcome] = useState<'YES' | 'NO' | undefined>(selected)
-  const [betAmount, setBetAmount] = useState<number | undefined>(undefined)
+  const windowSize = useWindowSize()
+  const initialOutcome =
+    windowSize.width && windowSize.width >= 1280 ? 'YES' : undefined
+  console.log('initialOutcome>', initialOutcome)
```
jahooma, line 185: console.log! haha

web/components/bet-panel.tsx
```diff
@@ -214,9 +232,13 @@ function BuyPanel(props: {
       .then((r) => {
         console.log('placed bet. Result:', r)
         setIsSubmitting(false)
-        setWasSubmitted(true)
         setBetAmount(undefined)
         if (onBuySuccess) onBuySuccess()
+        else {
+          toast('Trade submitted!', {
+            icon: <CheckIcon className={'text-primary h-5 w-5'} />,
+          })
```
jahooma, line 240: Nice touch!

web/components/bet-panel.tsx
```diff
@@ -281,92 +304,141 @@ function BuyPanel(props: {
...
+        <>
+          <YesNoSelector
+            className="mb-4"
+            btnClassName="flex-1"
+            selected={outcome}
+            onSelect={(choice) => mobileOnBetChoice(choice)}
+            isPseudoNumeric={isPseudoNumeric}
+          />
+        </>
+      )}
```
jahooma, line 328: Seems like these could be combined? You just need to change the `onSelect` prop based on `mobileView`

web/components/warning-confirmation-button.tsx
```diff
@@ -45,7 +53,7 @@ export function WarningConfirmationButton(props: {
           openModalButtonClass,
           isSubmitting && 'btn-disabled loading'
         ),
-        label: 'Submit',
+        label: amount ? `Wager ${formatMoney(amount)}` : 'Wager',
```
jahooma, line 56: Not a super fan of the word Wager, but we can think about alternatives.

jahooma (approved):
Very nice, Inga! This is a great improvement to the core of our app.

I've just left a few comments looking at only the code. @akrolsmir can feel free to add his thoughts.

jahooma:
Just checked out the branch to see how it looks, and I think it's great! Thanks Inga!

We should definitely ship this as soon as possible because it's a big improvement. We can iterate based on further feedback from Austin and/or the community as that feedback comes in.

Nitpicks that I have:
- The YES and NO buttons seem a little bit too close to the graph, maybe add a slight bit more margin? 
- I'm also not sure how clearly the YES/NO buttons are introduced, like maybe you'd want a label for them? I think users will understand once they click them, but not so sure otherwise.
- The QTY label for shares in the Sell dialog is too close to the number:
Screen Shot 2022-09-26 at 3 59 47 PM

We could also consider restyling the desktop bet panel to match this design, as a followup. Personally, I think we could even just use this exact design with YES and NO buttons on the center panel, and remove the whole third column on desktop, which is intimidating for causal users. I always thought having three columns didn't fit Manifold's vision of being accessible.

PR #912, opened 2022-09-21
More work on contract page and tabs

mqp:
Further cleanup. Now the comments are only required by the comments tab and the "top comment" thing on resolved markets (which is only on resolved markets, so we don't care about it as much.) As a result, we can comfortably remove the comments from the `getStaticProps` payload and load them async in the page.

Also making a dent in re-renderings.

On my laptop, this PR pretty much moves the needle on loading the biggest market pages (e.g. ACX discord market, elon musk twitter market) from "basically unusable" to "kind of OK" so that's cool.

### DIFF
--- a/web/pages/[username]/[contractSlug].tsx
+++ b/web/pages/[username]/[contractSlug].tsx
@@ -54,7 +54,7 @@ export const getStaticProps = fromPropz(getStaticPropz)
 export async function getStaticPropz(props: {
   params: { username: string; contractSlug: string }
 }) {
-  const { username, contractSlug } = props.params
+  const { contractSlug } = props.params
   const contract = (await getContractFromSlug(contractSlug)) || null
   const contractId = contract?.id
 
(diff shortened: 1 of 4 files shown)


mqp:
@jahooma because I know you have been trying to figure out how to keep these static props from showing stale data.

web/pages/[username]/[contractSlug].tsx
```diff
@@ -54,25 +51,14 @@ export const getStaticProps = fromPropz(getStaticPropz)
 export async function getStaticPropz(props: {
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:51]: Aside: I think we should kill off this 'getStaticPropz' pattern. We aren't running any private instances, I don't think we should be running any for the foreseeable future, and if we were, it would be better to use server side auth to keep the data private.

jahooma (approved):
Sweet! This is great. I like how we no longer need to thread comments all the way through, makes things simpler. Should reduce initial render time too if comments aren't rendered immediately.
Thanks, Marshall!

web/pages/[username]/[contractSlug].tsx
Comment (mqp) [reply web/pages/[username]/[contractSlug].tsx:51]: Talk to Austin about it, he thinks it's still useful.

PR #913, opened 2022-09-21
Fix group styles

sipec:
Before:
image

After:
image

Also, before the nav was hidden on some screen sizes.

### DIFF
--- a/web/components/nav/group-nav-bar.tsx
+++ b/web/components/nav/group-nav-bar.tsx
@@ -11,6 +11,7 @@ import { useUser } from 'web/hooks/use-user'
 import NotificationsIcon from '../notifications-icon'
 import router from 'next/router'
 import { userProfileItem } from './bottom-nav-bar'
+import Link from 'next/link'
 
 const mobileGroupNavigation = [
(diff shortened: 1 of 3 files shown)


web/components/nav/group-nav-bar.tsx
```diff
@@ -72,23 +76,23 @@ function NavBarItem(props: {
...
   return (
-    <button onClick={() => props.onClick(item.key ?? '#')}>
+    <Link href={item.href ?? '#'}>
       <a
         className={clsx(
           'block w-full py-1 px-3 text-center hover:bg-indigo-200 hover:text-indigo-700',
           currentPage === item.key && 'bg-gray-200 text-indigo-700'
         )}
-        onClick={track}
+        onClick={() => (onClick?.(item.key ?? '#'), track())}
```
sipec, line 91: imo using an onClick handler that calls Next router is an anti-pattern, and instead you should just put the correct href (like `groupSlug/market`)

jahooma:
I think this can be closed, since we switched back to tabs in groups.

PR #914, opened 2022-09-21
Clean up and fix stuff on answers panel

mqp:
This is more straightforward and fixes at least three small bugs:

1. It handles the case where the winning answer has no bets (weird but possible.)
2. It correctly prints "No answers yet" iff there are no answers (it previously also printed this if there was just one answer, which makes no sense.)
3. It correctly doesn't print the winning answer twice when there is exactly one answer which is winning and no other losing answers.

See https://manifold.markets/AnselFreniere/what-is-the-best-data-source-for-th for an example of a market that demonstrates both bugs 2 and 3.


jahooma (approved):
Great! This does look quite a bit simpler and fixes some bugs. Thanks Marshall!

PR #915, opened 2022-09-21
Fix up comment permalink stuff

mqp:
1. There's no reason to wait for a `useEffect` before looking at the router path, so we don't need to store `highlighted` in state.
2. The permalink scrolling broke when I started loading the comments on the client, so I needed to add additional code to scroll to highlighted comments when those comments actually first render. (The answer items are available immediately, but they are in the comments tab, which doesn't load until the comments are ready, so we need to do the same thing for them.)

### DIFF
--- a/web/components/feed/feed-answer-comment-group.tsx
+++ b/web/components/feed/feed-answer-comment-group.tsx
@@ -1,7 +1,7 @@
 import { Answer } from 'common/answer'
 import { FreeResponseContract } from 'common/contract'
 import { ContractComment } from 'common/comment'
-import React, { useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import { Col } from 'web/components/layout/col'
 import { Row } from 'web/components/layout/row'
 import { Avatar } from 'web/components/avatar'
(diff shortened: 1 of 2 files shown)


jahooma (approved):
Cool! Thanks for fixing this, I noticed it too.

I'm actually a little concerned that the experience for most contracts is a bit worse because the comments take a while to load.

What do you think about sending a subset of the comments through the static props, say, the last 20, so that for most markets the page appears fully loaded immediately?

I'm actually a fan of using getStaticProps as much as possible, because it's like the fastest possible way to retrieve data: cached from a CDN.

mqp:
> Cool! Thanks for fixing this, I noticed it too.
> 
> I'm actually a little concerned that the experience for most contracts is a bit worse because the comments take a while to load.
> 
> What do you think about sending a subset of the comments through the static props, say, the last 20, so that for most markets the page appears fully loaded immediately?
> 
> I'm actually a fan of using getStaticProps as much as possible, because it's like the fastest possible way to retrieve data: cached from a CDN.

Right now, I'm a fan of using `getStaticProps` as little as possible. The main reason is that because we always want live updates on everything, and the way we implement that is by re-fetching everything immediately on the client and also establishing a subscription with it. So using `getStaticProps` basically multiplies the total amount of work done to render the information by two, spreading that new work out over time. It also introduces a bunch of cache invalidation questions, like what you have been dealing with.

Regarding this specific issue, I notice that the comments are very slow to load (i.e. longer than a second) even when the number of comments being loaded is tiny. So there's something going wrong, that is not about it not being in static props. If we debug that and you're still unhappy with the loading experience I would be open to thinking about further work. But I don't think it would make sense to make a bunch of tough tradeoffs by using `getStaticProps` when we can just fix problems.

PR #918, opened 2022-09-21
Improve contract leaderboard computation

mqp:
Since bets now have the denormalized information that the leaderboard needs to render entries (name, username, avatar), we no longer need to load any additional data other than the bets in order to render the contract leaderboard.

Also fixed a small bug where the "top trades" logic was slightly broken in my new world where bet IDs and their associated comment IDs aren't always the same.


jahooma (approved):
Cool, thanks for the cleanup! LGTM

PR #920, opened 2022-09-21
Switch the user profile link in sidebar to go to stats tab

mqp:
It seems to me that this is likely the best tab, because the profit graph is likely to be the most interesting info.


jahooma:
Hmm, I don't think I agree. I believe most people go back to trades to check up on markets they already bet on.

mqp:
Since there's no consensus on this and no way to get consensus I guess we aren't changing it.

PR #922, opened 2022-09-22
Next.js 12.2.5 -> 12.3.1

mqp:
I don't know of anything important here in particular, but it's nice that they stabilized the image component we use, and it might be nice to try out the SWC minifier and see if it's better than the status quo.

https://nextjs.org/blog/next-12-3
https://github.com/vercel/next.js/releases/tag/v12.3.1


jahooma (approved):
LGTM! Good to be up to date with Nextjs

PR #926, opened 2022-09-22
Add a "Posts" tab to groups

FRCassarino:
Adds a Post section.

image
image


FRCassarino:
Think of this as the simplest way to give groups access to posts. What It'll end up actually looking like will be a "Frontpage" tab with a latest posts section, rather than a separate tab.

web/components/create-post.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+import { TextEditor, useTextEditor } from 'web/components/editor'
+import { createPost } from 'web/lib/firebase/api'
+import clsx from 'clsx'
+import Router from 'next/router'
+import { MAX_POST_TITLE_LENGTH } from 'common/post'
+import { postPath } from 'web/lib/firebase/posts'
+import { Group } from 'common/group'
+
+export default function CreatePost(props: { group?: Group }) {
```
FRCassarino, line 14: No major changes here, I just moved the file from pages to components, and it's showing up as if it was brand new

web/pages/group/[...slugs]/index.tsx
```diff
@@ -137,6 +149,12 @@ export default function GroupPage(props: {
   const group = useGroup(props.group?.id) ?? props.group
   const aboutPost = usePost(props.aboutPost?.id) ?? props.aboutPost
 
+  let groupPosts = usePosts(group?.postIds ?? []) ?? posts
+
+  if (aboutPost != null) {
+    groupPosts = [aboutPost, ...groupPosts]
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:155]: Might want to sort posts by newest

web/pages/group/[...slugs]/index.tsx
```diff
@@ -413,6 +445,81 @@ function GroupLeaderboard(props: {
...
+              <UserLink
+                className="text-neutral"
+                name={user?.name}
+                username={user?.username}
+              />
+              <span className="mx-1">•</span>
+              <span className="text-gray-500">{fromNow(post.createdTime)}</span>
+            </div>
+            <div className="text-lg font-medium text-gray-900">
+              {post.title}
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:514]: Seems like you'd want to show more than just the title. A preview of the body would be great, though difficult.
I think it would be a better alternative to show the whole post than just the title.

jahooma (approved):
Thanks Fede! It will be exciting to see how users use posts!

Austin has mentioned some other ideas for groups in the Monday checklist, including showing all comments, or all trades made for markets in the group.
Would be interesting to find features or use cases desired by people using groups today (e.g. in the Clearer Thinking tournament).

web/pages/group/[...slugs]/index.tsx
Comment (FRCassarino) [reply web/pages/group/[...slugs]/index.tsx:514]: I agree, but I think a preview is the right move (not the whole post, at least not in this context). It's tricky to show a preview though (because of how JSONContent  is setup), so I'm leaving it for a separate PR.

PR #927, opened 2022-09-23
Make DB text serialization consistent and efficient

mqp:
We have three kinds of rich text:

- Contract descriptions.
- Comment text.
- Post contents.

Previously, these were stored in two different ways:

- As plaintext strings.
- As structured ProseMirror JSON in Firestore.

We want to make all of these into strings containing serialized ProseMirror JSON. (We make the plaintext strings into ProseMirror JSON that just has one unadorned paragraph.) This is better than storing them as JSON objects in Firestore, because if you store them as JSON objects, the Firestore SDK does an incredible job puffing them up into [giant blobs](https://pastebin.com/LdgztEdF) over the wire with a ton of extra properties, inflating them by ~5-10x in size. So contracts for example get really bloated to fetch. Since we have no reason to ever want to query their internals, it's easy to just store them as strings and call it a day.

### DIFF
--- a/common/comment.ts
+++ b/common/comment.ts
@@ -1,5 +1,3 @@
-import type { JSONContent } from '@tiptap/core'
-
 export type AnyCommentType = OnContract | OnGroup | OnPost
 
 // Currently, comments are created after the bet, not atomically with the bet.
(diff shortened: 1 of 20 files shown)


jahooma:
This looks reasonable to me. Thanks for investigating, Marshall!

mqp:
This would still be a good idea in principle, but it's not worth the effort for me to actually do since in the Supabase world we won't care about this anymore.

PR #930, opened 2022-09-23
Move react-masonry-css dependency to web package.json

mqp:
This was just in the wrong one for no particular reason.


jahooma (approved):
Always a bit suspicious that it was working before haha

mqp:
> Always a bit suspicious that it was working before haha

The actual thing that makes most stuff work is the existence of packages in the root `node_modules`, so it's just random tools -- or us -- that might get confused if stuff is in the wrong `package.json`.

PR #932, opened 2022-09-24
basic sprig integration with possible page URL events

bgila:
i've given Austin, James, and Stephen access. Next up I will suggest some surveys

### DIFF
--- a/common/envs/prod.ts
+++ b/common/envs/prod.ts
@@ -3,6 +3,7 @@ export type EnvConfig = {
   firebaseConfig: FirebaseConfig
   amplitudeApiKey?: string
   twitchBotEndpoint?: string
+  sprigEnvironmentId: string
 
   // IDs for v2 cloud functions -- find these by deploying a cloud function and
   // examining the URL, https://[name]-[cloudRunId]-[cloudRunRegion].a.run.app
@@ -56,6 +57,7 @@ type FirebaseConfig = {
 export const PROD_CONFIG: EnvConfig = {
   domain: 'manifold.markets',
   amplitudeApiKey: '2d6509fd4185ebb8be29709842752a15',
+  sprigEnvironmentId: 'sQcrq9TDqkib',
 
   firebaseConfig: {
     apiKey: 'AIzaSyDp3J57vLeAZCzxLD-vcPaGIkAmBoGOSYw',
--- a/web/lib/service/analytics.ts
+++ b/web/lib/service/analytics.ts
@@ -12,6 +12,29 @@ init(ENV_CONFIG.amplitudeApiKey ?? '', undefined, { includeReferrer: true })
 
 export { track }
 
+// Integrate Sprig
+
+try {
+  (function(l,e,a,p) {
+    // @ts-expect-error
+    if (window.Sprig) return
+    // @ts-expect-error
+    window.Sprig = function(){S._queue.push(arguments)}
+    // @ts-expect-error
+    var S = window.Sprig;S.appId = a;S._queue = [];window.UserLeap=S
+    // @ts-expect-error
+    a=l.createElement('script')
+    // @ts-expect-error
+    a.async=1;a.src=e+'?id='+S.appId
+    // @ts-expect-error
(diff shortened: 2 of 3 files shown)


common/envs/prod.ts
```diff
@@ -3,6 +3,7 @@ export type EnvConfig = {
   firebaseConfig: FirebaseConfig
   amplitudeApiKey?: string
   twitchBotEndpoint?: string
+  sprigEnvironmentId: string
```
akrolsmir, line 6: Let's make this optional so other configs (dev, or custom instances) won't need it!

web/lib/service/analytics.ts
```diff
@@ -12,6 +12,29 @@ init(ENV_CONFIG.amplitudeApiKey ?? '', undefined, { includeReferrer: true })
 
 export { track }
 
+// Integrate Sprig
+
+try {
+  (function(l,e,a,p) {
+    // @ts-expect-error
```
akrolsmir, line 19: image

Currently getting these errors. Could fix, but I think installing the [Sprig NPM package](https://docs.sprig.com/docs/web-javascript#install-with-npmyarn-beta) might just be better

akrolsmir (commented):
Thanks for sending this out - excited to try out Sprig!

common/envs/prod.ts
bgila: will do. But wait -- we def want this on custom instances, is there same place i can set the config so that it takes effect there too?

bgila:
ready for another review, if you click the link you will see the sample survey on dev

web/lib/service/analytics.ts
```diff
@@ -12,6 +12,37 @@ init(ENV_CONFIG.amplitudeApiKey ?? '', undefined, { includeReferrer: true })
 
 export { track }
 
+// Integrate Sprig
```
jahooma, line 15: I'd put this in a new file under `web/lib/service` called `sprig.ts`. Then, import it here for effects, e.g. `import './sprig'`

jahooma: You can disable typescript for the whole file with `// @ts-nocheck` at the top instead of doing it per line.

web/lib/service/analytics.ts
```diff
@@ -33,10 +64,14 @@ export const withTracking =
 
 export async function identifyUser(userId: string) {
   setUserId(userId)
+  // @ts-expect-error Sprig doesn't yet have a native typescript snippet
+  window.Sprig.setUserId(userId)
```
jahooma, line 68: If you've moved sprig to its own file, you could export the Sprig variable and use it here.
E.g. at the bottom of `/sprig.ts`: `export const Sprig = window.Sprig as any`
Then, this line would be simply: `Sprig.setUserId(userId)`

jahooma (commented):
Sweet, thanks Barak for the Sprig integration! Just chiming in with a few thoughts.

Saw Austin's comment about using the NPM package and agree that would be better if you want to do that instead of following my comments.

One other note: this PR is adding a `package.lock` but it should be deleted since we don't use npm, we use `yarn` which has a `yarn.lock` file. You can type `yarn` into the command line to install packages.

bgila:
> Saw Austin's comment about using the NPM package and agree that would be better if you want to do that instead of following my comments.

The issue with NPM is that it won't automatically update to the latest version. 99% of Sprig customers are on the snippet installation..

Addressing all the comments now

bgila:
alright PTAnotherL @jahooma

jahooma (approved):
Awesome, looks great Barak! Thanks for making those changes.

Looking forward to seeing the results from this integration!

common/envs/prod.ts
jahooma: Custom instances currently are just on different branches. So you'd have to merge this whole PR into each of them.

I don't think I'd worry about it, but we could potentially get it on the Salem Center instance, I guess!

bgila:
feel free to merge whenever! the study is launched so that will start sending the study to eligible users (after they visit 4 more times)

PR #935, opened 2022-09-25
WIP rewrite contract probability graphs

mqp:
I got too fed up with the old ones so I decided to just finally throw out Nivo and rewrite them using d3. Now we will not have to constantly suffer every time we open a page that has a graph with a few hundred points.

TODO: Answers graph multi-line chart, testing, probably some details, add functionality that should be there like time selection.


mqp:
This is approaching its final form (the only missing functionality from the existing charts is tooltips on numeric/choice markets) so anyone who wants to look at it might want to take a look.

mqp:
Now this is feature-complete, fast, not ridiculously badly written, and also has time selection powers, so I will probably merge it soon after testing everything.

mqp:
OK, I don't know if there are any bugs or mistakes people will complain about, but the net is much better than the status quo, so I am merging this right away to avoid blocking anyone else's changes on contract graph kind of stuff.

PR #939, opened 2022-09-27
Expose multiple choice answer probabilities

gappleto97:
I think this is hidden unintentionally,  since the  relevant line is checking for the outcome type rather than for the DPM-2 flag. If you'd like, I could also just change that to check if it's DPM-2

I think I need approval in order to test this


akrolsmir:
Looks great, thank you so much!

PR #940, opened 2022-09-27
add to queue rather than invoking sprig object directly, as it's still being setup

bgila:
didn't read our own docs carefully enough and forgot about this issue

it's causing the initial "identify the user with their id/email" call to fail
Screen Shot 2022-09-27 at 10 10 55 AM

see the "html script tag" variation here https://docs.sprig.com/docs/web-javascript#values


akrolsmir:
Thanks Barak! (Merging without testing)

PR #942, opened 2022-09-27
Add ||spoilers||

sipec:
image


jahooma (approved):
Cool feature! This will make Eliezer happy.

Is this backwards compatible in that adding spoilers won't crash old clients?

Only other comment is that it might be a little hard to figure out how this works, at least until you've seen someone else do it, perhaps.

Oh, and if you turn it into plain text, does it strip out the spoiler words? I know that notifications show the description of new markets.

sipec:
this is not backwards compatible

currently spoilers when turned into text in og will just show their content :(
well actually, the notifications don't show up at all. I want to fix that before merging.

PR #945, opened 2022-09-27
Flag incorrectly resolved markets, warn about unreliable creators

FRCassarino:
image

image

image


FRCassarino:
Some quick upcoming followups:
- Adding an explanation under "Flag this market as incorrectly resolved"
- Adding a banner in the user profile

FRCassarino:
Something to keep in mind is that the thresholds and gral math that decides when we should flag a user is very tentative, and quite lenient atm. I'd like to get a few days/weeks of data in (by people reporting market), and then decide how to tweak it by looking at the data.

common/contract.ts
```diff
@@ -62,6 +62,7 @@ export type Contract<T extends AnyContractType = AnyContractType> = {
   featuredOnHomeRank?: number
   likedByUserIds?: string[]
   likedByUserCount?: number
+  resolutionReports?: string[]
```
mantikoros, line 65: This is a list of usernames? Maybe the name should be more explicit something like flaggedByUsernames

functions/src/update-metrics.ts
```diff
@@ -116,13 +116,33 @@ export async function updateMetricsCore() {
...
+          contract.resolutionReports.length / (contract.uniqueBettorCount ?? 1)
+
+        return contractRatio
+      })
+      .filter((ratio) => ratio > 0)
+    const badResolutions = contractRatios.filter(
+      (ratio) => ratio > BAD_RESOLUTION_THRESHOLD
+    )
+    const newCorrectResolutionPercentage =
+      (userContracts.length - badResolutions.length) / userContracts.length
```
mantikoros, line 137: Divide by 0 error when no userContracts

functions/src/update-metrics.ts
```diff
@@ -116,13 +116,33 @@ export async function updateMetricsCore() {
...
+          return 0
+        }
+        const contractRatio =
+          contract.resolutionReports.length / (contract.uniqueBettorCount ?? 1)
+
+        return contractRatio
+      })
+      .filter((ratio) => ratio > 0)
+    const badResolutions = contractRatios.filter(
+      (ratio) => ratio > BAD_RESOLUTION_THRESHOLD
```
mantikoros, line 134: reasonable heuristic for now

web/components/confirmation-button.tsx
```diff
@@ -91,7 +88,7 @@ export function ResolveConfirmationButton(props: {
     <ConfirmationButton
       openModalBtn={{
         className: clsx(
-          'border-none self-start',
+          'btn border-none self-start',
```
mantikoros, line 91: "btn" is a DaisyUI class. We're trying to avoid using it going forward.

web/components/contract/contract-report-resolution.tsx
```diff
@@ -0,0 +1,57 @@
...
+    return <></>
+  }
+  const userReported = contract.resolutionReports?.includes(user.id)
+
+  const onSubmit = async () => {
+    if (!user) {
+      return true
+    }
+    await updateContract(contract.id, {
+      resolutionReports: [...(contract.resolutionReports || []), user.id],
```
mantikoros, line 23: you can use my convenience function: bulidArray(contract.resolutionReports, user.id)

web/components/contract/contract-card.tsx
```diff
@@ -213,17 +214,20 @@ export function BinaryResolutionOrChance(props: {
...
+              contract={contract}
+              resolution={resolution}
+            />
           </div>
-          <BinaryContractOutcomeLabel
-            contract={contract}
-            resolution={resolution}
-          />
-        </>
+          <ContractReportResolution contract={contract} />
```
mantikoros, line 229: Probably shouldn't show the report flag in the contract card.

mantikoros: We also *should* show the report flag on FR markets.

mantikoros (changes requested):
Nice! I think this feature is almost ready to go.

functions/src/create-user.ts
```diff
@@ -69,6 +69,7 @@ export const createuser = newEndpoint(opts, async (req, auth) => {
     followerCountCached: 0,
     followedCategories: DEFAULT_CATEGORIES,
     shouldShowWelcome: true,
+    correctResolutionPercentageCached: 1,
```
jahooma, line 72: Maybe it should be a fraction instead of a percentage if it's from 0 to 1.
e.g. `correctResolutionFractionCached` or `fractionResolvedCorrectlyCached`.

I also think it would be ok to omit the Cached suffix: `fractionResolvedCorrectly`.

PR #946, opened 2022-09-28
Add /createcomment API endpoint

### LINKED ISSUE #821: Feature Request: Support commenting via the API
This is currently gating my issue https://github.com/gappleto97/ManifoldMarketManager/issues/5, so if I could comment via the API, my bot could reliably post explanations of why something is being resolved the way it is

akrolsmir:
Split up from #1035, which adds StableDiffusion

Fixes #821


web/pages/api/v0/dream.ts
```diff
@@ -0,0 +1,104 @@
...
+import {
+  CORS_ORIGIN_MANIFOLD,
+  CORS_ORIGIN_LOCALHOST,
+} from 'common/envs/constants'
+import { applyCorsHeaders } from 'web/lib/api/cors'
+
+export const config = { api: { bodyParser: true } }
+
+// Highly experimental. Proxy for https://github.com/vpzomtrrfrt/stability-client
+export default async function route(req: NextApiRequest, res: NextApiResponse) {
```
akrolsmir, line 15: @mqp Was wondering if I could get your help with configuring this API -- for some reason, the way I have this coded, it doesn't work when I pass in the header `Content-Type: application/json`, but if I leave it as `Content-Type: plain/text` everything parses correctly....

Am I doing something wrong?

mqp: Well, it's weird that you both have `bodyParser: true`, and you then also call `JSON.parse(req.body)`. Since it seems like you only want to accept JSON, So I would assume that `JSON.parse(req.body)` is going to fail if you call it with `application/json` because it will already be an object instead of a string.

I would just use the `bodyParser` and then complain in the endpoint if the body is something other than an object with the fields you want.

web/pages/api/v0/market/[id]/comment.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const config = { api: { bodyParser: true } }
+
+export default async function route(req: NextApiRequest, res: NextApiResponse) {
+  await applyCorsHeaders(req, res, {
+    origin: [CORS_ORIGIN_MANIFOLD, CORS_ORIGIN_LOCALHOST],
+    methods: 'POST',
+  })
+
+  const { id } = req.query
```
Comment (mqp) [inline web/pages/api/v0/market/[id]/comment.ts:19]: Kind of weird that you put the ID in a URL parameter instead of the path but I don't care at all about this.

web/pages/api/v0/market/[id]/comment.ts
```diff
@@ -0,0 +1,39 @@
...
+  const contractId = id as string
+
+  if (req.body) {
+    req.body.contractId = contractId
+
+    if (req.body.html && !req.body.content) {
+      req.body.content = htmlToRichText(req.body.html)
+    } else if (req.body.markdown && !req.body.content) {
+      req.body.content = htmlToRichText(micromark(req.body.markdown))
+    }
```
Comment (mqp) [inline web/pages/api/v0/market/[id]/comment.ts:29]: I don't like unnecessary divergence between the cloud functions and these fake Vercel endpoints. I would have changed the cloud function to accept these parameters if you think it's a better UX for users (which I agree with.)

web/components/editor/image-modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+  // TODO: Type this correctly?
+  upload: any
+  open: boolean
+  setOpen: (open: boolean) => void
+}) {
+  const { upload, open, setOpen } = props
+  return (
+    <Modal open={open} setOpen={setOpen}>
+      <Col className="gap-2 rounded bg-white p-6">
+        <Tabs
```
mqp, line 26: Why does this `Col` exist instead of applying a `className` to the `Tabs`?

web/components/editor/image-modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+          tabs={[
+            {
+              title: 'Upload file',
+              content: (
+                <Col>
+                  <FileUploadButton
```
mqp, line 32: Why does this `Col` exist?

web/pages/api/v0/dream.ts
```diff
@@ -0,0 +1,104 @@
...
+      // https://firebase.google.com/docs/storage/web/handle-errors
+      rejectPromise(error)
+      unsubscribe()
+    },
+    () => {
+      getDownloadURL(uploadTask.snapshot.ref).then((downloadURL) => {
+        resolvePromise(downloadURL)
+      })
+
+      unsubscribe()
```
mqp, line 99: The [documentation](https://firebase.google.com/docs/reference/js/storage.uploadtask.md#uploadtaskon) isn't super specific but it makes it look like you don't need to unsubscribe after the task is finished/errored.

web/pages/api/v0/dream.ts
```diff
@@ -0,0 +1,104 @@
...
+    contentType: 'image/png',
+  })
+
+  let resolvePromise: (url: string) => void
+  let rejectPromise: (reason?: any) => void
+
+  const promise = new Promise<string>((resolve, reject) => {
+    resolvePromise = resolve
+    rejectPromise = reject
+  })
```
mqp, line 83: This code looks kind of funny, I would have arranged it like this:

```typescript
const promisifiedUploadBytes = (...args: any[]) => {
  return new Promise((resolve, reject) => {
    const task = uploadBytesResumable(...args)
    task.on(
      'state_changed',
      null,
      (e: Error) => reject(e),
      () => getDownloadURL(task.snapshot.ref).then(resolve)
  })
}

// ...
return promisifiedUploadBytes(storageRef, buffer, {
  cacheControl: `public, max-age=${ONE_YEAR_SECS}`,
  contentType: 'image/png',
})
```

web/components/editor/image-modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }
+    const headers = {
+      'api-key': 'quickstart-QUdJIGlzIGNvbWluZy4uLi4K',
+    }
+    const response = await fetch(url, {
+      method: 'POST',
+      headers,
+      body: JSON.stringify(data),
+    })
+    const json = await response.json()
```
mqp, line 95: I note there's no error handling but whatever.

web/components/editor/image-modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+          Dream
+          {/* TODO: Charge M$5 with ({formatMoney(5)}) */}
+        </Button>
+      </Row>
+      {isDreaming && (
+        <div className="text-sm">This may take ~10 seconds...</div>
+      )}
+      <div className="pt-2 text-sm text-gray-400">Modifiers: {MODIFIERS}</div>
+
+      {/* Show the current imageUrl */}
```
mqp, line 128: In my experience using AI art generators, it's really valuable to have an interface that lets me freely explore without worrying about losing the stuff I already generated. I'm not suggesting you make this screen super complicated, but I would simply keep all of the images they generated and just pile them up here, one after another, and then let them pick which one they like the best. I think that would be much more fun and satisfying to use than just retaining the most recent generated image.

web/pages/api/v0/dream.ts
akrolsmir: Doh, that was the issue  -- using `JSON.parse(req.body)`. Thanks!

web/pages/api/v0/market/[id]/comment.ts
Comment (akrolsmir) [reply web/pages/api/v0/market/[id]/comment.ts:19]: Or just `/api/v0/comment` and everything else goes into the POST parameters?

web/components/editor/image-modal.tsx
akrolsmir: Tabs className only seems to modify the appearance of the tabs, not the body:

image

web/pages/api/v0/market/[id]/comment.ts
Comment (akrolsmir) [reply web/pages/api/v0/market/[id]/comment.ts]: Good point, done.

web/pages/api/v0/dream.ts
mqp: Why not type it correctly? I just put `...args` because I am lazy at writing GitHub comments and I thought you would write the real code. You can use [the `Parameters` type](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype).

PR #956, opened 2022-09-29
Remove tag parsing

sipec:
We used to parse #tags from descriptions and titles, and then we killed the tags feature, but some of the vestigial code remained

- old text-only markets with # will no longer link to (nonexistant) tag pages
- creating markets should be slightly faster


jahooma (approved):
LGTM! Thanks for the cleanup, Sinclair!

In `firestore.rules`, you might be able to remove 'tags' and 'lowercaseTags' from the modifiable keys of contracts.

sipec:
> 
> In `firestore.rules`, you might be able to remove 'tags' and 'lowercaseTags' from the modifiable keys of contracts.

I thought we still used tags array for something jank (the big groups?)
don't feel like touching it

PR #957, opened 2022-09-29
Hide spoiler content in emails / notifs

sipec:
opening a new pr instead of putting in #942 because I am more suspicious of this code and reverting new extension code in 942 can break things


sipec:
I just realized that we already do parsing like this for the @mentions, so the performance can't be thaat bad

sipec:
known issue: if the spoiler contains any formatting inside, like `||did Keltham _really_ eat the cake||`,  it will convert to multiple instances like "[spoiler][spoiler][spoiler]" because the json has a new object any time formatting changes (as it must).
the trivial solution to this is to check the previous sibling, if it has a spoiler mark then just convert current section to ""

PR #961, opened 2022-09-29
Overview page on groups

FRCassarino:
image
image

image
image
image
image
image
image
image

### DIFF
--- a/web/components/groups/group-overview.tsx
+++ b/web/components/groups/group-overview.tsx
@@ -77,7 +77,7 @@ export function GroupOverview(props: {
         maxResults={MAX_TRENDING_POSTS}
         defaultFilter={'all'}
         additionalFilter={{ groupSlug: group.slug }}
-        persistPrefix={`group-${group.slug}`}
+        persistPrefix={`group-trending-${group.slug}`}
       />
       <GroupAbout
         group={group}
@@ -181,22 +181,14 @@ function GroupOverviewPinned(props: {
               {element}
 
               {editMode && (
-                <div>
-                  <button
-                    className=" text-gray-500 hover:text-gray-700"
-                    onClick={() => {
-                      console.log(group.pinnedItems)
-                      const newPinned = group.pinnedItems.filter((item) => {
-                        return item.itemId !== group.pinnedItems[index].itemId
-                      })
-                      updateGroup(group, { pinnedItems: newPinned })
-                    }}
-                  >
-                    <div className="absolute top-0 left-0 right-0 bottom-0 bg-gray-200 bg-opacity-50">
-                      <XCircleIcon className="h-12 w-12 text-gray-600" />
-                    </div>
-                  </button>
-                </div>
+                <CrossIcon
+                  onClick={() => {
+                    const newPinned = group.pinnedItems.filter((item) => {
+                      return item.itemId !== group.pinnedItems[index].itemId
+                    })
+                    updateGroup(group, { pinnedItems: newPinned })
+                  }}
+                />
               )}
             </div>
           ))}
@@ -368,5 +360,19 @@ export function GroupAbout(props: {
         )}
       </Col>
     </>
+  )
+}
+
+function CrossIcon(props: { onClick: () => void }) {
+  const { onClick } = props
+
+  return (
+    <div>
+      <button className=" text-gray-500 hover:text-gray-700" onClick={onClick}>
+        <div className="absolute top-0 left-0 right-0 bottom-0 bg-gray-200 bg-opacity-50">
+          <XCircleIcon className="h-12 w-12 text-gray-600" />
+        </div>
+      </button>
+    </div>
   )
 }
--- a/web/components/groups/group-overview.tsx
+++ b/web/components/groups/group-overview.tsx
@@ -1,0 +1,372 @@
+import { track } from '@amplitude/analytics-browser'
+import {
+  ArrowSmRightIcon,
+  PlusCircleIcon,
+  XCircleIcon,
+} from '@heroicons/react/outline'
+
+import PencilIcon from '@heroicons/react/solid/PencilIcon'
+
+import { Contract } from 'common/contract'
+import { Group } from 'common/group'
+import { Post } from 'common/post'
+import { useEffect, useState } from 'react'
+import { ReactNode } from 'react'
+import { getPost } from 'web/lib/firebase/posts'
+import { ContractSearch } from '../contract-search'
+import { ContractCard } from '../contract/contract-card'
+
+import Masonry from 'react-masonry-css'
+
+import { Col } from '../layout/col'
+import { Row } from '../layout/row'
+import { SiteLink } from '../site-link'
+import { GroupOverviewPost } from './group-overview-post'
+import { getContractFromId } from 'web/lib/firebase/contracts'
+import { groupPath, updateGroup } from 'web/lib/firebase/groups'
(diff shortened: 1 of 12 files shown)


web/components/groups/group-overview.tsx
```diff
@@ -0,0 +1,369 @@
...
+        </SiteLink>
+      ) : (
+        <span className="text-xl">{content}</span>
+      )}
+      {children}
+    </Row>
+  )
+}
+
+export function GroupAbout(props: {
```
FRCassarino, line 274: This function was just moved (cut and pasted) from the group index page, no new code.

web/components/post-card.tsx
```diff
@@ -0,0 +1,82 @@
...
+import Link from 'next/link'
+import { useUserById } from 'web/hooks/use-user'
+import { postPath } from 'web/lib/firebase/posts'
+import { fromNow } from 'web/lib/util/time'
+import { Avatar } from './avatar'
+import { CardHighlightOptions } from './contract/contracts-grid'
+import { Row } from './layout/row'
+import { UserLink } from './user-link'
+
+export function PostCard(props: {
```
FRCassarino, line 13: Most (but not all) of this code was just moved from the group index page

web/components/groups/group-overview.tsx
```diff
@@ -0,0 +1,372 @@
...
+    await updateGroup(group, {
+      pinnedItems: [
+        ...group.pinnedItems,
+        ...(selectedItems as { itemId: string; type: 'contract' | 'post' }[]),
+      ],
+    })
+    setOpen(false)
+  }
+
+  return isEditable || pinned.length > 0 ? (
```
jahooma, line 143: Would be nice to show a loading indicator if the pinned items aren't loaded, instead of "No pinned items yet".

I guess for non-editors, it's a better experience when no items were pinned to show no loading indicator...

web/components/groups/group-overview.tsx
```diff
@@ -0,0 +1,372 @@
...
+          {pinned.map((element, index) => (
+            <div className="relative my-2">
+              {element}
+
+              {editMode && (
+                <div>
+                  <button
+                    className=" text-gray-500 hover:text-gray-700"
+                    onClick={() => {
+                      console.log(group.pinnedItems)
```
jahooma, line 188: Good to remove `console.log`s from production

web/components/groups/group-overview.tsx
```diff
@@ -0,0 +1,372 @@
...
+                No pinned items yet. Click the edit button to add some!
+              </p>
+            </div>
+          )}
+          {pinned.map((element, index) => (
+            <div className="relative my-2">
+              {element}
+
+              {editMode && (
+                <div>
```
jahooma, line 184: You could extract this 'X' button to a small component defined later in the file.

More small components is generally good style. It reduces nesting and makes the parent component & child component more readable.

web/components/groups/group-overview.tsx
```diff
@@ -0,0 +1,372 @@
...
+                    <div className="absolute top-0 left-0 right-0 bottom-0 bg-gray-200 bg-opacity-50">
+                      <XCircleIcon className="h-12 w-12 text-gray-600" />
+                    </div>
+                  </button>
+                </div>
+              )}
+            </div>
+          ))}
+          {editMode && group.pinnedItems && pinned.length < 6 && (
+            <div className=" py-2">
```
jahooma, line 204: Ditto for extracting the a pin button component from here.

web/components/pinned-select-modal.tsx
```diff
@@ -0,0 +1,164 @@
...
+    selectItem(contract.id, 'contract')
+  }
+
+  async function selectPost(post: Post) {
+    selectItem(post.id, 'post')
+  }
+
+  async function selectItem(itemId: string, type: string) {
+    setSelectedItem({ itemId: itemId, type: type })
+  }
```
jahooma, line 54: These don't need to be `async`

web/components/pinned-select-modal.tsx
```diff
@@ -0,0 +1,164 @@
...
+                hideQuickBet: true,
+                noLinkAvatar: true,
+              }}
+              highlightOptions={{
+                itemIds: [selectedItem?.itemId ?? ''],
+                highlightClassName:
+                  '!bg-indigo-100 outline outline-2 outline-indigo-300',
+              }}
+              additionalFilter={{ groupSlug: group.slug }}
+              persistPrefix={`group-${group.slug}`}
```
jahooma, line 138: This prefix should be unique within the app. Add like `pinned-select` or something.

web/pages/group/[...slugs]/index.tsx
```diff
@@ -258,14 +242,14 @@ export default function GroupPage(props: {
...
     },
-    {
-      title: 'About',
-      content: aboutTab,
-    },
     {
       title: 'Posts',
       content: postsPage,
     },
+    {
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:249]: I think it would be good if navigating to `/about` would show this overview tab. Helps with the ongoing tournaments for one.

I also think getting the routing right for switching between tabs is useful. E.g. clicking leaderboards should change the url to `/leaderboards` and if you hit back, it goes to the previous tab.  Probably a separate change though!

web/components/post-card.tsx
```diff
@@ -0,0 +1,82 @@
...
+  onPostClick?: (post: Post) => void
+  highlightOptions?: CardHighlightOptions
+}) {
+  const { post, onPostClick, highlightOptions } = props
+  const creatorId = post.creatorId
+
+  const user = useUserById(creatorId)
+  const { itemIds: itemIds, highlightClassName } = highlightOptions || {}
+
+  if (!user) return <> </>
```
jahooma, line 24: Wouldn't it be better to render most of the card while the user is loading?

web/components/pinned-select-modal.tsx
```diff
@@ -0,0 +1,164 @@
...
+  async function selectPost(post: Post) {
+    selectItem(post.id, 'post')
+  }
+
+  async function selectItem(itemId: string, type: string) {
+    setSelectedItem({ itemId: itemId, type: type })
+  }
+
+  async function onFinish() {
+    setLoading(true)
```
jahooma, line 57: Shouldn't this be inside the `if`?

web/components/groups/group-overview.tsx
jahooma: Thanks for letting me know!

web/components/groups/group-overview.tsx
```diff
@@ -0,0 +1,372 @@
...
+                    className=" text-gray-500 hover:text-gray-700"
+                    onClick={() => {
+                      console.log(group.pinnedItems)
+                      const newPinned = group.pinnedItems.filter((item) => {
+                        return item.itemId !== group.pinnedItems[index].itemId
+                      })
+                      updateGroup(group, { pinnedItems: newPinned })
+                    }}
+                  >
+                    <div className="absolute top-0 left-0 right-0 bottom-0 bg-gray-200 bg-opacity-50">
```
jahooma, line 195: Would it be better to put this in the top right?

jahooma (approved):
Very cool, thanks Fede! I like the direction this goes in for groups. An overview with many sections, including a curated pinned section, is a good idea.

Will be interested to see how the overview evolves and what sections users are interested in.

Might be worth seeking feedback from users about what they would want in the overview, or how they use groups in general.

PR #963, opened 2022-09-29
Redo add funds modal without daisy and actually use it

sipec:
image


web/components/add-funds-modal.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export function AddFundsModal(props: {
+  open: boolean
+  setOpen(open: boolean): void
+}) {
+  const { open, setOpen } = props
+
+  const user = useUser()
+  // in hook so not run in nextjs
+  const [location] = useState(window.location.href)
```
jahooma, line 17: This still could get run on the server. Only a function within a `useEffect` would not be run, but the main render method is executed as normal javascript.

web/components/add-funds-modal.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <div className="text-xl">{manaToUSD(amountSelected)}</div>
+      </div>
+
+      <div className="modal-action">
+        <Button color="gray-white" onClick={() => setOpen(false)}>
+          Back
+        </Button>
+
+        <form
+          action={checkoutURL(user?.id || '', amountSelected, location)}
```
jahooma, line 46: Actually, nothing fancy is needed to access the location here so long as it is inside a function that is invoked clientside (when the form is submitted).

I would refactor `checkoutURL` to directly return the string endpoint instead of a function that returns the string, and then change this to:
`action={() => checkoutURL(user?.id || '', amountSelected, window.location.href)}`

jahooma (approved):
Thanks Sinclair! The dialog does seem a lot better than sending you to a new page.

I had a couple comments about how to implement getting the current location, which should be addressed before merging.

web/components/add-funds-modal.tsx
mantikoros, line 17: I don't think this works. window.location is still evaluated by nextjs if it's not in a useEffect block

mantikoros (commented):
LGTM!

Looks like there's still some duplication between AddFundsModal and AddFundsPage...

PR #965, opened 2022-09-30
Inga/scroll to top

ingawei:
Screen Shot 2022-09-29 at 5 46 18 PM
Screen Shot 2022-09-29 at 5 46 28 PM

- added scroll to top button to replace the "predict" button at the bottom of comments/trades
- Mainly intended to work with new mobile betting, although I see how for multiple choice it may still be annoying. For multiple choice markets though, I do want to incorporate changes that will make all betting within the top screen.
- This will not appear when the side panel is there cause I think that just defeats the purpose


web/pages/[username]/[contractSlug].tsx
```diff
@@ -276,23 +272,9 @@ export function ContractPageContent(
           userBets={userBets}
           comments={comments}
         />
-
-        {!user ? (
-          <Col className="mt-4 max-w-sm items-center xl:hidden">
-            <BetSignUpPrompt />
-            <PlayMoneyDisclaimer />
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:283]: Does the PlayMoneyDisclaimer otherwise appear on mobile for signed out users? I think we could relocate it, but not remove it.

web/components/scroll-to-top-button.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [visible, setVisible] = useState(false)
+
+  useEffect(() => {
+    window.addEventListener('scroll', () => {
+      if (window.scrollY > 500) {
+        setVisible(true)
+      } else {
+        setVisible(false)
+      }
+    })
```
jahooma, line 17: This is a decent solution, but scroll event listeners are kind of slow.

(Also, you should remove the event listener when the component is destroyed, which you would do by returning a function in the useEffect: https://bobbyhadz.com/blog/react-remove-event-listener)

Another solution is to add our component `<VisibilityObserver />` at the top of the page, and show this button whenever it is not visible.

jahooma (changes requested):
I think this is a great UX improvement. Thanks, Inga!

Had a couple comments on the implementation.

web/components/scroll-to-top-button.tsx
```diff
@@ -0,0 +1,48 @@
+import { ArrowUpIcon } from '@heroicons/react/solid'
+import clsx from 'clsx'
+import { useEffect, useState } from 'react'
+import { Row } from './layout/row'
+
+export function ScrollToTopButton(props: { className?: string }) {
+  const { className } = props
+  const [visible, setVisible] = useState(false)
+
+  const onScroll = () => {
```
jahooma, line 10: I think I'd move this function inside the `useEffect`. One reason is it creates a new function each render here, while only the first is being referenced in the effect.

web/components/scroll-to-top-button.tsx
```diff
@@ -0,0 +1,48 @@
...
+    if (window.scrollY > 500) {
+      setVisible(true)
+    } else {
+      setVisible(false)
+    }
+  }
+
+  useEffect(() => {
+    window.addEventListener('scroll', onScroll, { passive: true })
```
jahooma, line 19: Awesome, passive is a good solution here!

jahooma (approved):
Thanks Inga!

PR #968, opened 2022-09-30
Experimental - change probability on binary contract page on chart hover

mqp:
This is an idea that Inga had. I don't personally think I like the feel of it, but I thought I would put it up here so that anyone can try it and play around with it and see if they want to make a case for it.

@ingawei feel free to check it out and see what you think. If you really really like it I will consider making the other charts work like this. (I guess the free response chart doesn't really have anything analogous, since the answer panel probably isn't even completely visible while you are hovering the chart...)


mqp:
Pretty sure we aren't doing this.

PR #969, opened 2022-09-30
getting rid of daisy buttons

ingawei:
this is so submit won't turn black, although now that I think about it closing the modal would've been a lot easier rip.
I basically just have replaced the daisy button for `ConfirmationButton`'s `openModalBtn` with our custom `Button` component, which ended up affecting quite a few other buttons.

### DIFF
--- a/web/components/confirmation-button.tsx
+++ b/web/components/confirmation-button.tsx
@@ -105,11 +105,7 @@ export function ResolveConfirmationButton(props: {
   return (
     <ConfirmationButton
       openModalBtn={{
-        className: clsx(
-          'border-none self-start',
-          openModalButtonClass,
-          isSubmitting && 'btn-disabled loading'
-        ),
+        className: clsx('border-none self-start', openModalButtonClass),
         label: 'Resolve',
         color: color,
         disabled: isSubmitting || disabled,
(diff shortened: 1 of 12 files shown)


web/components/confirmation-button.tsx
```diff
@@ -96,6 +111,9 @@ export function ResolveConfirmationButton(props: {
           isSubmitting && 'btn-disabled loading'
```
jahooma, line 111: I think you don't need this line

web/components/groups/create-group-button.tsx
```diff
@@ -82,11 +82,8 @@ export function CreateGroupButton(props: {
       openModalBtn={{
         label: label ? label : 'Create Group',
         icon: icon,
-        className: clsx(
-          isSubmitting ? 'loading btn-disabled' : 'btn-primary',
-          'btn-sm, normal-case',
-          className
-        ),
+        className: clsx(className),
```
jahooma, line 85: No need to wrap with clsx. This line could just be `className,`

jahooma (approved):
Thanks Inga! Very glad to have this fixed. Plus, getting rid of more Daisy UI components is the Lord's work.

PR #971, opened 2022-09-30
Chart visual style adjustment

mqp:
Now they look more like the portfolio charts; the color and area fill opacity match.

Not sure what to do with the top line... making it chunkier doesn't really look good to me. Feel free to play around with it; you would just need to fiddle with the `strokeWidth` property on the `AreaWithTopStroke` `LinePath` component.

FYI @ingawei


mqp:
Changed it a bit, looking better...

PR #974, opened 2022-09-30
Allow user to opt out of all unnecessary notifications

IanPhilips:
Users can opt out of all notifications except market closure notifications. I added a link when unsubscribing to an email type to unsubscribe to all unnecessary email notifications.
Screen Shot 2022-09-30 at 3 57 33 PM
Screen Shot 2022-09-30 at 3 57 27 PM


mqp:
Wow, if you merge this I will actually use notifications! That's the one kind of notification I want.

PR #976, opened 2022-09-30
Make small embeds into cards

sipec:
also some style fixes


web/pages/embed/[username]/[contractSlug].tsx
```diff
@@ -64,16 +66,31 @@ export default function ContractEmbedPage(props: {
...
     creatorId: contract.creatorId,
   })
 
+  const { height = 0 } = useWindowSize()
+
+  return height < 250 ? (
+    <ContractCard contract={contract} className="h-screen" />
+  ) : (
+    <ContractSmolView {...props} />
+  )
```
Comment (sipec) [inline web/pages/embed/[username]/[contractSlug].tsx:88]: I'm worried that putting rendering behind a hook might prevent Next from prefetching the content, but I haven't tested the alternative.

Comment (jahooma) [reply web/pages/embed/[username]/[contractSlug].tsx:88]: I think the Tailwind guys in general recommend rendering both and using media queries to show just one. This strategy works well with server-side rendering so that it shows the correct thing immediately no matter the client device / iframe size.

You might need custom breakpoints to do this here: https://tailwindcss.com/docs/screens

But I'd also be curious to see how the hooks version compares empirically. So I'm ok to try it out this way if you want.

Comment (jahooma) [reply web/pages/embed/[username]/[contractSlug].tsx:88]: In terms of prefetching, I think Nextjs would bundle all import dependencies for the page, so I think either way would be the same.

jahooma (approved):
This is great. Thanks Sinclair!

web/pages/embed/[username]/[contractSlug].tsx
Comment (sipec) [reply web/pages/embed/[username]/[contractSlug].tsx]: the hooks version subjectively felt slower so I'm doing it the media query way

PR #978, opened 2022-10-01
Don't render stuff whenever window size changes

mqp:
Stop doing this, it's no good, resizing should be fast.

### DIFF
--- a/web/components/bet-panel.tsx
+++ b/web/components/bet-panel.tsx
@@ -47,7 +47,6 @@ import { Modal } from './layout/modal'
 import { Title } from './title'
 import toast from 'react-hot-toast'
 import { CheckIcon } from '@heroicons/react/solid'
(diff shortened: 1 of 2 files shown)


web/components/bet-panel.tsx
```diff
@@ -179,12 +178,7 @@ export function BuyPanel(props: {
   const initialProb = getProbability(contract)
   const isPseudoNumeric = contract.outcomeType === 'PSEUDO_NUMERIC'
 
-  const windowSize = useWindowSize()
-  const initialOutcome =
-    windowSize.width && windowSize.width >= 1280 ? 'YES' : undefined
```
mqp, line 184: This code looked like it was doing something, but it wasn't, because `useWindowSize` always starts out returning `undefined` dimensions until its internal `useEffect` runs, so `windowSize.width` was always falsy and `initialOutcome` was always `undefined`.

PR #985, opened 2022-10-02
Rewrite stats graphs using new machinery

mqp:
Now the stats page uses the new graphs, so we remove the Nivo dependency, we get consistent look and feel, tooltips, zooming, speed, etc.

### DIFF
--- a/web/components/sized-container.tsx
+++ b/web/components/sized-container.tsx
@@ -1,0 +1,35 @@
+import { ReactNode, useEffect, useRef, useState } from 'react'
+
+export const SizedContainer = (props: {
+  fullHeight: number
+  mobileHeight: number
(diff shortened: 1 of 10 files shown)


web/components/sized-container.tsx
```diff
@@ -0,0 +1,35 @@
+import { ReactNode, useEffect, useRef, useState } from 'react'
+
+export const SizedContainer = (props: {
```
jahooma, line 3: Nice abstraction!

mqp:
FYI @jahooma since you seem to be the main hacker of the stats page.

jahooma (approved):
Whoa, super nice! Cool to use the better charts and get closer to ditching Nivo.

Code looks good!

PR #986, opened 2022-10-03
Rewrite portfolio graphs with new machinery

mqp:
Now all the charts are using the new stuff and Nivo is toast.

The behavior on this chart is slightly inferior in one way right now to the previous version -- there's no way to see the exact number at a time on mobile with the new tooltips -- but this is on the docket for me to fix in short order on all the charts. Meanwhile, other things are greatly improved (e.g. y-axis formatting, handling of profit graphs that go between negative and positive.)


jahooma:
Thanks Marshall, this was a big effort!

PR #989, opened 2022-10-03
Inga/prettier embed

ingawei:
Getting rid of inline betting in embeds, replacing with quickbets
Screen Shot 2022-10-03 at 1 45 33 PM

logged out/not binary market view
Screen Shot 2022-10-03 at 1 45 40 PM

weird iframe border :( help

### DIFF
--- a/web/pages/embed/[username]/[contractSlug].tsx
+++ b/web/pages/embed/[username]/[contractSlug].tsx
@@ -27,6 +27,9 @@ import {
   tradingAllowed,
 } from 'web/lib/firebase/contracts'
 import Custom404 from '../../404'
+import { useUser } from 'web/hooks/use-user'
+import { QuickBet } from 'web/components/contract/quick-bet'
+import { contractMetrics } from 'common/contract-details'
 
 export const getStaticProps = fromPropz(getStaticPropz)
 export async function getStaticPropz(props: {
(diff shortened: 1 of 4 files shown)


ingawei:
@akrolsmir need some help with figuring out if iFrame resizing is possible. Maybe we could talk about this in 1:1 tomorrow?

jahooma:
> @akrolsmir need some help with figuring out if iFrame resizing is possible. Maybe we could talk about this in 1:1 tomorrow?

I don't know if this would solve your problem, but you can measure the current iframe size with `useWindowSize()`

web/pages/embed/[username]/[contractSlug].tsx
```diff
@@ -74,66 +78,87 @@ export function ContractEmbed(props: { contract: Contract; bets: Bet[] }) {
...
 
   const [probAfter, setProbAfter] = useState<number>()
 
-  return (
-    <Col className="h-[100vh] w-full bg-white">
-      <Row className="justify-between gap-4 px-2">
-        <div className="text-xl text-indigo-700 md:text-2xl">
-          <SiteLink href={href}>{question}</SiteLink>
+  const user = useUser()
+  if (user && (isBinary || isPseudoNumeric)) {
```
Comment (jahooma) [inline web/pages/embed/[username]/[contractSlug].tsx:95]: What if the market is closed or resolved? You would probably want the `else` case for that.

jahooma (changes requested):
I think this change needs to be merged with Sinclair's change that shows the contract card on small iframe sizes.

I like this change overall! Quick bet arrows is better than having another custom UI for betting here.

PR #991, opened 2022-10-03
Mark scripts as modules to avoid global name collision

mqp:
@akrolsmir randomly checked this in busted and broke the build.


akrolsmir:
Whoops -- hadn't realized this could even be a thing. Thanks for catching and fixing @mqp!

PR #993, opened 2022-10-04
Make %mentions actually look like mentions

sipec:
Screen Shot 2022-10-03 at 5 28 14 PM
Screen Shot 2022-10-03 at 5 30 13 PM

### DIFF
--- a/web/components/contract/contract-mention.tsx
+++ b/web/components/contract/contract-mention.tsx
@@ -1,0 +1,43 @@
+import clsx from 'clsx'
+import { Contract } from 'common/contract'
+import Link from 'next/link'
+import { contractPath, getBinaryProbPercent } from 'web/lib/firebase/contracts'
+import { BinaryContractOutcomeLabel } from '../outcome-label'
+import { getColor } from './quick-bet'
+
+export function ContractMention(props: { contract: Contract }) {
+  const { contract } = props
+  const { outcomeType, resolution } = contract
+  const probTextColor = `text-${getColor(contract)}`
+
+  return (
+    <Link href={contractPath(contract)}>
+      <a
+        className="group inline whitespace-nowrap rounded-sm hover:bg-indigo-50 focus:bg-indigo-50"
+        title={outcomeType}
+      >
+        <span className="break-anywhere mr-0.5 whitespace-normal font-normal text-indigo-700">
+          {contract.question}
+        </span>
+        {outcomeType === 'BINARY' && (
+          <span
+            className={clsx(
+              probTextColor,
+              'rounded-full  px-1 font-semibold ring-1 ring-inset ring-indigo-100 group-hover:ring-0'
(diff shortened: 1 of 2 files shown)


sipec:
I know this is gonna be controversial. but the previous card mention experience wasn't actually that great when doing anything other than making a list of cards. I do think we should re-add it later..

web/components/contract/contract-mention.tsx
```diff
@@ -0,0 +1,54 @@
...
+export function ContractMention(props: { contract: Contract }) {
+  const { contract } = props
+  const { outcomeType, resolution } = contract
+  const probTextColor = `text-${getColor(contract)}`
+
+  return (
+    <Link href={contractPath(contract)}>
+      <a
+        className="group inline whitespace-nowrap rounded-sm hover:bg-indigo-50 focus:bg-indigo-50"
+        title={tooltipLabel(contract)}
```
akrolsmir, line 19: Can we use the tippy Tooltip instead of just title? For consistent formatting with all the rest of our tooltips

web/components/contract/contract-mention.tsx
```diff
@@ -0,0 +1,54 @@
...
+      >
+        <span className="break-anywhere mr-0.5 whitespace-normal font-normal text-indigo-700">
+          {contract.question}
+        </span>
+        {outcomeType === 'BINARY' && (
+          <span
+            className={clsx(
+              probTextColor,
+              'rounded-full  px-1 font-semibold ring-1 ring-inset ring-indigo-100 group-hover:ring-0'
```
akrolsmir, line 28: I like the look of this, though would add a bit more padding in the x direction to make it look more pill-like. px-2 or px-2.5 or px-3? Idk, play with it a bit

akrolsmir (approved):
Looks much neater inline, and love the little % pill at the end. Great idea!

I'm not sure I like how this breaks existing cards, but also probably not that many people have used contract cards/% mentions so far? Probably just give a heads-up in standup before merging this in, I think @SirSaltyy and @FRCassarino have used contract cards before.

Longer term, I'm not sure if I like the [Notion pattern](https://manifoldmarkets.notion.site/Notion-al-markets-master-plan-b6863be92e784fa6a28f18bea5ff3835) of having a second popup to choose whether you want to embed as a mention or link, because it adds more friction to selecting stuff. But I don't know of great UX alternatives either. Don't feel too strongly, I'll defer to your thoughts here

web/components/contract/contract-mention.tsx
akrolsmir: Saw your thoughts `[c] tooltips were too distracting, just made it a title attribute (yes I know this sucks)`

Which is fine, I guess. Actually the best thing would be to show the contract card in a popup on hover, gwern-style (which shouldn't be too hard)

PR #994, opened 2022-10-04
Inga/multiple colors

ingawei:
Screen Shot 2022-10-04 at 1 49 20 AM
Screen Shot 2022-10-04 at 1 49 26 AM
Screen Shot 2022-10-04 at 1 49 52 AM

when market is closed, there is no drop shadow and nothing is clickable
Screen Shot 2022-10-04 at 2 05 01 AM

Making multiple choice market options much smoller. I'm trying to make smaller PR's so this is just a part of a bigger change I want to make, where only the top 5 markets are colored and the rest are gray. I've picked lighter colors for @mqp to incorporate, but for now I've just used a lower opacity for the background colors.

### DIFF
--- a/web/components/answers/answers-panel.tsx
+++ b/web/components/answers/answers-panel.tsx
@@ -215,7 +215,7 @@ function OpenAnswer(props: {
         <Row className="z-20 -mb-1 justify-between gap-2 py-2 px-3">
           <Row>
             <Avatar
-              className="mr-2 inline h-5 w-5 border border-transparent transition-transform hover:border-none"
+              className="mt-0.5 mr-2 inline h-5 w-5 border border-transparent transition-transform hover:border-none"
               username={username}
               avatarUrl={avatarUrl}
             />
--- a/web/components/answers/answers-panel.tsx
+++ b/web/components/answers/answers-panel.tsx
@@ -210,7 +210,7 @@ function OpenAnswer(props: {
       </Modal>
 
       <Col
-        className="hover:bg-greyscale-1.5 bg-greyscale-1 relative w-full rounded-lg transition-colors"
+        className="bg-greyscale-1 relative w-full rounded-lg drop-shadow-sm transition-all hover:drop-shadow-md"
         onClick={() => setOpen(true)}
       >
         <Row className="z-20 -mb-1 justify-between py-2 px-3">
@@ -223,16 +223,20 @@ function OpenAnswer(props: {
               />
             </span>
             <Linkify
-              className="text-md cursor-pointer whitespace-pre-line"
+              className={clsx(
+                'text-md cursor-pointer whitespace-pre-line',
+                tradingAllowed(contract)
+                  ? 'text-greyscale-7'
+                  : 'text-greyscale-6'
+              )}
               text={text}
             />
           </div>
           <div
             className={clsx(
-              'my-auto text-xl '
-              // tradingAllowed(contract) ? 'text-greyscale-7' : 'text-gray-500'
+              'my-auto text-xl ',
+              tradingAllowed(contract) ? 'text-greyscale-7' : 'text-greyscale-6'
             )}
-            color={color}
           >
             {probPercent}
           </div>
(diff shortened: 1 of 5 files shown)


ingawei:
@mqp I'm pretty sure the colors match what's on the chart but if you could take a look/know a more better way I could do this please let me know!

web/components/answers/answers-panel.tsx
```diff
@@ -193,40 +206,37 @@ function OpenAnswer(props: {
...
+      <Col
+        className={clsx(
+          'bg-greyscale-1 relative w-full rounded-lg transition-all',
+          tradingAllowed(contract)
+            ? 'text-greyscale-7 drop-shadow-sm hover:drop-shadow-md'
+            : 'text-greyscale-5 pointer-events-none'
+        )}
+        onClick={() => setOpen(true)}
+      >
+        <Row className="z-20 -mb-1 justify-between py-2 px-3">
```
akrolsmir, line 218: It's not clear that you can click on these to purchase them (on desktop, at least -- and probably on mobile too.)

I'd at least add a `cursor-pointer`, and it would be really cool to have the word "Buy" or "Predict" or something float in from the left side when you're hovering over the bar!

Not sure how to make this more explicit on mobile...

web/components/charts/contract/choice.tsx
```diff
@@ -22,7 +22,7 @@ import { Row } from 'web/components/layout/row'
 import { Avatar } from 'web/components/avatar'
 
 // thanks to https://observablehq.com/@jonhelfman/optimal-orders-for-choosing-categorical-colors
-const CATEGORY_COLORS = [
+export const CATEGORY_COLORS = [
```
akrolsmir, line 25: I still think these colors are super ugly :( Gavriel suggests some alternatives https://discord.com/channels/915138780216823849/915138780653051910/1013157893593370664, and I like the look of https://projects.susielu.com/viz-palette

akrolsmir: (but can change this in a different PR. it's mostly ugly on the chart, which isn't changing in this PR)

web/components/answers/answers-panel.tsx
```diff
@@ -128,7 +133,12 @@ export function AnswersPanel(props: {
           )}
         >
           {answerItems.map((item) => (
-            <OpenAnswer key={item.id} answer={item} contract={contract} />
```
akrolsmir, line 131: I kinda like `gap-1` in line 131 now, wdyt
image

web/tailwind.config.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
       colors: {
         'red-25': '#FDF7F6',
         'greyscale-1': '#FBFBFF',
+        'greyscale-1.5': '#F4F4FB',
```
akrolsmir, line 21: Hm, is this used? (If not, possibly remove it -- I'm a bit skeptical of adding in-between shades but it's your call)

akrolsmir (approved):
Looks great, I really like that the colors match up to the chart now, as well as the overall look and feel of the bars!

web/components/answers/answers-panel.tsx
ingawei: Hmm yeah I thought maybe having a small dropshadow made them look more clickable, I'll continue thinking about it

web/components/charts/contract/choice.tsx
ingawei: haha I agree, Marshall and I are actually thinking of only displaying the top 5 answers as colors, the reset in a gray color under 'Other'
I found this palette that is specifically colorblind friendly
image

web/components/answers/answers-panel.tsx
mqp: I agree with Austin, I think this looks good but nobody will have any clue how to bet. Why not put a button saying "buy" to the right of the answer?

ingawei: added a buy button!

web/tailwind.config.js
ingawei: I've been wanting it for a while, the difference between greyscale-1 and greyscale-2 is a bit too jarring. Yeah 1.5 is kind of annoying but I didn't want to shift everything over

PR #999, opened 2022-10-04
Memoize `FeedBet` component

mqp:
This is valuable because this bets list has to re-render when the LPs come in, even though most of the content is the same bets and contract. Unfortunately it doesn't save us when we subscribe to the bets and get the same list that was in the static props.


jahooma (approved):
Sounds good!

PR #1000, opened 2022-10-04
Change `Tipper` interface, memoize `FeedComment`

mqp:
Similar to #999. This isn't perfect, because it still re-renders when the comments are all re-loaded from the subscription, but it does prevent re-rendering of (most, untipped) comments when the tips come in.

### DIFF
--- a/web/components/feed/feed-comments.tsx
+++ b/web/components/feed/feed-comments.tsx
@@ -192,7 +192,7 @@ export const FeedComment = memo(function FeedComment(props: {
           {onReplyClick && (
             <button
               className="font-bold hover:underline"
-              onClick={() => onReplyClick?.(comment)}
+              onClick={() => onReplyClick(comment)}
             >
               Reply
             </button>
--- a/web/components/feed/feed-comments.tsx
+++ b/web/components/feed/feed-comments.tsx
@@ -1,11 +1,14 @@
+import React, { memo, useEffect, useRef, useState } from 'react'
+import { Editor } from '@tiptap/react'
+import { useRouter } from 'next/router'
+import { sum } from 'lodash'
+import clsx from 'clsx'
+
 import { ContractComment } from 'common/comment'
 import { Contract } from 'common/contract'
-import React, { useEffect, useRef, useState } from 'react'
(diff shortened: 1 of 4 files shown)


web/components/feed/feed-comments.tsx
```diff
@@ -180,20 +192,26 @@ export function FeedComment(props: {
           {onReplyClick && (
             <button
               className="font-bold hover:underline"
-              onClick={onReplyClick}
+              onClick={() => onReplyClick?.(comment)}
```
jahooma, line 195: `onReplyClick` should be defined here

web/components/tipper.tsx
```diff
@@ -13,25 +12,28 @@ import { Row } from './layout/row'
...
+  comment: Comment
+  myTip: number
+  totalTip: number
+}) {
+  const { comment, myTip, totalTip } = prop
 
   const me = useUser()
-  const myId = me?.id ?? ''
-  const savedTip = tips[myId] ?? 0
+  const savedTip = myTip
```
jahooma, line 23: Why have both `savedTip` and `myTip`?

jahooma (approved):
Sweet, hope this saves us a bunch of rerenders!

web/components/tipper.tsx
mqp: I have no idea, this component code looks ridiculous at a glance but I don't feel like trying to deal with it right now so I just preserved existing logic.

web/components/feed/feed-comments.tsx
mqp: Can you clarify what you think should be changed here?

jahooma: Sure — you don't need the `?` in `onReplyClick?.(comment)` since `onReplyClick` is truthy in the `&&` condition above.

web/components/tipper.tsx
jahooma: You could delete this line and just refer to `myTip`

mqp: I grudgingly admit that I can at least do that.

PR #1014, opened 2022-10-07
Call updatemetrics v2 cloud function from scheduled function

jahooma:


functions/src/update-metrics.ts
```diff
@@ -20,13 +21,32 @@ import {
...
+  .onRun(async () => {
+    const response = await fetch(getFunctionUrl('updatemetrics'), {
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      method: 'POST',
+      body: JSON.stringify({}),
+    }).then((res) => res.json())
+
+    console.log(await response)
```
mqp, line 40: I might have written a `console.error` for `!res.ok` responses.

mqp:
Seems fine to me, I don't care that there's no auth.

mqp:
Was this the only problem with the old `updateMetrics`? I recall you mentioning something that sounded like the Firestore query result sets getting too big?

jahooma:
> Was this the only problem with the old `updateMetrics`? I recall you mentioning something that sounded like the Firestore query result sets getting too big?

Oh yeah. In order to get all the best, I had to switch to this:
```
  const contractBets = await batchedWaitAll(
    contracts
      .filter((c) => c.id)
      .map(
        (c) => () =>
          getValues<Bet>(
            firestore.collection('contracts').doc(c.id).collection('bets')
          )
      ),
    100
  )
  const bets = contractBets.flat()
```

It's possible we'll have to use something similar if, say, the contracts collection get too big.

jahooma:
Thanks for the review!

PR #1015, opened 2022-10-08
Autosave post, market, comment rich text

sipec:
It turns out the editing speed bottleneck was spellcheck, not saving to localstorage.

I didn't bother autosaving the title and other settings like close time.

known issue: after submit is successful but before the router redirect, the content becomes cleared, which will strike fear in the user's heart that their data has been lost. The solution is probably a custom editor command to `clearAutosave` that doesn't also clear content - but I don't feel like doing it rn


web/components/editor.tsx
```diff
@@ -75,19 +81,34 @@ export function useTextEditor(props: {
...
-    '[&_.ProseMirror-selectednode]:outline-dotted [&_*]:outline-indigo-300' // selected img, emebeds
+    '[&_.ProseMirror-selectednode]:outline-dotted [&_*]:outline-indigo-300' // selected img, embeds
   )
 
   const editor = useEditor({
-    editorProps: { attributes: { class: editorClass } },
+    editorProps: {
+      attributes: { class: editorClass, spellcheck: simple ? 'true' : 'false' },
+    },
+    onUpdate: key ? ({ editor }) => save(editor.getJSON()) : undefined,
```
jahooma, line 111: Is getting the JSON expensive? Could also be re-jiggered to only get the JSON after the debounce.

web/components/editor.tsx
```diff
@@ -75,19 +81,34 @@ export function useTextEditor(props: {
   defaultValue?: Content
   disabled?: boolean
   simple?: boolean
+  key?: string // unique key for autosave. If set, plz call `clearContent(true)` on submit to clear autosave
```
jahooma, line 84: Hmm, having to call `editor.commands.clearContent(true)` is not that great. But I'm not sure of a better solution.

Slightly better would be a top level `reset()` returned from the hook?
Or maybe the caller should change the key after submitting?
Otherwise, this is not blocking.

jahooma (approved):
Cool, this is a big improvement! Thanks Sinclair!

web/components/editor.tsx
sipec: doesn't seem expensive at all

PR #1016, opened 2022-10-10
refactor text input into one component

sipec:
Refactors all our `input input-bordered` text inputs to use a shared component. This will make it easier to restyle all of them to match our rich-text editors and `<select>` menus (although actually doing that is left as an exercise for ~inga~ later)

This fixes a ~bug~ feature in iOS safari where focusing an `<input type="text" />` with font size less than 16px will cause the browser to zoom in. This became an issue after my SEO changes (removing `max-scale=1`)


sipec:
@ingawei fyi

PR #1022, opened 2022-10-11
Simplify rich text to string parsing logic

sipec:
instead of copy the json, dfs, then pass to `generateText`, we just generate the correct text inside of `generateText`

this is now possible after #1021 since these extensions are just used for the text conversion, so we can customize them just for this.

### DIFF
--- a/common/util/parse.ts
+++ b/common/util/parse.ts
@@ -51,8 +51,8 @@ export function parseMentions(data: JSONContent): string[] {
   return uniq(mentions)
 }
 
-// can't just do [StarterKit, Image...] because it doesn't work with cjs imports
 const stringParseExts = [
+  // StarterKit extensions
   Blockquote,
   Bold,
   BulletList,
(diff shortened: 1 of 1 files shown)


common/util/parse.ts
```diff
@@ -69,38 +69,20 @@ const stringParseExts = [
...
   Link,
+  Image.extend({ renderText: () => '[image]' }),
   Mention, // user @mention
   Mention.extend({ name: 'contract-mention' }), // market %mention
-  Iframe,
-  TiptapTweet,
-  TiptapSpoiler,
+  Iframe.extend({
+    renderText: ({ node }) =>
+      '[embed]' + node.attrs.src ? `(${node.attrs.src})` : '',
```
jahooma, line 79: Maybe put the url within the embed brackets? `[embed ${node.attrs.src}]`

jahooma (approved):
Thanks Sinclair, looks like a much cleaner solution!

PR #1024, opened 2022-10-11
Featured items to homepage

FRCassarino:
Adds a featured section to homepage.

At this time, it's only visible to Manifold admins, so we can ship the basics and add more stuff before showing it to everyone. 

image
image
image
image


web/pages/home/index.tsx
```diff
@@ -200,17 +210,31 @@ function renderSections(
     score: CPMMBinaryContract[]
   }
 ) {
+  const HOME_SECTIONS_IDS = HOME_SECTIONS.map((s) => s.id)
+  type sectionTypes =
```
FRCassarino, line 214: I don't like having to duplicate this from HOME_SECTION_IDS, but the alternative was adding "featured" to sectionContracts for now reason (as it doesn't make conceptual sense). Open to other options if you can think of any @jahooma

common/homeConfig.ts
```diff
@@ -0,0 +1,3 @@
+export type HomeConfig = {
```
FRCassarino, line 1: Is there a better place to store this? It feels sort of pointless to create a new root collection just for this, but I don't know where else we'd put it. I thought about adding a "homepage" group and using the existing "pinnedItems" field from that group, but it seems too hacky. I thought about adding it to some other existing root collection, but none of them really fit conceptually ('stats' is the only static one, but it doesn't make a lot of sense).

web/pages/home/index.tsx
```diff
@@ -324,6 +347,86 @@ function SearchSection(props: {
...
+function FeaturedSection() {
+  const [pinned, setPinned] = useState<JSX.Element[]>([])
+  const [posts, setPosts] = useState<Post[]>([])
+  const [homeConfig, setHomeConfig] = useState<HomeConfig | undefined>(
+    undefined
+  )
+  useEffect(() => {
+    getHomeConfig().then(setHomeConfig)
+    getAllPosts().then(setPosts)
+  }, [])
```
jahooma, line 359: This would only start loading when the component is rendered. The other sections load above from hooks.

One way to improve this is to extract this loading into a hook, load the data above, and pass it down into the section.

Another idea would be to load these in the getStaticProps, although that won't always be fresh data, e.g. if you just edited the pinned items.

web/pages/home/index.tsx
```diff
@@ -324,6 +347,86 @@ function SearchSection(props: {
...
+
+  async function onSubmit(selectedItems: { itemId: string; type: string }[]) {
+    if (homeConfig == null) return
+    await updateHomeConfig(homeConfig, {
+      pinnedItems: [
+        ...(homeConfig?.pinnedItems ?? []),
+        ...(selectedItems as { itemId: string; type: 'contract' | 'post' }[]),
+      ],
+    })
+    getHomeConfig().then(setHomeConfig)
```
jahooma, line 403: This is ok, but if someone else edited pinnedItems, it would only fetch it after you submit.

You don't have to change it, but a better way to do this is to listen for home config above with a hook. Then you wouldn't need to fetch it again after setting home config.

jahooma (approved):
Cool! Can't wait to see some pinned posts on our home page.

web/pages/home/index.tsx
FRCassarino: Nice!

PR #1025, opened 2022-10-11
New implementation of market card embeddings

FRCassarino:
The previous implementation of market card embeddings worked by embedding a page that showed a market grid via an iframe.

That worked fine, but we had no way of knowing in advance what the height of the iframe should be, so it resulted in either a scrollable grid of market cards, or a grid of market cards with unnecessary padding.

This new implementation embeds the ContractsGrid react component directly, which sidesteps the issue. 

I'm not deleting the old iframe pages just yet to not break existing posts that use them (like Jack's nuclear post), but all new market grid embeds will now use this new implementation.

image

### DIFF
--- a/web/components/editor.tsx
+++ b/web/components/editor.tsx
@@ -22,6 +22,8 @@ import { FileUploadButton } from './file-upload-button'
 import { linkClass } from './site-link'
 import { DisplayMention } from './editor/mention'
 import { DisplayContractMention } from './editor/contract-mention'
+import ReactComponent from './editor/tiptap-grid-cards'
+
 import Iframe from 'common/util/tiptap-iframe'
 import TiptapTweet from './editor/tiptap-tweet'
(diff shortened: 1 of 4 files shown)


web/components/editor.tsx
```diff
@@ -72,6 +74,7 @@ export const editorExtensions = (simple = false): Extensions => [
   DisplayLink,
   DisplayMention,
   DisplayContractMention,
+  ReactComponent,
```
sipec, line 77: i think we should have a more descriptive name for this extension, like TipTapMarketGrid or just MarketGrid

sipec (approved):

jahooma:
Thanks for reviewing @sipec!

web/components/editor.tsx
FRCassarino: Ah missed that, definitely

PR #1033, opened 2022-10-12
Fix some hydration issues

mqp:
I don't know if it matters very much, but I sure as heck am not gonna work on the contract page with the console spewing errors at me about it.

### DIFF
--- a/web/hooks/use-is-client.ts
+++ b/web/hooks/use-is-client.ts
@@ -1,0 +1,7 @@
+import { useEffect, useState } from 'react'
+
+export const useIsClient = () => {
+  const [isClient, setIsClient] = useState(false)
+  useEffect(() => setIsClient(true), [])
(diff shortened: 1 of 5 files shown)


web/hooks/use-is-client.ts
```diff
@@ -0,0 +1,7 @@
+import { useEffect, useState } from 'react'
+
+export const useIsClient = () => {
```
jahooma, line 3: Nice! I'll remember this hook.

jahooma (approved):
Thanks Marshall! I think the latest version of React has gotten more strict with the console errors.

mqp:
> Thanks Marshall! I think the latest version of React has gotten more strict with the console errors.

Yes, I don't think anything broke, I think it's just that React 18 actually produces error messages for problems which 17 was just ignoring.

PR #1040, opened 2022-10-12
Inga/colorful answer comments

ingawei:
Screen Shot 2022-10-12 at 3 40 48 PM

### DIFF
--- a/web/components/contract/contract-tabs.tsx
+++ b/web/components/contract/contract-tabs.tsx
@@ -38,6 +38,8 @@ import TriangleDownFillIcon from 'web/lib/icons/triangle-down-fill-icon'
 import Curve from 'web/public/custom-components/curve'
 import { Answer } from 'common/answer'
 import { AnswerCommentInput } from '../comment-input'
+import { getAnswerColor } from '../answers/answers-panel'
+import { useChartAnswers } from '../charts/contract/choice'
 
 export function ContractTabs(props: {
   contract: Contract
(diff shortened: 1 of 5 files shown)


sipec (approved):

web/components/contract/contract-tabs.tsx
```diff
@@ -139,95 +137,27 @@ const CommentsTabContent = memo(function CommentsTabContent(props: {
...
+        <FreeResponseComments
+          contract={contract}
+          answerResponse={answerResponse}
+          onCancelAnswerResponse={onCancelAnswerResponse}
+          topLevelComments={topLevelComments}
+          commentsByParent={commentsByParent}
+          tips={tips}
+        />
+      )}
+      {contract.outcomeType != 'FREE_RESPONSE' &&
```
jahooma, line 159: If you're using `===` above, might as well use `!==` here. I think it's good to default to the extra equals anyway.

The only differences are roughly `null == undefined` `'1' == 1` evaluate true, but evaluate false with `===`.

web/components/feed/feed-answer-comment-group.tsx
```diff
@@ -1,42 +1,158 @@
...
+  const {
+    contract,
+    answerResponse,
+    onCancelAnswerResponse,
+    topLevelComments,
+    commentsByParent,
+    tips,
+  } = props
+  const answersArray = useChartAnswers(contract).map(
+    (answer, _index) => answer.text
```
jahooma, line 92: No need to have the `_index` param

jahooma (approved):
Thanks Inga!

PR #1045, opened 2022-10-13
Tipping in posts

FRCassarino:


sipec (approved):

common/contract.ts
```diff
@@ -66,6 +67,7 @@ export type Contract<T extends AnyContractType = AnyContractType> = {
   flaggedByUsernames?: string[]
   openCommentBounties?: number
   unlistedById?: string
+  itemType: 'contract'
```
jahooma, line 70: If you're adding this field as a non-optional type, you need to backfill existing contracts to have it by writing a script.

jahooma: I actually don't think this field is necessary. The only place it's used is `likeItem`.

You could simply pass in the itemType to that function. There's no need to store it in the document, it's redundant because we know which collection the document is in.

Another idea is to have functions `likeContract` and `likePost` which call `likeItem` with the appropriate itemType.

common/post.ts
```diff
@@ -13,6 +13,11 @@ export type Post = {
   creatorName: string
   creatorUsername: string
   creatorAvatarUrl?: string
+
+  likedByUserIds?: string[]
+  likedByUserCount?: number
+
+  itemType: 'post'
```
jahooma, line 20: Ditto.

jahooma (approved):
Thanks Fede! This is a highly requested feature.

LGTM, pending the removal the `itemType` field.

common/contract.ts
FRCassarino: Yea I had the script to backfill it setup, was gonna run it before shipping. 

But anyway, yep that makes sense, just updated to pass the itemType along.

PR #1046, opened 2022-10-13
Don't break build on unused import

sipec:


mqp:
I really wish we could make a big red X without failing the Vercel build. If there's no big red X, that just means "people will constantly put in unused imports and then whoever is paying attention gets to remove them." That's frustrating. I would rather be doing something I like instead of cleaning up other people's messes.

Unused imports are actually bad, because our framework chooses how to divide up JS between bundles based on the dependency tree, so if the dependency tree is wrong it will do suboptimal code splitting.

That's my opinion.

sipec:
@mqp ok now "Check PRs / Static analysis (pull_request)" correctly red X's but the dev/prod deploys correctly do not

also people wont have to manually fix unused imports regardless - we have a github action that auto fixes that on pushes to main.

jahooma:
IIRC, changing to warn will disable the auto-remove GitHub action for some reason.

In theory it is better to make it a warning, and then have the gh action automatically fix unused imports, if you can make it work.

jahooma:
I actually don't like the red x. I don't want to even think about imports if a script can just fix it.

sipec:
> IIRC, changing to warn will disable the auto-remove GitHub action for some reason.

that's weird. When I run `yarn lint` in the root dir (which is what lint.yml runs) it correctly removes the unused import

web/components/input.tsx
```diff
@@ -1,6 +1,5 @@
 import clsx from 'clsx'
 import React from 'react'
-import { noop } from 'lodash'
```
sipec: look it works

jahooma (approved):
Thanks, it's a lot better if there's no error from unused imports!

PR #1047, opened 2022-10-14
added cancelling tipping and lil coin

ingawei:
Screen Shot 2022-10-13 at 8 04 33 PM
Screen Shot 2022-10-13 at 8 04 37 PM

### DIFF
--- a/web/components/contract/like-market-button.tsx
+++ b/web/components/contract/like-market-button.tsx
@@ -5,11 +5,12 @@ import { useUserLikes } from 'web/hooks/use-likes'
 import toast from 'react-hot-toast'
 import { formatMoney } from 'common/util/format'
 import { likeContract } from 'web/lib/firebase/likes'
-import { LIKE_TIP_AMOUNT } from 'common/like'
+import { LIKE_TIP_AMOUNT, TIP_UNDO_DURATION } from 'common/like'
 import { firebaseLogin } from 'web/lib/firebase/users'
 import { useMarketTipTxns } from 'web/hooks/use-tip-txns'
 import { sum } from 'lodash'
 import { TipButton } from './tip-button'
+import { TipToast } from '../tipper'
 
 export function LikeMarketButton(props: {
   contract: Contract
@@ -35,8 +36,20 @@ export function LikeMarketButton(props: {
     if (!user) return firebaseLogin()
 
     setIsLiking(true)
-    likeContract(user, contract).catch(() => setIsLiking(false))
-    toast(`You tipped ${contract.creatorName} ${formatMoney(LIKE_TIP_AMOUNT)}!`)
+    const timeoutId = setTimeout(() => {
+      likeContract(user, contract).catch(() => setIsLiking(false))
+    }, 3000)
+    toast.custom(
+      (t) => (
+        <TipToast
+          userName={contract.creatorUsername}
+          onUndoClick={() => {
+            clearTimeout(timeoutId)
+          }}
+        />
+      ),
+      { duration: TIP_UNDO_DURATION }
+    )
   }
 
   return (
--- a/web/components/contract/like-market-button.tsx
+++ b/web/components/contract/like-market-button.tsx
@@ -3,7 +3,6 @@ import { Contract } from 'common/contract'
 import { User } from 'common/user'
 import { useUserLikes } from 'web/hooks/use-likes'
 import toast from 'react-hot-toast'
-import { formatMoney } from 'common/util/format'
 import { likeContract } from 'web/lib/firebase/likes'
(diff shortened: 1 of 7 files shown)


ingawei:
@sinclairity @mqp Looks like you guys were the last people to touch tipping, feel free to take a look and make sure I didn't screw anything up :)

web/components/contract/like-market-button.tsx
```diff
@@ -35,8 +35,20 @@ export function LikeMarketButton(props: {
     if (!user) return firebaseLogin()
 
     setIsLiking(true)
-    likeContract(user, contract).catch(() => setIsLiking(false))
-    toast(`You tipped ${contract.creatorName} ${formatMoney(LIKE_TIP_AMOUNT)}!`)
+    const timeoutId = setTimeout(() => {
+      likeContract(user, contract).catch(() => setIsLiking(false))
+    }, 3000)
```
jahooma, line 40: Did you mean to put TIP_UNDO_DURATION instead of 3000?

web/components/tipper.tsx
```diff
@@ -67,30 +59,75 @@ export function Tipper(prop: {
...
-  useEffect(() => () => void saveTip.flush(), [saveTip])
 
   const addTip = (delta: number) => {
-    setLocalTip(localTip + delta)
-    me && saveTip(me, comment, localTip - myTip + delta)
-    toast(`You tipped ${comment.userName} ${formatMoney(LIKE_TIP_AMOUNT)}!`)
+    setUserTipped(true)
+    const timeoutId = setTimeout(() => {
+      me && saveTip(me, comment, delta)
+    }, 3000)
```
jahooma, line 69: Ditto, did you mean TIP_UNDO_DURATION?

web/components/tipper.tsx
```diff
@@ -19,24 +20,15 @@ export function Tipper(prop: {
 }) {
   const { comment, myTip, totalTip } = prop
 
-  const me = useUser()
-
-  const [localTip, setLocalTip] = useState(myTip)
+  // This is a temporary tipping amount that we can cancel. Used just for rendering :)
+  const [userTipped, setUserTipped] = useState(false)
```
jahooma, line 24: `didTip` might be slightly more recognizable as a boolean...

web/components/tipper.tsx
```diff
@@ -67,30 +59,75 @@ export function Tipper(prop: {
...
         toId: comment.userId,
       })
-    }, 1500)
+    }
   )
-  // instant save on unrender
-  useEffect(() => () => void saveTip.flush(), [saveTip])
 
   const addTip = (delta: number) => {
-    setLocalTip(localTip + delta)
```
jahooma, line 76: Hmm, did this allow for larger tips in one transaction if you clicked multiple times?

Might want to support that use case, especially if you have to wait a few seconds for one tip to go through.

(A side note: I think the info popup appears over the market tip button! Which makes it hard to tip multiple times rapidly.)

jahooma (approved):
Sweet, I like the coin and giving users an undo button is fabulous!

web/components/tipper.tsx
```diff
@@ -67,30 +59,75 @@ export function Tipper(prop: {
...
+          }}
+        />
+      ),
+      { duration: TIP_UNDO_DURATION }
+    )
   }
 
   const canUp =
-    me && comment.userId !== me.id && me.balance >= localTip + LIKE_TIP_AMOUNT
+    me && comment.userId !== me.id && me.balance >= myTip + LIKE_TIP_AMOUNT
```
jahooma, line 84: I think you don't want to add `myTip` here. That's all the previous tips that already went through

web/components/tipper.tsx
```diff
@@ -19,24 +20,15 @@ export function Tipper(prop: {
...
-  useEffect(() => {
-    if (myTip && !initialized.current) {
-      setLocalTip(myTip)
-      initialized.current = true
-    }
-  }, [myTip])
+  const me = useUser()
 
-  const total = totalTip - myTip + localTip
+  const total = totalTip - myTip
```
jahooma, line 28: Delete this line, and pass `totalTip` as a prop below

web/components/contract/like-market-button.tsx
ingawei: no, I figured timeout should be slightly longer than when the toast is on screen to avoid a race condition, but I'll make it not a magic number lol

web/components/tipper.tsx
ingawei: I don't think it did that, but yeah that info popup is really annoying in general for mobile, will get rid of in another PR

PR #1053, opened 2022-10-14
Prevent comments from a bot username from notifying

akrolsmir:


IanPhilips (commented):
The best thing to do would be to move the relevant code into another function called handleCommentOnContract() prior to the notification generation. Then you'd check if the bot user name is included so that you could skip all the logic before your check, but it looks like it'll work as is. At least you could put the check before all the unnecessary notif generation logic.

IanPhilips:
I will fix this in a separate commit

IanPhilips:
e158a1c5d9ca647ac8ca8f33f7c3b52237a5ac09 does this, but I recently saw v commenting on a market, so maybe we should just exclude manifold dream? https://manifold.markets/SG/will-elon-musk-buy-twitter-this-yea#VbOgpaneOKAaBZqRnyGI

PR #1055, opened 2022-10-14
Refactoring

jahooma:


akrolsmir:
Thanks for cleaning up our codebase, investing in making our code easy to navigate pays dividends in the future!

I have a (weak) aesthetic preference for smaller refactors where possible (aka commit-by-commit instead of a single large PR like this). Partly that's just general arguments in favor of small-cls (revertability, easier to read diffs, fewer mistakes) and partly refactoring specifically generates a lot of commit churn, where it's harder to follow the git history.

jahooma:
> Thanks for cleaning up our codebase, investing in making our code easy to navigate pays dividends in the future!
> 
> I have a (weak) aesthetic preference for smaller refactors where possible (aka commit-by-commit instead of a single large PR like this). Partly that's just general arguments in favor of small-cls (revertability, easier to read diffs, fewer mistakes) and partly refactoring specifically generates a lot of commit churn, where it's harder to follow the git history.

I see, I guess you're right. Just saw this. May do so going forward!

PR #1063, opened 2022-10-17
Fix issues with chart y-axis zooming

mqp:
This was super suspicious before -- because it was mutating props without using React state, there were weird interaction bugs.

Now it's pretty decent, really the only problem is that we should write some clever formatting code so that the y-axis displays an appropriate amount of significant figures on the ticks for the range of values being displayed. (EDIT: I just addressed this by making it so that you can't zoom in tighter than 4% on percentage graphs or 10 bucks on portfolio graphs.)


mqp:
OK, IMO this y-axis zoom now basically works nicely in every case.

mqp:
@FRCassarino you're lucky I think this is a nice feature or I would be annoyed at you tossing such a half assed version over the fence so that I had to fix all the details! But it is a nice feature and it's much better than not having it.

PR #1064, opened 2022-10-17
Use new React `useId` hook in chart code

mqp:
This is new in React 18 and is a nicer solution than the `nanoid` version. It also works with SSR, which is nice.


jahooma (approved):
LGTM! 

Love this new hook. I was thinking about how `useId()` combines well as the key for `usePersistentState`...

[Unrelated] I also don't know if you saw but I created an in-memory storage option for `usePersistentState`: https://github.com/manifoldmarkets/manifold/commit/4e5b78f4ee5d34923a2928d917e81e078f49a102

PR #1067, opened 2022-10-18
Clean up contract comment stuff

mqp:
Just tidying up and fixing the performance here to avoid re-rendering comments on subscription load. (See similar #1000.)

### DIFF
--- a/web/components/feed/feed-comments.tsx
+++ b/web/components/feed/feed-comments.tsx
@@ -38,8 +38,19 @@ export function FeedCommentThread(props: {
   const { contract, threadComments, tips, parentComment } = props
   const [replyTo, setReplyTo] = useState<ReplyTo>()
   const [seeReplies, setSeeReplies] = useState(true)
-
+  const [highlightedId, setHighlightedId] = useState<string>()
   const user = useUser()
+
+  const router = useRouter()
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


web/components/feed/feed-comments.tsx
```diff
@@ -38,8 +38,20 @@ export function FeedCommentThread(props: {
   const { contract, threadComments, tips, parentComment } = props
   const [replyTo, setReplyTo] = useState<ReplyTo>()
   const [seeReplies, setSeeReplies] = useState(true)
-
+  const [highlightedId, setHighlightedId] = useState<string>()
   const user = useUser()
+
+  const router = useRouter()
+  useEffect(() => {
+    const parts = router.asPath.split('#')
```
jahooma, line 46: I think this is a tricky case. Essentially, you need to check the route path after `router.isReady` is true.

You can't reproduce any problem with this in localhost, but in prod, getStaticProps pages will render first with router.isReady = false and the `asPath` missing any query params. This is because it is loading the same static page for everyone and there could be different query params.

See the third Caveat here: https://nextjs.org/docs/advanced-features/automatic-static-optimization#caveats

mqp: Yep, I think you're right. I tend to think to myself that the URL should be available immediately both on server and client and forget about the SSG case.

PR #1095, opened 2022-10-25
Fix some small bugs mostly related to stats charts

mqp:
As context, the stats chart is unusual because the last data point is exactly at the end of the chart range. We don't extend the chart out to "now."

1. The tooltip code wasn't working right when you hovered over a data point that was at the very right edge, because `interpolateY` wasn't handling the case where `x0 === x1`.
2. The y-axis scale-on-zoom code wasn't correctly taking the last data point into account when you zoomed into a time range including the last data point.
3. For most of our charts, it makes conceptual sense for the tooltip to show the "last data point that happened before time T." But this works very poorly on the stats charts, since the data points are the metric computations, and they aren't actually happening at the time they are for. It's also very hard to see the last data point in the tooltip since it's only at the very right edge! It makes more sense on the stats chart, when you hover over it, to show the nearest data point to time T.

### DIFF
--- a/web/components/charts/contract/binary.tsx
+++ b/web/components/charts/contract/binary.tsx
@@ -31,14 +31,15 @@ const getBetPoints = (bets: Bet[]) => {
 }
 
 const BinaryChartTooltip = (props: TooltipProps<Date, HistoryPoint<Bet>>) => {
-  const { data, x, xScale } = props
+  const { prev, x, xScale } = props
   const [start, end] = xScale.domain()
   const d = xScale.invert(x)
+  if (!prev) return null
   return (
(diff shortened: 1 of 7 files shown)


jahooma (commented):
I can't say I follow all the code, but thanks for looking into this and fixing. LGTM

mqp:
> I can't say I follow all the code, but thanks for looking into this and fixing. LGTM

It's basically the code that says, given that you are hovering over some pixel with the mouse, where should we render the little blue line (i.e. what is the y-coordinate at that point) and what should we show in the tooltip (i.e. what data points are related to where you are hovering.) So the math is fiddly.

PR #1098, opened 2022-10-26
Fix scrollbar on market page

jahooma:


jahooma:
@ingawei Sorry, had to undo this change to perfect the highlight rectangle, as it was causing issues triggering horizontal and vertical scrollbar on mobile.

PR #1100, opened 2022-10-26
Split update-metrics into three separate functions

mqp:
WIP. This breaks up some of the work at the cost of creating more total work between all three (they more or less all need to load ~all contracts and ~all bets -- only ones in groups for `updateGroupMetrics`), but now we can go to town making the individual parts more efficient since they don't depend on shared "load the whole world" data structures.


functions/src/update-contract-metrics.ts
```diff
@@ -0,0 +1,98 @@
...
+const firestore = admin.firestore()
+export const scheduleUpdateContractMetrics = functions.pubsub
+  .schedule('every 15 minutes')
+  .onRun(async () => {
+    try {
+      console.log(await invokeFunction('updatecontractmetrics'))
+    } catch (e) {
+      console.error(e)
+    }
+  })
```
jahooma, line 27: What do you think about the idea of having one scheduler function that calls all three of the update cloud functions?

It's nice that currently the scheduled function is co-located, but it might be simpler overall to have one scheduler process. I guess I could go either way here.

jahooma (approved):
Thanks Marshall! This looks like a clean way to split up `updateMetrics`. Will be a lot more maintainable.

Code all looks correct. LGTM!

I'm ok with the cost increase, especially if we'll be able to shortly decrease it through some optimizations.

mqp:
I further adjusted this to not load all bets at once in `updateContractMetrics` and instead load them in batches to satisfy a set of contracts, because the efficiency loss should be minimal but the decreased memory pressure is huge.

mqp:
Hmm, I have a bug to fix here before merging.

mqp:
OK, I fixed a few things:

- A couple efficiency fixups.
- Using `batchedWaitAll` to do less total stuff at once, relieving memory pressure.
- Fixed a bug in `updateUserMetrics`. We need to load all contracts the user bet on, not just all contracts they created. Since there is a huge amount of overlap between users, it's best to load them all up front.

mqp:
Merging and deploying this one, I am not going to 100% guarantee there are no bugs but it seems to be working to me.

PR #1103, opened 2022-10-27
More bots for the list

NcyRocks:
Added a few more bots, some outdated.


jahooma (approved):
Thanks for the additions!

PR #1107, opened 2022-10-27
Blocking content

IanPhilips:
You can block users (and be blocked by them). 

If you block them:
- You won't see their markets from search and homepage, you won't see their comments and they won't see your comments, and they won't be able to make new comments nor submit new answers on your markets. You won't get notified when they tag you in comments or markets. Their profile will look blank to you. 

If they block you:
- You won't see their markets from search and homepage, you won't see their comments and they won't see your comments, and you won't be able to make new comments on their markets. 

You can hide groups and hide individual markets to remove them from your search & homepage results.
 
The hide group button isn't styled red bc it seems less eye-catching and you might be on that page more than another's user page and market info dialog. I actually think it might belong back in the 'About' tab bc for example the destiny.gg folks might live in their group and they'd rather not be prompted to hide the group all the time they're there.

Screen Shot 2022-10-27 at 4 37 48 PM
Screen Shot 2022-10-27 at 4 37 41 PM
Screen Shot 2022-10-27 at 4 38 20 PM

Screen Shot 2022-10-27 at 4 59 16 PM

Screen Shot 2022-10-27 at 4 54 24 PM


jahooma (approved):
Impressive work, Ian! I think users will appreciate this feature (or set of blocking features!).

Looks like a reasonable implementation too. And thanks for the screenshots!

Couple points of feedback:
- The red block button on the user's page would probably be better behind a '...' menu or some advanced setting. It's not that nice to see a red block button.
- The Hide button for markets might not be clear. Oh wait, I see you've changed it to say Block market. That seems clearer.
- I think it's hard to fit the hide button in the group header on mobile. For long group names, there's not much room. I think putting the block button in About is a good solution.

This is cool. I might add a way to choose more groups to block from the homepage after this merges.

IanPhilips:
Thanks for the review and suggestions, James! Fixed it up real nice now.

PR #1109, opened 2022-10-28
Ingawei/man 100 iterate on home page design

ingawei:
image


web/components/bet/quick-bet.tsx
```diff
@@ -150,57 +154,137 @@ export function QuickBet(props: {
...
+  const {
+    onClick,
+    direction,
+    user,
+    contract,
+    hover,
+    onMouseEnter,
+    onMouseLeave,
+  } = props
+  const userBets = useUserContractBets(user.id, contract.id)
```
jahooma, line 202: This hook unfortunately fetches new data every time.

Maybe just pass in `userBets` since the parent has it.

web/components/bet/quick-bet.tsx
```diff
@@ -150,57 +154,137 @@ export function QuickBet(props: {
...
+    )
+    position = yesWinnings - noWinnings
+    outcome = position < 0 ? 'NO' : 'YES'
+  }
+  let hasInvestment = false
+  if (direction === 'UP') {
+    hasInvestment = outcome === 'YES' && invested != null && floor(invested) > 0
+  } else {
+    hasInvestment = outcome === 'NO' && invested != null && floor(invested) > 0
+  }
```
jahooma, line 227: I think you can get this with:
 `const { invested } = getContractBetMetrics(contract, userBets)`
and
` const { hasYesShares, hasNoShares } = useSaveBinaryShares(contract, userBets)`

web/components/contract/contract-card.tsx
```diff
@@ -378,80 +382,78 @@ export function ContractCardProbChange(props: {
...
+    const excludeSales = bets.filter((b) => !b.isSold && !b.sale)
+    yesWinnings = sumBy(excludeSales, (bet) =>
+      calculatePayout(contract, bet, 'YES')
+    )
+    noWinnings = sumBy(excludeSales, (bet) =>
+      calculatePayout(contract, bet, 'NO')
+    )
+    position = yesWinnings - noWinnings
+    outcome = position < 0 ? 'NO' : 'YES'
+  }
```
jahooma, line 413: I would also recommend `useSaveBinaryShares` here.

The other benefit of this function is the data is cached to local storage, which makes for a much better loading experience.

jahooma (approved):
Cool, thanks for iterating, Inga! LGTM.

I think it's nice to see your total invested next to the quick arrows.

PR #1113, opened 2022-10-28
Add even more bots

diomidov:


sipec (commented):
hmm these are all run by humans though

PR #1121, opened 2022-10-31
Refactor loan code a little bit

mqp:
This is just some cleanup subsequent to the changes I made in #1100. There's no efficiency advantage for the code in `loans.ts` to run the loan computations in a batch, so it's cleaner for its interface to not be batched.


jahooma (approved):
LGTM!

Might just merge this, so I can avoid conflicts switching to v2 function that is called by a scheduled function.

PR #1122, opened 2022-10-31
Make `updateGroupMetrics` use precomputed contract metrics

mqp:
WIP. Now `updateGroupMetrics` doesn't have to load all bets, so it's basically good enough for now.


mqp:
Well, there's an issue here, which is that the new numbers this is producing don't match the numbers the current group computation is producing. What that means there's an issue with, I don't know.

mqp:
OK, I fixed everything up, and now the numbers match.

PR #1124, opened 2022-10-31
clicking comment automatically switches tabs

ingawei:


ingawei:
@jahooma This is a short boi but I feel as if I might've done it in a hacky way, would love any suggestions to make it cleaner if you have them :)

jahooma:
> @jahooma This is a short boi but I feel as if I might've done it in a hacky way, would love any suggestions to make it cleaner if you have them :)

Cool, I see how this works. I would say this is not exactly the best pattern haha.

React is often about where you store state in the hierarchy. Usually when there's a tricky case like this it means you need to move the state up into the parent(s).

In fact, the old `Tabs` implementation got this wrong by managing the active tab state itself (which is why it's deprecated), and you should try using `ControlledTabs`, which has you pass in the active tab.

This case is particularly tricky because I think you need to have the active tab state all the way up in the ContractPageContent component, and pass it down (along with the setter) to ContractTabs and then ControlledTabs. I'd give that a shot and see if you think it's cleaner.

ingawei:
> > @jahooma This is a short boi but I feel as if I might've done it in a hacky way, would love any suggestions to make it cleaner if you have them :)
> 
> Cool, I see how this works. I would say this is not exactly the best pattern haha.
> 
> React is often about where you store state in the hierarchy. Usually when there's a tricky case like this it means you need to move the state up into the parent(s).
> 
> In fact, the old `Tabs` implementation got this wrong by managing the active tab state itself (which is why it's deprecated), and you should try using `ControlledTabs`, which has you pass in the active tab.
> 
> This case is particularly tricky because I think you need to have the active tab state all the way up in the ContractPageContent component, and pass it down (along with the setter) to ContractTabs and then ControlledTabs. I'd give that a shot and see if you think it's cleaner.

@jahooma you right this is so much better lol. Let me know if you still see any issues!

jahooma (approved):
Great! Looks good, Inga!

PR #1125, opened 2022-10-31
Update metrics for resolved contracts also

mqp:
This is necessary for a variety of future work, incl. #1122.

It would be better for it to only update metrics for resolved contracts exactly once, after resolution; we will probably add this complication soon.


jahooma (approved):
Thank, Marshall!

PR #1126, opened 2022-10-31
Make type styles more consistent

sipec:
- comments in old format (string) vs json are now the same size (14px)
- @mention and link now are now the same font weight (300px)
- post editor now has between-paragraph space, just like posts
- comment editor now same font size as comments

@ingawei I don't feel opinionated about any of these changes, I just think it should be consistent. To change the styles just change the `proseClass` function in editor.tsx. Previously the styles were split across multiple components in some cases but now its more consolidated

This adds a size property to `Content` and `useTextEditor` to make styling easier and more consistent.
@mantikoros now posts are just `Content` with `size='lg'` rather than `PostContent`

### DIFF
--- a/web/components/activity-log.tsx
+++ b/web/components/activity-log.tsx
@@ -178,11 +178,7 @@ export const CommentLog = memo(function FeedComment(props: {
           <RelativeTimestamp time={createdTime} />
         </div>
       </Row>
-      <Content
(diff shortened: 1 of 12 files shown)


ingawei (approved):
Looks good to me! Thank you Sinclair :)

PR #1128, opened 2022-10-31
Make `updateContractMetrics` only look at recent bets

mqp:
Now this only has to look at bets for the window of time it's computing probability changes for (one month atm), and unfilled limit bets for the purposes of computing elasticity.


jahooma (approved):
Thanks, looks like this improves code quality too! LGTM

PR #1130, opened 2022-11-01
made buttons for graph

ingawei:
Screen Shot 2022-10-31 at 7 46 15 PM

### DIFF
--- a/web/components/portfolio/portfolio-value-section.tsx
+++ b/web/components/portfolio/portfolio-value-section.tsx
@@ -1,18 +1,21 @@
 import clsx from 'clsx'
 import { formatMoney } from 'common/util/format'
 import { last } from 'lodash'
-import { memo, useState } from 'react'
+import { memo, ReactNode, useState } from 'react'
 import { usePortfolioHistory } from 'web/hooks/use-portfolio-history'
 import { Col } from '../layout/col'
 import { Row } from '../layout/row'
 import { GraphMode, PortfolioGraph } from './portfolio-value-graph'
 import { SizedContainer } from 'web/components/sized-container'
+import { Period } from 'web/lib/firebase/users'
+import { useEvent } from 'web/hooks/use-event'
+import PlaceholderGraph from 'web/lib/icons/placeholder-graph'
 
 export const PortfolioValueSection = memo(
   function PortfolioValueSection(props: { userId: string }) {
     const { userId } = props
-
-    const portfolioHistory = usePortfolioHistory(userId, 'allTime')
+    const [currentTimePeriod, setCurrentTimePeriod] = useState<Period>('weekly')
+    const portfolioHistory = usePortfolioHistory(userId, currentTimePeriod)
     const [graphMode, setGraphMode] = useState<GraphMode>('profit')
     const [graphDisplayNumber, setGraphDisplayNumber] = useState<
       number | string | null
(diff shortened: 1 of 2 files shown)


mqp:
1. Wouldn't it have been nicer to put these inside the box of the graph, maybe at the top right?
2. Why put these on the portfolio page but not on all graphs?

ingawei:
> 1. Wouldn't it have been nicer to put these inside the box of the graph, maybe at the top right?
> 2. Why put these on the portfolio page but not on all graphs?

1. I think that'll lead to a lot of complications with the tooltip, also having it closer to the X axis makes sense since that's what it's modifying
2. I forget who but someone mentioned it might make sense for portfolio, and not markets. Markets tend to have a shorter lifetime/I imagine there's more reason for people to be interested in the long term patterns of a market, whereas most people want to see short term portfolio changes.

web/lib/icons/placeholder-graph.tsx
```diff
@@ -0,0 +1,18 @@
+// from Feather: https://feathericons.com/
+export default function PlaceholderGraph(props: React.SVGProps<SVGSVGElement>) {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      // width="18"
+      // height="18"
```
jahooma, line 7: Remove?

web/components/portfolio/portfolio-value-section.tsx
```diff
@@ -1,111 +1,256 @@
...
     if (!portfolioHistory || !lastPortfolioMetrics) {
-      return <></>
+      return (
+        <PortfolioValueSkeleton
+          graphMode={graphMode}
+          onClickNumber={onClickNumber}
+          currentTimePeriod={currentTimePeriod}
+          setCurrentTimePeriod={(timePeriod: Period) =>
+            setCurrentTimePeriod(timePeriod)
+          }
```
jahooma, line 40: Could just be `setCurrentTimePeriod={setCurrentTimePeriod}`

web/components/portfolio/portfolio-value-section.tsx
```diff
@@ -1,111 +1,256 @@
...
-          )}
-        </SizedContainer>
-      </>
+      <PortfolioValueSkeleton
+        graphMode={graphMode}
+        onClickNumber={onClickNumber}
+        currentTimePeriod={currentTimePeriod}
+        setCurrentTimePeriod={(timePeriod: Period) =>
+          setCurrentTimePeriod(timePeriod)
+        }
```
jahooma, line 75: Ditto

web/components/portfolio/portfolio-value-section.tsx
```diff
@@ -1,111 +1,256 @@
...
+        </Row>
+      </Row>
+      <SizedContainer fullHeight={200} mobileHeight={100}>
+        {graphElement}
+      </SizedContainer>
+      <PortfolioTimeSelection
+        currentTimePeriod={currentTimePeriod}
+        setCurrentTimePeriod={(timePeriod: Period) =>
+          setCurrentTimePeriod(timePeriod)
+        }
```
jahooma, line 186: Ditto

jahooma (approved):
Cool, thanks Inga! LGTM.

Nice job with the skeletons. Code also looks good.

I do think the portfolio graph is the most important case to have the buttons and shorter default time range.

But it seems like the time range buttons would be valuable for the market page graphs too. Probably "All" is the best default for them, but I imagine many want to jump into the last 24 hours or 7 days to examine activity in detail, etc.

On mobile, there currently isn't a way to do that. And drag or pinch-to-zoom will never be that obvious.

I'm ok with that being a followup change though!

PR #1131, opened 2022-11-01
Some easy cleanups on `updateUserMetrics`

mqp:
It's going to be a slog to make this actually efficient, but we can easily do some small things.

1. We don't need to reload contracts for each user, we already loaded them up front.
2. Firestore's `BulkWriter` seems more effective at getting updates out over the wire than making `WriteBatch`es directly. (Doing all the writes actually takes a small majority of the time of the current version, so it's worth optimizing.)


mqp:
Getting this in so I can see the performance of `BulkWriter` compared to self-assembling `WriteBatch`es in production.

PR #1132, opened 2022-11-01
Use `BulkWriter` in other metrics update functions

mqp:
This is definitely much better than the previous approach, so let's do it.

I would have changed `updateLoans` too but it's a whole other ball of wax to be tackled later.


jahooma (approved):
Sweet! Seems like a better tool. LGTM

PR #1136, opened 2022-11-01
reset graph on time period change

ingawei:


ingawei:
another short boi but would like a check on code cleanliness please :)

web/components/charts/generic-charts.tsx
```diff
@@ -312,28 +319,46 @@ export const MultiValueHistoryChart = <P extends MultiPoint>(props: {
...
   margin: Margin
   xScale: ScaleTime<number, number>
   yScale: ScaleContinuousNumeric<number, number>
+  viewXScale: ScaleTime<number, number, never> | undefined
+  setViewXScale: Dispatch<
+    SetStateAction<ScaleTime<number, number, never> | undefined>
+  >
+  viewYScale: ScaleContinuousNumeric<number, number, never> | undefined
+  setViewYScale: Dispatch<
+    SetStateAction<ScaleContinuousNumeric<number, number, never> | undefined>
```
jahooma, line 338: What do you think about bundling these 4 props within one object? e.g. `viewScaleProps`

You could export the type for this object from this file as well. Could make it easier to read when passed through other components.

jahooma (approved):
Great! Cool to see you applying this pattern in another place.

I think this is right. It's a conceptually simple way to fix the bug.

PR #1140, opened 2022-11-02
move y axis on right side of graph

ingawei:
image


ingawei:
@mqp please let me know if there was a better way I could've done this

mqp (approved):
> @mqp please let me know if there was a better way I could've done this

Nope, looks like 100% the way I would have done it.

PR #1141, opened 2022-11-02
fixed gap in nav bar, also profile now highlights

ingawei:
image


web/components/nav/bottom-nav-bar.tsx
```diff
@@ -50,36 +60,21 @@ export function BottomNavBar() {
...
-            noLink
-          />
-        </NavBarItem>
-      )}
+          user={user}
+        />
+      ))}
       <div
-        className="w-full select-none py-1 px-3 text-center hover:cursor-pointer hover:bg-indigo-200 hover:text-indigo-700"
+        // className="w-full select-none py-1 px-3 text-center hover:cursor-pointer hover:bg-indigo-200 hover:text-indigo-700"
```
jahooma, line 76: Remove?

web/components/nav/bottom-nav-bar.tsx
```diff
@@ -50,36 +60,21 @@ export function BottomNavBar() {
     return null
   }
 
-  const navigationOptions =
-    user === null ? signedOutNavigation : getNavigation()
+  const navigationOptions = !user ? signedOutNavigation : getNavigation(user)
```
jahooma, line 63: This line is different when `user` is `undefined`. (Which is often true on the first render.)

I think it might be better to default to the signed in navigation since that is most common.

jahooma (approved):
Cool, looks good! I like the new order with your profile as the middle option.

PR #1146, opened 2022-11-03
Add caching for daily/monthly/weekly traders

FRCassarino:


functions/src/on-create-bet.ts
```diff
@@ -205,6 +206,13 @@ const updateUniqueBettorsAndGiveCreatorBonus = async (
         trans.update(contractDoc, {
           uniqueBettorIds: newUniqueBettorIds,
           uniqueBettorCount: newUniqueBettorIds.length,
+          cachedUniqueBettorCount: [
+            ...(previousCachedUniqueBettorCount ?? []),
+            {
+              time: Date.now(),
+              value: newUniqueBettorIds.length,
+            },
+          ],
```
jahooma, line 215: This is a creative strategy to store the unique bettors over time.

However, we generally want to store as little data on the contract as possible. The most popular contracts get hundreds unique of bettors already.

Moving this to a subcollection is a good idea.

The other way I can think of doing this is to directly count the bets from unique users within each time period.
So `updatecontractmetrics` could do these counts for the last day, week, month for each contract (since it has all bets loaded, this shouldn't be a problem).
And then `updateusermetrics` would add those up, for each period, for all contracts a user created.
The benefit of this approach is no backfill is necessary!

jahooma (commented):
Cool, thanks for diving into this task! It's a bit harder than I realized haha.

I left a comment suggesting to either move the unique better counts array to a subcollection, or try a different approach of caching the period unique bettor counts on each contract.

PR #1150, opened 2022-11-03
Remove inline betting in embeds

sipec:
just cleaning some stuff up while refactoring how we bet

even if we want some easy way for people to bet from the embed, we ought to use the new quick bet interface.


jahooma (approved):
LGTM!

PR #1154, opened 2022-11-04
Merge twitch bot

PhilBladen:
Merged Twitch bot codebase into twitch-bot directory in the root of the Manifold monorepo. History of both repos should be preserved, but please check I've not done something stupid.


jahooma (approved):
Thanks Phil! Happy to merge this into our repo.

Over time, we can try to make this code more uniform with the rest of our codebase, dedup types and components, use the same prettier formatting, etc. But for now, this is fine.

Thanks for your work so far!

PR #1158, opened 2022-11-05
Clean up additional mana options, add to modal

sipec:
- Now clicking on the cards redirects you (if that makes sense)
- The options have also been added to the add funds modal on a separate tab
  - Add funds modal is adjusted on iOS, similar to add funds page
Screen Shot 2022-11-04 at 9 40 19 PM
Screen Shot 2022-11-04 at 9 40 32 PM
erate tab


IanPhilips (approved):
I like it! Thanks for sprucing up this page, Sinclair!

PR #1160, opened 2022-11-05
Filter markets you've swiped on already

sipec:


web/hooks/use-user-bets.ts
```diff
@@ -59,3 +61,13 @@ export const useGetUserBetContractIds = (userId: string | undefined) => {
 
   return contractIds
 }
+
+export const useUserSwipes = () => {
+  const user = useUser()
+  const [swipes, setSwipes] = useState<string[]>()
```
jahooma, line 67: A simple improvement would be to switch `useState` for

```
usePersistentState<string[] | undefined>(undefined, {
    store: inMemoryStore(),
    key: `user-swipes-${user.id}`,
  })
```
Which would keep the data loaded in case they navigate back.

web/lib/firebase/api.ts
```diff
@@ -78,6 +78,10 @@ export function sellBet(params: any) {
   return call(getFunctionUrl('sellbet'), 'POST', params)
 }
 
+export function logSwipe(params: any) {
```
jahooma, line 81: I would put the intended type here.

functions/src/log-swipe.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return await firestore.runTransaction(async (trans) => {
+    const contractDoc = firestore.doc(`contracts/${contractId}`)
+    const userDoc = firestore.doc(`private-users/${auth.uid}`)
+    const [contractSnap, userSnap] = await trans.getAll(contractDoc, userDoc)
+    if (!contractSnap.exists) throw new APIError(400, 'Contract not found.')
+    if (!userSnap.exists) throw new APIError(400, 'User not found.')
+
+    const ref = firestore
+      .collection(`/private-users/${auth.uid}/seenMarkets`)
```
jahooma, line 26: I'm not sure if this should be called seen markets unless there are plans to log visits to markets here as well?

Actually, I like this collection, keyed by contract id for general use.

It could be reasonable to set `visited: true` and `lastVisitTime` on the doc `seenMarkets/{contractId}` if you navigate to a market page (or create the doc if it doesn't exist).

Either plan for this broader use case, or change the collection name to `swipes`.

jahooma (approved):
Left few comments, otherwise LGTM!

functions/src/log-swipe.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const firestore = adminFirestore()
+
+const bodySchema = z.object({
+  contractId: z.string(),
+  amount: z.number().optional(),
+  outcome: z.enum(['YES', 'NO', 'SKIP']),
+})
+
+export const logswipe = newEndpoint({}, async (req: any, auth: AuthedUser) => {
```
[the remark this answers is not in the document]
sipec: Hmm this literally did not occur to me. Yes, doing it clientside makes wayy more sense

functions/src/log-swipe.ts
sipec: leaning towards broader use case. though you don't need `visited: true`, just the timestamp will do

PR #1162, opened 2022-11-05
Add senate races

LivInTheLookingGlass:
Note the Oklahoma market is a derivative of the markets for each individual seat


mantikoros (approved):
LGTM. Thanks for your help!

PR #1165, opened 2022-11-07
Fix market card for resolved pseudonumeric markets

diomidov:
The market cards for resolved pseudonumeric markets were displaying "Resolved as X%" instead of "Resolved as Y".

This PR fixes this, and also cleans up related code in bet feed (Please don't nest so many ternaries! :scream: Especially when there already is a helper function that does exactly that)


jahooma (approved):
Thanks Yev! This looks correct and simplifies things.

Could I bother you to also change the `getMappedValue` and `getFormattedMappedValue`  functions not to be curried? I.e. instead of
`getMappedValue = (contract: Contract) => (p: number)`
Do
`getMappedValue = (contract: Contract, p: number)`

PR #1168, opened 2022-11-07
Certs on Manifold

akrolsmir:
See https://manifoldmarkets.notion.site/Certs-4319707df7be43489b0f0620ed095511


akrolsmir:
Mostly looking for a review on the way the data is structured here, through a log of Txns rather than Bets; and secondarily a review of the Uniswapv2 mechanism -- thanks @mantikoros @mqp!

The UI will continue to change a lot, so don't worry too much about it

common/calculate/cert.ts
```diff
@@ -0,0 +1,60 @@
...
+  // 3) Pay out that much M$ to each holder (assume that all pool shares belong to cert creator)
+  const ownership = getCertOwnershipUsers(providerId, txns)
+  const totalShares = sum(Object.values(ownership))
+  const dividendPerShare = totalDividend / totalShares
+  const payouts = Object.entries(ownership).map(([ownerId, shares]) => ({
+    userId: ownerId,
+    payout:
+      shares * dividendPerShare -
+      // Set a negative total payout for the provider
+      (ownerId === providerId ? totalDividend : 0),
```
mantikoros, line 57: What if the owner and dividend payer are different people? Might be easier to deduct dividend payment in a different transaction.

common/calculate/uniswap2.ts
```diff
@@ -0,0 +1,35 @@
...
+export function afterSwap(
+  pool: { [outcome: string]: number },
+  token: 'M$' | 'SHARE',
+  amount: number
+) {
+  const k = pool['M$'] * pool['SHARE']
+  const other = token === 'M$' ? 'SHARE' : 'M$'
+  return {
+    [token]: pool[token] + amount,
+    [other]: k / (pool[token] + amount),
```
mantikoros, line 26: Check for overflow/underflow if pool[token] + amount is close to 0.

common/contract.ts
```diff
@@ -107,6 +111,21 @@ export type CPMM = {
   }
 }
 
+export type Uniswap2 = {
+  mechanism: 'uniswap-2'
+  // outcome can be e.g. 'M$' or a 'SHARE'
+  pool: { [outcome: string]: number }
```
mantikoros, line 117: Why not just use fields for shares and mana? Multivariable case probably is not very common.

common/contract.ts
```diff
@@ -107,6 +111,21 @@ export type CPMM = {
...
+  // Unused; just for making Typescript happy, because code paths
+  // assume this thing falls under DPM-2
```
mantikoros, line 119: There must be a way to get typescript to not complain. Adding an extra field is bad.

mantikoros (approved):
Always check for pool overflow with Uniswap (ie make sure each item in the pool stays above some epsilon value), but otherwise LGTM.

common/contract.ts
akrolsmir: Yeah, will look into getting rid of the field!

common/contract.ts
akrolsmir: Wanted to leave open the mechanism in the future to eg swap two different shares/coins, I think? But maybe YAGNI. I don't think it matters much either way, here.

PR #1169, opened 2022-11-08
Ingawei/man 103 design a new signed out home page

ingawei:
image
- made animation thingies
- added quick bet back to signed out page, clicking prompts sign in


jahooma:
This is great, Inga! The animation looks really nice and the sections of content are much better than before. It's cool that it works on mobile as well.

Here are my thoughts on how to improve it:
- Might be too many sections. Also, I don't think featured and latest posts are that great. I would axe both of those. Then it's pretty simple and clean with Trending + Live.
- Might want another Get started button at the bottom of the page?
- The current Get started button doesn't stand out that much? Maybe if there were more blank space around it, or the light blue background around it was even lighter for more contrast.
- Even though the second slide says "play money", I don't know if most people will catch that. Maybe the `get M$500` should say `get M$500 play money to start predicting`?
- Should the Profit arrow be green instead of red?

web/components/landing-page-panel.tsx
```diff
@@ -1,64 +1,451 @@
...
+
+  useEffect(() => {
+    if (currTimeoutId) {
+      clearTimeout(currTimeoutId)
+    }
+    const newTimeoutId = setTimeout(
+      () => setPageNumber(getNextPageNumber(pageNumber)),
+      5500
+    )
+    setCurrTimeoutId(newTimeoutId)
```
jahooma, line 68: I think a cleaner way to clear the timeout is to return a cleanup function for `useEffect` instead of saving `currTimeoutId` as state:

```
return () => clearTimeout(newTimeoutId)
```

web/components/landing-page-panel.tsx
```diff
@@ -1,64 +1,451 @@
...
+          <LandingPageManifoldMarketsLogo isMobile={isMobile} />
+          <div
+            className="absolute bottom-16 right-8 z-30 md:right-12"
+            onMouseEnter={() => setIsButtonHovered(true)}
+            onMouseLeave={() => setIsButtonHovered(false)}
+          >
+            <Button
+              className={clsx(
+                'absolute z-10 transition-all ease-in-out focus:bottom-0.5 focus:-left-0.5',
+                isButtonHover ? 'bottom-2 -left-2' : 'bottom-1.5 -left-1.5'
```
jahooma, line 141: I believe you can accomplish this hover effect with css. This is desirable because javascript listeners are slower and more laggy.

https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state

In short, mark the parent div with the `group` class, then add the `group-hover:` prefix to classes that should apply on hover.

web/components/landing-page-panel.tsx
```diff
@@ -1,64 +1,451 @@
...
+}
+
+export function LandingPage0(props: { isMobile: boolean }) {
+  const { isMobile } = props
+  const text = 'Ask any question'
+  return (
+    <>
+      <div
+        className={clsx(
+          'sm: absolute z-10 text-xl text-white',
```
jahooma, line 184: Remove `sm:`? Don't think it's doing anything.

web/components/landing-page-panel.tsx
```diff
@@ -1,64 +1,451 @@
...
+          'font-major-mono sm:text-greyscale-7 text-sm lowercase text-white sm:whitespace-nowrap'
+        )}
+      >
+        Manifold Markets
+      </div>
+    </Row>
+  )
+}
+
+export function LandingPage1(props: { isMobile: boolean }) {
```
jahooma, line 269: Impressive effort!

jahooma (approved):
Thanks Inga! The code looks pretty good, nice work!

Left a couple comments.

web/components/landing-page-panel.tsx
ingawei: Oh this is really helpful!

web/components/landing-page-panel.tsx
ingawei: :)

PR #1171, opened 2022-11-08
Efficient user metrics

mqp:
Now all the work involving bets in `updateUserMetrics` only needs bets on unresolved markets, instead of all bets ever, so that's a lot better.

- You can only have outstanding loans on unresolved markets.
- We only need to compute the "investment value" for unresolved markets, since the investment value for resolved markets is 0.
- We only need to compute contract metrics for unresolved markets, if we make sure we compute it once at resolution, since they can't change post-resolution.


mqp:
Oh damn I just realized I made a dumb mistake with this one. Closing until I can figure out what to do about it.

mqp:
OK, fixed it up. I have to rely on `uniqueBettorIds` to get "all bets you have made on unresolved markets" but that should be OK.

PR #1172, opened 2022-11-08
Make `updateUserMetrics` only consider bets on unresolved markets

mqp:
GitHub won't let me reopen https://github.com/manifoldmarkets/manifold/pull/1171 but here it is!

Now all the work involving bets in `updateUserMetrics` only needs bets on unresolved markets, instead of all bets ever, so that's a lot better.

- You can only have outstanding loans on unresolved markets.
- We only need to compute the "investment value" for unresolved markets, since the investment value for resolved markets is 0.
- We only need to compute contract metrics for unresolved markets, if we make sure we compute it once at resolution, since they can't change post-resolution.

### DIFF
--- a/functions/src/update-user-metrics.ts
+++ b/functions/src/update-user-metrics.ts
@@ -184,9 +184,10 @@ const loadUserContractBets = async (userId: string, contractIds: string[]) => {
   const betDocs = await batchedWaitAll(
     contractIds.map((c) => async () => {
       return await firestore
-        .collectionGroup('bets')
+        .collection('contracts')
+        .doc(c)
+        .collection('bets')
         .where('userId', '==', userId)
-        .where('contractId', '==', c)
         .get()
     }),
     100
--- a/functions/src/update-user-metrics.ts
+++ b/functions/src/update-user-metrics.ts
@@ -51,7 +51,6 @@ export async function updateUserMetrics() {
   const contractsById = Object.fromEntries(contracts.map((c) => [c.id, c]))
   log(`Loaded ${contracts.length} contracts.`)
 
-  log('Computing metric updates...')
   const now = Date.now()
   const monthAgo = now - DAY_MS * 30
   const writer = firestore.bulkWriter({ throttling: false })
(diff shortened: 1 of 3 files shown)


jahooma (approved):
Great, thanks Marshall! This will reduce cost quite a bit.

functions/src/update-user-metrics.ts
```diff
@@ -57,21 +57,21 @@ export async function updateUserMetrics() {
...
-      const currentBets = (
-        await firestore
-          .collectionGroup('bets')
-          .where('userId', '==', user.id)
-          .get()
-      ).docs.map((d) => d.data() as Bet)
-      const betsByContractId = groupBy(currentBets, (b) => b.contractId)
+      const unresolvedBetContracts = contracts.filter(
+        (c) => !c.isResolved && c.uniqueBettorIds?.includes(user.id)
+      )
```
jahooma, line 62: Actually, there's another tricky thing here.

We actually save changes in daily, weekly, and monthly profit in the docs of `contract-metrics`. Technically, those need to be updated until 30 days after the market resolves, when 30-day profit would be 0.

The quick fix here is to update all markets that haven't resolved or have been resolved within the last 30 days.

functions/src/on-update-contract.ts
```diff
@@ -55,6 +57,20 @@ async function handleResolvedContract(contract: Contract) {
...
+  await Promise.all(
+    Object.entries(betsByUser).map(([userId, userBets]) => {
+      const metrics = getContractBetMetrics(contract, userBets)
+      return firestore
+        .collection('users')
+        .doc(userId)
+        .collection('contract-metrics')
+        .doc(contract.id)
+        .set(metrics)
+    })
```
jahooma, line 71: So, `getContractBetMetrics` doesn't include all the fields, there's `from` which has the daily weekly monthly profit changes.

If we're updating markets for up to 30 days after resolution then I guess we could remove this code anyway.

functions/src/update-user-metrics.ts
mqp: Hmm. I see your point. I guess just applying the quick fix for now sounds OK.

functions/src/on-update-contract.ts
mqp: That's a good point. I think I will just remove it. That would have been simpler in the first place.

mqp:
By the way, I was curious, so I actually ran a profiler and I can verify that the thing taking up time in this function is more or less the network stuff, not, like, some accidentally slow code we wrote inside the metrics updating somewhere.

PR #1179, opened 2022-11-09
Implement multiple choice with new mechanism

jahooma:
Ok, seems to be all working!

Follow up work:
- Sell shares dialog for answer shares
- Ability to add liquidity

Screen Shot 2022-11-19 at 4 07 26 PM


jahooma:
This is mostly ready for review!

Still needs a few things
- Bet status in list and table for short selling
- Sell button, dialog, and back end
- Testing

common/add-liquidity.ts
```diff
@@ -18,7 +18,6 @@ export const getNewLiquidityProvision = (
     contractId: contract.id,
     amount,
     pool,
-    p,
```
mantikoros, line 21: Removing p is good.

common/calculate-fixed-payouts.ts
```diff
@@ -18,21 +20,59 @@ export function calculateFixedCancelPayout(bet: Bet) {
...
+        ? resolutionProbability
+        : getProbability(contract)
+
+    const betP = outcome === 'YES' ? p : 1 - p
+
+    return betP * shares
+  }
+
+  const { resolutions, pool } = contract
+  const resolutionsSum = resolutions ? sum(Object.values(resolutions)) : 100
```
mantikoros, line 55: Resolutions sum to 100 instead of 1?

common/calculate-cpmm-multi.ts
```diff
@@ -36,7 +38,9 @@ export function buy(
   const kMissingOutcome = getK(tempPool)
   const shares = maxShares - Math.exp(k - kMissingOutcome)
 
-  tempPool[outcome] = maxShares - shares
+  const newShares = maxShares - shares
+  tempPool[outcome] = Math.max(MIN_POOL_SHARES, newShares)
```
mantikoros, line 42: It really should reject the trade if if the number of shares goes below MIN_POOL_SHARES rather just setting a min on the pool size. Theoretically, this is printing money that could be exploited. In practice, I don't think it will be that big of an issue.

web/components/answers/answers-panel.tsx
```diff
@@ -235,16 +251,36 @@ function OpenAnswer(props: {
...
-                size="2xs"
-                color="gray-outline"
-                onClick={() => setOpen(true)}
-                className="my-auto"
-              >
-                BUY
-              </Button>
-            )}
+            {tradingAllowed(contract) &&
+              (isFreeResponse ? (
```
mantikoros, line 255: You need to handle DPM multiple choice markets as well.

mantikoros (changes requested):
Mostly good. You also need to handle DPM MC markets in the answer panel UI, and update the all items in the market info dialog for this new market type.

common/calculate-fixed-payouts.ts
jahooma: Yeah, and it's very annoying

common/calculate-cpmm-multi.ts
jahooma: I was thinking it was losing them money, but I guess you're right. This could increase `k` as well, which breaks the invariant.

I guess I will change it.

web/components/answers/answers-panel.tsx
jahooma: ok, good catch

common/calculate-cpmm-multi.ts
jahooma: Added a TODO for later. I don't think it's bad to ship it this way.

PR #1181, opened 2022-11-14
Make API /bets endpoint be able to look stuff up by ID

mqp:
This is important so that we can deprecate returning full comments and bets by default from the `/market` endpoint.


jahooma (approved):
LGTM! Good feature for our api to have

PR #1182, opened 2022-11-14
Add `/comments` endpoint for querying comments by contract

mqp:
Like https://github.com/manifoldmarkets/manifold/pull/1181, this is important so that people can query market data without getting a firehose of bets and comments.

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -701,6 +701,17 @@ Parameters:
 - `html`: The comment to post, formatted as an HTML string, OR
 - `markdown`: The comment to post, formatted as a markdown string.
 
+### `GET /v0/comments`
+
+Gets a list of comments for a contract, ordered by creation date descending.
+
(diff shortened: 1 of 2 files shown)


mqp:
I'm going to fix this up with a few features to work like /bets so that it isn't quite so half-assed.

docs/docs/api.md
```diff
@@ -701,6 +701,17 @@ Parameters:
 - `html`: The comment to post, formatted as an HTML string, OR
 - `markdown`: The comment to post, formatted as a markdown string.
 
+### `GET /v0/comments`
+
+Gets a list of comments for a contract, ordered by creation date descending.
+
+Parameters:
+
+- `contractId`: Optional. Which contract to read comments for. Either an ID or slug must be specified.
```
PatMyron, line 710: Any chance of being able to get comments across all contracts instead of just one?

PR #1184, opened 2022-11-14
Make contract bet loading actually efficient

mqp:
Yesterday James wrote https://github.com/manifoldmarkets/manifold/commit/c1f2dff1a813f3a32da0345d08e3341bf03a9ff0, the intent of which was to only establish a subscription to new bets, rather than re-subscribe to all the bets already in the static props.

Unfortunately, this fix wasn't working, because `listAllBets` doesn't return bets in sorted order. (`useBets` does.) So it was still loading some random, probably large subset of the old bets. Now the fix should be actually working.


jahooma (approved):
Thanks Marshall! Maybe this will make a big difference!

PR #1185, opened 2022-11-14
More accurately produce the first N bets

mqp:
This wasn't right before because it wasn't sorting in the query, so it could end up with something that wasn't a prefix of the bets.


jahooma (approved):
Thanks Marshall!

PR #1187, opened 2022-11-15
Make `onCreateBet` trigger more efficient

mqp:
I don't think this is the worst problem we have, but clearly this bet loading was performing totally needless carnage on our Firestore reads.

### DIFF
--- a/functions/src/helpers/handle-referral.ts
+++ b/functions/src/helpers/handle-referral.ts
@@ -50,22 +50,15 @@ export async function handleReferral(user: User, eventId: string) {
     }
     console.log(`referredByGroup: ${referredByGroup}`)
 
-    const txns = (
-      await firestore
+    const txns = await transaction.get(
+      firestore
         .collection('txns')
         .where('toId', '==', referredByUserId)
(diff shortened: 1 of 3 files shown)


functions/src/helpers/handle-referral.ts
```diff
@@ -50,22 +50,15 @@ export async function handleReferral(user: User, eventId: string) {
...
-        console.error('error getting txns:', err)
-        throw err
-      })
+    )
+    if (txns.size > 0) {
       // If the referring user already has a referral txn due to referring this user, halt
-      if (
-        referralTxns.map((txn) => txn.data()?.description).includes(user.id)
-      ) {
+      if (txns.docs.map((txn) => txn.data()?.description).includes(user.id)) {
```
mqp, line 61: This is still a bad piece of code; we should add the referral user ID directly to the referral txn object so that we can filter on it in the query.

jahooma: Yep

jahooma (approved):
Thanks, this is a nice cleanup, and not fetching all the bets for no reason is actually pretty important. LGTM!

PR #1188, opened 2022-11-15
Make `onCreateCommentOnContract` trigger more efficient

mqp:
Like #1187, I am not sure if this is super important, but if I want to find something missing, and it's hard to find, I start cleaning my house and see if it turns up.

### DIFF
--- a/functions/src/on-create-comment-on-contract.ts
+++ b/functions/src/on-create-comment-on-contract.ts
@@ -10,7 +10,6 @@ import {
 } from './utils'
 import { ContractComment } from '../../common/comment'
 import { Bet } from '../../common/bet'
-import { Answer } from '../../common/answer'
 import { getLargestPosition } from '../../common/calculate'
 import {
   createCommentOrAnswerOrUpdatedContractNotification,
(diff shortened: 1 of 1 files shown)


jahooma:
Sorry, I told @IanPhilips about this and he made his own improvements, so you'll have to merge.

I think though that you might have found different things to optimize than him.

jahooma (approved):
Thanks Marshall. LGTM!

IanPhilips (approved):
LGTM! In my other commit 6516bae I opted for reducing the computation needed to calculate investment value of bets at the expense of 'with shares in' being more of a educated guess than a certainty

IanPhilips:
Oh it looks like it's still timing out. Hmmmm I might just increase the timeout 
Screen Shot 2022-11-16 at 8 26 21 AM

IanPhilips:
I made a couple more adjustments: lengthening the timeout, adding logs each step of the way, and using the uniqueBettorIds field rather than finding the unique bettors. On that topic though, the popular contracts are now carrying around 1500 strings with them. Perhaps it's time to move those unique bettor ids to a subcollection...

IanPhilips:
Looks like we're in the clear!
Screen Shot 2022-11-16 at 9 28 42 AM

jahooma:
> Looks like we're in the clear! A new comment on dan-stock from memestiny finished in under 20 seconds (1600 traders) screenshot removed b/c it included emails

Thanks @IanPhilips!

PR #1189, opened 2022-11-15
Don't bother fetching follow docs before writing them

mqp:
Follows happen whenever bets happen, so they shouldn't do extra work.

### DIFF
--- a/functions/src/follow-market.ts
+++ b/functions/src/follow-market.ts
@@ -6,30 +6,22 @@ export const addUserToContractFollowers = async (
   contractId: string,
   userId: string
 ) => {
-  const followerDoc = await firestore
-    .collection(`contracts/${contractId}/follows`)
-    .doc(userId)
-    .get()
-  if (followerDoc.exists) return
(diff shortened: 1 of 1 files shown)


functions/src/follow-market.ts
```diff
@@ -6,30 +6,22 @@ export const addUserToContractFollowers = async (
...
   contractId: string,
   userId: string
 ) => {
-  const followerDoc = await firestore
-    .collection(`contracts/${contractId}/follows`)
-    .doc(userId)
-    .get()
-  if (!followerDoc.exists) return
-  await firestore
+  return await firestore
```
jahooma, line 24: Would this code now throw an error if the doc doesn't exist (different behavior)?

Or does `delete()` just become a no op if there's no doc.

jahooma (approved):
LGTM

functions/src/follow-market.ts
mqp: According to the docs it's the latter.

PR #1193, opened 2022-11-15
Backfill bet flags into DB so that we can use them in queries

mqp:
Pretty self-explanatory. There are certainly lots of redemptions in particular so this is reasonably valuable.

### DIFF
--- a/common/antes.ts
+++ b/common/antes.ts
@@ -88,8 +88,10 @@ export function getAnteBets(
     probBefore: p,
     probAfter: p,
     createdTime,
-    isAnte: true,
     fees: noFees,
+    isAnte: true,
+    isRedemption: false,
+    isChallenge: false,
   }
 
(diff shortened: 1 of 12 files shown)


common/antes.ts
```diff
@@ -67,8 +67,11 @@ export function getAnteBets(
     probBefore: p,
     probAfter: p,
     createdTime,
-    isAnte: true,
     fees: noFees,
+    isAnte: true,
+    isLiquidityProvision: false,
```
jahooma, line 72: This field is deprecated and not used anywhere, I believe? (Liquidity has its own subcollection now)

I think we can remove it.

jahooma (approved):
Thanks Marshall!

common/antes.ts
mqp: I will do this in a subsequent PR after checking to see whether any bets with this flag exist.

PR #1197, opened 2022-11-15
replace mana symbol

ingawei:
Screen Shot 2022-11-15 at 3 39 33 PM


ingawei:
Found a more plug and play solution, by replacing a special character that doesn't exist on readex pro, Ṁ. Couple of questions
1. @jahooma should I be replacing the M$ in token within transact?
    for example:
```  if (LIKE_TIP_AMOUNT > 0) {
    result = await transact({
      amount: LIKE_TIP_AMOUNT,
      fromId: user.id,
      fromType: 'USER',
      toId: item.creatorId,
      toType: 'USER',
      token: 'M$',
      category: 'TIP',
      data: { contractId: item.id },
      description: `${user.name} liked ${itemType}${item.id} for M$ ${LIKE_TIP_AMOUNT} to ${item.creatorId} `,
    })```
3. @PhilBladen I replaced the M$ with Ṁ in your twitch code, please just make sure I didn't break anything!

File: common/util/format.ts [hunk 1]
```diff
@@ -26,6 +26,21 @@ export function formatMoney(amount: number) {
   return ENV_CONFIG.moneyMoniker + formatter.format(newAmount).replace('$', '')
 }
 
+export function formatMoneyNumber(amount: number) {
```
jahooma, line 29: There should be a way not to duplicate all the logic from the above function here.

The above function could just call this one.

web/components/add-funds-modal.tsx
```diff
@@ -84,7 +84,7 @@ function BuyManaTab(props: { onClose: () => void }) {
   return (
     <>
       <div className="mt-6 mb-4">
-        Buy mana (M$) to trade in your favorite markets.
+        Buy mana (Ṁ) to trade in your favorite markets.
```
jahooma, line 87: Rather than changing it to the new symbol, you should use the `ENV_CONFIG.moneyMoniker` constant, so it's easier to change in the future.

Ditto for all the other references.

web/styles/globals.css
```diff
@@ -0,0 +1,40 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
+
+@layer base {
+  html {
+    -webkit-tap-highlight-color: transparent;
```
jahooma, line 7: Are we sure we want to turn this off? When I tap a contract card on android, the blue highlight gives me some feedback about what I did.

web/styles/globals.css
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+@font-face {
+  font-family: 'icomoon';
+  src: url('../fonts/icomoon.eot?v49ui9');
+  src: url('../fonts/icomoon.eot?v49ui9#iefix') format('embedded-opentype'),
+    url('../fonts/icomoon.ttf?v49ui9') format('truetype'),
+    url('../fonts/icomoon.woff?v49ui9') format('woff'),
+    url('../fonts/icomoon.svg?v49ui9#icomoon') format('svg');
```
jahooma, line 16: Are you supposed to specify the `src` prop twice?

web/styles/globals.css
```diff
@@ -0,0 +1,40 @@
...
+  font-weight: normal;
+  font-style: normal;
+  font-display: block;
+}
+
+[class^='icon-'],
+[class*=' icon-'] {
```
jahooma, line 23: This matches all `icon-` classes. Why do we want to do that?

I'm not actually not sure how this change works overall haha. How can it apply a different font just for the money character? Oh I see you said it's for a character that readex pro doesn't have.

jahooma (approved):
Thanks Inga for figuring out a solution for a better mana symbol! LGTM.

web/styles/globals.css
ingawei: I feel like the feedback should be that the card's dropshadow gets bigger. I just think the blue highlight is ugly in general and we should just make better focus states to give feedback. I was wondering if there were maybe some accessibility considerations I wasn't thinking about though

PR #1206, opened 2022-11-17
Normalize section height on the home page and (usually) prevent unbalanced columns

SubsequentlySneeds:
Changes the logic for ContractsGrid rendering to always render in a rectangular-ish area on the screen, so more contracts can be shown regardless of whether the first or last contracts have an image to display. 

Not sure if this is wanted (the screen looks "busier", there is sometimes unused data transferred, and the method is ugly - inserting a fake div to trick the Masonry into putting more elements on the right), but it does remove the gap at the end of most sections that have an image on the first contract, and causes the 3rd highest scored contract to display above the 4th, whereas previously when a single image was shown for the first contract, the 3rd would display below the image, in the same line as the 6th.

In a section with an image, the changed layout can fit 7 contracts in the same space the current one fits 6:

chrome_31a4kNGqFh

If there is no image for the top contract in the section, the behavior is the same as before, except that an odd number of contract inputs gets rounded down to the nearest even number to create an (approximate) rectangle. If there is one image (the first contract has an image but the last one does not), the number is rounded down to the nearest odd number, and the 3rd contract is pushed to the right to balance the extra space taken by the image. If there is both a first and last image, the total number of contracts is reduced by 2, to compensate for the extra line taken by the images.

This also changes the default values for how many contracts are shown in a home section and group section respectively - the minimums (2 image case) are the same as before, but in the other cases one or two more contracts can be shown in the same amount of space.


jahooma (commented):
Thanks SubsequentlySneeds for your submission! 

It looks like a good attempt that works to prevent unbalanced columns on the home page, and we're thankful for that.

However, I've left a comment on why this kind of very specific code may not be net-positive in the long run, so I don't think I will approve it.

PR #1208, opened 2022-11-18
'before' and 'limit' paramters for /groups endpoint

mia-z:
Following the conventions used in other endpoints that fetch large amounts of data, I decided to add the same parameters to the /groups endpoint. 

- Added a hard limit of 100 to groups.

### DIFF
--- a/web/lib/firebase/groups.ts
+++ b/web/lib/firebase/groups.ts
@@ -6,11 +6,15 @@ import {
   doc,
   getDocs,
   onSnapshot,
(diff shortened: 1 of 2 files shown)


mqp:
This is cool, but the code in `pages/groups.tsx` needs to be adjusted to still load all groups (which sucks, but that's how it was designed.)

sipec:
we now fetch 500 groups and in the endpoint and allow pagination via `beforeTime` so this pr is no longer needed

PR #1214, opened 2022-11-21
Load additional bets on broken embeds after initial load

mqp:
This is necessary to support "embeds pointing to the main contract page", which load the contract page static props (only including a prefix of bets) and then render the embed view afterward. Otherwise, embeds with more than 2.5k bets will just show a wrong chart forever. You can see an example on the Donald Trump market shown at https://astralcodexten.substack.com/p/mantic-monday-twitter-chaos-edition.


jahooma (approved):
Cool, looks like a good improvement!

sipec (approved):

PR #1216, opened 2022-11-21
More flexible bet filter machinery, use it everywhere

mqp:
Now more places can use the `BetFilter` query stuff to filter out (typically) redemptions and antes in the query (e.g. the API, the activity feed, the user profile.)

### DIFF
--- a/web/hooks/use-bets.ts
+++ b/web/hooks/use-bets.ts
@@ -1,11 +1,9 @@
 import { useEffect, useState } from 'react'
-import { Contract } from 'common/contract'
 import {
   Bet,
   BetFilter,
   listenForBets,
   listenForUnfilledBets,
-  withoutAnteBets,
 } from 'web/lib/firebase/bets'
 import { LimitBet } from 'common/bet'
 import { inMemoryStore, usePersistentState } from './use-persistent-state'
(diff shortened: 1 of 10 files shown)


web/hooks/use-bets.ts
```diff
@@ -37,9 +29,12 @@ export const useBetsWithoutAntes = (contract: Contract, initialBets: Bet[]) => {
...
   )
   useEffect(() => {
-    return listenForBets(contract.id, (bets) => {
-      setBets(withoutAnteBets(contract, bets).sort((b) => b.createdTime))
-    })
+    return listenForBets(
+      (bets) => {
+        setBets(withoutAnteBets(contract, bets).sort((b) => b.createdTime))
+      },
+      { contractId: contract.id }
```
jahooma, line 36: I think you can add the `filterAntes: true` filter and then delete the `withoutAnteBets` function. I believe it's from a time long ago when ante bets used to be just the first two bets.

jahooma (approved):
Nice! Seems like a solid cleanup. Makes it easier to change the bet querying infra in the future.

web/hooks/use-bets.ts
mqp: OK, I wasn't confident about what the deal was with that first two bets business. If you think it's OK I will happily clean this up.

PR #1217, opened 2022-11-21
Move onboarding emails to scheduled function

IanPhilips:
In order to respect users' email preferences we move from scheduling the emails on mailgun to a scheduled cloud function

Screenshot 2022-11-21 at 11 52 24 AM

### DIFF
--- a/functions/src/onboarding-emails.ts
+++ b/functions/src/onboarding-emails.ts
@@ -1,0 +1,126 @@
+import * as functions from 'firebase-functions'
+import * as dayjs from 'dayjs'
+import * as utc from 'dayjs/plugin/utc'
+dayjs.extend(utc)
+
+import * as admin from 'firebase-admin'
+import { PrivateUser, User } from '../../common/user'
+import {
+  sendCreatorGuideEmail,
+  sendInterestingMarketsEmail,
+  sendPersonalFollowupEmail,
+} from './emails'
+import { getNotificationDestinationsForUser } from '../../common/user-notification-preferences'
+import { getTrendingContracts } from './weekly-markets-emails'
+const firestore = admin.firestore()
+
+//TODO: We'll probably need a longer timeout for this function, so find out how to
+// limit function calls via other backend functions only
+// TODO2: we'll want to stop scheduling the other onboarding emails before this goes into action
+export const onboardingEmails = functions
+  .runWith({ secrets: ['MAILGUN_KEY'], timeoutSeconds: 540 })
+  // run every day at 10am PST
+  .pubsub.schedule('0 10 * * *')
+  .timeZone('America/Los_Angeles')
+  .onRun(async () => {
+    // Users created between 24-48 hours ago
+    await send1DayInterestingMarketsEmails()
+    // Users created between 48-72 hours ago
+    await send2DayFollowupEmails()
+    // Users created between 72-96 hours ago
+    await send3DayCreatorGuideEmails()
+  })
+
+export const send2DayFollowupEmails = async () => {
+  const { users, idToPrivateUsers } = await getUsersCreatedBetweenDays(2, 3)
+  await Promise.all(
+    users.map(async (user) => {
+      const privateUser = idToPrivateUsers.get(user.id)
+      if (!privateUser) return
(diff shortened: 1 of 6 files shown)


IanPhilips:
@mqp Do you have any idea how I could make a v2 function whose permissions only allow another cloud function to call it? E.g. what you have for the update user metrics but instead of no auth required, **only** allow a pubsub function to call the v2 function.

mqp:
> @mqp Do you have any idea how I could make a v2 function whose permissions only allow another cloud function to call it? E.g. what you have for the update user metrics but instead of no auth required, **only** allow a pubsub function to call the v2 function.

The most obvious way I would implement this is to make a secret that only those two functions have access to, which the first seconds to the second, and the second checks.

functions/src/onboarding-emails.ts
```diff
@@ -0,0 +1,126 @@
...
+  const day = dayjs().utc().day()
+  // don't run on sun or mon
+  if (day === 0 || day === 1) return
+  const contracts = await getTrendingContracts()
+  const { users, idToPrivateUsers } = await getUsersCreatedBetweenDays(1, 2)
+  await Promise.all(
+    users.map(async (user) => {
+      const privateUser = idToPrivateUsers.get(user.id)
+      // If it's a wednesday we might get users created on monday that've already gotten the weekly email
+      if (!privateUser || privateUser.weeklyTrendingEmailSent) return
```
mantikoros, line 62: I'm assuming weeklyTrendingEmailSent gets reset sometime.

functions/src/onboarding-emails.ts
```diff
@@ -0,0 +1,126 @@
...
+  const maxDaysTime = dayjs().subtract(maxDaysAgo, 'days').valueOf()
+
+  console.log(
+    `Getting users created between ${minDaysAgo} and ${maxDaysAgo} days ago`
+  )
+  const users = (
+    await firestore
+      .collection('users')
+      .where('createdTime', '<', minDaysTime)
+      .where('createdTime', '>', maxDaysTime)
```
mantikoros, line 109: Did you make sure an firestore index was created for this on prod?

functions/src/onboarding-emails.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  const usersAndPrivateUsersMap = new Map<string, PrivateUser>()
+  await Promise.all(
+    users.map(async (user) => {
+      const privateUser = await firestore
+        .collection('private-users')
+        .doc(user.id)
+        .get()
+        .then((doc) => doc.data() as PrivateUser)
+      usersAndPrivateUsersMap.set(user.id, privateUser)
```
mantikoros, line 121: It's bad practice to put a mutable expression like usersAndPrivateUsersMap.set() in normally immutable function like map(). You could have the function return a [string, PrivateUser] and then construct an object from the entries instead.

mantikoros (approved):
The logic looks alright to me. And I suppose this does provide additional functionality (following user email preferences), but adding a new brittle process that loads all users into memory every day is a little less robust than the current implementation.

functions/src/onboarding-emails.ts
IanPhilips: I will!

functions/src/onboarding-emails.ts
IanPhilips: Yup, on monday mornings

IanPhilips:
> > @mqp Do you have any idea how I could make a v2 function whose permissions only allow another cloud function to call it? E.g. what you have for the update user metrics but instead of no auth required, **only** allow a pubsub function to call the v2 function.
> 
> The most obvious way I would implement this is to make a secret that only those two functions have access to, which the first sends to the second, and the second checks.

Cool, that sounds reasonable & easy. Was thinking there could be some other way I was missing.

IanPhilips:
One thought: can you un-schedule a mailgun email via email address?

IanPhilips:
Oh nvm I looked this up previously, you can only dequeue them all

IanPhilips:
Did this: 04932cae239f965ded986148394cd7c3f221d063

PR #1218, opened 2022-11-22
Make popularity score based on 7 day window so we don't have to load all bets

mqp:
@jahooma do you think this is reasonable? If not, we can add a 3-day metric for unique bettors on the contract. But I figure maybe 7 days is just as good for this purpose anyway.

### DIFF
--- a/functions/src/score-contracts.ts
+++ b/functions/src/score-contracts.ts
@@ -1,7 +1,5 @@
 import * as functions from 'firebase-functions'
 import * as admin from 'firebase-admin'
-import { uniq } from 'lodash'
-import { Bet } from '../../common/bet'
 import { Contract } from '../../common/contract'
 import { log } from './utils'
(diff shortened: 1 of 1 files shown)


jahooma:
Hmm, I don't think that would be good, in fact, I think the trending algorithm should go in the other direction.

How about sorting by unique traders in the last 24 hours?

To solve for some groups where many markets would have 0 traders in this timeframe, you could add 1/10 of the 7-day traders.

mqp:
> Hmm, I don't think that would be good, in fact, I think the trending algorithm should go in the other direction.
> 
> How about sorting by unique traders in the last 24 hours?
> 
> To solve for some groups where many markets would have 0 traders in this timeframe, you could add 1/10 of the 7-day traders.

OK, if that's what you think is good I will happily implement it like that.

PR #1220, opened 2022-11-22
Somewhat improve `changeuserinfo`

mqp:
This fixes a bug where people could enter invalid display names and propagate those invalid names across all the denormalized stuff. (It wasn't super obvious because we also sanitize this on the client.) It also fixes gratuitous inefficiency. Note that all this code is still incorrect -- it will race to update denormalized data in any case where you quickly update your info twice in succession. It's just a little less bad.

### DIFF
--- a/functions/src/change-user-info.ts
+++ b/functions/src/change-user-info.ts
@@ -54,6 +54,9 @@ export const changeUser = async (
     avatarUrl?: string
   }
 ) => {
+  if (update.username) update.username = cleanUsername(update.username)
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Thanks, looks a bit cleaner!

mqp:
Also fixed an issue where the denormalized data gets out of sync (which it should, since the code is wrong) then it will be broken forever. Now if you update your info again at least you will fix it.

mqp:
Also fixed an issue where we were throwing away error messages, and an issue where we were failing to update the denormalized user data on contract answers.

PR #1221, opened 2022-11-22
Contract card UI Fixes

sipec:
This was supposed to be just a fix to the avatar/group link regression but man, soo many tiny style issues

Fixes MAN-215


jahooma (approved):
Great, thanks for the cleanup, Sinclair!

Good to reduce complexity in ContractSearch, and fix bugs like having the user be clickable from a card.

PR #1222, opened 2022-11-22
Remove dead code related to tracking contract views

mqp:
This is from an older iteration of our analytics stuff and hasn't been recording data for a long time.


jahooma (approved):
LGTM!

PR #1224, opened 2022-11-22
Update API changelog

Zentrik:
This seems useful. Sorry if this is not wanted.


mqp:
Cool, thanks. I'm afraid to say that we definitely do not and will not soon update that changelog with every user-visible change, so I don't like it very much. But as long as it's there, I would rather have any given important thing on it.

PR #1225, opened 2022-11-22
Change loan text when viewing another user's bets

### LINKED ISSUE #1210: Portfolios for other users say "you" have gotten a loan instead of saying they got a loan
image

SubsequentlySneeds:
Fixes #1210

chrome_iDX0uUW1U6


jahooma (approved):
Sweet, thanks a lot for the fix!

PR #1226, opened 2022-11-22
Remove unnecessary creator volume computation

mqp:
I believe this was left over from when we used to have leaderboards ranking this, but after we removed them we forgot to clean it up.


jahooma (approved):
LGTM! Thanks for the cleanup

PR #1229, opened 2022-11-23
Remove precomputed user profit rank, clean up leaderboards code

mqp:
Motivation: trying to eliminate complicated and expensive batch job work. Profit rank is almost guaranteed to change for ~every user ~every time we update, so it makes us update every user doc every 15 minutes for sure.

- We no longer need to precompute the profit rank in `updateUserMetrics`.
- The leaderboards page no longer re-downloads the static props for no obvious reason.
- We no longer censor our bot from the leaderboard. (Actually, we previously didn't bother to do any metrics updates on our bot, which is pretty bizarre. It just had a bunch of out of date info in there.) If someone really loves this I can put it back in. We will need to add a field in the database like `isEligibleForLeaderboard`. I think it's much better to have it on the leaderboard.

### DIFF
--- a/common/user.ts
+++ b/common/user.ts
@@ -26,13 +26,6 @@ export type User = {
     allTime: number
   }
 
-  profitRankCached?: {
-    daily: number
-    weekly: number
-    monthly: number
-    allTime: number
-  }
-
   creatorTraders: {
     daily: number
     weekly: number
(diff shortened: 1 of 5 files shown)


IanPhilips (approved):
This looks much cleaner, thanks Marshall! This is also a good starting point to add a users' unique trader & follower ranks on the leaderboards. If you feel like adding those in that would be good, too! If not, no worries, I will take what you've written and extend it.

PR #1230, opened 2022-11-23
Don't load redemption bets in `updateContractMetrics`

mqp:
Redemptions shouldn't affect the probability, volume, or unique bettors, so they are useless to consider in these metrics.


jahooma (approved):
LGTM!

PR #1231, opened 2022-11-23
Remove `volume7Days` precomputed metric

mqp:
I'm going to claim we don't need this for anything, and once I replace the "unique bettors in time interval" and "probability change over time" metric computations to be more efficient (both of which I plan to do), then we will no longer need to load as many bets in `updateContractMetrics`.

### DIFF
--- a/web/lib/firebase/contracts.ts
+++ b/web/lib/firebase/contracts.ts
@@ -199,7 +199,7 @@ export function listenForInactiveContracts(
 const newContractsQuery = query(
   contracts,
   where('isResolved', '==', false),
-  where('volume7Days', '==', 0),
+  where('volume', '==', 0),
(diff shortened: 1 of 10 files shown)


web/lib/firebase/contracts.ts
```diff
@@ -199,7 +199,7 @@ export function listenForInactiveContracts(
 const newContractsQuery = query(
   contracts,
   where('isResolved', '==', false),
-  where('volume7Days', '==', 0),
+  where('volume', '==', 0),
   where('createdTime', '>', Date.now() - 7 * DAY_MS)
```
mqp, line 203: This seems equivalent, since if the contract was created in the last 7 days then `volume` should `===` `volume7Days` (except for batch processing delay.)

jahooma (approved):
LGTM!

PR #1232, opened 2022-11-23
Invalidate contract page props on description and question changes

mqp:
This broken invalidation surely causes many bugs, but it seems one of the worst bugs is with the description editor, which doesn't correctly handle the case where the description changes after loading the page. As a result everyone on Discord is mad, because they will change a description and the change basically won't work until someone bets.

I don't want to deal with trying to fix the bug in that editor right now, so I am doing the simplest thing that I think will make people less mad.

### DIFF
--- a/functions/src/on-update-contract.ts
+++ b/functions/src/on-update-contract.ts
@@ -27,8 +27,15 @@ export const onUpdateContract = functions.firestore
       await handleUpdatedCloseTime(previousContract, contract, eventId)
     }
 
-    // maybe we should do this more often, but at least if someone bets
-    if (previousContract.volume !== contract.volume) {
+    // mqp: if you are here trying to figure out why the contract page is
+    // showing random incorrect data, it's probably because we don't do this on
+    // every contract change, and only do it sometimes instead.  complain to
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Thanks Marshall. This has been a very annoying bug, so good to have at least a partial solution. This might require them to refresh the page to get the new description.

But I have alerted Sinclair to the bug, so hopefully it will be fixed clientside as well

mqp:
The only bug which I understand (and I just glanced at it so maybe I just
don't know) is

1. If you edit a description,
2. The contract static props wouldn't get updated,
3. So then the next time you load the page, the editor will be initialized
with the old description,
4. And it happens to be that when you click to edit in the editor, it will
always use the first description it was initialized with, instead of any
subsequent update.
5. So then when you load the page, first you will see the old description,
then you will see the new description, and then if you click "edit" you
will see the old one again. (And if you save it will re-save the old one
over the new one.)

This change should totally fix that bug since the static props won't be out
of date. I don't know if there are some other bugs that only client-side
fixes will address. (Well, there's the obvious bug if you have two browser
windows open, but I am guessing that's not what people were reporting.)

PR #1233, opened 2022-11-23
Make % contract mention use algolia

sipec:
Fixes MAN-46
Fixes MAN-312


web/components/editor/contract-mention-suggestion.ts
```diff
@@ -15,12 +13,11 @@ export const contractMentionSuggestion: Suggestion = {
...
-      ),
-      [(c) => [c.question].some((s) => beginsWith(s, query))],
-      ['desc']
-    ).slice(0, 5),
+    (
+      await trendingIndex.search(query, {
+        hitsPerPage: 5,
+        removeStopWords: true,
+      })
+    ).hits,
```
jahooma, line 21: Cool that was easier than one might think. Should have done this earlier haha.

jahooma: Seems like the search index `searchClient.initIndex(searchIndexName)` would be better than the trending index?

jahooma (approved):
LGTM!

PR #1239, opened 2022-11-24
Introduce `contract-metrics` `lastBetTime`, optimize `updateContractMetrics`

mqp:
The goal of this field is to make it easy to compute "unique bettors between time T0 and T1." (I intend to remove the contract unique bettor fields entirely in a later PR, but adding this field is the first step.)

Now we don't need to load all the bets from the last month in `updateContractMetrics`; just the bets from the last day, for the purpose of computing daily volume.

Later I may transition to maintaining this field in the `onCreateBet` trigger so that it doesn't have a 15-minute lag time, but for now let's populate it in the same place that we populate the other contract metrics, since that's easy.

### DIFF
--- a/functions/src/update-contract-metrics.ts
+++ b/functions/src/update-contract-metrics.ts
@@ -1,15 +1,13 @@
 import * as functions from 'firebase-functions'
 import * as admin from 'firebase-admin'
+import { CollectionReference } from 'firebase-admin/firestore'
+import { sumBy } from 'lodash'
 
 import { getValues, invokeFunction, log } from './utils'
 import { Bet, LimitBet } from '../../common/bet'
 import { Contract, CPMM } from '../../common/contract'
 import { DAY_MS } from '../../common/util/time'
-import {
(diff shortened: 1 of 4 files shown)


functions/src/update-contract-metrics.ts
```diff
@@ -133,8 +116,48 @@ export async function updateContractMetrics() {
...
+}
+
+async function getUniqueBettors(contractId: string, since: number) {
+  return (
+    await firestore
+      .collectionGroup('contract-metrics')
+      .where('contractId', '==', contractId)
+      .where('lastBetTime', '>', since)
+      .count()
+      .get()
```
jahooma, line 161: Whoa! This is very clever.

jahooma (approved):
Wow, what a great way to query for a probability or the number of unique bettors!

Seems like this will save us a lot of money. Thanks Marshall!

IanPhilips:
Just fyi, I've got a PR to update contract-metrics on create bet: #1256

PR #1242, opened 2022-11-24
Remove suspicious `payUser` function, clean some things up

mqp:
As you can see, this function was a kind of anti-pattern that made the Stripe webhook and the loan update code worse.

EDIT: I backed out the refactoring because some stuff changed, so now this PR just fixes the Stripe webhook to be correct.


jahooma (approved):
Thanks for the cleanup! LGTM

PR #1247, opened 2022-11-25
Bump grid.js and gridjs-react to 5.1

mqp:
This isn't important, it's just nice because it's marked as React-18-compatible so yarn will stop yelling at us.


jahooma (approved):
LGTM!

PR #1250, opened 2022-11-26
always include current username in share-market-URLs

### LINKED ISSUE #1161: The "Share your market" link does not contain the "referrer"
At the bottom of my market there is a box that says "Share your market! Earn a M$250 referral bonus if a new user signs up using the link.". The link that I get when clicking on "Copy link" does not contain the "referrer=USERNAME".

speezepearson:
Fixes #1161, if that's regarded as a bug worth fixing! If not, feel free to ignore this.

Testing:

- Before: Screenshot from 2022-11-25 22-52-56

- After: Screenshot from 2022-11-25 22-53-04

(If you'd like unit tests for this, I'd be happy to add them! But I'll need a pointer about where to put them.)

### DIFF
--- a/common/util/share.ts
+++ b/common/util/share.ts
@@ -3,7 +3,5 @@ import { ENV_CONFIG } from 'common/envs/constants'
 
 export const getShareUrl = (contract: Contract, username: string | undefined) =>
   `https://${ENV_CONFIG.domain}/${contract.creatorUsername}/${contract.slug}${
-    username && contract.creatorUsername !== username
-      ? '?referrer=' + username
-      : ''
+    username ? '?referrer=' + username : ''
   }`


akrolsmir:
Hey @speezepearson -- thanks for sending this out! I _think_ the choice not to include the referrer URL was deliberate, because we'll actually award the referrer bonus by default to the market creator (in the case when there's no referrer attached), so it's a slight optimization to make the URL a little prettier.

I'll check with the team on Monday about this, anyways. It's possible that including the referrer anyways would be more clear/consistent!

(also: we don't have any unit tests in our entire codebase atm, lol)

jahooma:
I wasn't sure before, but now I think it is better to include it, so that the person sharing the link has more confidence they will get credit.

jahooma:
I'm just going to merge this, and if you disagree, we can undo it.

mqp:
I strongly agree with this change and I think the reasoning Austin described would be a poor reason to do it the other way.

PR #1251, opened 2022-11-27
Supabase read replica prototype

mqp:
This implements replication of some database data to Supabase. General outline:

- We have a Supabase database for both dev and prod.
- Each Supabase database has tables for the Firestore collections we want to replicate, e.g. contracts.
- Each replicated collection table contains `(id, data)` pairs, where data is a Postgres JSONB blob.
- Most writes should wind up in Supabase within half a second or so unless there is a service interruption.

To perform the replication, we introduce a Firestore transaction log that uses Firesore triggers to record all writes on replicated collections into the `transactionLog` collection. Each write is stored with a timestamp, the document ID, the document type, and the new post-write data in the document. We can optionally truncate this log in the future if it bloats our storage costs.

Another trigger observes transaction log entries and sends them to Supabase's `incoming_writes` table. Right now, if this trigger fails (i.e. because Supabase is unreachable), we'll be missing data in Supabase. Once this is not a prototype, we should add a scheduled job that re-sends missed writes from the transaction log. This should be easy because "putting things in the `incoming_writes` table is idempotent, so we can just re-send a bunch of them. Or we can identify missing ones by comparing the contents of our transaction log to the contents of Supabase's `incoming_writes` table.

The `incoming_writes` table has a trigger which looks at the incoming writes and updates the corresponding documents in the Supabase data tables. Importantly, the write will only be applied if it is the most recent write we have seen for that document. Because of this property, we can do a live import from Firestore while the transaction log replication is running by just sending `create` writes for snapshots of all current documents with a conservatively early timestamp, and once we are done it will be caught up and consistent. This is tricky so please let me know if you noticed a problem I didn't.

The general plan is to get this up and running, start migrating reads that aren't time-sensitive or inside transactions to use Supabase, and then switch over to use Supabase exclusively for some subset of our collections.

Please contact me for an invite to our Supabase organization.


functions/src/replication/import.ts
```diff
@@ -0,0 +1,107 @@
...
+// strategy for live importing collection C without dropping data (times are firestore server times)
+// 1. optional - clear supabase table for collection C
+// 2. start replication of new collection C transaction log entries to supabase starting at T0
+// 3. establish import starting timestamp T1 such that T0 <= T1 <= now
+// 4. read collection C from firestore with effective timestamp T1 and send supabase writes
+// 5. supabase should be caught up
+
+async function getServerTimestamp() {
+  // firestore doesn't seem to have an API to just ask for the time, so we do this kludge
+  const result = await admin.firestore().collection('temp').doc('time').get()
```
jahooma, line 23: We have a `globalConfig` collection if you want to create the `time` doc there.

jahooma: Or, I guess the doc doesn't have to actually exist?

functions/src/replication/utils.ts
```diff
@@ -0,0 +1,24 @@
+import { createClient, PostgrestResponse } from '@supabase/supabase-js'
+
+import { DEV_CONFIG } from '../../../common/envs/dev'
+import { PROD_CONFIG } from '../../../common/envs/prod'
+import { isProd } from '../utils'
+
+export function createSupabaseClient() {
```
jahooma, line 7: Shouldn't this file be under the `supabase` directory?

functions/supabase/seed.sql
```diff
@@ -0,0 +1,101 @@
+create table if not exists users (
```
jahooma, line 1: Cool, a sql file!

functions/supabase/seed.sql
```diff
@@ -0,0 +1,101 @@
...
+    return new;
+  end if;
+  if new.write_kind = 'create' or new.write_kind = 'update 'then
+    execute format(
+      'insert into %1$I (id, data, fs_updated_time) values (%2$L, %3$L, %4$L)
+       on conflict (id) do update set data = %3$L, fs_updated_time = %4$L
+       where %1$I.fs_updated_time <= %4$L
+       returning id;',
+      dest_table, new.doc_id, new.data, new.ts
+    ) into replicated_id;
```
jahooma, line 83: I don't actually know sql. What does this do?

functions/src/replication/import.ts
```diff
@@ -0,0 +1,107 @@
...
+  snap: QueryDocumentSnapshot,
+  docKind: DocumentKind,
+  ts: Date
+) {
+  return {
+    doc_kind: docKind,
+    write_kind: 'create',
+    doc_id: snap.id,
+    data: snap.data(),
+    ts: ts.toISOString(),
```
jahooma, line 37: So, timestamps are strings? Not milliseconds like we were using in firestore?

common/transaction-log.ts
```diff
@@ -0,0 +1,21 @@
...
+  | 'contractBet'
+  | 'contractComment'
+
+export type WriteKind = 'create' | 'update' | 'delete'
+
+export type WriteDocument = { [k: string]: any }
+
+export type TLEntry<T extends WriteDocument = WriteDocument> = {
+  eventId: string
+  docKind: DocumentKind
```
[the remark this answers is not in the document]
mqp: Technically they are already not 1:1 with collections. Bets come from many independent Firestore collections. I am also not sure about the future. In general, the important thing here is the type of the document, not where it came from, because the type is mostly what is going to make me want to make a separate SQL table for it.

functions/src/replication/import.ts
mqp: Correct, the doc doesn't exist.

functions/src/replication/utils.ts
mqp: I think as long as we have a `src` directory, ource code should be in it and non-source code shouldn't. If this was in the `supabase` directory, we would have to change the build script to build it and package it.

functions/src/replication/import.ts
mqp: In SQL there are specific date/time types that support date/time operations (see here: https://www.postgresql.org/docs/current/datatype-datetime.html) and reading/writing them is conventionally via ISO strings.

functions/supabase/seed.sql
mqp: This inserts a new row into the replicated table with the ID, data, and timestamp provided. If there is already a row with that ID (i.e. a write for that document is already in the system) then it updates it iff the new update time is later than the update time on that existing row.

mqp:
The part where it replays failed writes is in, so if I have correctly reasoned, then this should be a complete solution for having an eventually consistent replica with no errors.

mqp:
I think this basically works, although I don't know if the kinks are worked out. I'm going to deploy it onto dev and poke around at it a lot to exercise it and become confident that it works, and fix any problems. Then I'll deploy it onto prod. Then we can start writing real queries to Supabase in our app.

PR #1256, opened 2022-11-28
Make contract metrics good for the front end

IanPhilips:
updates contract metrics onCreateBet, create and backfill contractMetrics with denormalized user data

### DIFF
--- a/functions/src/on-create-bet.ts
+++ b/functions/src/on-create-bet.ts
@@ -317,7 +317,8 @@ const updateContractMetrics = async (contract: Contract, user: User) => {
   const bets = betSnap.docs.map((doc) => doc.data() as Bet)
   const newMetrics = calculateMetricsByContract(
     { [contract.id]: bets },
-    { [contract.id]: contract }
+    { [contract.id]: contract },
+    user
   )
 
   console.log('newMetrics', newMetrics)
--- a/functions/src/on-create-bet.ts
+++ b/functions/src/on-create-bet.ts
@@ -76,11 +76,13 @@ export const onCreateBet = functions
     // They may be selling out of a position completely, so only add them if they're buying
     if (bet.amount > 0 && !bet.isSold)
       await addUserToContractFollowers(contractId, bettor.id)
-    await updateBettingStreak(bettor, bet, contract, eventId)
     await updateUniqueBettorsAndGiveCreatorBonus(contract, eventId, bettor)
     await notifyFills(bet, contract, eventId, bettor)
-    await handleReferral(bettor, eventId)
     await updateContractMetrics(contract, bettor)
+    // Referrals should always be handled before the betting streak bc they both use lastBetTime
+    handleReferral(bettor, eventId).then(async () => {
+      await updateBettingStreak(bettor, bet, contract, eventId)
+    })
   })
 
 const updateBettingStreak = async (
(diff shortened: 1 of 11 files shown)


mqp:
I think this is probably a good change for right now, although I don't know if it's going to scale well for very long. But why are you still recomputing them in `updateUserMetrics`?

functions/src/on-create-bet.ts
```diff
@@ -305,3 +307,23 @@ const notifyFills = async (
 const currentDateBettingStreakResetTime = () => {
   return new Date().setUTCHours(BETTING_STREAK_RESET_HOUR, 0, 0, 0)
 }
+
+const updateContractMetrics = async (contract: Contract, user: User) => {
+  const betSnap = await firestore
+    .collection(`contracts/${contract.id}/bets`)
+    .where('userId', '==', user.id)
+    .get()
```
mqp, line 315: Does this include the current bet that triggered the function? It does, doesn't it?

functions/src/scripts/backfill-contract-metrics-user-data.ts
```diff
@@ -0,0 +1,64 @@
...
+      await Promise.all(
+        contractMetricsWithoutUserData.map(async (cm) => {
+          await firestore
+            .doc(`users/${u.id}/contract-metrics/${cm.contractId}`)
+            .set({
+              ...cm,
+              userId: user.id,
+              userUsername: user.username,
+              userName: user.name,
+              userAvatarUrl: user.avatarUrl,
```
mqp, line 45: This needs to be fixed up when user info changes.

web/hooks/use-contract-metrics.ts
```diff
@@ -0,0 +1,62 @@
...
+export function listenForBinaryContractMetricsOnContract(
+  contractId: string,
+  count: number,
+  outcome: 'YES' | 'NO',
+  setMetrics: (metrics: ContractMetric[]) => void
+) {
+  const yesQuery = query(
+    collectionGroup(db, 'contract-metrics'),
+    where('contractId', '==', contractId),
+    where(outcome === 'YES' ? 'hasYesShares' : 'hasNoShares', '==', true),
```
mqp, line 56: Are these fields useful? Could the query equally well just be checking for a positive number of shares? Just curious.

mqp:
I'll be waiting on this PR with the presumption that it will cause `lastBetTime` to get updated immediately upon bets, and then I'll likely be getting rid of all the precomputed `uniqueBettor` fields (at least `uniqueBettorIds` which is very heavy) by querying `lastBetTime`.

functions/src/on-create-bet.ts
IanPhilips: I think it must, too!

functions/src/scripts/backfill-contract-metrics-user-data.ts
IanPhilips: Yeah :(

web/hooks/use-contract-metrics.ts
IanPhilips: I will try that! If that works then that would solve the FR case

IanPhilips:
> I think this is probably a good change for right now, although I don't know if it's going to scale well for very long. But why are you still recomputing them in `updateUserMetrics`?

Just haven't gotten around to thinking about that yet, but thanks for the tip! I'll remove it there.

web/hooks/use-contract-metrics.ts
IanPhilips: Okay, so there's a clear path for non-binary queries now but the above queries are useful bc they cut down on the filtering needed by the client

PR #1260, opened 2022-11-29
hotfix market grid ssr (wip)

sipec:
renders the market grid as iframe. I am really not happy with this solution

we could merge it if we need to hotfix the world cup dashboard but otherwise I'd hold out on actually figuring out how to render the react node views server side


sipec:
no longer needed because of #1282

PR #1262, opened 2022-11-29
Optimize size of emoji list (stop using tiptap extension)

sipec:
Now the emoji list is only 51K


web/components/editor/emoji/emoji-extension.ts
```diff
@@ -2,28 +2,19 @@ import { Extension } from '@tiptap/core'
...
-const emojis: EmojiItem[] = [
-  ...ttEmojis,
-  {
-    name: 'mana',
-    shortcodes: ['m$, mana'],
-    tags: ['money'],
-    emoji: ENV_CONFIG.moneyMoniker,
-  },
-]
+// {
```
akrolsmir, line 8: Let's remove this unused code now

akrolsmir (approved):
Looks good, thanks! I'm a bit sad that things like :) won't automatically convert to 😄 anymore, but it's not that important (we can patch it in later, if we want)

mqp:
Thanks for improving this, it was on my list of "stuff I wanted to deal with at some point."

web/components/editor/emoji/emoji-suggestion.ts
```diff
@@ -21,26 +26,41 @@ const rank = (c: string) => {
   return r < 0 ? 101 : r
 }
 
+const emojiArr = Object.entries(invertBy(shortcodes)).map(
```
mqp, line 29: Kind of odd to store the file as a `shortcode: codepoint` map if you are immediately going to transform it into a `codepoint: [shortcodes]` list. Why not just store it how you want it?

PR #1267, opened 2022-11-29
Split contract metrics and positions into separate docs

mqp:
Still kind of WIP, I probably made some mistake and I haven't tested yet. It's also a bit of a mess. I also need to write a backfill script.

Now `users/{userId}/contract-positions/{contractId}` contains real-time updated information on user share balances on contracts, whereas `users/{userId}/contract-metrics/{contractId}` contains the traditional batch information that is computed every 15 minutes.

### DIFF
--- a/common/calculate.ts
+++ b/common/calculate.ts
@@ -158,13 +158,53 @@ function getDpmInvested(yourBets: Bet[]) {
   })
 }
 
+export function getContractPositions(yourBets: Bet[]) {
+  let loan = 0
+  const totalShares: { [outcome: string]: number } = {}
+
+  for (const bet of yourBets) {
+    const { loanAmount, shares, sharesByOutcome, outcome } = bet
+
+    if (sharesByOutcome) {
+      for (const [o, s] of Object.entries(sharesByOutcome)) {
+        totalShares[o] = (totalShares[o] ?? 0) + s
+      }
+    } else {
+      totalShares[outcome] = (totalShares[outcome] ?? 0) + shares
+    }
+
+    loan += loanAmount ?? 0
+  }
+
+  const hasShares = Object.values(totalShares).some(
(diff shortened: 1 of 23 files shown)


common/calculate.ts
```diff
@@ -158,13 +158,53 @@ function getDpmInvested(yourBets: Bet[]) {
...
+
+  const { YES: yesShares, NO: noShares } = totalShares
+  const hasYesShares = yesShares >= 1
+  const hasNoShares = noShares >= 1
+  const lastBetTime = Math.max(...yourBets.map((b) => b.createdTime))
+  const maxSharesOutcome = hasShares
+    ? maxBy(Object.keys(totalShares), (outcome) => totalShares[outcome])
+    : null
+
+  return {
```
jahooma, line 191: I think `invested` should be included as part of your position and not as a metric, because it's not affected by other people's bets.

web/hooks/use-saved-contract-metrics.ts
```diff
@@ -1,36 +1,52 @@
...
+import {
+  useUser,
+  useUserContractMetrics,
+  useUserContractPositions,
+} from './use-user'
 import { useEffectCheckEquality } from './use-effect-check-equality'
 import { safeLocalStorage } from 'web/lib/util/local'
 
-export const useSavedContractMetrics = (
+export const useSavedContractInfo = (
```
jahooma, line 19: Thanks for updating this. I know the code's not great

jahooma (approved):
Cool, thanks Marshall! This design seems much better.

I wonder if we could not load all the bets in `update-user-metrics` and just use user positions. Seems like it could work for contract-metrics and loan calculations?

mqp:
I'm not going to get around to testing and merging this anytime soon.

PR #1268, opened 2022-11-30
Use points instead of bets on binary & pseudonumeric graphs (inlcuding embeds). Always include user in contract leaderboards

IanPhilips:
It looks like this halves the large graphs' (e.g. will donald trump tweet, will destiny get to 600k subs) data from ~4mb to ~1.5mb. We lose the avatar url on the graph tooltip but the probability that you see an avatar you know is probably ~1% anyways so this seems fine. This works great for embeds bc you don't likely know the avatars either.

### DIFF
--- a/web/components/charts/contract/binary.tsx
+++ b/web/components/charts/contract/binary.tsx
@@ -3,7 +3,6 @@ import { last, sortBy } from 'lodash'
 import { scaleTime, scaleLinear } from 'd3-scale'
 import { curveStepAfter } from 'd3-shape'
 
-import { Bet } from 'common/bet'
 import { getProbability, getInitialProbability } from 'common/calculate'
 import { BinaryContract } from 'common/contract'
 import { DAY_MS } from 'common/util/time'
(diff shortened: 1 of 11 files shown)


web/pages/[username]/[contractSlug].tsx
```diff
@@ -75,9 +88,10 @@ export async function getStaticPropz(props: {
   return {
     props: {
       contract,
-      bets,
+      bets: bets.slice(0, 100),
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:91]: I think the client assumes ascending order of bets? You could reverse this.

Or, change the client to assume descending order of bets! (i.e. listen for bets created after the first)

web/components/charts/contract/binary.tsx
```diff
@@ -16,28 +15,29 @@ import {
...
+const BinaryChartTooltip = (
+  props: TooltipProps<Date, HistoryPoint<BetPoint>>
+) => {
   const { prev, x, xScale } = props
   const [start, end] = xScale.domain()
   const d = xScale.invert(x)
   if (!prev) return null
   return (
     <Row className="items-center gap-2">
-      {prev.obj && <Avatar size="xs" avatarUrl={prev.obj.userAvatarUrl} />}
```
jahooma, line 40: I'll miss the avatar. Seems like we should be able to keep it for small markets?

Could include the avatar url as optional in BetPoint and set it if there are e.g. less than 500 bets.

web/pages/[username]/[contractSlug].tsx
```diff
@@ -63,8 +66,18 @@ export async function getStaticPropz(props: {
   const contract = (await getContractFromSlug(contractSlug)) || null
   const contractId = contract?.id
   const bets = contractId
-    ? await listBets({ contractId, limit: 4000, ...CONTRACT_BET_FILTER })
+    ? await listBets({
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:69]: With no limit, seems like we could still hit the Vercel size limit and have static props error out.

Maybe just increase the limit. 10k or 12k?

web/pages/[username]/[contractSlug].tsx
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:91]: Another thing is that that some parts of the client may assume it has all the bets. I think you may need to update, e.g. the contract leaderboards code, or the bets tab to not show an incorrect count.

These cases seem blocking to me.

jahooma (changes requested):
Awesome change, Ian! This will speed up loading markets and decrease our firebase bill.

I have a few concerns about how the client will be handling the subset of bets passed to it.

web/components/charts/contract/binary.tsx
mqp: I think it's probably worth keeping in. Something I mentioned to Ian is that another Actually Good Change for the whole website would be to make avatar URLs a function of the user ID, which would be very easy, and then we wouldn't have to be storing and sending bulky avatar URLs around everywhere.

Of course, with 10k bets, then a 20 character user ID with every bet is still 200k characters coming over the wire. So at some point we are just going to have to redesign the charts to not download every data point from forever by default.

mqp:
I thought the idea was to make the contract page not need all the bets at once at all. This change is a lot less appealing if the contract page still needs to load all the bets.

web/pages/[username]/[contractSlug].tsx
Comment (IanPhilips) [reply web/pages/[username]/[contractSlug].tsx:69]: good idea

IanPhilips:
Actually lol can't we get the contract leaderboard from the positions?

web/pages/[username]/[contractSlug].tsx
Comment (IanPhilips) [reply web/pages/[username]/[contractSlug].tsx]: just reversed it, thanks!

IanPhilips:
Comes with new and improved contract leaderboards!
Screenshot 2022-11-30 at 7 57 38 AM

web/pages/[username]/[contractSlug].tsx
Comment (IanPhilips) [reply web/pages/[username]/[contractSlug].tsx]: done!

web/components/charts/contract/binary.tsx
IanPhilips: I'm now adding the avatar for contracts with less than 1000 bets. I could optionally include some avatars in the most recent 1000 bet points 🤔

IanPhilips:
I think I'll just merge this and add pagination once/if people complain

PR #1274, opened 2022-12-01
Migrate placebet to Vercel function

jahooma:
Copied the place bet logic into `api/v0/bet`.

Copied most of the machinery Marshall made for setting up API endpoints.

Based in part on Austin's PR: https://github.com/manifoldmarkets/manifold/pull/1104

Not quite finished!

- Still need to run it in local dev. Not sure how to get the right environment variable with the firebase secret.


web/components/bet/bet-panel.tsx
```diff
@@ -229,7 +230,7 @@ export function BuyPanel(props: {
   }
 
   async function submitBet() {
-    if (!user || !betAmount) return
+    if (!user || !betAmount || !outcome) return
```
mqp, line 233: What case is this covering? Shouldn't the bet placement UI be disabled when they haven't picked an outcome?

web/lib/api/call-endpoint.ts
```diff
@@ -0,0 +1,21 @@
+import { auth } from 'web/lib/firebase/users'
+import { PlaceBetParams } from 'web/pages/api/v0/bet'
+
+export function placeBet(params: PlaceBetParams) {
+  return call('/api/v0/bet', params)
+}
```
mqp, line 6: I guess no matter what direction we go with running these API endpoints, we could put some types for them in `common` and use them like this.

web/components/bet/bet-panel.tsx
jahooma: Oh, I think the submit button is disabled so it wouldn't happen, but the type of outcome is `string | undefined`, so I added this case to appease the types (plus it's a good idea anyway).

web/lib/api/endpoint.ts
```diff
@@ -0,0 +1,85 @@
...
+  return async (req: NextApiRequest, res: NextApiResponse) => {
+    try {
+      initAdmin()
+      await applyCorsHeaders(req, res, {
+        origin: [
+          CORS_ORIGIN_MANIFOLD,
+          CORS_ORIGIN_VERCEL,
+          CORS_ORIGIN_LOCALHOST,
+        ],
+        methods: 'POST',
```
mqp, line 26: I think you need `allowedHeaders: 'Authorization'` here. But I don't know why it was working at all without it.

web/lib/api/call-endpoint.ts
jahooma: Yeah, I aways thought it was bad that we didn't type our api calls.

web/pages/api/v0/bet/index.ts
```diff
@@ -1,24 +1,308 @@
...
-    methods: 'POST',
-    allowedHeaders: 'Authorization',
+  DocumentReference,
+  FieldValue,
+  Query,
+  Transaction,
+} from 'firebase-admin/firestore'
+import { groupBy, mapValues, sumBy, uniq } from 'lodash'
+
+import { redeemShares } from 'functions/src/redeem-shares'
```
jahooma, line 11: Not sure what to do about this. Could copy the redemption code here.

Or, if we move it into `common`, then common has to depend on Firebase admin, which seems not ideal.

I could also add a TODO, because it won't be a problem if we decide to not proceed with Vercel functions. (Or, if we decide to migrate them all.)

mqp, line 11: Here's the import I was complaining about in our call.

web/lib/api/endpoint.ts
jahooma: Thanks. I haven't got as far as testing it yet haha.

web/pages/api/v0/bet/index.ts
```diff
@@ -1,24 +1,308 @@
...
+import { log, validate } from 'web/lib/api/utils'
+import { newEndpoint } from 'web/lib/api/endpoint'
+
+export const config = { api: { bodyParser: true } }
+
+const placeBet = newEndpoint(async (req, userId: string) => {
+  log(`Inside endpoint handler for ${userId}.`)
+  const { amount, contractId } = validate(bodySchema, req.body)
+
+  const result = await firestore.runTransaction(async (trans) => {
```
mqp, line 36: Do you happen to know what the typical latency between Vercel and Firestore is? That's another kind of potential pitfall with running this as a Vercel function, since we make many queries in serial. If latency between Vercel and Firestore were 50ms round trip, then that would be an increase of 30ms over the status quo. It seems that would make all binary contract bets take 100-200ms longer, which would be pretty sad. (Note that a round trip is necessary to start the transaction.)

If it's true that transaction retries are a big problem for us, increased latency inside the transaction will severely exacerbate that problem.

web/pages/api/v0/bet/index.ts
```diff
@@ -1,24 +1,308 @@
...
+
+const numericSchema = z.object({
+  outcome: z.string(),
+  value: z.number(),
+})
+
+export type PlaceBetParams = z.infer<
+  typeof bodySchema &
+    (typeof binarySchema | typeof freeResponseSchema | typeof numericSchema)
+>
```
mqp, line 232: I guess we could dream of a more precise type here, since this isn't capturing the correlation between the body content and the type of the contract-type-specific parameters.

web/pages/api/v0/bet/index.ts
```diff
@@ -1,24 +1,308 @@
...
+
+const firestore = admin.firestore()
```
mqp, line 234: Running this on file import is an anti-pattern. It means that nobody can import this file unless the admin SDK is already initialized. Recommend moving it into functions. There's no work here that needs to be "cached".

web/components/bet/bet-panel.tsx
mqp: For just appeasing the types I suggest using the Typescript `!` non-null assertion operator. Whether you think it's good style to handle it here I leave up to your judgement. I am guessing I would not put it, because I think it's confusing to readers for the code to defensively handle cases that are logically impossible.

web/pages/api/v0/bet/index.ts
mqp: I would say it's totally unacceptable for `common` to depend on `firebase-admin`, because that would pull it into our website bundle. I think you should just factor out the "meat" of the redemption code as appropriate, or since this is just an experiment for now just copying it seems fine. That's much better than having this weird cross-dependency.

PR #1275, opened 2022-12-01
Replace generateHTML with custom tiptap parser

sipec:
Fixes tweets, market grids
Prettier %mentions, at mentions, spoilers
Regression: formatting in spoilers chops it up

possible improvements:
- [x] dedupe node view logic
- error boundaries? this code feels very fragile :/


jahooma:
Thanks for working on this fix! It does seem a bit complex to render tiptap server side. Hopefully with this, the main issues will be solved...

PR #1276, opened 2022-12-01
Ingawei/man 463 notifications user suggestions

ingawei:
image
image

made it so notifications don't automatically become 'read' upon clicking. I think this was bugged anyhow, as it seemed to mark notifications that weren't even visible as read after refresh.

There are 3 ways to mark a notification as read:
1. Mark all as read button
2. click the notification
3. choose to mark it as read in the '...' menu

Setting up this '...' button next to notifications to allow notification settings configuration on notifications page later, but that's for future Inga to worry about

### DIFF
--- a/web/components/notification-settings.tsx
+++ b/web/components/notification-settings.tsx
@@ -195,70 +195,7 @@ export function NotificationSettings(props: {
     const [emailEnabled, setEmailEnabled] = useState(previousEmailValue)
     const [mobileEnabled, setMobileEnabled] = useState(previousMobileValue)
     const [error, setError] = useState<string>('')
-    const loading = 'Changing Notifications Settings'
-    const success = 'Changed Notification Settings!'
     const highlight = navigateToSection === subscriptionTypeKey
-
-    const attemptToChangeSetting = (
-      setting: 'browser' | 'email' | 'mobile',
-      newValue: boolean
-    ) => {
-      const necessaryError =
-        'This notification type is necessary. At least one destination must be enabled.'
-      const necessarySetting =
-        NOTIFICATION_DESCRIPTIONS[subscriptionTypeKey].necessary
-      if (
-        necessarySetting &&
-        setting === 'browser' &&
-        !emailEnabled &&
-        !newValue
-      ) {
-        setError(necessaryError)
-        return
-      } else if (
-        necessarySetting &&
-        setting === 'email' &&
-        !inAppEnabled &&
-        !newValue
-      ) {
-        setError(necessaryError)
-        return
-      }
-      // Mobile notifications not included in necessary destinations yet
-
-      changeSetting(setting, newValue)
-    }
-
-    const changeSetting = (
-      setting: 'browser' | 'email' | 'mobile',
-      newValue: boolean
-    ) => {
-      toast
-        .promise(
-          updatePrivateUser(privateUser.id, {
-            notificationPreferences: {
-              ...privateUser.notificationPreferences,
-              [subscriptionTypeKey]: destinations.includes(setting)
-                ? destinations.filter((d) => d !== setting)
-                : uniq([...destinations, setting]),
-            },
-          }),
-          {
-            success,
-            loading,
-            error: 'Error changing notification settings. Try again?',
-          }
-        )
-        .then(() => {
-          if (setting === 'browser') {
-            setInAppEnabled(newValue)
-          } else if (setting === 'email') {
-            setEmailEnabled(newValue)
-          } else if (setting === 'mobile') {
-            setMobileEnabled(newValue)
-          }
-        })
-    }
 
     return (
       <Row
@@ -275,7 +212,19 @@ export function NotificationSettings(props: {
             {!browserDisabled.includes(subscriptionTypeKey) && (
               <SwitchSetting
                 checked={inAppEnabled}
-                onChange={(newVal) => attemptToChangeSetting('browser', newVal)}
+                onChange={(newVal) =>
+                  attemptToChangeSetting(
+                    'browser',
+                    newVal,
+                    subscriptionTypeKey,
+                    privateUser,
+                    destinations,
+                    setInAppEnabled,
+                    emailEnabled,
+                    inAppEnabled,
+                    setError
+                  )
+                }
                 label={'Web'}
                 disabled={optOutAll.includes('browser')}
               />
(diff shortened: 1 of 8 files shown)


web/components/notification-settings.tsx
```diff
@@ -477,3 +450,100 @@ export function NotificationSettings(props: {
     </div>
   )
 }
+
+export const notificationIsNecessary = (
```
IanPhilips, line 454: Why the move and export if it's still only used in this file?

web/components/notification-settings.tsx
```diff
@@ -477,3 +450,100 @@ export function NotificationSettings(props: {
...
+  ) {
+    if (setError) {
+      setError(necessaryError)
+    }
+    return true
+  }
+  return false
+}
+
+export const attemptToChangeSetting = (
```
IanPhilips, line 485: Same here, why is it exported?

web/components/notification-settings.tsx
```diff
@@ -477,3 +450,100 @@ export function NotificationSettings(props: {
...
+    setting,
+    newValue,
+    privateUser,
+    subscriptionTypeKey,
+    destinations,
+    setEnabled
+  )
+}
+
+export const changeSetting = (
```
IanPhilips, line 520: Another export

web/components/notifications/income-summary-notifications.tsx
```diff
@@ -89,7 +89,7 @@ export function IncomeNotificationGroupItem(props: {
   const combinedNotifs = combineNotificationsByAddingNumericSourceTexts(
     notifications.filter((n) => n.sourceType !== 'betting_streak_bonus')
   )
-  const [highlighted, setHighlighted] = useState(
+  const [highlighted, _setHighlighted] = useState(
```
IanPhilips, line 92: is this `_setHighlighted` used somewhere?

web/components/notifications/notification-helpers.tsx
```diff
@@ -47,7 +51,7 @@ export function PrimaryNotificationLink(props: {
   }
   return (
     <span className="font-semibold transition-colors hover:text-indigo-500">
-      {truncateText(text, truncatedLength ?? 'xl')}
+      {truncatedLength ? truncateText(text, truncatedLength ?? 'xl') : text}
```
IanPhilips, line 54: I don't think the truncations in the middle of the text is helpful. I feel pretty strongly that it should only truncate text at the end, and I believe James agrees.

web/pages/notifications.tsx
```diff
@@ -56,46 +62,71 @@ export default function Notifications() {
...
     <Page>
       <div className={'px-2 pt-4 sm:px-4 lg:pt-0'}>
-        <Title text={'Notifications'} className={'hidden md:block'} />
+        <Row
+          className={clsx('w-full justify-between', isMobile ? 'hidden' : '')}
+        >
+          <Title text={'Notifications'} className="grow" />
+          <div className="my-auto">
+            {!isMobile && activeIndex === 0 && (
+              <MarkAllAsReadButton
```
IanPhilips, line 71: Hmmmm I am pretty unsure that we need this functionality. I think it's pretty good that notifications are assigned as read on reaching the notification page. This mirrors how twitter's notifications work. I get dozens of notifications per day and I do not want to have to click anything to mark them as read.

web/pages/notifications.tsx
```diff
@@ -237,65 +265,52 @@ function NotificationGroupItem(props: {
...
   const onExpandHandler = (event: React.MouseEvent<HTMLDivElement>) => {
     if (event.ctrlKey || event.metaKey) return
     setExpanded(!expanded)
   }
 
-  useEffect(() => {
-    if (expanded) setHighlighted(false)
-  }, [expanded, setHighlighted])
-
+  const shownArray = expanded
```
IanPhilips, line 295: > grouped notifications now previews all unread notifications, even if unread notifications exceeds 3 notifications

I don't think it's a good idea to show more than 3 unread notifications on a market at a time. Otherwise markets with a lot of activity will take up the whole page.

IanPhilips (commented):
Inga, thanks for the style updates and experimenting with new notifications designs! 

I do have concerns about the new proposed functionalities, though. I personally don't want to have to click a button to mark my notifications as read, and we previously chose to only show 3 unread notifications per grouping because highly active markets can blow up your notifications page. Am I interpreting your proposal as previewing all unread notifications per market grouping, i.e. 10, 20, etc.?

Also, I still think we should do away with the mid-line truncation. I think truncation should only happen at the end of the line.

IanPhilips:
> made it so notifications don't automatically become 'read' upon clicking. I think this was bugged anyhow, as it seemed to mark notifications that weren't even visible as read after refresh.

We decided to mark all notifications as read, even those that weren't visible, bc if a user has more notifications than fit on a page they'll have to go back in history and 'view' all of them to silence the notification bell. Your new button would fix this but as I said above, I'd rather not have to click something every time I visit the notifications page.

jahooma:
Agree with all of Ian's comments.

Clicking a button to mark as read helps you carefully look through every notification you missed, but I think users would much prefer to have them automatically marked read. Notifications are not as important as, like, email.

I don't think anyone wants to mark individual notifications as read/unread either. So maybe even the '...' menu is not useful.

I'm also not sure about showing all new notifications — just think of a Destiny market where 50 people write spammy comments. RIP, your notifications inbox. It means you might not see other important notifications from other markets.

web/components/notification-settings.tsx
ingawei: I want to use it in the future for setting preferences from the notifications tab

ingawei:
Reverted double read/seen notifications, all notifications are marked as seen upon entering page and also am keeping just 3 notifications regardless of how many are read within a group. Good catch I didn't think about if a market is exploding haha

@IanPhilips could you review the notification preferences logic and make sure it's sound?
also there's some weirdness going on with the new Like notifications, it's showing above all others even though it occurred earlier. I think it also needs to be added to preferences
image

IanPhilips:
I was kind of interested in trying out the facebook 2-read style notifications! What led you to abandon that track?

ingawei:
> I was kind of interested in trying out the facebook 2-read style notifications! What led you to abandon that track?

I agreed with everyone in that it's tedious to have to press a button to mark everything as read. I implemented and played around with it and idk it just felt extra confusing. I think with FB it works because your unread notifications will always be at the top... they don't have grouped notifications/daily income hierarchy and stuff like that. Notifications are already pretty complicated so I was meh about it.

Also I think the only benefit from having a notification marked as read/unread is if you could organize/sort it. FB has an unread notifications section. Maybe if we implemented that I would feel differently but having grouped notifications also complicates that

common/user-notification-preferences.ts
```diff
@@ -203,6 +203,7 @@ const notificationReasonToSubscriptionType: Partial<
   resolution_on_contract_with_users_comment: 'resolutions_on_watched_markets',
   reply_to_users_answer: 'all_replies_to_my_answers_on_watched_markets',
   reply_to_users_comment: 'all_replies_to_my_comments_on_watched_markets',
+  loan_income: 'loan_income',
```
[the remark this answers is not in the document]
ingawei: oh weird. It didn't seem to be working until I added it?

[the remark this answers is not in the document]
IanPhilips: Could you not unsubscribe? I've recently tested this with different preferences and you shouldn't need to add anything there and, but if it's not working that's a separate bug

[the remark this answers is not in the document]
ingawei: oh no I was just implementing it dumb, didn't know notificationToPreferences wasn't for all of them. Thanks Ian!

PR #1279, opened 2022-12-01
Display error and disable submit for default amount if user has insufficient funds

### LINKED ISSUE #1277: Submit bet button is enabled when selecting an outcome even if the user has insufficient funds
The button is enabled after an outcome is selected with the default bet of 10, even though the user has less than 10 in funds. Clicking the submit button triggers an APIError.
chrome_l4WC2xKwaM chrome_KINnn9Q98X

This happens for binary markets, numeric markets, and when creating a new answer for a multiple-choice market.

SubsequentlySneeds:
Fixes #1277 

**Previous behavior:** If the user has less than the default bet amount (10), and selects an outcome without changing the bet amount, the submit button is enabled and no error is shown (until the user submits the bet, which returns an APIError).

**New behavior:** As soon as the bet panel is "ready" (i.e. outcome or bucket selected), errors are displayed and the submit button is disabled if there is an error, regardless of whether the user has manually changed the bet amount.

chrome_XvJLzRmlGU

When creating a new answer for a multiple choice market, users with less than 10 balance will see "Not enough funds" as soon as they start typing their answer, which is at least better than not notifying them until they try to submit. Maybe the default "new answer" bet could be changed to use the user's balance as the default if it is lower than 10.


mqp:
@jahooma FYI this seems modestly related to the bet UI code you were just looking at in #1274.

jahooma (approved):
Thanks, this looks like a reasonable fix!

PR #1282, opened 2022-12-01
Vert custom tiptap ssr - with bug fixes

sipec:
line breaks were breaking cuz they can't have children
now `undefined` in generateHTML produces nothing and children are only inserted if there's a `0`


sipec:
ok I'm reasonably sure I tested out all possible nodes now
https://prod-git-custom-ssr-tiptap-2-mantic.vercel.app/post/rich-text-test

PR #1284, opened 2022-12-02
Native market embeds

sipec:
Use react instead of iframes for the markets. Resolves MAN-86

I'm not sure if this is actually faster. There's no static props so each market loads its bets client-side. At least like the iframes there isn't any CLS because I gave it a fixed height.

### DIFF
--- a/common/util/parse.ts
+++ b/common/util/parse.ts
@@ -58,7 +58,7 @@ export const stringParseExts = [
       '[embed]' + node.attrs.src ? `(${node.attrs.src})` : '',
   }),
   skippableComponent('gridCardsComponent', '[markets]'),
-  skippableComponent('staticReactEmbedComponent', '[map]'),
+  skippableComponent('market-embed', '[market]'),
   TiptapTweet.extend({ renderText: () => '[tweet]' }),
   TiptapSpoiler.extend({ renderHTML: () => ['span', '[spoiler]', 0] }),
 ]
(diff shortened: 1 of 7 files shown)


sipec:
@IanPhilips fyi - wonder if you have any opinion on this since you just made the market page / market iframes faster with the bet points

IanPhilips:
Oh shoot why can't we use static props? The client loading all the bets is a very bad UX for popular markets.

mqp:
I am not a big fan of the static props stuff, but I am even less of a big fan of making static props our de facto high performance cache that we depend on, and then having things not use that cache.

Is there a reason you can't request the static props and use that to populate the embed, just like the embed page would? I don't really see the downside.

sipec:
@mqp afaik static props only works for pages, not for components. We'd have to fetch the data at the page level and pass it down, probably via a react Context. Alternatively, use [Next 13's server components](https://beta.nextjs.org/docs/rendering/server-and-client-components)

mqp:
> @mqp afaik static props only works for pages, not for components. We'd have to fetch the data at the page level and pass it down, probably via a react Context. Alternatively, use [Next 13's server components](https://beta.nextjs.org/docs/rendering/server-and-client-components)

Static props just means that Next.js generates JSON files on their webserver, why can't you fetch the JSON file and populate the component props with it? e.g. https://manifold.markets/_next/data/L1-2ogSb8FrmwfASnagY4/Yev/will-the-biggest-shareholder-own-no.json

I am not sure how to get the Next.js build ID at runtime but it can't be too hard.

sipec:
closing. we should re-do this with /app if we do it all

PR #1285, opened 2022-12-02
WIP continued Supabase improvements

mqp:
Continuing to productionize this work and add tools for monitoring, failure recovery, etc.

### DIFF
--- a/functions/supabase/seed.sql
+++ b/functions/supabase/seed.sql
@@ -51,16 +51,15 @@ create table if not exists incoming_writes (
   processed boolean not null default false
 );
 alter table incoming_writes enable row level security;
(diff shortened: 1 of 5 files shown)


functions/supabase/seed.sql
```diff
@@ -69,33 +74,75 @@ begin
...
   end if;
-  new.processed := true;
+  return true;
+end
+$$;
+
+create or replace function replicate_writes_process_all()
+  returns table(succeeded boolean, n bigint)
+  language plpgsql
+as
```
mqp, line 117: This can be manually used to recover from the case where Firestore writes made it to the Supabase `incoming_writes` table, but failed to be replicated to the data tables. This will only happen if we wrote a bug (either in the trigger or in the cloud function populating `incoming_writes`) but we still need to be able to fix it when it happens.

PR #1288, opened 2022-12-02
Use supabase search for at mentions

sipec:
more of a proof of concept, dipping toes in water. Resolves MAN-473

- [x] replicate prod db, and test this on it
- [ ] supabase types using https://supabase.com/docs/reference/javascript/typescript-support
- [x] env vars for api keys?

later maybe
- use full text search instead of `like` operator.

@jahooma @mqp fyi


web/lib/supabase/stuff.ts
```diff
@@ -0,0 +1,19 @@
...
+const SUPABASE_READ_KEY =
+  'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6Im1mb2RvbnpueWZ4bGxjZXp1ZmdyIiwicm9sZSI6ImFub24iLCJpYXQiOjE2Njc5ODgxNjcsImV4cCI6MTk4MzU2NDE2N30.RK8CA3G2_yccgiIFoxzweEuJ2XU5SoB7x7wBzMKitvo'
+
+const supabase = createClient(DEV_CONFIG.supabaseUrl ?? '', SUPABASE_READ_KEY)
+
+export async function searchUsers(prompt: string) {
+  const { data } = await supabase
+    .from('users')
+    .select('data')
+    .or(`data->>username.ilike.%${prompt}%,data->>name.ilike.%${prompt}%`)
```
mqp, line 15: FWIW, since we only have like 10k users this will probably be fine, but you should get in the habit of adding appropriate SQL indexes to serve queries like this.

web/lib/supabase/stuff.ts
```diff
@@ -0,0 +1,19 @@
+import { createClient } from '@supabase/supabase-js'
+import { DEV_CONFIG } from 'common/envs/dev'
+
+// TODO: rewrite entirely. Set up types, put keys as env vars or set up row level access
+
+const SUPABASE_READ_KEY =
+  'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6Im1mb2RvbnpueWZ4bGxjZXp1ZmdyIiwicm9sZSI6ImFub24iLCJpYXQiOjE2Njc5ODgxNjcsImV4cCI6MTk4MzU2NDE2N30.RK8CA3G2_yccgiIFoxzweEuJ2XU5SoB7x7wBzMKitvo'
+
+const supabase = createClient(DEV_CONFIG.supabaseUrl ?? '', SUPABASE_READ_KEY)
```
mqp, line 9: If you want to put the Supabase "anonymous access keys" into the dev and prod config, that would seem totally sensible to me.

web/lib/supabase/stuff.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export async function searchUsers(prompt: string) {
+  const { data } = await supabase
+    .from('users')
+    .select('data')
```
mqp, line 14: Is it not possible to select only the fields you want from inside the JSON doc using the JS API? That would be a big problem if so.

mqp:
The Supabase prod DB should be populated now, so you can try this out against that.

yarn.lock
```diff
@@ -1108,6 +1115,18 @@
     "@supabase/storage-js" "^2.0.0"
     cross-fetch "^3.1.5"
 
+"@supabase/supabase-js@2.1.1":
```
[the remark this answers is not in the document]
sipec: hmm I wonder if there is an easy way to have packages in `common` be automatically included `web` and `function`, so we wouldn't have to import packages like this 2 or 3 times.

PR #1293, opened 2022-12-03
Don't query resolved contracts unnecessarily

mqp:
Self-explanatory.


IanPhilips:
I thought isResolved was optional? Total assumption and not at my comp rn but that's why I didn't add that. If it's not then this is perfect!

IanPhilips (approved):
Looks like `isResolved` is required so this is perfect! Thanks for looking into this.

PR #1294, opened 2022-12-03
Use `array.some` instead of `.map` followed by `.includes`

ForrestWeiswolf:
Avoids iterating through the same array twice. Also slightly clearer to read, in my opinion.


jahooma (approved):
Thanks for the improvement! Will merge this.

Unfortunately, for outside contributions like this, I can't know how much to trust you, so I have to check every line. For a small style improvement, that means it isn't actually worth the time cost. So, just a heads up, future PRs like this may be closed.

PR #1301, opened 2022-12-07
Recommended markets page [WIP]

IanPhilips:
Result of my hackathon project. Shows recommends markets similar to those you've bet on and viewed (via their contract page), showing 100 markets by similar creators, similar groups, similar bettors' interests, but won't show you markets you've already bet on or viewed. Austin brought up the point that perhaps we shouldn't reshow you markets you've scrolled past too many times. Could be an interesting addition to add. Using this page we could start logging markets you've seen and not show them more than 2/3 times. 

It could go on the homepage and should probably have an infinite scroll.

Screenshot 2022-12-07 at 10 54 42 AM

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -46,6 +46,7 @@ service cloud.firestore {
                           // && (get(/databases/$(database)/documents/users/$(request.resource.data.referredByUserId)).referredByUserId == resource.data.id);
       match /events/{eventId} {
         allow create: if (request.auth == null && userId == 'NO_USER') || userId == request.auth.uid;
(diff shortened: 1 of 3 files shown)


IanPhilips:
Oh yeah James and I merged this expecting James to change the majority of it for the new homepage. I don't think any of that stuff will stay on the front end.

jahooma:
Yeah, the plan was for me to change it. Maybe a good idea to move to a separate file than use user, yeah.

But actually, because it is only loaded if you scroll to the bottom of the homepage, I just shipped it with my last commit. You can take a look!

PR #1306, opened 2022-12-08
Add quarterfinal markets to worldcup

SirSaltyy:


SirSaltyy:
@akrolsmir do u know why the masonry on this market is weird? The 3rd market is going under the market with the image rather than going into the empty space horizontal to it.

PR #1318, opened 2022-12-09
Ingawei/man 447 UI for likes including showing who liked

ingawei:
image

- access this by either clicking `& n more` in tooltip, or holding on mobile
image


web/components/widgets/tooltip.tsx
```diff
@@ -16,14 +16,15 @@ import { ReactNode, useRef, useState } from 'react'
...
 
 export function Tooltip(props: {
-  text: string | false | undefined | null
+  text: string | false | undefined | null | ReactNode
   children: ReactNode
   className?: string
   placement?: Placement
   noTap?: boolean
   noFade?: boolean
+  externalOpen?: boolean
```
ingawei, line 25: @sipec I added this to basically override the tooltip disappearing upon hovering over it, let me know if you have a less hacky way to do it lol

sipec: yes, you can use SafePolygon: https://floating-ui.com/docs/usehover#handleclose
you use it like `useHover(context, { mouseOnly: noTap, handleClose: safePolygon() }),`

web/components/multi-user-transaction-link.tsx
```diff
@@ -48,27 +48,41 @@ export function MultiUserTransactionModal(props: {
...
-              username={userInfo.username}
-              avatarUrl={userInfo.avatarUrl}
-            />
-            <UserLink name={userInfo.name} username={userInfo.username} />
-          </Row>
-        ))}
+      <Col className="relative items-start gap-4 rounded-md bg-white p-6">
+        <span className={'sticky top-0 text-xl'}>{modalLabel}</span>
+        {hasUsers && (
+          <Col className="max-h-96 w-full gap-4 overflow-y-scroll">
```
sipec, line 60: should be `overflow-y-auto`

image

sipec (commented):
left some comments. also `useLongTouch` is cool!

web/components/widgets/tooltip.tsx
```diff
@@ -72,11 +73,17 @@ export function Tooltip(props: {
       </span>
       {/* conditionally render tooltip and fade in/out */}
       <Transition
-        show={open}
+        show={externalOpen ? externalOpen : open}
```
jahooma, line 76: Equivalent to `externalOpen || open`

web/hooks/use-long-touch.ts
```diff
@@ -0,0 +1,59 @@
...
+export default function useLongTouch(
+  onLongTouch = () => {},
+  onClick = () => {},
+  ms = 500
+) {
+  const [startLongTouch, setStartLongTouch] = useState(false)
+  const [mouseState, setMouseState] = useState<
+    null | 'startPress' | 'endPress' | 'startTouch' | 'endTouch'
+  >(null)
+  const [timerId, setTimerId] = useState<
```
jahooma, line 18: I think this could be a `useRef` instead of `useState`, because you don't need to re-render. It's fine though.

web/hooks/use-long-touch.ts
```diff
@@ -0,0 +1,59 @@
...
+  useEffect(() => {
+    if (mouseState === 'endPress') {
+      onClick()
+    } else if (mouseState === 'endTouch') {
+      clearTimeout(timerId)
+      if (!startLongTouch) {
+        onClick()
+      }
+      setStartLongTouch(false)
+    } else if (mouseState === 'startTouch') {
```
jahooma, line 31: Is it possible to get multiple 'startTouch' events before an 'endTouch'? Would make the clearTimeout not work b/c of a lost timerId.

I guess I wouldn't worry too much if this code seems like it's working.

jahooma (approved):
Thanks Inga! Looking forward to using this feature.

I'm not a super fan of the long press, but happy to try it out. Seems like most mobile users wouldn't discover it.

web/hooks/use-long-touch.ts
ingawei: Good point! There was a case of that if you open the modal from your phone, added 'onMouseLeave' condition to fix it

PR #1319, opened 2022-12-09
Greatly ameliorate badness of user profile follow machinery

mqp:
Previously this was a kind of total disaster, as I wrote in Discord:

> 1. when you visit a user page it makes a `<ProfilePublicStats>`
> 2. the `<ProfilePublicStats>` makes a `<FollowersButton>` and a `<FollowingButton>`
> 3. both those buttons independently do this (i.e. all this work is duplicated twice. note this happens immediately on page load, it's not "if you click the button" or something)
>   a. call both `useFollows` and `useFollowers` to fetch and subscribe to the list of user IDs who are following and followers
>   b. make a `<FollowsDialog>` with those lists
>   c. the `<FollowsDialog>` calls `useDiscoverUsers`
>   d. `useDiscoverUsers` calls `useFollows` again to fetch and subscribe to the list of user IDs who are following the target
>   e. it also lists all the bets of the target (!!!) to get their contract IDs (at least it doesn't subscribe to them!)
>   f. it then gets all those contracts (!!!)
>   g. the `<FollowsDialog>` then makes `<FollowList>s` for both your follows and followers
>   h. each `<FollowList>` calls `useFollows` on you to fetch and subscribe to your followers
>   i. each `<FollowList>` also calls `useUserById` for each person in the list to fetch and subscribe to the user object for that person!
>   j. believe it or not, it does even more if it's your own profile, but i will cut it short here
>   k. note all of this work was done twice, every time you clicked a user profile, before you did anything on the page!!!!

This fixes a lot of that stuff, specifically:
- All the duplicated fetching is gone.
- The business where it calls `useDiscoverUsers` on other people's profile even though all that information is pointless is gone.

The main remaining performance problem with this is that it still loads all the user objects for the follow lists on page load. This should be ameliorated by denormalizing the user names and avatars into the follow documents, but I am busy with Supabase stuff so I'm not going to do that at this instant.


jahooma (approved):
LGTM, thanks for the cleanup! Will be good to not load so much data by default on the user page, haha.

PR #1323, opened 2022-12-10
Load only what we need on embed page

mqp:
Previously:

- The embed page would query the most recent 10k bets in `getStaticProps`.
- It would pass down to the client either: (for choice contracts) all the bets, or: (for other contracts) both the most recent 100 bets plus chart points for all bets.
- Regardless of what it got, the client would use `useBets` to make a subscription and extend its lists of bets and points to newer bets and points.
- The client would go on to render either a small embed card or a large embed chart. Notably, the embed page has no purpose for bets unless it's specifically rendering a choice contract chart.

Now:

- The embed page still queries all the bets.*
- It only passes down either (for choice contracts) all the bets, or (for other contracts) all the points.
- It doesn't call `useBets` -- I don't think it is important to embed users to have live-updating embeds, and embeds really want to be lightweight, because there may be many of them on a webpage.
- It renders the stuff with what it has.

*Technically for old numeric contracts it doesn't even need the bets.

So in the common case of "a random large binary contract embed", the new version basically saves downloading 100 bets in the static props.

There's still something pretty unfortunate, which is that we fetch the bets and/or points even for small "card" embeds which have no use for them. It would be much better to control the "card" embed presentation through the URL rather than through the measured size of the embed.

### DIFF
--- a/web/pages/embed/[username]/[contractSlug].tsx
+++ b/web/pages/embed/[username]/[contractSlug].tsx
@@ -2,7 +2,6 @@ import { Bet } from 'common/bet'
 import { Contract } from 'common/contract'
 import { DOMAIN } from 'common/envs/constants'
 import { useEffect } from 'react'
-import { first } from 'lodash'
 import {
   BinaryResolutionOrChance,
   ContractCard,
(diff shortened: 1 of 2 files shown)


web/pages/embed/[username]/[contractSlug].tsx
```diff
@@ -26,54 +25,55 @@ import { contractPath, getContractFromSlug } from 'web/lib/firebase/contracts'
...
-        (bet) =>
-          ({
-            x: bet.createdTime,
-            y: bet.probAfter,
-          } as HistoryPoint<Partial<Bet>>)
-      )
-    : []
-
+  if (contract == null) {
+    return { notFound: true, revalidate: 60 }
```
Comment (IanPhilips) [inline web/pages/embed/[username]/[contractSlug].tsx:72]: Thanks for removing all the conditionals! This is much cleaner

IanPhilips (approved):
This seems reasonable to me! I like light-er weight embeds especially when they come in groups like on Mantic Mondays.

PR #1327, opened 2022-12-13
added dropdown items

ingawei:
image

### DIFF
--- a/web/components/buttons/follow-market-button.tsx
+++ b/web/components/buttons/follow-market-button.tsx
@@ -74,16 +74,16 @@ export const FollowMarketButton = (props: {
 }
 
 export async function unfollowMarket(
-  contractid: string,
-  contractslug: string,
+  contractId: string,
+  contractSlug: string,
   user: User
 ) {
-  await unFollowContract(contractid, user.id)
+  await unFollowContract(contractId, user.id)
   toast("You'll no longer receive notifications from this market", {
     icon: <CheckIcon className={'h-5 w-5 text-teal-500'} />,
   })
   track('Unwatch Market', {
-    slug: contractslug,
+    slug: contractSlug,
   })
 }
 
--- a/web/components/buttons/follow-market-button.tsx
+++ b/web/components/buttons/follow-market-button.tsx
@@ -38,21 +38,9 @@ export const FollowMarketButton = (props: {
         onClick={async () => {
           if (!user) return firebaseLogin()
           if (followers?.includes(user.id)) {
-            await unFollowContract(contract.id, user.id)
-            toast("You'll no longer receive notifications from this market", {
-              icon: <CheckIcon className={'h-5 w-5 text-teal-500'} />,
-            })
-            track('Unwatch Market', {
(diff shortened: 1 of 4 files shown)


web/hooks/use-follows.ts
```diff
@@ -31,9 +31,11 @@ export const useFollowers = (userId: string | undefined) => {
...
-export const useContractFollows = (contractId: string) => {
+export const useContractFollows = (contractId: string | undefined) => {
   const [followIds, setFollowIds] = useState<string[] | undefined>()
-
+  if (!contractId) {
+    return
+  }
   useEffect(() => {
     return listenForContractFollows(contractId, setFollowIds)
   }, [contractId])
```
jahooma, line 41: Hmm, the early return would violate one of the rules of hooks: every code path must evaluate all the hooks.

https://reactjs.org/docs/hooks-rules.html

I would instead add an `if` within the `useEffect` to make sure `contractId` is defined.

web/components/notifications/notification-dropdown.tsx
```diff
@@ -73,15 +77,46 @@ function useNotificationPreferenceItem(notification: Notification) {
...
+
+function useNotificationFollowItem(notification: Notification) {
+  const sourceContractId = notification.sourceContractId
+  const sourceContractSlug = notification.sourceContractSlug
+  const followers = useContractFollows(sourceContractId)
+  const user = useUser()
+  if (!user || !sourceContractId || !sourceContractSlug || !followers) {
+    return []
+  }
+  const isFollowing = followers?.includes(user.id)
```
jahooma, line 100: `followers` is confusing if it refers to the contractIds you are watching.

Wait, I think `isFollowing` should be checking if your contract follows includes `sourceContractId`, not `user.id`

web/components/buttons/follow-market-button.tsx
```diff
@@ -84,3 +72,31 @@ export const FollowMarketButton = (props: {
     </Tooltip>
   )
 }
+
+export async function unfollowMarket(
+  contractid: string,
+  contractslug: string,
```
jahooma, line 78: I'd use camelCase here.

jahooma (approved):
Thanks for quickly building this, Inga! Our users will love it.

Just a couple comments for you to address.

web/components/notifications/notification-dropdown.tsx
ingawei: is that right? I thought followers is the list of people following a contract, so you're looking for yourself in that list

ingawei: yee it's how FollowMarketButton uses it
```
export const FollowMarketButton = (props: {
  contract: Contract
  user: User | undefined | null
}) => {
  const { contract, user } = props
  const followers = useContractFollows(contract.id)

  const watching = followers?.includes(user?.id ?? 'nope')
```

jahooma: Oh, you were right

PR #1330, opened 2022-12-14
Move Supabase replicator to separate GCR service

mqp:
1. If we try to use v2 Firestore trigger cloud functions to do all the actual Supabase calls, then our Supabase instance CloudFlare proxy/pg_bouncer can't handle all the connections at once during bursty times (e.g. metrics updates), so it will spin up a gazillion instances, drop a ton of writes, etc. and it's a huge pain in the ass.

2. If we try to use a v2 Firestore pub/sub subscriber function to handle Supabase calls, and have the Firestore trigger publish to it, that works way better, but unfortunately `firebase-tools` is broken right now and v2 pub/sub subscriber functions don't work right without a bunch of effort.

Since we don't like Firebase functions anyway, I decided I would just switch this replication stuff over to a separate Cloud Run service that can subscribe to the Firestore write pub/sub topic. I think this will work well and not have any big problem, and it's good practice for my plan to switch the API over to a very similar looking structure and deployment code running on GCP.


mqp:
This isn't deployed and hooked up quite yet, but merging this so that the organizational changes are in.

PR #1335, opened 2022-12-15
Track card views via callback

IanPhilips:
Card views are only being tracked in the your feed section, and filtering them out after they've been seen 3 times.


web/hooks/use-is-visible.ts
```diff
@@ -0,0 +1,31 @@
...
+  useEffect(() => {
+    const element = ref.current
+    if (!element) return
+
+    const observer = new IntersectionObserver(
+      (entries) => {
+        entries.forEach((entry) => {
+          if (entry.isIntersecting) {
+            cb()
+            if (once) {
```
IanPhilips, line 14: We might not need this once bit or we might want to track how long the card was viewed for.

web/hooks/use-user.ts
```diff
@@ -145,6 +140,28 @@ export const useUserRecommendedMarkets = (userId = '_', count = 500) => {
...
+  const viewedMarketCardEvents = useFirestoreQueryData<ContractCardView>(
+    ['user-recently-viewed-market-cards', userId, count],
+    viewedMarketCardsQuery
+  )
+  const viewedMarketCardIds =
+    viewedMarketCardEvents.data?.map((e) => e.contractId) ?? []
+  // get the count the number of times each market card was viewed
+  const marketCardViewCounts = countBy(viewedMarketCardIds)
+  // filter out markets that were viewed 3 times
+  const viewedMultipleTimesMarketIds = uniq(viewedMarketCardIds).filter(
```
IanPhilips, line 161: maybe change to 4 or 5 times? I kind of like seeing cards a few times before I'm bored of them.

web/pages/home/index.tsx
```diff
@@ -413,7 +414,22 @@ export const YourFeed = (props: { user: User; count: number }) => {
   const contracts = savedContracts ?? computedContracts
 
   if (!contracts) return <LoadingIndicator />
-  return <ContractsGrid contracts={contracts} showImageOnTopContract />
+  return (
+    <ContractsGrid
+      contracts={contracts}
+      showImageOnTopContract
+      onViewCard={(contract) => {
+        track('view contract card', {
```
IanPhilips, line 422: Could change this to an optional boolean to track card views

web/hooks/use-is-visible.ts
jahooma: Yup. Seems good for now though

web/hooks/use-user.ts
jahooma: Not sure. But I bet there's some better way eventually than just a hard cutoff after a certain number of views.

common/events.ts
```diff
@@ -0,0 +1,16 @@
...
+  name: 'view market'
+  slug: string
+  timestamp: number
+  contractId: string
+  creatorId: string
+  groupSlugs: string[]
+}
+
+export type ContractView = {
+  name: 'view market card'
```
jahooma, line 11: The `name` strings here seem to be swapped.

jahooma (approved):
Cool stuff, Ian! Seems like an important ingredient to build the best possible feed.

Was talking with @mqp, and it seems like we might want to replicate all these collections in Supabase, and then just write SQL queries to fetch a personalized feed from the client.

Would allow us to rapidly iterate on the feed while still being performant. And is a great way to learn SQL.
If we start doing this, we will probably have opinions on what data to store exactly so we can write efficient queries.

common/events.ts
IanPhilips: good eye, thanks!

PR #1336, opened 2022-12-15
Bet orders

IanPhilips:


jahooma:
```
export const fillOrders = (
  contract: CPMMBinaryContract | PseudoNumericContract,
  orders: Order[],
  unfilledBets: LimitBet[],
  balanceByUserId: { [userId: string]: number }
) => {
  const bets = []
  const allMakers = []
  const ordersToCancel = []
  let pool = contract.pool

  for (const order of orders) {
    const {
      newBet,
      newPool,
      makers,
      ordersToCancel,
      newUnfilledBets,
      newBalanceByUserId,
    } = getBinaryCpmmBetInfo(
      order.outcome,
      order.amount,
      contract,
      pool,
      order.limitProb,
      unfilledBets,
      balanceByUserId
    )
  }
  bets.push(newBet)
  pool = newPool
  allMakers.push(...makers)
  ordersToCancel.push(...ordersToCancel)
  unfilledBets = newUnfilledBets
  balanceByUserId = newBalanceByUserId
}
```

IanPhilips:
Goal: Make placing bets work well under high load in anticipation of a market going viral. Test case is using ` ts-node place-many-bets.ts [apikey] 50` that tries to place 50 bets as quickly as possible. This currently results in the requests taking ~10 seconds to resolve with ~50% of them failing.

This implementation has clients create open orders in subcollections of contracts in firebase. The trigger, `onCreateOrder` calls a v2 function that runs a transaction (the parent transaction) that gets all the open orders, running child transactions to create user bets for each order. The parent transaction updates each order sequentially (sorted by earliest order first) upon bet creation. This doesn't work though and creates more bets than it should bc of contention in the child transactions, leading to retrying those with the same open orders query from the parent transaction. James and Stephen and I spent a while talking about alternatives. James and I think the best way to do it would be to run one transaction that gets all open orders, keeping track of all new bets, balance updates, and final contract state in memory, then executing all the transaction updates once at the end of the function using the in memory data. Stephen thinks this is a bad idea, a waste of time, and we should fail more, earlier before we optimize `placebet` any more. @mqp do you have any other ideas for alternatives or opinions on this idea?

jahooma:
> Test case is using ts-node place-many-bets.ts [apikey] 50 that tries to place 50 bets as quickly as possible. This currently results in the requests taking ~10 seconds to resolve with ~50% of them failing.

This is ok, but not great. That's about 25 bets per 10 seconds or one per 400ms.

My revised opinion is there's probably a better medium-term architecture, using something like the PubSub architecture @mqp set up to process batches of bets. We should do a thing where we can process, say, 100 bets on a single market within one transaction by fetching all the data upfront and computing all trades at once.

A second revised opinion is that this high volume, real time use-case is not currently important to our product and is unlikely to be important in the next couple months. That's because I have less faith in the Twitch bot / streamers use-case, and it's hard to think of another use case that would require this.

As such, I'd recommend punting on this, even though some work has been already done.

IanPhilips:
Sounds good, will be nice to avoid the extra complexity as well.

PR #1338, opened 2022-12-16
Save notifs page

IanPhilips:
This has been annoying me recently: when I go back in the app after exploring an older notifications page I don't go back a page, instead I go to where I was before I was on the notifications page. I'm curious if I'm the only one annoyed, though.


IanPhilips:
oops this branch was created from the wrong base

PR #1339, opened 2022-12-16
Track notifs page in query

IanPhilips:
This has been annoying me recently: when I go back in the app after exploring an older notifications page I don't go back a page, instead I go to where I was before I was on the notifications page. I'm curious if I'm the only one annoyed, though.


ingawei (approved):
This is great, thanks Ian! One nitpicky thing is that I think the page number should align with what the UI says, right now it starts at 0 when I think it should start at p=1

PR #1340, opened 2022-12-16
Tidying up some dependencies

mqp:
Should have no user-visible changes, unless one of the transitive dependency upgrades has a bug and randomly breaks something.

Just a quick audit for unnecessary cruft. This change throws out about 100M of crap from `node_modules` on my box, so I guess it should probably stave off disaster for the build machine for a little bit.


jahooma (approved):
Thanks a lot Marshall for looking into this and cleaning up unused packages!

PR #1344, opened 2022-12-17
Ingawei/man 508 improve group visually

ingawei:
image
image
image

This is what happens when you click Read more
image

Change banner:
image
image
image

Known bugs:
About section doesn't update automatically upon changing, must refresh


jahooma:
Thanks, Inga!

I took a look at the new preview site. It does looks cleaner and more usable!

- The group banner will be a nice bit of customization along with the about description.
- I like the tab for Posts, will be interesting to see how people use it.
- The hide group button is well-placed behind the '...'. Looks a lot better than the former red button.

Feedback:

- Should it say don't *show* questions from this group? "See" doesn't sound right.
Screen Shot 2022-12-19 at 12 51 59 AM
- The way the last line of the about is a bit cut off looks like a bug. Wasn't there some kind of fade out thing? Or, it would be fine to just show two whole lines, because of the Read more button.
Screen Shot 2022-12-19 at 1 01 23 AM

- I can add a post to a closed group I don't follow:
Screen Shot 2022-12-19 at 12 55 20 AM

- We might need to restrict the ability to edit the group. E.g. Destiny users might be changing the cover image inappropriately.
- Do we need closed groups? Not being able to follow a group doesn't seem very important. Maybe it should be about whether the group can be modified by other users. Otherwise we should get rid of this concept.
- On desktop, you could consider moving the leaderboards out of the tab and into the third column of the page. So you'd see Top traders on the right and if you scroll further down, Top creators.

Overall seems like a strict upgrade! Will review the code separately.

web/components/groups/group-banner-image.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+    setFileLoading(true)
+
+    await uploadImage(user.username, file)
+      .then(async (url) => {
+        setFileUrl(url)
+        setBannerSelection(url)
+        setFileLoading(false)
+      })
+      .catch(() => {
```
jahooma, line 93: Did you mean `.finally(...)`? This would swallow the error.
Or, you could use `setError` here to inform the user.

web/components/widgets/collapsible-content.tsx
```diff
@@ -65,13 +65,12 @@ export function CollapsibleContent(props: {
       }
     }
   }, [contentRef.current?.offsetHeight])
-
+  console.log(shouldAllowCollapseOfContent)
```
jahooma, line 68: Can remove

web/components/groups/group-overview-post.tsx
```diff
@@ -1,113 +1,246 @@
...
 }
 
-function RichEditGroupAboutPost(props: { group: Group; post: Post | null }) {
+export function GroupOverviewPost(props: {
+  group: Group
+  isEditable: boolean
+  post: Post | null
+}) {
+  const { group, isEditable, post } = props
+  // const post = usePost(group.aboutPostId) ?? props.post
```
jahooma, line 64: You can remove this. I think you listen for about post changes higher up.

web/components/groups/group-overview-post.tsx
```diff
@@ -1,113 +1,246 @@
...
-    </div>
+    <Col className={clsx(MODAL_CLASS, SCROLLABLE_MODAL_CLASS)}>
+      <div className="text-xl">About {groupName}</div>
+      <Content content={content} />
+    </Col>
   )
 }
 
-function RichEditGroupAboutPost(props: { group: Group; post: Post | null }) {
+export function GroupOverviewPost(props: {
```
jahooma, line 58: Shouldn't this be merged with the `GroupAboutSection` component?

web/components/groups/group-banner-image.tsx
```diff
@@ -0,0 +1,259 @@
...
+  return DEFAULT_BANNERS.some((path) => path === url)
+}
+export default function BannerImage(props: {
+  group: Group
+  user: User | undefined | null
+  isEditable: boolean
+}) {
+  const { group, user, isEditable } = props
+  const [groupBannerUrl, setGroupBannerUrl] = useState(
+    group.bannerUrl ?? DEFAULT_BANNERS[0]
```
jahooma, line 31: If you want a random default banner, you could do:
`DEFAULT_BANNERS[genHash(group.id) % DEFAULT_BANNERS.length]`

jahooma (approved):
Wow, thanks for this effort, Inga! LGTM.

Your code quality is good (as is the design). Only minor comments.

This fixes most of the rough edges groups had. There are still a few outstanding problems with the way groups work, e.g. the free-for-all nature of tagging markets with groups.

PR #1345, opened 2022-12-17
Fix deploy with problematic packages

jahooma:


jahooma:
Welp, I didn't get very far here. The Vercel suggestions are not that great. 

Clearing the cache with `yarn clean cache && yarn install` just led to a typescript error for not having Node types. Which I don't understand because we have a dev dependency for that. Plus it takes a lot of extra time, so it is not a good solution in anycase.

PR #1346, opened 2022-12-17
Use Supabase to power "add user to group" dropdown

mqp:
Getting this in as a kind of proof of concept. Related to but not the same as #1288. The UI on this whole "add user to group" dialog still just kind of looks and works very badly. But now at least it doesn't also download our whole users DB when you click it.


mqp:
Funnily enough this only powers two pieces of dead code now (add user to group, and select referral user) but I might as well merge it anyway. If we're going to have dead code it can be good dead code.

PR #1348, opened 2022-12-19
Maniswipe: scroll + swipe

jahooma:


sipec:
This would resolve [MAN-504](https://linear.app/manifoldmarkets/issue/MAN-504/make-swiping-more-satisfying-clear)

web/components/contract/swipe-card.tsx
```diff
@@ -0,0 +1,202 @@
...
+            followsLinkClass
+          >
+            {question}
+          </SiteLink>
+          <Percent
+            contract={contract}
+            amount={amount}
+            outcome={swipeDirection}
+          />
+          {/* TODO: use editor excluding widgets */}
```
sipec, line 85: well, we do want formatted text and mentions but we don't want to show images, tweets, embeds, videos

probably we just use the rich text parser, but change the extensions such that the block level things don't render at all?

PR #1357, opened 2022-12-22
Query bets as client needs them via supabase

IanPhilips:
Queries bets via supabase as the user pages through the trades tab. We could do some clever ascending/descending sorting if they jump to the end but the latency is actually pretty good without it so perhaps we can just keep it simple for now.


supabase-replicator/supabase/seed.sql
```diff
@@ -46,6 +46,7 @@ create table if not exists bets (
 alter table bets enable row level security;
 drop policy if exists "public read" on bets;
 create policy "public read" on bets for select using (true);
+create index concurrently bets_data_gin on bets using GIN (data);
```
IanPhilips, line 49: Is this the correct index to use a contains query?

web/lib/supabase/bets.ts
```diff
@@ -0,0 +1,22 @@
...
+import { JsonData } from 'web/lib/supabase/json-data'
+
+export async function getOlderBets(
+  contractId: string,
+  beforeTime: number,
+  limit: number
+) {
+  const { data } = await run(
+    db
+      .from('bets')
```
mqp, line 13: Note that bets moved to the table `contract_bets` in #1375.

IanPhilips: Thanks!

PR #1365, opened 2022-12-30
the live feed is a list

Gurkenglas:
so it shouldn't list unlisted markets.


jahooma (approved):
Thanks Gurkenglas! Looks good, thanks for the fix.

PR #1368, opened 2022-12-31
Update create-comment.ts

firstuserhere:
1. MAX_COMMENT_JSON_LENGTH -> It's better to validate the length before stringify than stringify and then checking length.

2. Does removeUndefinedProps handle null?


jahooma:
Hi @firstuserhere, thanks for contributing!

This change doesn't appear to do anything different than before.

`removeUndefinedProps` requires a non-null object argument.

Issue #1363, opened 2022-12-29
[FR] Surface user posts as its own tab


mcw0933:
Presently, Posts are seen as a conditional section under the Comments tab.  Adding a first-class posts tab, with similar "no X yet" default behavior (like markets and comments have), would add discoverability to the feature.

PR #1371, opened 2023-01-01
Add posts tab, change icon for comments tab

mcw0933:
This PR aims to address https://github.com/manifoldmarkets/manifold/issues/1363. This was inspired by [this comment thread](https://manifold.markets/MattCWilson/do-you-think-creating-trading-in-an#D4st3ITRUqEl8fVQUmzx).

I did notice that new users are prohibited from creating posts for the first 30 days of use. This PR changes nothing there, although I did wonder if making the posts feature less discoverable was intentional for this reason.

I also followed the [Contributing](https://manifold.markets/MattCWilson/do-you-think-creating-trading-in-an#D4st3ITRUqEl8fVQUmzx) section as best I could. I didn't see anywhere to add or extend UI tests, so I just manually tested these four cases:

    Viewing a user with existing Posts (@ian)
    Viewing a user with no existing Posts (@ACX_Bot)
    Viewing my own user account
    Attempting to create a post as a new user

Everything looks great, no obvious signs of anything breaking.

One other note - it looks like PostCardList is used a number of places, so rather than extending it to do the "no posts yet" capability, I just instead inlined that into the new Posts tab alone. Wanted to go for minimally invasive, but if it's preferable to have that in the component (as it is for UserCommentsList), happy to adjust.

Excited to potentially contribute back to the community! Thanks for your consideration!

### DIFF
--- a/web/components/user-page.tsx
+++ b/web/components/user-page.tsx
@@ -1,5 +1,6 @@
 import {
   DocumentIcon,
+  ChatIcon,
   FolderIcon,
   PencilIcon,
   ScaleIcon,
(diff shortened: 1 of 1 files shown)


mcw0933:
Updated my repo but that closed #1364 and wouldn't let me reopen it; created a new PR 😕

SirSaltyy:
Thanks for taking the initiative to create a PR for this!
Unfortunately, I'm not sure how likely it is to be merged. The reason we only have 3 tabs is that on certain mobile resolutions a 4th tab causes it to overflow onto a 2nd row which looks bad and takes up vertical space.

I'd be happy to send you a small manalink though either way (I can't do it right now, so @ me here if I forgot and don't send u an email with it in the next couple of days)

I've added @ingawei to review it though as she is the person who has been working on this.

mcw0933:
Aha - Figured there might be good reason.  Thanks for the reply!  I wonder
if the third icon could become a More button with a tab menu or something?

jahooma:
There's also another reason. There's a product decision on whether we want to promote Posts as a thing every user should do (in which case it should be an option on their user page).

I lean against giving posts more prominence because I think most users won't want to create one. It shouldn't be a tab on the same level as Markets, Portfolio, and Comments because it's not that important, and we need to focus on the core product.

It's a bad thing to present more options to the user and more things to do. That's why posts are kind of hidden at the moment while we work on other things.

PR #1374, opened 2023-01-02
Web: Fix N/A resolution button for numeric markets

AnnikaCodes:
This fixes [bug #1083](https://discord.com/channels/915138780216823849/943628693958459462/1059514377532887090) by not disabling the "Resolve" button if the user has selected to cancel the market.


mantikoros:
Thanks for the fix!

PR #1377, opened 2023-01-03
[In progress] Recommended markets using matrix factorization

jahooma:


jahooma:
Merging since the recommendation back end is running, and I want to start using the util method `mapAsync`

PR #1378, opened 2023-01-03
Show a graph on certs

akrolsmir:


web/components/charts/contract/cert.tsx
```diff
@@ -0,0 +1,91 @@
...
+    [props.certPoints]
+  )
+  const data = useMemo(
+    () => [
+      { x: start, y: startP },
+      ...certPoints,
+      { x: end ?? Date.now() + DAY_MS, y: endP },
+    ],
+    [certPoints, start, startP, end, endP]
+  )
```
mqp, line 68: This seems busted, `end` is going to vary on every render (you set it to `Date.now()`) so this `useMemo` is not doing anything.

FWIW I don't see why you have an `end` or `endP` or create an "ending point" for this chart at all.

web/components/charts/contract/cert.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { scaleLinear, scaleTime } from 'd3-scale'
+import { CertTxn } from 'common/txn'
+import { Row } from 'web/components/layout/row'
+
+const MARGIN = { top: 20, right: 40, bottom: 20, left: 10 }
+const MARGIN_X = MARGIN.left + MARGIN.right
+const MARGIN_Y = MARGIN.top + MARGIN.bottom
+
+const CertChartTooltip = (
+  props: TooltipProps<Date, HistoryPoint<Partial<CertTxn>>>
```
mqp, line 23: This and all the other `HistoryPoint<...>` types in this file should just be `HistoryPoint`, right? You don't actually put the `CertTxn` in the point.

PR #1392, opened 2023-01-05
Web: Fetch total bet count from Supabase

AnnikaCodes:


web/lib/supabase/bets.ts
```diff
@@ -20,3 +20,19 @@ export async function getOlderBets(
...
+      .select('*', { count: 'exact', head: true })
+      .eq('contract_id', contractId)
+      .eq('data->isChallenge', false)
+      .eq('data->isRedemption', false)
+      .eq('data->isAnte', false)
+  )
+
+  // should never happen
+  if (count === null) throw new Error("Couldn't fetch bet count from database")
+  return count
```
mqp, line 37: I agree this condition makes no sense so why are you checking it? Seems like you should just cast `count` to `number`.

(P.S. I am definitely going to work on making this stuff have better Typescript support)

AnnikaCodes: Sure, casting works fine too.

web/lib/supabase/bets.ts
```diff
@@ -20,3 +20,17 @@ export async function getOlderBets(
...
 }
+
+export async function getTotalBetCount(contractId: string) {
+  const { count } = await run(
+    db
+      .from('contract_bets')
+      .select('*', { count: 'exact', head: true })
+      .eq('contract_id', contractId)
+      .eq('data->isChallenge', false)
+      .eq('data->isRedemption', false)
```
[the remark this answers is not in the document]
IanPhilips: likewise for isChallence and isAnte

sipec:
We now pass the total bet count through the contract page's static props! check out `getTotalBetCount` in `common/src/supabase/bets.ts`. So this PR is no longer needed. Thanks though!

PR #1403, opened 2023-01-06
Omnisearch

sipec:
click search button in sidebar or keyboard shortcut cmd/ctrl k to bring up search modal
resolves MAN-647, MAN-500, MAN-634

- [x] showing results in a table
- [x] showing users, groups, etc (MAN-647)

### DIFF
--- a/web/components/nav/search-button.tsx
+++ b/web/components/nav/search-button.tsx
@@ -1,0 +1,26 @@
+import { SearchIcon } from '@heroicons/react/outline'
+import { useIsClient } from 'web/hooks/use-is-client'
+import { isMac } from 'web/lib/util/device'
+import { useSearchContext } from '../search/search-context'
+
+export const SearchButton = () => {
+  const { setOpen } = useSearchContext() ?? {}
+  const isClient = useIsClient()
+
+  if (!setOpen) {
+    return null
+  }
+
+  return (
+    <button
+      onClick={() => setOpen(true)}
+      className="mb-5 flex w-full items-center rounded-md border border-gray-300 bg-white p-2 text-sm text-gray-400 hover:border-indigo-300"
+    >
+      <SearchIcon className="mr-3 h-6 w-6 text-gray-400" />
+      <span className="text-md">Search</span>
+      <span className="ml-auto mr-1">
+        {isClient && isMac() ? '⌘' : 'Ctrl '}K
+      </span>
+    </button>
+  )
+}
--- a/web/components/nav/search-button.tsx
+++ b/web/components/nav/search-button.tsx
@@ -1,9 +1,10 @@
 import { SearchIcon } from '@heroicons/react/outline'
+import clsx from 'clsx'
 import { useIsClient } from 'web/hooks/use-is-client'
 import { isMac } from 'web/lib/util/device'
 import { useSearchContext } from '../search/search-context'
 
-export const SearchButton = () => {
+export const SearchButton = (props: { className?: string }) => {
   const { setOpen } = useSearchContext() ?? {}
   const isClient = useIsClient()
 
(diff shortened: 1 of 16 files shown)


mqp:
I think this is really promising and I like the idea. Some comments on my initial impression of the current version as of today (which I don't know how "complete" it's supposed to be or not -- I notice that groups aren't in yet)

- Right now, when I am looking at a list of suggestions and I type a key, it loads the suggestions async, and then when they are ready it puts them in place. But my cursor in the list doesn't move. That means that if I typed a query, the request is a little slow, and in the meantime I try to pick a suggestion, the query can come in right before I pick my suggestion and fool me into picking the wrong thing. Relatedly, when I type a query, the different suggestion sources come in at different rates, so if I (for example) typed "Austin", the markets came in first, and I tried to click on a market, I could get sniped by the users coming in right underneath my mouse. My suggestions are:

1. Don't show the new suggestions until they are all in.
2. If a user types something but then starts keying through the list, they probably want to pick something they see, so stop showing new results that come in -- leave their list alone until they go back to type more stuff.

- The UX when I actually select an item is a little fishy. Firstly, it visibly clears my query from the box, which is weird looking. Secondly, it closes the box immediately, but because our page loads are often rather slow, it looks like nothing is happening (this is most visible if you click on a big contract.) It feels like something broke and your search did nothing. I'm not sure what I suggest doing about this (other than speeding up page loads, which would be good) but it's pretty bad, so I think we should do something. One approach would be to leave the modal up and put it into a visible "loading" state until the new page's first render is ready.

- You should make the search results `user-select: none` and `cursor: pointer` -- right now it looks weird when you mouse over them and you get the text select cursor.

- It feels strange that the thing on the left looks like a search box, but it's really just a glorified button -- you can never type into it. I don't know what design I recommend exactly instead, but it's strange.

- The "blog" link on the bottom is broken.

I hope this gets really good and then we don't have to keep reinventing mediocre-feeling search boxes all over!

web/pages/_app.tsx
```diff
@@ -76,8 +77,10 @@ function MyApp({ Component, pageProps }: AppProps<ManifoldPageProps>) {
       <AuthProvider serverUser={pageProps.auth}>
         <NativeMessageListener />
         <QueryClientProvider client={queryClient}>
-          <Welcome />
-          <Component {...pageProps} />
+          <SearchProvider>
```
mqp, line 80: I don't see why this is a context instead of just a component? Nothing ever uses the context.

web/components/search/query-contracts.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const index = searchClient.initIndex(searchIndexName)
+
+export function useMarketSearchResults(query: string) {
+  const [results, setResults] = useState([] as Contract[])
+  useEffect(() => {
+    index
+      .search(query, { hitsPerPage: 20 })
+      .then((data) => setResults(data.hits as any))
+  }, [query])
```
mqp, line 13: This should have e.g. the same nonce technique as the users query to avoid the problem of racing results showing the wrong result.

web/components/search/query-users.ts
```diff
@@ -0,0 +1,18 @@
+import { User } from 'common/user'
+import { useEffect, useState } from 'react'
+import { searchUsers } from 'web/lib/supabase/users'
+
+// use nonce to make sure only latest result gets used
+let nonce = 0
```
mqp, line 6: It seems a little bit spooky to write it like this, because if you used `useUserSearchResults` in more than one place at the page, everything would be busted. It would make more sense to store this as a `useRef` inside `useUserSearchResults`.

web/components/search/query-users.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function useUserSearchResults(query: string) {
+  const [results, setResults] = useState([] as User[])
+  useEffect(() => {
+    ++nonce
+    const thisNonce = nonce
+    searchUsers(query, 2).then((users) => {
+      if (thisNonce === nonce) setResults(users)
+    })
```
mqp, line 15: I note that it would be reasonable to have a cache in these hooks of previously-fetched results per query, to make backspacing really fast, but searching is pretty fast anyway so I wouldn't say it's mandatory.

web/components/nav/search-button.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  return (
+    <button
+      onClick={() => setOpen(true)}
+      className="mb-5 flex w-full items-center rounded-md border border-gray-300 bg-white p-2 text-sm text-gray-400 hover:border-indigo-300"
+    >
+      <SearchIcon className="mr-3 h-6 w-6 text-gray-400" />
+      <span className="text-md">Search</span>
+      <span className="ml-auto mr-1">
+        {isClient && isMac() ? '⌘' : 'Ctrl '}K
```
mqp, line 22: This seems like it would be creating a hydration error on initial render. I would use a `useEffect` to populate the placeholder text if you think this behavior is important.

sipec: the console doesn't seem to complain about it

web/components/search/query-contracts.ts
sipec: I tried the nonce technique and it was just super slow for some reason

web/pages/_app.tsx
sipec: the button in the sidebar uses it to open the modal.

I suppose it could be a component in the sidebar, and if we want it to share state across pages later (like for search history or cached results) we could use local storage or turn sidebar into a next 13 layout

PR #1407, opened 2023-01-06
Add portfolio history table

IanPhilips:


functions/src/scripts/supabase-import.ts
```diff
@@ -107,6 +107,14 @@ async function importDatabase(kinds?: string[]) {
 
   if (shouldImport('user'))
     await importCollection(client, firestore.collection('users'), 'user', 500)
+  if (shouldImport('portfolioHistory'))
```
mqp, line 110: By convention this would be named `userPortfolioHistory` but I don't care that much about the convention.

mqp:
Looks correct :+1:

functions/src/scripts/supabase-import.ts
IanPhilips: Changed it to match, thanks!

PR #1409, opened 2023-01-07
Tweak recommendation SQL to be a little better

mqp:
1. I didn't appreciate this until now, but it seems that if you are writing a small function that you are going to call a lot in SQL, it's good for it to be `language sql`, because then it's eligible for inlining. There's a great Stack Overflow answer explaining some differences between `language sql` and `language plpgsql` [here](https://stackoverflow.com/a/24771561).
2. Added an index on `user_events` by `user_id` and `name` to accelerate "have I viewed this market" query and because it just seems like an obviously useful index to have.

Together these changes bring the performance of the thing from about 80ms to about 30ms for my user.

I also fixed a bug, which is that if you asked for `get_recommended_contracts('myid', 10)` you wouldn't actually get 10, because it was first getting the top 10 scores, and then filtering those down to valid recommendations (i.e. open markets.) Now it will get the top 10 valid recommendations, like you asked for.

### DIFF
--- a/supabase-replicator/supabase/seed.sql
+++ b/supabase-replicator/supabase/seed.sql
@@ -425,16 +425,14 @@ create or replace function dot(
   crf contract_recommendation_features
 ) returns real
 immutable parallel safe
-language plpgsql as $$
-begin
-  return (
+language sql as $$
+  select (
     urf.f0 * crf.f0 +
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Great! Thanks for these fix ups!

PR #1413, opened 2023-01-09
Portfolio history -> supabase, from 1600 ms to 200ms

IanPhilips:
waiting for the import-supabase to finish


IanPhilips:
Supabase seems to think this sorting business is expensive so I sorted the histories on the client.

Screenshot 2023-01-09 at 10 35 18 AM

I'm probably misunderstanding something though, does that look right @mqp ?

PR #1419, opened 2023-01-11
Move portfolio metrics queries to supabase

IanPhilips:
This is the dumbest way to move the portfolio queries over but on the plus side it only takes ~1000ms to load vs previous 3000ms-4000ms (on my profile) and it has no custom complex sql queries.

### DIFF
--- a/functions/src/scripts/supabase-import.ts
+++ b/functions/src/scripts/supabase-import.ts
@@ -93,13 +93,19 @@ async function importCollectionGroupAtTimestamp(
   const t1 = await getServerTimestamp()
   const n = (await source.count().get()).data().count
   if (startTime === 0) {
-    startTime = await source.orderBy(timePropName, 'asc').limit(1).get().then((snap) => {
-      return snap.docs[0].data()[timePropName]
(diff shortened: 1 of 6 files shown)


IanPhilips:
@mqp I reduced the delta between timestamps because I was hitting 400k objects per day and timing out

functions/src/scripts/supabase-import.ts
```diff
@@ -93,13 +93,19 @@ async function importCollectionGroupAtTimestamp(
...
+        return snap.docs[0].data()[timePropName]
+      })
   }
-  log(`Documents to import: ${n}. Timestamp: ${t1.toISOString()}. Starting from: ${startTime}.`)
+  log(
+    `Documents to import: ${n}. Timestamp: ${t1.toISOString()}. Starting from: ${startTime}.`
+  )
   // go from startTime to t1 via 1 day chunks
-  const delta = 1 * DAY_MS
+  const delta = 0.25 * DAY_MS
```
IanPhilips, line 108: The delta here I mean

PR #1423, opened 2023-01-11
add Cavendish Labs

ag8:


akrolsmir (commented):
Hey @ag8! The code looks good, thanks for submitting this.

The only hangup is that I don't see a way for us to actually send money to Cavendish Labs. We're happy to send money to any 501c3, but we'd need some method of actually doing so (usually, credit card purchase, Venmo or Paypal are ways that we've set up before). Do you happen to know how we could forward donations to Cavendish (are you affiliated with them?)

jmcavanagh:
Does Wise work? @akrolsmir

akrolsmir:
> Does Wise work? @akrolsmir

I don't believe we have Wise set up to send payments, unfortunately...

jmcavanagh:
That's OK. I've created a paypal for us. The link to donate is https://www.paypal.com/donate/?hosted_button_id=ZMCEJNMYXEXFE

PR #1428, opened 2023-01-13
Ingawei/man 250 maniswipe mobile app design

ingawei:
image

image


jahooma:
Got NaN% for this market:
https://prod-git-ingawei-man-250-maniswipe-mobile-app-design-mantic.vercel.app/NathanpmYoung/will-ftxff-spend-more-than-600mn-in
Screen Shot 2023-01-17 at 3 39 23 PM

Overall, looks good!

- Too bad you can't scroll back to previous markets.
- I like the way you read more of the description in the bottom popup. Can probably do similar for comments at some point.
- The description seems to truncate before filling 3 full lines? Did you use `line-clamp-3`?

I'll let @mantikoros (Stephen) do the code review!

ingawei:
@jahooma Oh no I just did a character count instead of a line-clamp, I like the idea of the "read more" being inline but lmk if that's dumb

jahooma:
"Read more" inline seems good!

common/util/format.ts
```diff
@@ -40,12 +40,22 @@ export function manaToUSD(mana: number) {
...
   // Show 1 decimal place if <2% or >98%, giving more resolution on the tails
-  const decimalPlaces = zeroToOne < 0.02 || zeroToOne > 0.98 ? 1 : 0
+  const decimalPlaces = getPercentDecimalPlaces(zeroToOne)
   return (zeroToOne * 100).toFixed(decimalPlaces) + '%'
 }
 
+export function formatPercentNumber(zeroToOne: number) {
+  // Show 1 decimal place if <2% or >98%, giving more resolution on the tails
+  const decimalPlaces = getPercentDecimalPlaces(zeroToOne)
+  return Math.round((zeroToOne * 100 * 1) ^ decimalPlaces) / (1 ^ decimalPlaces)
```
mantikoros, line 56: This is a divide by zero error! Remember, ^ is XOR not exponentiation (which is ** in typescript)

web/components/contract/like-button.tsx
```diff
@@ -30,7 +30,9 @@ export const LikeButton = memo(function LikeButton(props: {
   contract: Contract
   contentText: string
   className?: string
-  size?: 'md' | 'xl'
+  size?: 'md' | 'lg' | 'xl'
```
mantikoros, line 33: We should have export a button size type somewhere...

web/components/contract/like-button.tsx
```diff
@@ -170,7 +180,12 @@ export const LikeButton = memo(function LikeButton(props: {
           short={true}
         />
       )}
-    </>
+      {showTotalLikesUnder && (
+        <div className="mx-auto -mt-1.5 h-6 text-sm text-white drop-shadow-sm disabled:opacity-50">
+          {totalLikes > 0 ? totalLikes : ''}
```
mantikoros, line 185: Might include a label "likes"? I believe TikTok uses labels, but I'm not sure about this case...

web/components/swipe/swipe-bet-panel.tsx
```diff
@@ -0,0 +1,169 @@
...
+        >
+          NO
+        </button>
+        {swipingLeft && (
+          <div className="absolute -left-9 z-30">
+            {betStatus === 'loading' && (
+              <LoadingIndicator size="md" spinnerClassName="border-white" />
+            )}
+            {betStatus === 'success' && (
+              <CheckCircleIcon className={'h-7 w-7 text-white'} />
```
mantikoros, line 95: Nice touch

web/components/swipe/swipe-card.tsx
```diff
@@ -0,0 +1,376 @@
...
+function getSwipeAction(
+  mx: number,
+  my: number,
+  cappedDist: number
+): SwipeAction {
+  if (cappedDist >= horizontalSwipeDist) {
+    return mx < 0 ? 'left' : 'right'
+  }
+  if (my <= verticalSwipeDist) {
+    return 'up'
```
mantikoros, line 42: Why not support swiping back to the previous card?

web/components/swipe/swipe-card.tsx
```diff
@@ -0,0 +1,376 @@
...
+        {...bind()}
+        className={clsx(
+          'user-select-none pointer-events-auto absolute inset-1 max-w-lg touch-none'
+        )}
+        style={{ x, y }}
+        onClick={(e) => e.preventDefault()}
+      >
+        <SwipeCard
+          key={
+            contract.description + contract.question + contract.creatorUsername
```
mantikoros, line 203: Why not contract.id?

web/components/swipe/swipe-card.tsx
```diff
@@ -0,0 +1,376 @@
...
+    <Col className="flex flex-col items-center justify-center">
+      <LikeButton
+        contentId={contract.id}
+        contentCreatorId={contract.creatorId}
+        user={user}
+        contentType={'contract'}
+        totalLikes={contract.likedByUserCount ?? 0}
+        contract={contract}
+        contentText={contract.question}
+        size={'lg'}
```
mantikoros, line 368: Make bigger?

web/components/swipe/swipe-widgets.tsx
```diff
@@ -0,0 +1,106 @@
...
+        className={clsx(
+          'text-8xl transition-all',
+          !outcome && '[text-shadow:#4337c9_0_8px]',
+          outcome === 'YES' &&
+            '[text-shadow:#14b8a6_-6px_4px,#0f766e_-12px_8px]',
+          outcome === 'NO' && '[text-shadow:#FF2400_6px_4px,#991600_12px_8px]'
+        )}
+      >
+        {outcome === 'YES' && (
+          <CountUp
```
mantikoros, line 36: the count up effect is cool 😄

web/pages/swipe/index.tsx
```diff
@@ -206,45 +104,58 @@ export default function Swipe() {
...
-          className="w-full max-w-lg touch-none"
-          style={{ x, y }}
-        >
-          {cards.map((c, i) => (
+        {cards.length > 0 && (
+          <>
+            <div className="z-30">
+              <PrimarySwipeCard
+                key={
+                  cards[0].description +
```
mantikoros, line 115: cards[0].id ?

common/util/format.ts
mantikoros: This function doesn't make that much sense even with exponents. Don't you just want
`(zeroToOne * 100).toFixed(decimalPlaces)`  without the percent sign at the end?

mantikoros (approved):
Overall, I like the new redesign a lot!

- The new betting interface and mechanics are really good!
- The bet confirmation with checkmark is really good.
- Still would prefer more explicitness, using the label "bet" somewhere. (Maybe just a "bet size" label above the bet size toggler?)
- The like button should be bigger.
- The layout looks a little less good using iPhone XR / iPhone 12 dimensions than the screenshot you have here—spacing between percentage and question is a little awkward. 
- I'd prefer to be able to swipe back to the previous card.
- I like the "see more" for the market description.
- Fix formatPercentNumber

mantikoros:
iPhone screenshot...looks much less good than the screenshot above.
Screen Shot 2023-01-17 at 6 00 46 PM

mantikoros:
Wait, this one is a little worse. 
Screen Shot 2023-01-17 at 6 33 06 PM

mantikoros:
Yeah, I'm not sure I like having such a big gap between the question, probability, and description. It makes it a little harder to read versus the old version: 
Screen Shot 2023-01-17 at 6 40 05 PM

common/util/format.ts
ingawei: Oh it has to be a number to have the cool counting down/up effect when the probability changes

ingawei: I just converted it and used toFixed

web/components/contract/like-button.tsx
ingawei: Tiktok labels it if there's no likes yet, I kind of don't like the way it looks and I think it's self explanatory enough?

web/components/swipe/swipe-card.tsx
ingawei: because we must always be looking forward

ingawei: lol I guess I could but I'll just have to think out logistic ux things, like will the card behind change depending on whether you swipe up or down (prev vs. next)

web/components/swipe/swipe-card.tsx
ingawei: because of brain fart, changed to id!

web/components/swipe/swipe-card.tsx
ingawei: I think it'll look reasonably sized once we get other functionalities alongside it

web/pages/swipe/index.tsx
ingawei: yepyep

ingawei:
> iPhone screenshot. Looks much less good than the screenshot above.
> 
> Screen Shot 2023-01-17 at 6 33 06 PM

oof, updated view
image

PR #1429, opened 2023-01-13
add Camfed to the list of charities

purple-hockeystick:


akrolsmir:
This looks great, thanks for the PR @purple-hockeystick

PR #1434, opened 2023-01-14
Market search and filter/sort at the same time

sipec:
Adds a sort "relevance" which is the same as not having a sort (pure algolia search, trending if empty)
Aside from that, if user sets a sort then algolia will search with that sort.

The previous UI where the filters disappear when typing was very magic, but I think instead we should have omni be very magic while market search be powerful

### DIFF
--- a/web/pages/group/[...slugs]/index.tsx
+++ b/web/pages/group/[...slugs]/index.tsx
@@ -64,14 +64,6 @@ export async function getStaticPropz(props: { params: { slugs: string[] } }) {
   const memberIds = group && (await listMemberIds(group))
   const creatorPromise = group ? getUser(group.creatorId) : null
 
-  const contracts =
-    (group && (await listContractsByGroupSlug(group.slug))) ?? []
-  const now = Date.now()
-  const suggestedFilter =
-    contracts.filter((c) => (c.closeTime ?? 0) > now).length < 5
-      ? 'all'
-      : 'open'
-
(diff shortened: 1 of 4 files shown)


web/pages/group/[...slugs]/index.tsx
```diff
@@ -64,14 +64,6 @@ export async function getStaticPropz(props: { params: { slugs: string[] } }) {
...
   const creatorPromise = group ? getUser(group.creatorId) : null
 
-  const contracts =
-    (group && (await listContractsByGroupSlug(group.slug))) ?? []
-  const now = Date.now()
-  const suggestedFilter =
-    contracts.filter((c) => (c.closeTime ?? 0) > now).length < 5
-      ? 'all'
-      : 'open'
-
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:74]: This logic was kind of good for groups. Means most long running groups default to open filter.

Also, we wanted the default sort to be trending.

It's worth keeping these IMO.

jahooma (approved):
Thanks Sinclair, this is a good change, I think! Had one comment though for search on groups.

web/pages/group/[...slugs]/index.tsx
Comment (jahooma) [reply web/pages/group/[...slugs]/index.tsx:74]: Appears that Relevance is equivalent to Trending with no query, so that's reasonable then.

I guess the open filter is mostly unnecessary when sorting by Trending as well, so I retract my comment.

PR #1435, opened 2023-01-14
Omni search: Show results after all subqueries loaded.

jahooma:
I think this improves the user experience of search because the results don't change so many times for one keypress.


mqp:
Strong agree :+1:

sipec (approved):

PR #1440, opened 2023-01-17
Implement fancy Typescript stuff for Supabase queries

mqp:
Introducing two new helpers:

- `selectFrom(db, tableName, ...fields)`, which selects the fields in `fields` from the `data` JSON blob in `tableName`, and produces a query returning `Pick<TData, TKeys>` where `TData` is the type of the blob (e.g. `User`, `Contract`) and `TKeys` is the union of the fields.

- `selectJson(db, tableName)`, which just selects the whole `data` blob and produces a query returning `{ data: TData }` where `TData` is the type of the blob.

These should cut down a lot on the type wrangling associated with querying the JSON blob.

The main limitation is that these helpers don't provide an interface to also select other stuff, like computed fields, or joined tables, or whatever. (They could, they just don't.) So they will only help with queries that are just selecting a bunch of fields out of the JSON blob associated with some rows. That's most of our queries though.

### DIFF
--- a/web/lib/supabase/groups.ts
+++ b/web/lib/supabase/groups.ts
@@ -1,5 +1,5 @@
 import { db } from './db'
-import { run } from 'common/supabase/utils'
+import { run, selectFrom } from 'common/supabase/utils'
 import { Group } from 'common/group'
 export type SearchGroupInfo = Pick<
   Group,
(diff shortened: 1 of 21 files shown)


web/lib/supabase/reactions.ts
```diff
@@ -24,13 +28,13 @@ export async function getLikedContracts(userId: string) {
 }
 
 export async function getLikedContractsCount(userId: string) {
-  const { data } = await run(
+  const { count } = await run(
     db
       .from('user_reactions')
-      .select('count')
+      .select('*', { head: true, count: 'exact' })
```
IanPhilips, line 34: What's the reasoning behind the change here? I thought selecting 'count' worked fine.

web/lib/supabase/groups.ts
```diff
@@ -13,43 +13,54 @@ export type SearchGroupInfo = Pick<
...
 export async function searchGroups(prompt: string, limit: number) {
-  const query = db
-    .from('groups')
-    .select(
-      'id, data->name, data->about, data->slug, data->totalMembers, data->totalContracts, data->anyoneCanJoin'
-    )
+  const query = selectFrom(
+    db,
+    'groups',
+    'id',
```
IanPhilips, line 19: I think James might prefer the property names in an object as the third parameter rather than just a list of parameters.

web/lib/supabase/reactions.ts
mqp: The way this Javascript library works, is it looks at the string in your `select` to figure out what the return type of your query is, by comparing it against the database schema (i.e. what rows are in the table and what the types of the rows are.) Since `count` isn't a row, it gets confused and spits out a type error.

(Selecting `count` is also very quirky -- I had no idea that was even legal PGSQL until I saw you do it. The normal way would be to select `count(*)`, which is what this Javascript will end up producing.)

web/lib/supabase/groups.ts
mqp: Hmm, I see your guys' point. I think I will change it. That will also make it doable to select both inside-of JSON fields and other fields in the same query and keep everything straight and correctly typed.

PR #1442, opened 2023-01-18
(try to) Infinite scroll related markets

IanPhilips:
This gets markets that share the group slugs with the current market as well as related markets by similar bettor activity.


IanPhilips:
This doesn't work well when the market is in a large group with lots of traders and a small, more specific group as it favors the more popular markets rather than true similarity. This could be solved perhaps by selecting for the smaller group but likely we'll use some word factorization scheme soon.

PR #1448, opened 2023-01-19
Fix contract group search SQL

mqp:
1. These functions weren't using any index -- the GIN index on the top-level JSON data doesn't "reach in" and index the contents of arrays and objects inside of it. You have to create an index on individual sub-objects that you want to query by. So they were really slow.

2. `immutable` on `get_time` was just wrong and would result in the wrong time being used in some cases. Please see the documentation [here](https://www.postgresql.org/docs/current/xfunc-volatility.html) to understand these annotations.

3. Cleaned up some unnecessary cruft. (The part of the cleanup where I rearranged the `jsonb_array_elements` stuff was necessary to even use the index.)

### DIFF
--- a/supabase-replicator/supabase/seed.sql
+++ b/supabase-replicator/supabase/seed.sql
@@ -684,7 +684,7 @@ select array_agg(data) from (
     from contracts
     where data->'groupSlugs' ?| elem
     and is_valid_contract(data)
-    order by (to_jsonb(data) ->> 'uniqueBettors7Days')::int desc, to_jsonb(data) ->> 'slug'
+    order by (data->'uniqueBettors7Days')::int desc, data->'slug'
     offset start limit lim
     ) as search_contracts
 $$;
@@ -717,8 +717,8 @@ select array_agg(data) from (
     from contracts
     where data->'groupSlugs' ?| elem
       and is_valid_contract(data)
-      and to_jsonb(data)->>'creatorId' = creator_id
-    order by (to_jsonb(data) ->> 'uniqueBettors7Days')::int desc, to_jsonb(data) ->> 'slug'
+      and data->'creatorId' = creator_id
+    order by (data->'uniqueBettors7Days')::int desc, data->'slug'
     offset start limit lim
 ) as search_contracts
 $$;
--- a/supabase-replicator/supabase/seed.sql
+++ b/supabase-replicator/supabase/seed.sql
@@ -642,6 +642,23 @@ as $$
   ) as rec_contracts
 $$;
 
+create or replace function get_time()
+    returns bigint
+    language sql
+    stable parallel safe
+as $$
+select (extract(epoch from now()) * 1000)::bigint;
+$$;
+
+create or replace function is_valid_contract(data jsonb)
+    returns boolean
+as $$
+select
+        not (data->>'isResolved')::boolean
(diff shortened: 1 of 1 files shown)


mqp:
@IanPhilips please note these changes -- these functions were consuming 20 times more resources than e.g. "all writes to the database combined!"

They took multiple seconds to run, so it was totally obvious something was broken. To a close approximation, no query you ever write should take a second to run. Now they should take a couple milliseconds, which is normal.

Please notice if something takes a couple seconds -- it means something is really wrong and needs fixing!

supabase-replicator/supabase/seed.sql
```diff
@@ -680,12 +681,10 @@ create or replace function search_contracts_by_group_slugs(group_slugs text[], l
...
     select data
-    from contracts,
-      jsonb_array_elements(data -> 'groupSlugs')
-          as elem
-    where elem ?| group_slugs
+    from contracts
+    where data->'groupSlugs' ?| elem
     and is_valid_contract(data)
-    order by (to_jsonb(data) ->> 'uniqueBettors7Days')::int desc, to_jsonb(data) ->> 'slug'
+    order by (data->'uniqueBettors7Days')::int desc, data->'slug'
```
IanPhilips, line 687: I thought we had to use `to_jsonb` in order to use unindexed columns in the json blob. Although looking at `user_events` I see that was mistaken. I have to use `to_jsonb` in the queries in the sql editor though I thought...

supabase-replicator/supabase/seed.sql
```diff
@@ -702,26 +701,23 @@ $$ language sql;
...
     language sql
 as $$
 select array_agg(data) from (
     select data
-    from contracts,
-         jsonb_array_elements(data -> 'groupSlugs')
-             as elem
-    where elem ?| group_slugs
+    from contracts
+    where data->'groupSlugs' ?| elem
```
IanPhilips, line 717: If you hadn't indexed the groupSlugs prop I would need that `jsonb_array_elements` or is that the same deal as the rest of my mistaken `json_b` calls?

IanPhilips:
Thanks for speeding these up! I did notice they were a bit slow but when I analyzed their cost it said something on the order of 700 so I thought it couldn't be too bad. I considered indexing the props but I know over indexing can be detrimental. Next time I'm just going to ask you what you think!

supabase-replicator/supabase/seed.sql
mqp: I wrote some stuff in Discord explaining this. Basically, you only need to use `to_jsonb` in queries if the index is on `to_jsonb` stuff.

However, it's basically a moot point here, because this ordering isn't important for an index to be serving. The important part of the query is finding the contracts that are in the specific groups -- that's the part that needs an index in order to be fast.

supabase-replicator/supabase/seed.sql
mqp: The `jsonb_array_elements` is just a weird way of writing this. If you really wanted it to work like that, you would have had to create an index like, `create index ians_weird_index on contracts (jsonb_array_elements(data->'groupSlugs')). I think that would work?

mqp:
P.S. did you not deploy any of these schema changes on dev? I will take care of it.

PR #1451, opened 2023-01-20
Add Fred Hollows Foundation to charities

rougee:
SS from localhost


akrolsmir:
Awesome, thanks for adding this new charity @rougee!

PR #1453, opened 2023-01-20
Create sql queries for user portfolio

IanPhilips:


IanPhilips:
limit bet performance for jack:
Screenshot 2023-01-20 at 4 36 07 PM
for ian:
Screenshot 2023-01-20 at 4 36 59 PM

Planning on asking @mqp to help me understand if these are normal response times or if I should create any indexes or one of those newfangled clustering tables. It seems like the sort or grouping takes a lot of time? But honestly I could also use some pointers on how to read these tables.

IanPhilips:
I also tried out using distinct contract_ids to left join with contracts but it seems like postgres does that for us and it affords no performance gain.

mqp:
Regarding the limit order query:

### Understanding the query

- For reference, here is the query you wrote:
```sql
select bets.contract_id, array_agg(bets.data) as bets, contracts.data as contract
from (
  select * from contract_bets
  where (data->>'userId') = uid and
        (data->>'isFilled')::boolean = false and
        (data->>'isCancelled')::boolean = false
        order by (data->>'createdTime') desc
  ) as bets
left join contracts
on contracts.id = bets.contract_id
group by contract_id, contracts.data
limit count
```
- It can be hard to read a complicated execution plan in the Supabase editor like that because it's truncating the indentation. One easy way to see it if you don't have a good editor is by using a plan visualizer tool, e.g. if you paste the output of `explain (analyze, format json) ...` into https://explain.dalibo.com/. You can see this one here: https://explain.dalibo.com/plan/cedb82f02b6e8707

- So, basically what's happening is:

1. Postgres uses the `contract_bets_user_id` index to look up all bets for the user ID you specified.
2. It filters those by checking each bet to see if it's an unfilled uncancelled limit order bet.
3. Now it has the "inner part" of your query, inside the `from`, all loaded up.
4. Then it reads the whole contracts table and does a loop that is basically doing this:
```typescript
for (let contract in contracts) {
  const bets = betsFromStep2ByContractId[contract.id]
  if (bets) {
    result.push([contract, bets])
}
```
5. It does the sorting and grouping and limit that you said.

### Improving the query

- As you can see in the plan visualizer, most of the work is "getting all the bets by user ID", because there are really quite a few bets for popular users like Jack. Then matching them up to the contracts is some additional work. In general, "looking up lots of things scattered around a big table" is pretty much usually the most expensive part of any Postgres query, so that makes sense.
- Since we happen to know that most bets aren't limit order bets, we know a lot of these bet lookups are wasted work that are about to get filtered out. This suggests a useful possibility, which is to make an index that can just get us limit order bets, rather than having to get all the bets and then check whether they are limit order bets.
- I created for example this index which specifically helps identify limit orders:
```sql
create index contract_bets_user_outstanding_limit_orders on contract_bets (
  (data->>'userId'),
  ((data->'isFilled')::boolean),
  ((data->'isCancelled')::boolean))
```
which made that part of the query much faster, speeding up the whole query by ~2x.

(If you want to continue using this index please put it into seed.sql and deploy it to dev also.)

- I also thought that other parts of the query were a little odd; specifically, the left join, and the fact that you do the grouping on the outside, instead of on the inside. When I rewrote those parts to be more "natural", I saw that it made another substantial improvement. If you look at the old query, it's kind of saying "you have to join multiple bet rows for the same contract, to the contracts table." But in the new query, it groups the bet rows and it only has to join one for each contract to the contracts table. I'm not sure if this really explains the difference, but it may.

So the end result is this query which I like better and which is about 10 times faster for Jack's data:

```sql
select contract_id, bets.data as bets, contracts.data as contract
from (
  select contract_id, array_agg(data order by (data->>'createdTime') desc) as data from contract_bets
  where (data->>'userId') = uid and
        (data->'isFilled')::boolean = false and
        (data->'isCancelled')::boolean = false
  group by contract_id
) as bets
join contracts on contracts.id = bets.contract_id
limit count
```

I hope this makes sense as a thinking process you could do. Except for the last part where it just seemed fishy so I changed some stuff. But at least I hope that two thirds of it made sense.

mqp:
Regarding the contract metrics query, it's substantially simpler, so I won't write such a big manifesto about it.

- The main cost of the query is fetching the documents out of the big `contract_metrics` table.
- One kind of silly thing the query is doing, is fetching all the `contract_metrics` from the index, ordering them by `lastBetTime`, and then since it's doing `limit N`, it's throwing away a lot of the stuff it fetched.
- So a simple way to make it much faster is to create this index, and then slightly modify the query to use it by doing `order by (data->'lastBetTime')::bigint desc`:
```sql
create index user_contract_metrics_recent_bets on user_contract_metrics (user_id, ((data->'lastBetTime')::bigint) desc)
```
- After doing this the query gets the first 10 results for Jack in like a millisecond or two, which should make sense, since it doesn't have to do much.

Here is how I would write the final query, cleaned up per my stylistic preferences:

```
select ucm.contract_id, ucm.data as metrics, c.data as contract
from user_contract_metrics as ucm
join contracts as c on c.id = ucm.contract_id
where ucm.user_id = uid
order by ((ucm.data)->'lastBetTime')::bigint desc
limit count
```

mqp:
By the way, in theory you should be able to write these using the Supabase JS API without putting them in RPCs, at least the contract metrics one... I empathize if you find it easier to just write whatever SQL you want instead of trying to figure out how to write it in the JS API mini language though.

IanPhilips:
This is awesome, thank you for the excellent write up, Marshall! I feel really lucky to have you on the team to walk me through this stuff, it's super helpful.

IanPhilips:
Also deployed the indexes and functions to dev.

PR #1459, opened 2023-01-24
Ingawei/man 721 add roles to groups

ingawei:
Yay now our groups will be more structured and constrained!

ROLE MANAGEMENT:
Can now view group members and their respective roles, big groups will load in 50 members at a time. Group creators cannot leave the group and are automatically admins
image

Admins will have access to this dropdown to change the role of people within the group
image

Members can also remove themselves from a role
image

These update live yay!

ADDING/REMOVING CONTRACTS FROM GROUP PAGE:
The add button will only appear if you are a member. If you're not an admin or moderator, clicking the add button will open a create market modal within the group:
image

If you are an admin or moderator, you will get the option to add an existing market
image
Markets that are already in the group should not appear in this list, yay!

If you're an admin or moderator, you can remove markets from your groups with this dropdown:
image

I rewrote these functions to the backend for security (see `add-contract-to-group.ts`)

ADDING/REMOVING GROUPS FROM CONTRACT PAGE:
I changed the look of the market groups modal
image

If you're a group admin or moderator or the market creator, you'll have the option to remove the groups upon hover
image

You'll only see groups that you are an admin or moderator of. If you are the market creator, you will see all the groups you are a member of
image

Same for the create market page
image

OTHER SMOL CHANGES:
- changed button name from 'follow/unfollow' to 'join/leave'
image
- got rid of anyonecanjoin permissions and checks for now

KNOWN WEIRD BUGS:
I think there were one or two groups where I was just bombarded with permission errors when I got on them, not sure if it's a problem with my code or it was just some legacy test group. Will list the slugs here if I come across them again.


web/hooks/use-group-supabase.ts
```diff
@@ -0,0 +1,196 @@
...
+    channel.on(
+      'postgres_changes',
+      {
+        event: '*',
+        schema: 'public',
+        table: 'group_members',
+        filter: `group_id=eq.${groupId}`,
+      },
+      (payload) => {
+        if (payload.new.member_id === user?.id) {
```
ingawei, line 34: is there an easy fix for this bug?

ingawei:
@jahooma sorry this is kind of a big boi :(
@mantikoros I would appreciate a sanity check on supabase things I wrote!

common/supabase/schema.ts
```diff
@@ -69,6 +69,23 @@ export interface Database {
           fs_updated_time?: string
         }
       }
+      contract_embeddings: {
```
mantikoros, line 72: This is for the OpenAI semantic embeddings and has nothing to do with this PR, or am I very mistaken?

firestore.rules
```diff
@@ -216,18 +216,33 @@ service cloud.firestore {
...
+        allow write: 
+        if (
+        request.auth.uid == get(/databases/$(database)/documents/groups/$(groupId)).data.creatorId || 
+        isAdmin() || 
+        (
+        isGroupMember() && 
+        (
+        get(/databases/$(database)/documents/contracts/$(contractId)).data.creatorId == request.auth.uid || 
+        get(/databases/$(database)/documents/groups/$(groupId)/groupMembers/$(request.auth.uid)).data.role == 'admin' ||
+        get(/databases/$(database)/documents/groups/$(groupId)/groupMembers/$(request.auth.uid)).data.role == 'contributor'
```
mantikoros, line 232: Doesn't matter that much, but I believe this will do the query twice. Instead, you could do something like
`['admin', 'contributor'].includes(get/ ......))`

mantikoros: Might want to move this logic to a function too

functions/src/add-contract-to-group.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+    const newGroupLinks = [
+      ...(contract.groupLinks ?? []),
+      {
+        groupId: group.id,
+        createdTime: Date.now(),
+        slug: group.slug,
+        userId: auth.uid,
+        name: group.name,
+      } as GroupLink,
```
mantikoros, line 81: This is potentially a lot of data to be storing on the contract if there are many groups...

web/components/groups/add-market-modal.tsx
```diff
@@ -0,0 +1,90 @@
...
+          {group.name}
+        </div>
+        {userRole === 'member' && (
+          <Col className="w-full pt-4">
+            <NewContractFromGroup group={group} user={user} />
+          </Col>
+        )}
+        {(userRole === 'admin' || userRole === 'moderator') && (
+          <Col className="w-full pt-4">
+            <UncontrolledTabs
```
mantikoros, line 42: The tabs are a much nicer solution than before.

web/hooks/use-group-supabase.ts
```diff
@@ -0,0 +1,196 @@
...
+          setTranslatedMemberRole(groupId, isManifoldAdmin, setUserRole, user)
+        }
+      }
+    )
+    channel.subscribe(async (status) => {})
+  }, [db, user])
```
mantikoros, line 40: Shouldn't the hook return a function to unsubscribe from the channel?

`return () => db.removeChannel(channel)`

web/hooks/use-group-supabase.ts
```diff
@@ -0,0 +1,196 @@
...
+        schema: 'public',
+        table: 'group_members',
+        filter: `group_id=eq.${groupId}`,
+      },
+      (payload) => {
+        fetchGroupMembers()
+      }
+    )
+    channel.subscribe(async (status) => {})
+  }, [db])
```
mantikoros, line 120: Same thing. Needs to unsubscribe.

mantikoros (approved):
LGTM.

common/supabase/schema.ts
ingawei: Yes it's generated from the database so doesn't really matter all that much

web/components/groups/add-market-modal.tsx
ingawei: :)

web/hooks/use-group-supabase.ts
ingawei: u rite

web/hooks/use-group-supabase.ts
```diff
@@ -0,0 +1,202 @@
...
+        filter: `group_id=eq.${groupId}`,
+      },
+      (payload) => {
+        fetchGroupMembers()
+      }
+    )
+    channel.subscribe(async (status) => {})
+    return () => {
+      db.removeChannel(channel)
+    }
```
jahooma, line 125: Nice! Cool to see this working. Seems a little awkward to fetch again when there's a change (payload might contain what changed?), but I guess this is simpler.

web/hooks/use-group-supabase.ts
jahooma: What is the bug? Updating your role when you just joined a group?

jahooma (approved):
Thank you Inga! This is nice change — fleshes out a more concrete framework for groups.

I'm excited to see what communities this will enable if moderators can be appointed.

The code looks pretty good overall, not many comments to make. Thanks for sorting out the group permissions, and checking out the realtime supabase channels!

My biggest comment would be to try to split up changes like this into smaller PRs if possible.

E.g. Maybe the realtime supabase part could have been a follow-up piece. Or the create market UI could be implemented in tabs first, and later the ability to add existing markets could be restricted based on role.

It's worth separating out changes like this (just commit, create a new branch, undo the last commit and delete changes not specific to the one PR you have in mind). Can actually make things go faster overall.

PR #1460, opened 2023-01-25
Include Topos Institute and Haskell Foundation as available charities.

ekmett:
I saw a bounty on the discord server for adding charities, and while the bounty was already claimed, there are two 501(c)3 charities I'm actively involved with that I would love to be able to donate to through Manifold. I've included them here.


akrolsmir:
Hey @ekmett, thanks for adding these two charities!

PR #1461, opened 2023-01-25
Display hourly updated daily changed via contract-metrics sum

IanPhilips:
What do you think of this persistent, revalidated state strategy? I took `setState` out of the return because I'm ignoring undefined states because when `savedValue` was read with a non-null saved value, the `useEffect` would overwrite it with undefined. In usePersistentState this wouldn't happen but I'm not sure why.

I'm going to merge this now for the improvement to the daily changed status but still curious to absorb your knowledge on this code.

### DIFF
--- a/web/hooks/use-persistent-state.ts
+++ b/web/hooks/use-persistent-state.ts
@@ -1,9 +1,14 @@
-import { useEffect } from 'react'
+import { useEffect, useState } from 'react'
 import { useStateCheckEquality } from './use-state-check-equality'
 import { NextRouter, useRouter } from 'next/router'
 import { useHasLoaded } from './use-has-loaded'
 
 export type PersistenceOptions<T> = { key: string; store: PersistentStore<T> }
+export type RevalidationOptions<T> = {
+  every: number
+  // callback is async with results of the revalidation
+  callback: () => Promise<T>
+}
 
 export interface PersistentStore<T> {
   get: (k: string) => T | undefined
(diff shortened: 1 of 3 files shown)


web/hooks/use-persistent-state.ts
```diff
@@ -141,3 +146,81 @@ export const usePersistentState = <T>(
...
+    hasLoaded,
+    store,
+    key,
+    lastUpdateTime,
+    every,
+    callback,
+    stringStore,
+    lastUpdateTimeKey,
+    setState,
+  ])
```
jahooma, line 201: This effect needs to be run after `every` time has elapsed. I'd remove the whole deps list so it runs on every re-render. Currently, it's mostly running just the first time you render a component with the hook, I think.

web/hooks/use-persistent-state.ts
```diff
@@ -141,3 +146,81 @@ export const usePersistentState = <T>(
...
+    const router = useRouter()
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useEffect(() => {
+      if (router.isReady) {
+        const savedValue = key != null ? store.get(key) : undefined
+        setState(savedValue ?? initial)
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [router.isReady])
+  }
```
jahooma, line 223: Typically you would just move the if statement inside the `useEffect` so the rules of hooks are not violated.
I guess in this case there's an invariant that store.readUrl would not change so it's logically correct.

jahooma (commented):
Cool stuff! Seems like a reasonable caching strategy.

I guess I would only reach for this after all the simpler strategies don't work. E.g.  Always loading the data once per component; Loading the data once per reload of the app.

Because we currently have many instances of this (once per maniswipe card) this seems reasonable.

Maybe a better solution would be to only render the component once (e.g. on top of all the cards, fixed to the top right corner) and then you could switch to the strategy of fetching new data once for each instance of the component.

web/hooks/use-persistent-state.ts
IanPhilips: Yeah the revalidation is kind of low stakes in this case imo so I didn't want to run it too often. But will probably be best if others use it that want more assurance of revalidation.

IanPhilips:
> Maybe a better solution would be to only render the component once (e.g. on top of all the cards, fixed to the top right corner) and then you could switch to the strategy of fetching new data once for each instance of the component.

I didn't like this just bc it would still mean we're getting essentially the same data over and over as the user moves around the site - in this case to their profile and back to swipe, (and to where ever we add the component in the future). This hook could be useful elsewhere, too: we can add a data-hungry component wherever we want without worrying about re-fetching data on every re-load.

PR #1465, opened 2023-01-26
Add 1d1w1m switcher to market page

sipec:
Resolves MAN-371

- just binary and pseudonumeric for now
- ended up not adding 1H as most markets don't have hourly updates
- I'm not happy with the look of having the entire thing be disabled with just "ALL" selectable when you create it

also includes some a few style changes / refactors to the market header to make it look nicer
- thinner close date, market tags
- add group cta more prominent when there are no groups
these don't really belong in this PR but at this point I can't be bothered to untangle it out

### DIFF
--- a/web/components/buttons/follow-button.tsx
+++ b/web/components/buttons/follow-button.tsx
@@ -92,7 +92,10 @@ export function MiniUserFollowButton(props: {
   if (justFollowed) {
     return (
       <CheckCircleIcon
-        className={clsx('text-highlight-blue h-5 w-5 rounded-full bg-white')}
+        className={clsx(
+          'text-highlight-blue h-5 w-5 rounded-full bg-white',
+          className
+        )}
         aria-label="followed"
       />
     )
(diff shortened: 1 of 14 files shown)


ingawei (commented):
This doesn't have to be addressed in this PR but just a thought:
image
^ I really like the numbers on the left, date picker on the right, do you think we could do this for the profile graph as well and just put the get more M buttons somewhere else?

web/components/buttons/follow-button.tsx
```diff
@@ -61,8 +61,11 @@ export function UserFollowButton(props: { userId: string }) {
   )
 }
 
-export function MiniUserFollowButton(props: { userId: string }) {
-  const { userId } = props
+export function MiniUserFollowButton(props: {
```
ingawei, line 64: The formatting for this is messed up when you click follow:
https://www.loom.com/share/94fcfc8d7033410d856e957d938a4221

PR #1466, opened 2023-01-26
Fix duplicate bet IDs

mqp:
It makes no sense and makes our API hard to use, since the API exposes the ID in the data blob, not the Firestore document ID.


mqp:
FYI @jahooma since I know you are aware of this phenomenon.

PR #1468, opened 2023-01-27
Removed logging "Received native event"

TrueMilli:
I noticed console logging on manifold.markets and thought that might not be intended.

image

### DIFF
--- a/web/hooks/use-native-messages.ts
+++ b/web/hooks/use-native-messages.ts
@@ -13,7 +13,6 @@ export const useNativeMessages = (
       return
     }
     const { type, data } = event
-    console.log('Received native event: ', event)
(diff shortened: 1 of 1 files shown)


akrolsmir:
Thanks for taking this out! Ideally we'd have a lint check that prevents us from checking in random console.log messages by accident, haha

mqp:
FWIW I think console logging (including on prod) is totally fine iff it's logging something useful to see, no idea whether this was useful or not. So I wouldn't make that lint.

IanPhilips:
Hm, this is useful to me during mobile app dev. If none of our ft engineers had a problem with it cluttering the console I'd rather keep it in.

PR #1476, opened 2023-01-31
Pay out users with txns on contract resolution

IanPhilips:
Pays users with txns upon contract resolution

### DIFF
--- a/functions/src/resolve-market.ts
+++ b/functions/src/resolve-market.ts
@@ -1,6 +1,6 @@
 import * as admin from 'firebase-admin'
 import { z } from 'zod'
-import { mapValues, groupBy, sumBy, uniqBy, sum } from 'lodash'
+import { mapValues, groupBy, sumBy, sum } from 'lodash'
 
 import {
   Contract,
@@ -15,8 +15,7 @@ import {
   getValues,
   isProd,
   log,
-  payUsers,
-  payUsersMultipleTransactions,
+  payUsersTransactions,
   revalidateStaticProps,
 } from './utils'
 import {
@@ -31,7 +30,6 @@ import { APIError, newEndpoint, validate } from './api'
 import { getContractBetMetrics } from '../../common/calculate'
 import { createContractResolvedNotifications } from './create-notification'
 import { CancelUniqueBettorBonusTxn, Txn } from '../../common/txn'
-import { runTxn, TxnData } from './transact'
 import {
   DEV_HOUSE_LIQUIDITY_PROVIDER_ID,
   HOUSE_LIQUIDITY_PROVIDER_ID,
@@ -39,6 +37,7 @@ import {
 import { User } from 'common/user'
 import { updateContractMetricsForUsers } from './helpers/user-contract-metrics'
 import { computeContractMetricUpdates } from './update-contract-metrics'
+import { runTxn, TxnData } from './run-txn'
 
 const bodySchema = z.object({
   contractId: z.string(),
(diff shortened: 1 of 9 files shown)


functions/src/run-txn.ts
```diff
@@ -0,0 +1,48 @@
+import * as admin from 'firebase-admin'
```
IanPhilips, line 1: This refactor is just a move, no changes were made

functions/src/transact.ts
```diff
@@ -37,44 +31,4 @@ export const transact = newEndpoint({}, async (req, auth) => {
   })
 })
 
-export async function runTxn(
```
IanPhilips, line 40: I had to remove this function bc the emulator was complaining about the transact function in this file.

functions/src/run-txn.ts
```diff
@@ -0,0 +1,67 @@
...
+import { User } from '../../common/user'
+import { FieldValue } from 'firebase-admin/firestore'
+import { removeUndefinedProps } from '../../common/util/object'
+import { ContractResolutionPayoutTxn, Txn } from '../../common/txn'
+
+export type TxnData = Omit<Txn, 'id' | 'createdTime'>
+
+export async function runTxn(
```
IanPhilips, line 9: This function is unchanged

functions/src/run-txn.ts
```diff
@@ -0,0 +1,67 @@
...
+  fbTransaction.create(newTxnDoc, removeUndefinedProps(txn))
+  fbTransaction.update(fromDoc, {
+    balance: FieldValue.increment(-amount),
+    totalDeposits: FieldValue.increment(-amount),
+  })
+
+  return { status: 'success', txn }
+}
+
+export async function runContractPayoutTxn(
```
IanPhilips, line 49: This is the new function

functions/src/resolve-market.ts
```diff
@@ -202,18 +201,10 @@ export const resolveMarket = async (
...
-    await firestore.runTransaction(async (transaction) => {
-      payUsers(transaction, payouts)
-      transaction.update(contractDoc, contract)
-    })
-  } else {
-    await payUsersMultipleTransactions(payouts)
-    await contractDoc.update(contract)
-  }
+  await payUsersTransactions(payouts, contractId)
+  await contractDoc.update(contract)
```
jahooma, line 207: A nice feature of the previous code was that it tried to update the contract in the same transaction that pays out users. So if paying one user failed or writing  the contract failed, it would not do any update. 

But I think this is fine, since it's simpler, and it didn't work for more than 500 uses anyway.

mantikoros (commented):
I think this looks technically correct, but I'm not wild about this change overall. Seems to be adding more technical debt that we don't need right now. 

- Payouts from a contract are currently of several different types (share payouts, returned liquidity, fees (which we don't currently have)). Having one txn type like ContractResolutionPayout doesn't really make sense.
- If we do create a new txn type, we probably should backfill with previous payouts and think about how we'd handle re-resolution.
- On a meta level, I think it's actively bad to try to add rigorous accounting while we're still making fundamental changes to how our product works. This just means more busy work and technical debt in the near term. The right approach is to design a rigorous txn system a year or two from now when we know which features will stand the test of time.

jahooma:
I agree with Stephen, though I think it could be good if we plan to use this data very soon.

With this change, I think the balance change log could be minimally viable by just using the txn data: streak bonuses, unique bettor bonuses, market payouts, loans, and maybe bets (that subtract from balance). Perhaps as an option somewhere on the portfolio tab / user page.

IanPhilips:
What in the hail, you want to wait a year or two to do this? We get requests for more insight into where people's mana comes from at least once per week. This is a minimal change to the implementation, and like James said, the mvp for showing a balance change log. I would hardly describe this as technical debt, we're just sending a txn instead of a balance change

If you'd like I could add a field on the `data` prop to pass what type of contract resolution payout it is: i.e. liquidity vs shares

IanPhilips:
It'd make this easy too: https://discord.com/channels/915138780216823849/915138780653051910/1072145960811106365

functions/src/utils.ts
```diff
@@ -314,22 +317,38 @@ export const payUsers = (
...
+            category: 'CONTRACT_RESOLUTION_PAYOUT',
+            fromType: 'CONTRACT',
+            fromId: contractId,
+            toType: 'USER',
+            toId: userId,
+            amount: payout,
+            token: 'M$',
+            description: 'Contract payout for resolution: ' + contractId,
+          } as ContractResolutionPayoutTxn
+          await runContractPayoutTxn(transaction, payoutTxn, deposit ?? 0)
```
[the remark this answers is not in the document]
IanPhilips: Awesome, thanks

PR #1483, opened 2023-02-04
Bump Tiptap, fix performance bug messing up comment rendering

mqp:
TipTap's `generateText`, `generateJSON`, and `generateHTML` helpers are written poorly and do the majority of their work just processing the extensions you hand them every time. Generating rich text from the comment to track `LikeButton` clicks was taking up the majority of rendering time when loading a popular contract page. It's not a good idea to call these.

I replaced `generateText` with an efficient version so this is no longer an issue. I had to bump TipTap to do this due to https://github.com/ueberdosis/tiptap/issues/3435.


mqp:
FYI @sipec

PR #1484, opened 2023-02-04
Clean up small issues with the contract list

mqp:
Pretty self-explanatory.


web/components/contract/contracts-list.tsx
```diff
@@ -64,9 +65,9 @@ function DesktopPopover(props: { contract: Contract }) {
         />
       }
     >
-      <span tabIndex={0}>
```
mqp, line 67: @akrolsmir I am not sure how this `tabIndex={0}` got here, but it was breaking "tab through the search results".

PR #1488, opened 2023-02-04
Use `floating-ui` for contract list popup

mqp:
The main motivation behind this was because I want the `ContractCard` to only render conditionally on hover (it's quite heavy, e.g. it subscribes to Firestore for the contract data) and I didn't see any reasonable way to use Tippy to do that.

As far as I can tell `floating-ui` is a better library and there's no reason for us to write code against Tippy.


mqp:
FYI @akrolsmir (I put back the quick betting because now we don't render all the cards unconditionally.)

PR #1489, opened 2023-02-04
Make mobile like list work again

mqp:
I didn't know this existed, so it didn't work after I made my changes in #1485. Now it works.


sipec:
@mqp on desktop:

image
image

otoh its aligned on other markets

PR #1496, opened 2023-02-08
Fix up recommended contracts efficiency further

mqp:
It sucks if this is slow so let's make it reasonable. Now it doesn't fetch all the contracts for the "default user" unless you actually don't have enough for yourself.


mqp:
FYI @jahooma.

PR #1505, opened 2023-02-10
added restricted groups

ingawei:
Added public/restricted groups:
image

all groups are public by default. Also set up some scaffolding for private groups.
People will now be able to add their own markets to public groups even if they're not members


backend/functions/src/add-contract-to-group.ts
```diff
@@ -90,3 +87,23 @@ export const addcontracttogroup = newEndpoint({}, async (req, auth) => {
 })
 
 const firestore = admin.firestore()
+
+export function canUserAddGroupToMarket(props: {
```
jahooma, line 91: Nice job extracting this function! I was just about to comment that you could do this.

common/src/group.ts
```diff
@@ -26,8 +26,10 @@ export type Group = {
   }
   pinnedItems: { itemId: string; type: 'post' | 'contract' }[]
   bannerUrl?: string
+  privacyStatus?: PrivacyStatusType
```
jahooma, line 29: It would actually be better to not have this as optional, and add a third option for 'public'.
- It's clearer and requires fewer cases in code
- You can do firestore queries for privacyStatus === 'public' (otherwise you literally can't query for it)

To make this work you'll need a script to update all existing groups to 'public'

web/components/groups/contract-groups-list.tsx
```diff
@@ -31,23 +31,29 @@ export function ContractGroupsList(props: {
...
+  const [adminGroups, setAdminGroups] = useState<Group[]>([])
+  const [publicGroups, setPublicGroups] = useState<Group[]>([])
+
+  useEffect(() => {
+    if (user) {
+      getGroupsWhereUserHasRole(user.id).then((g) =>
+        setAdminGroups(g.map((gp: { group_data: any }) => gp.group_data))
+      )
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jahooma, line 43: I don't think you need this anymore? The eslint should be disabled. Also below.

web/lib/supabase/groups.ts
```diff
@@ -92,3 +92,32 @@ export async function getGroupsWhereUserIsMember(userId: string) {
...
+export async function getNonPublicGroupsWhereUserHasRole(userId: string) {
+  const groupThings = await run(
+    db
+      .from('group_role')
+      .select('group_data')
+      .eq('member_id', userId)
+      .or(
+        'group_data->>privacyStatus.eq.private,group_data->>privacyStatus.eq.restricted'
+      )
+      .or('role.eq.admin,role.eq.moderator')
```
jahooma, line 105: I could be wrong, but shouldn't these be `.and` queries?

web/components/new-contract-panel.tsx
```diff
@@ -220,13 +223,22 @@ export function NewContractPanel(props: {
...
+    getNonPublicGroupsWhereUserHasRole(creator.id)
+      .then((g) =>
+        setAdminGroups(g.map((gp: { group_data: any }) => gp.group_data))
+      )
+      .catch((e) => console.log(e))
   }, [creator.id])
 
+  useEffect(() => {
+    getPublicGroups()
+      .then((g) => setPublicGroups(g.map((gp: { data: any }) => gp.data)))
```
jahooma, line 238: If you do this transformation everywhere, shouldn't the `getPublicGroups` function do it for you?

jahooma: Same for `getNonPublicGroupsWhereUserHasRole`

jahooma (approved):
Nice change, Inga!

I have a few comments. The biggest one is that I think the group privacy status should not be optional and it should include 'public'.

mqp:
Sorry for the merge conflict churn on this. I changed very little, just moved the files around a little.

jahooma:
Also,  a "Curated" group is way better sounding than a "Restricted" group, and helps people understand the use-case.

web/lib/supabase/groups.ts
ingawei: no the whole thing is an or statement between the two conditions

PR #1510, opened 2023-02-11
Og market chart

sipec:
image

~~I'm not happy with the JSON serialization for the points - leads to really big tag / image url and it doesn't always deserialize correctly. It also limits how far back the history can go~~

~~Also I have no idea why build is failing. was working before I merged main into this branch~~

Edit: now it's like this:

image


web/components/og/graph.tsx
```diff
@@ -81,3 +79,39 @@ const computeColorStops = (
...
+  aspectRatio?: number
+  min: number
+  max: number
+}) {
+  const { data, height: h, aspectRatio = 1, min, max } = props
+  const w = h * aspectRatio
+  const visibleRange = [data[0].x, data[data.length - 1].x]
+  const curve = curveStepBefore
+
+  const xScale = scaleTime(visibleRange, [4, w - 4])
```
mqp, line 96: This seems like a pretty weird way to make a margin, why not make the `scaleTime` range 0 to width and then use CSS to make a margin?

web/components/og/og-market.tsx
```diff
@@ -0,0 +1,177 @@
+import { OgCardProps } from 'common/contract-details'
+import clsx from 'clsx'
+import { Sparkline, ogPoint } from './graph'
+import { formatPercent } from 'common/util/format'
+
+// Notes for working with this:
+// - Some css elements are missing or broken (e.g. 'gap' and 'text-ellipsis' and 'line-clamp')
+// - I also can't make things overflow hidden in only one direction
+// - Every element should have `flex` set
```
mqp, line 9: FYI, I know this comment says this stuff, but I don't really trust it. I would trust the actual documentation: https://github.com/vercel/satori

So for example I bet that `gap` works (which would let you simplify some things) and I bet that `text-ellipsis` works.

Maybe I am being too cynical, I just think that if this comment author had known what they were talking about, they would have at least linked to the docs.

common/src/util/og.ts
```diff
@@ -0,0 +1,17 @@
+import { DOMAIN } from 'common/envs/constants'
+import { removeUndefinedProps } from 'common/util/object'
+
+export function buildOgUrl<P extends Record<string, string | undefined>>(
+  props: P,
+  endpoint: string
+) {
+  const generateUrlParams = (params: P) =>
+    new URLSearchParams(removeUndefinedProps(params) as any).toString()
```
mqp, line 9: I don't really think it makes sense for this function to take `foo: undefined` parameter pairs and then throw them directly in the trash. I would have filtered those out outside of this function.

mqp, line 9: I don't see the use of making this anonymous function just so you can call it once below, instead of just writing
```typescript
const queryString = new URLSearchParams(params).toString()
```

mqp:
> I'm not happy with the JSON serialization for the points - leads to really big tag / image url and it doesn't always deserialize correctly. It also limits how far back the history can go

I agree with this criticism, so like...why did you do that then?

common/src/contract-details.ts
```diff
@@ -9,46 +8,7 @@ import {
...
-    (closeTime
-      ? ` • ${closeTime > Date.now() ? 'Closes' : 'Closed'} ${dayjs(
-          closeTime
-        ).format('MMM D, h:mma')}`
-      : '') +
-    ` • ${volumeLabel}` +
-    (groupHashtags ? ` • ${groupHashtags.join(' ')}` : '')
-  )
-}
+import { formatPercent, formatWithCommas } from './util/format'
```
mqp, line 11: If we are sticking with Vercel and using it for stuff like this, we're probably going to have to figure out some general strategy for keeping track of "which of our code can be run in edge functions".

mqp:
> Also I have no idea why build is failing. was working before I merged main into this branch

image

It's failing because you are bringing in lodash, which has edge-function-incompatible code.

mqp:
I really like the idea of this design, but something about the line looks kind of ugly... unfortunately, I'm too stupid an artist to know how could be better without experimenting. Maybe try anti-aliasing it, and maybe make it a smooth curve? I think you can do that with some SVG attributes, although I don't know if `satori` will like it.

IanPhilips:
> 

I support changing this from the json format! iI worked fine for the weekly summaries on my portfolio but after deploying it I found out everyone has different numbers of portfolio history points and some people have too many points to fit well. I also support rounding the points - I thought about doing this but didn't seem necessary at the time. If we want to fit more points though, we should def do this.

I'd change the json formatting and and rounding myself if you weren't working on this bc some peoples' weekly summary og images are broken.

IanPhilips:
I love the general idea of this PR but am also not convinced about this being the best image to go with. As it is now, at least for this market, it looks a bit boring. I liked the idea of putting it in the background with the title in the foreground, or maybe we could condense the graph to a box on the side with the title on the other? Or really put the graph in any section with the title wrapping around it. Although the css rules in this case are rather limiting. Also you could try changing the curve to `curveLinear` as well. 

I kind of like how google adds a hint of shading to parts of the graph
Screenshot 2023-02-12 at 11 38 55 AM

web/components/og/graph.tsx
```diff
@@ -29,27 +31,23 @@ export function Graph(props: {
   const stops = computeColorStops(data, color, px)
 
   return (
-    <svg width={w * (scaleX ?? 1)} height={h} viewBox={`0 0 ${w} ${h}`}>
-      <clipPath id={clipId}>
-        <rect x={0} y={0} width={w * (scaleX ?? 1)} height={h} />
-      </clipPath>
+    <svg width={w} height={h} viewBox={`0 0 ${w} ${h}`}>
```
IanPhilips, line 34: Have you tried these changes with the weekly summary og previews?

mqp:
> I liked the idea of putting it in the background with the title in the foreground

I also think this is a great idea if you can manage to find the perfect colors/transparency/positioning so that the text is readable, the general gist of the shape of the graph is visible at a glance, and it looks clean. But that's asking a lot, I don't actually know if those are all possible to do.

mqp:
> I kind of like how google adds a hint of shading to parts of the graph

I like this look too and it would be interesting to play with it on our graphs and see if they could instantly look cooler.

web/components/og/og-market.tsx
sipec: gap didn't work when I tried. I'll look into it

web/components/og/graph.tsx
sipec: not yet. where do I see weekly summaries?

IanPhilips: It's at the /update endpoint, you should be able to see that by seeing usages of the buildOgUrl

web/components/og/graph.tsx
sipec: purpose was to leave enough room in svg for the final points not to be cut off (since the line is 4px thick)

web/components/og/graph.tsx
sipec: left a comment on discord - I know where the code is, I just doin't know what an og image url for the weekly profit in the wild looks like because I don't know where to see my or other people's weekly profit

PR #1512, opened 2023-02-13
Add some convenience imports to `web` project

mqp:
Now you can do e.g. `import { useEvent } from 'hooks/use-event'` instead of `web/hooks/use-event`.


mqp:
It seems like we want to go in the opposite direction, where we have less build complexity and less fancy imports.

PR #1514, opened 2023-02-13
Undo misresolved contracts

IanPhilips:
This doesn't undo all trades, just the final payouts and sets the contract back to its last, unresolved state.


IanPhilips:
oops I have some changes from a different state

PR #1515, opened 2023-02-13
Undo contract resolution

IanPhilips:
This doesn't undo all trades, just the final payouts and sets the contract back to its last, unresolved state.

### DIFF
--- a/backend/scripts/undo-resolution.ts
+++ b/backend/scripts/undo-resolution.ts
@@ -1,0 +1,53 @@
+import * as admin from 'firebase-admin'
+
+import { initAdmin } from 'shared/init-admin'
+initAdmin()
+import { ContractResolutionPayoutTxn } from 'common/txn'
+import { chunk } from 'lodash'
+import { undoContractPayoutTxn } from 'shared/run-txn'
+
+const firestore = admin.firestore()
+
+const undoResolution = async (contractId: string) => {
+  // get all the txns with category CONTRACT_RESOLUTION_PAYOUT
+  // and fromType CONTRACT and fromID contractId
+  const txns = await firestore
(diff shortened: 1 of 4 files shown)


backend/scripts/undo-resolution.ts
```diff
@@ -0,0 +1,51 @@
...
+    await firestore.runTransaction(async (transaction) => {
+      for (const txn of chunk) {
+        undoContractPayoutTxn(transaction, txn)
+      }
+    })
+  }
+  console.log('reverted txns')
+
+  await firestore.doc(`contracts/${contractId}`).update({
+    isResolved: false,
```
mantikoros, line 33: set closeTime to now

backend/shared/src/run-txn.ts
```diff
@@ -48,20 +48,43 @@ export async function runTxn(
...
+  const toDoc = firestore.doc(`users/${toId}`)
+  fbTransaction.update(toDoc, {
+    balance: FieldValue.increment(-amount),
+    totalDeposits: FieldValue.increment(-(deposit ?? 0)),
+  })
+  const txnDoc = firestore.doc(`txns/${id}`)
+
+  fbTransaction.update(txnDoc, {
+    data: {
+      ...(data ?? {}),
```
mantikoros, line 83: I think you can just do ...data

It will work even if data is undefined.

mantikoros (approved):
LGTM! We should add an admin button on the resolve panel to support un-resolution too.

backend/shared/src/run-txn.ts
mantikoros: Hmm, is data not undefined?

PR #1518, opened 2023-02-13
Add `pg_vector` and `pg_repack` extensions

mqp:
`pg_vector` provides a `vector` data type that provides efficient storage over large multidimensional vectors and fast indexes for looking up vectors ordered by e.g. cosine similarity to a target vector. You can see [this Supabase blog post](https://supabase.com/blog/openai-embeddings-postgres-vector).

`pg_repack` provides a way to do the equivalent of `cluster` and `vacuum full` (i.e. keep tables in an efficient physical order) without locking, so we can do it regularly. This should greatly improve the performance of common operations on large tables like `contract_bets` and `user_seen_events`. See the documentation [here](https://github.com/reorg/pg_repack).

### DIFF
--- a/backend/supabase/seed.sql
+++ b/backend/supabase/seed.sql
@@ -7,6 +7,12 @@
 /* allow our backend to have a long statement timeout */
 alter role service_role set statement_timeout = '120s';
 
+/* for clustering without locks */
+create extension if not exists pg_repack;
+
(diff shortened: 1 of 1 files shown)


mqp:
FYI @jahooma and @mantikoros you can feel free to go nuts with the OpenAI embeddings stuff now.

PR #1520, opened 2023-02-14
Fix bad feeling loading stuff on landing page

mqp:
The landing page loading used to feel really janky. This tries to fix it up a little by having the footer not appear until the content is loaded (so then it doesn't get yeeted to the bottom when the content loads) and by having the width stay consistent throughout loading.


mqp:
FYI @sipec

PR #1524, opened 2023-02-15
Make `localStorage` access safer

mqp:
Now this covers the case where accessing the property throws (I think this happens when a security policy restricts local storage use in Chromium) and it covers the case where `set` and `remove` don't work (I don't know of a case where this happens, but it makes sense that it could happen for "not allowing trackers to set stuff" reasons.)

### DIFF
--- a/web/components/auth-context.tsx
+++ b/web/components/auth-context.tsx
@@ -64,7 +64,7 @@ export function AuthProvider(props: {
 
   useEffect(() => {
     if (serverUser === undefined) {
-      const cachedUser = localStorage.getItem(CACHED_USER_KEY)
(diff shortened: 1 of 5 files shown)


mqp:
@sipec my guess is this should fail more gracefully in the case we were looking at on Monday where local storage was being blocked in incognito cross-origin iframes.

sipec (commented):
The code looks fine but this doesn't fix the embed in incognito when I tested it. (last 2 embeds in https://manifoldmarkets.notion.site/Test-85b6befbd1d04d69bba3d69b5a0ee610)
even the /embed/ link redirects to the homepage for some reason

PR #1531, opened 2023-02-16
Discord bot

IanPhilips:
Screenshot 2023-02-17 at 1 39 21 PM


discord-bot/deploy.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+SERVICE_NAME="manifold-discord-bot"
```
mqp, line 3: I would have just called it `discord-bot`.

discord-bot/deploy.sh
```diff
@@ -0,0 +1,43 @@
...
+ENV=${1:-dev}
+case $ENV in
+    dev)
+      ENVIRONMENT=DEV
+      GCLOUD_CPU=1
+      GCLOUD_PROJECT=dev-mantic-markets ;;
+    prod)
+      ENVIRONMENT=PROD
+      GCLOUD_CPU=4
+      GCLOUD_PROJECT=mantic-markets ;;
```
mqp, line 14: You should just stick to 1 CPU. 4 CPUs was because the replicator actually has a lot of work to do.

discord-bot/deploy.sh
```diff
@@ -0,0 +1,43 @@
...
+         --set-env-vars ENVIRONMENT=${ENVIRONMENT} \
+         --set-env-vars GOOGLE_CLOUD_PROJECT=${GCLOUD_PROJECT} \
+         --set-secrets SUPABASE_PASSWORD=SUPABASE_PASSWORD:latest \
+         --execution-environment gen2 \
+         --cpu ${GCLOUD_CPU} \
+         --memory 2Gi \
+         --concurrency 1000 \
+         --min-instances 1 \
+         --no-allow-unauthenticated \
+         --no-cpu-throttling
```
mqp, line 39: `no-cpu-throttling` is the setting for "allocate CPU even when not serving HTTP requests" so I think we are all set on that score.

discord-bot/deploy.sh
```diff
@@ -0,0 +1,43 @@
...
+# note that we deploy this onto us-east-4, nearest to supabase, not us-central like most stuff
+
+yarn build && \
+  gcloud builds submit . \
+         --tag gcr.io/${GCLOUD_PROJECT}/${SERVICE_NAME} \
+         --project ${GCLOUD_PROJECT} && \
+  gcloud beta run deploy ${SERVICE_NAME} \
+         --image gcr.io/${GCLOUD_PROJECT}/${SERVICE_NAME} \
+         --project ${GCLOUD_PROJECT} \
+         --region us-east4 \
```
mqp, line 29: You should move this to `us-central`.

discord-bot/deploy.sh
```diff
@@ -0,0 +1,43 @@
...
+         --cpu ${GCLOUD_CPU} \
+         --memory 2Gi \
+         --concurrency 1000 \
+         --min-instances 1 \
+         --no-allow-unauthenticated \
+         --no-cpu-throttling
+
+# to establish subscription to service, e.g.:
+# gcloud pubsub subscriptions create supabaseReplicationSubscription --topic firestoreWrite --ack-deadline 600 --push-endpoint https://supabase-replicator-w3txbmd3ba-uc.a.run.app
+# and make sure subscription push service account is authed to call service
```
mqp, line 43: This comment is not useful.

discord-bot/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "scripts": {
+    "build": "tsc --project tsconfig.json",
+    "start": "yarn run build && nodemon --watch src --exec ts-node src/index.ts"
+  },
+  "type": "module",
+  "dependencies": {
+    "discord.js": "^14.3.0",
+    "manifold-sdk": "2.0.1"
```
mqp, line 9: I 100% would not depend on this unless you think that we should plan on owning this library and taking responsibility for it in the future.

discord-bot/package.json
```diff
@@ -0,0 +1,23 @@
+{
```
mqp, line 1: I didn't even know it was legal to have a package.json with no package name.

discord-bot/yarn.lock
```diff
@@ -0,0 +1,1945 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
mqp, line 1: I don't see any upside to not putting this in the main yarn workspace.

discord-bot/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+    "compilerOptions": {
+        "rootDir": "src",
+        "composite": true,
+        "target": "es2021",
+        "module": "es2022",
```
mqp, line 6: Interesting, I didn't know Node could run with ES modules. I will have to look into this.

discord-bot/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+        "lib": ["es2021"],
+        "noImplicitReturns": true,
+        "moduleResolution": "node",
+        "outDir": "./lib",
+        "tsBuildInfoFile": "lib/tsconfig.tsbuildinfo",
+        "sourceMap": true,
```
mqp, line 12: I wonder if there's any use to producing source maps. I don't know of one.

discord-bot/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+        "baseUrl": "./src",
```
mqp, line 13: This seems not good and will not work if you want to start using `common` etc.

discord-bot/.gitignore
```diff
@@ -0,0 +1,109 @@
...
+lerna-debug.log*
+
+# Diagnostic reports (https://nodejs.org/api/report.html)
+report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
+
+# Runtime data
+pids
+*.pid
+*.seed
+*.pid.lock
```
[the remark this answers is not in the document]
IanPhilips: lol I just forked this from arae, I should trim this down 👍

discord-bot/deploy.sh
IanPhilips: done, thanks!

discord-bot/package.json
IanPhilips: Please don't tell the cops.

discord-bot/.gcloudignore
```diff
@@ -0,0 +1,5 @@
+*.log
+.git
```
mqp, line 2: Seems like this could go away.

discord-bot/.gitignore
```diff
@@ -0,0 +1,19 @@
...
+node_modules/
+
+# TypeScript cache
+*.tsbuildinfo
+
+# Optional eslint cache
+.eslintcache
+
+# Yarn Integrity file
+.yarn-integrity
```
mqp, line 15: I don't think these last two exist?

discord-bot/deploy.sh
```diff
@@ -0,0 +1,41 @@
...
+    prod)
+      ENVIRONMENT=PROD
+      GCLOUD_CPU=1
+      GCLOUD_PROJECT=mantic-markets ;;
+    *)
+      echo "Invalid environment; must be dev or prod."
+      exit 1
+esac
+
+# note that we deploy this onto us-east-4, nearest to supabase, not us-central like most stuff
```
mqp, line 20: This comment isn't true anymore.

discord-bot/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "name": "manifold-discord-bot",
+  "description": "Discord bot that uses reactions to place bets on Manifold Markets.",
+  "version": "0.1.0",
+  "scripts": {
+    "build": "tsc --project tsconfig.json && cp ../yarn.lock ./dist/yarn.lock",
+    "start": "ENVIRONMENT=PROD yarn run build && nodemon --watch src --exec ts-node src/index.ts",
+    "dev": "ENVIRONMENT=DEV yarn run build && nodemon --watch src --exec ts-node src/index.ts"
```
mqp, line 8: I don't think you should need `yarn run build` here.

discord-bot/src/commands/market-react-bets.ts
```diff
@@ -0,0 +1,147 @@
...
+  .addStringOption((option) =>
+    option
+      .setName('link')
+      .setDescription('The link to the market to bet on')
+      .setRequired(true)
+  )
+
+export async function execute(interaction: ChatInputCommandInteraction) {
+  const link = interaction.options.getString('link')
+  if (!link || !link.startsWith('https://manifold.markets/')) {
```
mqp, line 32: Interesting that this currently is only working with the main production domain. Makes me think of whether it should be more general so it could talk to dev, a self-hosted API, etc.

discord-bot/src/register-api-key.ts
```diff
@@ -0,0 +1,37 @@
...
+  await message.reply({
+    content: 'Checking API key...',
+  })
+  let failed = false
+  try {
+    me = await fetch('https://manifold.markets/api/v0/me', {
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Key ${key}`,
+      },
```
mqp, line 22: It seems like this should be in parallel with creating the "Checking API key..." message above.

discord-bot/src/storage.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+  const key = process.env.SUPABASE_KEY
+  if (!key) {
+    throw new Error("Can't connect to Supabase; no process.env.SUPABASE_KEY.")
+  }
+  const url = `https://${getInstanceHostname(instanceId)}`
+  return createClient(url, key)
+}
+
+export const supabase = createSupabaseClient()
```
mqp, line 28: Seems like a lot of this stuff you could get from `common`. Note that `backend/shared` also has Supabase code.

discord-bot/src/storage.ts
```diff
@@ -0,0 +1,68 @@
...
+    .insert({ discord_user_id: discordUserId, api_key: apiKey })
+  console.log('write api key error', error)
+  return error
+}
+export const getApiKeyFromDiscordId = async (discordUserId: string) => {
+  const { data, error } = await supabase
+    .from('discord_users')
+    .select('api_key')
+    .eq('discord_user_id', discordUserId)
+  console.log('got api key for user from supabase', data, 'error:', error)
```
mqp, line 66: If you want this to actually show up at `ERROR` level in the GCP log viewer, check out the replicator code for how to log stuff with log levels that it can read. You have to produce some kind of JSON log lines.

discord-bot/src/server.ts
```diff
@@ -0,0 +1,16 @@
+import * as http from 'http'
+import * as process from 'process'
+
+const hostname = '0.0.0.0'
+const port = process.env.PORT ?? '8080'
+
+const server = http.createServer((req, res) => {
+  res.statusCode = 200
+  res.setHeader('Content-Type', 'text/plain')
+  res.end('Hello World')
```
mqp, line 10: There should probably be some comment in here (or a response) explaining that this server just exists because Cloud Run things need to host a webserver so that Cloud Run can see that they are alive when it spins them up.

discord-bot/src/index.ts
```diff
@@ -0,0 +1,202 @@
...
+  console.log('Loading commands... ')
+  const commandsPath = path.join(
+    path.dirname(fileURLToPath(import.meta.url)),
+    'commands'
+  )
+  const extension = commandsPath.includes('lib') ? '.js' : '.ts'
+  console.log('Commands path: ', commandsPath)
+  const commandFiles = fs
+    .readdirSync(commandsPath)
+    .filter((file: string) => file.endsWith(extension))
```
mqp, line 59: I think this whole thing where it automatically enumerates all the command files and imports them dynamically is just worse than if you didn't do that. At least you are breaking typechecking, etc. on the interface for commands.

discord-bot/src/index.ts
```diff
@@ -0,0 +1,202 @@
...
+  } catch (error) {
+    console.error(error)
+    await interaction.reply({
+      content: 'There was an error while executing this command :(',
+      ephemeral: true,
+    })
+  }
+})
+
+client.login(token)
```
mqp, line 110: I think you should `await` this before trying to do stuff like refresh the slash commands.

discord-bot/src/index.ts
```diff
@@ -0,0 +1,202 @@
...
+}
+
+client.on(Events.MessageReactionAdd, handleOldReaction)
+
+client.on(Events.MessageCreate, async (message) => {
+  // Here you check for channel type
+  // We only need direct messages here, so skip other messages
+  if (!message.channel.isDMBased()) return
+  if (message.author.id === client.user?.id) return
+  await registerApiKey(message)
```
mqp, line 198: It's kind of a shame that it will assume anyone talking to it is sending an API key and produce a correspondingly weird error message if someone tries to e.g. use the /market command via DM.

discord-bot/src/commands/market-react-bets.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+  const slug = getSlug(link)
+  if (!slug) {
+    await interaction.reply(
+      'Invalid market link, could not find slug from link'
+    )
+    return
+  }
+  const market = await getOpenBinaryMarketFromSlug(slug, (error) =>
+    interaction.reply(error)
```
mqp, line 46: The error callback and the fact that we return without awaiting this are pretty weird.

discord-bot/src/commands/market-react-bets.ts
```diff
@@ -0,0 +1,147 @@
...
+  // Let client listener know we've this message in memory
+  messagesHandledViaInteraction.add(message.id)
+
+  // Add emoji reactions
+  for (const emoji of emojis) {
+    if (customEmojis.includes(emoji)) {
+      // TODO: this only works on my guild rn
+      const reactionEmoji = interaction.guild?.emojis.cache.find(
+        (e) => e.id === emoji
+      )
```
mqp, line 94: I wonder if this cache is populated right away on startup?

discord-bot/src/helpers.ts
```diff
@@ -0,0 +1,235 @@
...
+  marketEmbed.setDescription(getCurrentMarketDescription(market))
+  await message.edit({ embeds: [marketEmbed], files: [] })
+}
+
+const getThread = async (
+  channel: TextChannel,
+  marketName: string,
+  title?: string
+) => {
+  const name = marketName.slice(0, 40) + '...'
```
mqp, line 163: This seems like it's going to run into trouble, on markets that are e.g. like "Will blah blah blah by 2023", "Will blah blah blah by 2024", etc.

You could store a market message ID -> thread message ID mapping in Supabase.

discord-bot/src/helpers.ts
```diff
@@ -0,0 +1,235 @@
...
+    if (!apiKey) {
+      if (sale) return
+      await user.send(registerHelpMessage)
+      const userReactions = message.reactions.cache.filter((reaction) =>
+        reaction.users.cache.has(user.id)
+      )
+      try {
+        for (const reaction of userReactions.values()) {
+          await reaction.users.remove(user.id)
+        }
```
mqp, line 94: Kind of weird that this literally removes all their reacts, not just the one they just made.

discord-bot/src/index.ts
```diff
@@ -0,0 +1,202 @@
...
+    : reaction.message
+
+  console.log('got old message id', message.id)
+  const emojiKey = getAnyHandledEmojiKey(reaction)
+  console.log('got emoji key', emojiKey)
+  if (!emojiKey) return
+
+  const user = pUser.partial ? await pUser.fetch() : pUser
+
+  const channel = await client.channels.fetch(message.channelId)
```
mqp, line 150: I would have expected the channel to be in the client cache.

discord-bot/src/index.ts
IanPhilips: We can add more functionalities for DMing the bot later. I don't want to force the user to use a slash command to register their key as they feel pretty foreign, at least to me.

discord-bot/src/helpers.ts
IanPhilips: good point! Fixed

discord-bot/src/helpers.ts
IanPhilips: Good idea! Done.

discord-bot/src/commands/market-react-bets.ts
IanPhilips: Added a domain to the config file!

discord-bot/src/index.ts
IanPhilips: Yup, weird way of doing things! Not sure why the documentation does that.

IanPhilips: Changed this to import commands and export them as an array from commands/index

discord-bot/package.json
IanPhilips: 🪓

IanPhilips:
I added a supabase table for keeping track of markets and threads via message ids but this means old markets that aren't in the table will be ignore. Perhaps worth waiting until deploying the new version until there's a lull or add back in the old code to handle the old markets for a while.

discord-bot/src/commands/market-react-bets.ts
IanPhilips: Not sure! Either way we have to figure out how to use the same emojis across different guilds

PR #1534, opened 2023-02-17
Use localhost or vercel preview domain in embed, og image, share links 

sipec:
now `LIVE_DOMAIN` is the same thing as `ENV_CONFIG.domain`, used for /backend and for compatibility with existing code that wants the domain without the protocol

otoh `BASE_URL` will automatically be the same as whatever environment you're in, whether that's manifold.markets, localhost, or a preview branch. To do this I added some vercel env vars according to  https://github.com/vercel/next.js/discussions/16429#discussioncomment-1302156


common/src/envs/constants.ts
```diff
@@ -33,7 +33,9 @@ export function isManifoldId(userId: string) {
   return userId === 'IPTOzEqrpkWmEzh6hwvAyY9PqFb2'
 }
 
-export const DOMAIN = ENV_CONFIG.domain
```
akrolsmir, line 36: Would add a note about when to use this, e.g. "// Use this instead of BASE_URL for /backend code and when you want the domain without the protocol".

common/src/envs/constants.ts
```diff
@@ -33,7 +33,9 @@ export function isManifoldId(userId: string) {
...
+export const LIVE_DOMAIN = ENV_CONFIG.domain
+// like http://localhost:3000 https://manifold.markets or https://prod-420gg69xd-mantic.vercel.app
```
akrolsmir, line 37: And probably mention that BASE_URL is preferred?

akrolsmir (approved):
I think this looks good? I understand the rationale (to enable localhost/vercel preview tests), but in general the split seems kind of confusing, I notice that I'm still not sure when to use which one after thinking about it for 15 minutes...

common/src/envs/constants.ts
sipec: to be clear, I am not sure that you absolutely need to use it for /backend code, as opposed to BASE_URL. I am just more sure this won't break anything

PR #1538, opened 2023-02-18
adding group member backend functionality

ingawei:


web/lib/supabase/users.ts
```diff
@@ -47,3 +47,57 @@ export async function searchUsers(prompt: string, limit: number) {
     limit
   )
 }
+
+export async function searchUsersExcludingArray(
```
jahooma, line 51: Hmm, might be better to have `searchUsers` call this function with an empty array to reduce code duplication

jahooma (approved):
Thanks, Inga! Code looks really good. Seems like a useful feature.

As discussed, could remove the add user button as it's still accessible in menue.

PR #1565, opened 2023-03-01
adding market depth chart to limit orders popup

RomanHN:


jahooma (changes requested):
Thanks for your contribution, Roman!

Looks pretty good. Made a few comments. We can merge after resolving the chart library question.

jahooma (approved):
Cool. Plan is to merge and then migrate the chart library at some point. An intermediate step I may take is to dynamically load the Victory library.

sipec:
closing the loop: as of 2594a35483d23fd28e15569d2ce6388fb8c51298 the depth chart uses our chart components instead of victory

PR #1574, opened 2023-03-02
Load contract page and embed bets from Supabase again

mqp:
We need to make progress on the Firebase bill soon, so let's turn this on and see how it goes.


mqp:
I'm just gonna push this for value of information, even though there is totally obvious additional work to do (e.g. fix the contract page to be in a state where we can request only the x and y coordinate fields for the points.)

PR #1577, opened 2023-03-03
Do `updateRecommended` in Supabase

mqp:
This seems like a very big source of Firestore reads and it also is slow as shit (the loading process takes longer than half an hour) so let's use Supabase to make that not the case.


jahooma (approved):
Very nice! Should speed things up and save us some money.

Code looks correct. Thanks Marshall!

mqp:
@jahooma FYI -- I easily made the matrix factorization stuff 5-10x faster, it could probably be faster but it would get uglier, so I think this is fine for now.

PR #1580, opened 2023-03-05
More Supabase migration

mqp:
Migrated `updatecontractmetrics` entirely and tidied up `updategroupmetrics` and `updaterecommended` a little bit, loading the last of their data from Supabase.

### DIFF
--- a/backend/functions/src/api/resolve-market.ts
+++ b/backend/functions/src/api/resolve-market.ts
@@ -32,7 +32,6 @@ import {
 } from 'common/antes'
 import { User } from 'common/user'
 import { updateContractMetricsForUsers } from 'shared/helpers/user-contract-metrics'
-import { computeContractMetricUpdates } from '../scheduled/update-contract-metrics'
 import { runTxn, TxnData } from 'shared/run-txn'
 
 import { ContractResolutionPayoutTxn } from 'common/txn'
@@ -195,7 +194,7 @@ export const resolveMarket = async (
     resolutionProbability
   )
 
-  let contract = {
+  const contract = {
     ...unresolvedContract,
     ...removeUndefinedProps({
       isResolved: true,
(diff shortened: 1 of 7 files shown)


backend/functions/src/api/resolve-market.ts
```diff
@@ -210,8 +209,9 @@ export const resolveMarket = async (
     subsidyPool: 0,
   } as Contract
 
-  const updates = await computeContractMetricUpdates(contract, Date.now())
-  contract = { ...contract, ...(updates as any) }
+  // mqp: it would be nice to do this but would require some refactoring
+  // const updates = await computeContractMetricUpdates(contract, Date.now())
+  // contract = { ...contract, ...(updates as any) }
```
mqp, line 214: @jahooma how much do you think it should do this on market resolution?

Since the updating code is all batched now, it would slightly complicate it to make it so that it can update individual contracts on demand -- basically we would have to go make all the SQL queries parameterized on a subset of contract IDs. Not a big deal, but if we don't have to then it would be nice not to.

Just by looking at it, I don't see why "resolving a market" is a special case where we need to do this update any more than "placing a bet" or "opening a limit order" or whatever. (Of course, now that we have Postgres, we probably actually could go keep the contract metrics up to date in realtime everywhere, if we wanted to, and perhaps we should.)

jahooma: It's decently important. The use case is once the market resolves everyone gets an email/notification immediately and clicks through to the market. It should show correct profit figures or else they get upset.

jahooma: One alternative is to compute these metrics on the fly for each user that visits a market they are invested in.

mqp: Is this related to the profit figures? The only things it was recomputing were:
- 24h volume
- Elasticity
- Unique bettor period counts
- Historical probability changes

mqp: Oh, I guess it was also updating the current cached contract probability, but it's not the case that we are e.g. using that in the market resolve email or something.

jahooma: Oh, I was under the impression it was also updating user contract metrics, so it would show your own profit correctly.

If it's not, then, no big deal.

mqp: Nah, there's another part of the code that does that.

PR #1589, opened 2023-03-07
Don't fetch contract follows in notification dropdown

mqp:
This is fetching a huge amount of data for large contracts.

Since they got the notification, we can reasonably assume as a default that they are following the market.


mqp:
FYI @ingawei

PR #1590, opened 2023-03-07
Ingawei/move group search to supabase

ingawei:


web/lib/supabase/contract-groups.ts
```diff
@@ -0,0 +1,235 @@
...
+      run(
+        db
+          .from('groups')
+          .select('*')
+          .or('data->>privacyStatus.eq.public,data->>privacyStatus.eq.curated')
+          .ilike('data->>name', `%${prompt}%`)
+          .order('data->totalMembers', { ascending: false } as any)
+          .limit(GROUPS_LIST_SIZE)
+      ),
+    ])
```
jahooma, line 99: I think a helper function could make this easier.

You could build most of the query once except for the `.ilike()` and then use a helper function to do the parallel queries, e.g.:

```
function (q, field, prompt) {
  return await Promise.all([
  q.ilike(field, prompt),
  q.ilike(field, `{prompt}%`),
  q.ilike(field, `%{prompt}`%),
}
```

web/components/activity-log.tsx
```diff
@@ -37,14 +37,17 @@ export function ActivityLog(props: {
...
-    )
+  useEffect(() => {
+    if (user) {
+      getShouldBlockDestiny(user.id).then((result) =>
+        setShouldBlockDestiny(result)
+      )
+    } else if (!shouldBlockDestiny) {
+      setShouldBlockDestiny(true)
+    }
+  }, [user?.id])
```
jahooma, line 50: You could make this a hook `useShouldBlockDestinty`. And then use it in `use-feed.ts` as well. 
This wouldn't block this PR though.

common/src/supabase.schema.ts
```diff
@@ -0,0 +1,1113 @@
+export type Json =
```
jahooma, line 1: I don't think we need this copy of the schema here, right? You can delete this file.

jahooma (approved):
Looks good, Inga! Just a couple comments.

web/lib/supabase/contract-groups.ts
ingawei: could not do :(

PR #1592, opened 2023-03-10
Internal Ads

sipec:
feed of ads without the ones you redeemed/skipped

later:
- currently no way to ability to see ad spend / cancel it on the post
- liking ads
- write some templates? (currently too much power)

Resolves https://www.notion.so/manifoldmarkets/In-House-Ads-b9b19ee3e379434085a49c331af07426


common/src/txn.ts
```diff
@@ -20,7 +20,8 @@ type AnyTxnType =
   | QfPayment
   | QfAddPool
   | QfDividend
-type SourceType = 'USER' | 'CONTRACT' | 'CHARITY' | 'BANK'
+  | AdRedeem
+type SourceType = 'USER' | 'CONTRACT' | 'CHARITY' | 'BANK' | 'AD'
```
jahooma, line 24: The destination could be 'AD' as well if there's a txn for depositing into the ad.

web/pages/ad/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+function Ad(props: { ad: AdType; onNext: () => void }) {
+  const { ad, onNext } = props
+  const { costPerView, content } = ad
+
+  const counter = useCounter()
+  const timeLeft = WAIT_TIME - counter
+
+  const comments = useCommentsOnPost(ad.id) ?? []
+  const tips = useTipTxns({ postId: ad.id })
```
jahooma, line 74: Tips 😆

web/pages/ad/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                Claim {formatMoney(costPerView)} in {timeLeft + 1} seconds
+              </span>
+              <Button
+                color="override"
+                className="hover:bg-ink-500/50 z-10 shadow-none"
+                onClick={skip}
+              >
+                Skip
+              </Button>
+              <TimerBar duration={WAIT_TIME} />
```
jahooma, line 122: nit: I'd probably pass in 'absolute' as a className, since that layout depends on the parent divs. You don't have to change it though.

jahooma (approved):
Awesome work, Sinclair! Looks like a fully working feature.

Left a couple comments on the code, but in general it's written very well.

Minor design gripes:
- Something about the skip button's position doesn't seem intuitive. Perhaps you could make it a separate button next to the progress bar? The progress bar kind of looks like a button as well.
- Very wide create ad button is not great haha.
- Actually, maybe the max width of the whole ad page should just be smaller. If it were as narrow as the home page and centered, I think the whole design would look wayyy better on desktop. This is something I discovered for the date docs as well.
- Title missing ink class for dark mode.
Screenshot 2023-03-09 at 9 22 03 PM

I'm extremely excited for this feature. I think everyone will immediately find it very fun and create all sorts of wacky ads. I think we should ship this and continue iterating based on feedback. Also, I definitely would have taken a cut for manifold, but I see now that it's not obviously necessary.

backend/functions/src/api/create-post.ts
```diff
@@ -34,23 +35,38 @@ const contentSchema: z.ZodType<JSONContent> = z.lazy(() =>
...
+        question: z.string(),
+      }),
+      z.object({
+        type: z.literal('ad'),
+        totalCost: z.number().min(0),
+        costPerView: z.number().min(0),
+      }),
+      z.object({
+        type: z.literal('base'),
+      }),
```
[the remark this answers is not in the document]
sipec: ok just removing it insetead

PR #1593, opened 2023-03-11
Run API as separate Cloud Run service

mqp:
**WIP - need to test and fix bugs.**

Here's what happened:
- Moved all the endpoints from `backend/functions/src/api/` to `backend/api/src`.
- `backend/api` now has a new Cloud Run service that runs an Express app serving those endpoints (similar to the old serve.ts.)
- Now that's our API instead of having separate Cloud Functions for each endpoint.

Consequences:
- The new API service is using Node 19, but it doesn't matter much.
- Code that was shared between API endpoints and other cloud functions was moved into `backend/shared`.
- `firebase deploy` will no longer redeploy the API. To deploy the API use `backend/api/deploy.sh [prod|dev]`. It will build a new Dockerfile and deploy it to the service. This is much faster (i.e. maybe takes a minute, vs. who knows how long for the old one.)
- This should save us a ton of money, since we have one box, instead of separate `minInstances: 1` services for every endpoint humming along every day. Probably around $50/day if you add up both dev and prod.


mqp:
FYI @akrolsmir @jahooma @IanPhilips - any feedback on this?

I will probably do a kind of "cleanup pass" after this merge to tidy things up.

If you still want to move to using Vercel functions after doing this, then go nuts and I will no longer make any complaining about it.

jahooma:
Wow, this is great! I was just thinking about the backend architecture today (https://www.notion.so/manifoldmarkets/Backend-architecture-7399131953954568a9c76bee1f996431). Would love to get your thoughts on that as well.

Having a server kind of makes sense. One reason I like it is because we can fulfill backend requests with a mix typescript code and SQL rather than just a SQL function RPC.

There's obviously a lot of other reasons to have a server, from cost to better latency to unlocking other tech (like websockets). I even think the dev experience could be better if it's simple to run locally.

Excited for this change!

backend/api/src/validate-iap.ts
```diff
@@ -23,9 +23,8 @@ const PRODUCTS_TO_AMOUNTS: { [key: string]: number } = {
 }
 
 const IAP_TYPES_PROCESSED = 'apple'
-const opts = { secrets: ['MAILGUN_KEY'] }
 
-export const validateiap = newEndpoint(opts, async (req, auth) => {
+export const validateiap = authEndpoint(async (req, auth) => {
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const iap = require('@flat/in-app-purchase')
```
IanPhilips, line 29: Looks like this was removed from `package.json`

IanPhilips: Oh good, it was added back in to `api/package.json`

PR #1594, opened 2023-03-11
Ingawei/man 929 create separation of group pages for

ingawei:
Reorganized groups page so that certain static props are only fetched conditional on group privacy
image
- GroupPageContent just fetches group params from supabase, and has static props as a backup if they exist
- there aren't really any permission checks on this, this is written out assuming RLS will eventually be implemented

would like a sanity check on the architecture since we will likely have to also do this for private markets/posts/etc.
- You can't currently make a private group through the front end, I'd recommend just changing groupPrivacy in firebase to test it out


web/hooks/use-group.ts
```diff
@@ -205,3 +207,18 @@ export function useGroupContractIds(groupId: string) {
 export function useGroups(groupIds: string[]) {
   return useStoreItems(groupIds, listenForGroup, { loadOnce: true })
 }
+
+export function useIsGroupMember(groupSlug: string) {
+  const [isMember, setIsMember] = useState<any>(undefined)
+  const user = useUser()
+  useEffect(() => {
+    if (user) {
+      setTimeout(() => {
```
jahooma, line 216: Other people that use this hook might not want a delay haha, probably should either rename the hook, or add a param to put the delay.

Also, I'm skeptical of delays like this. People want access ASAP. An unlocking animation might be the kind of thing you'd want to see once.

web/hooks/use-group.ts
```diff
@@ -205,3 +207,18 @@ export function useGroupContractIds(groupId: string) {
 export function useGroups(groupIds: string[]) {
   return useStoreItems(groupIds, listenForGroup, { loadOnce: true })
 }
+
+export function useIsGroupMember(groupSlug: string) {
+  const [isMember, setIsMember] = useState<any>(undefined)
```
jahooma, line 212: Wouldn't this be `boolean | undefined` instead of `any`?

web/hooks/use-post-supabase.ts
```diff
@@ -0,0 +1,57 @@
...
+      const channel = db.channel('post-realtime')
+      channel.on(
+        'postgres_changes',
+        {
+          event: '*',
+          schema: 'public',
+          table: 'posts',
+          filter: `id=eq.${postId}`,
+        },
+        (payload) => {
```
jahooma, line 34: I'm pretty sure payload contains the updated post... Did you console.log it?
Then you would not need to fetch it at all.

web/hooks/use-group-supabase.ts
```diff
@@ -238,3 +241,49 @@ export async function setTranslatedMemberRole(
...
+    channel.on(
+      'postgres_changes',
+      {
+        event: '*',
+        schema: 'public',
+        table: 'groups',
+        filter: `data->>slug=eq.${groupSlug}`,
+      },
+      (payload) => {
+        fetchGroup()
```
jahooma, line 270: I'm pretty sure payload contains the updated group... Did you console.log payload?
Then you would not need to fetch it.

web/lib/supabase/group.ts
```diff
@@ -123,3 +124,21 @@ export async function searchUserInGroup(
     'member_id'
   ).slice(0, limit)
 }
+
+export async function getGroupPrivacyBySlug(groupSlug: string) {
```
jahooma, line 128: General comment: We should probably switch to using group ids everywhere instead of having two ways to identify groups.

Oh, I see this is used from the group page when reading the url slug. That's the one acceptable usage haha.

web/pages/group/[...slugs]/index.tsx
```diff
@@ -1,198 +1,212 @@
...
     groupId: group?.id,
   })
 
-  const [writingNewAbout, setWritingNewAbout] = useState(false)
-  const bannerRef = useRef<HTMLDivElement | null>(null)
-  const bannerVisible = useIntersection(bannerRef, '-120px', useRef(null))
-  const isMobile = useIsMobile()
-  if (group === null || !groupSubpages.includes(page) || slugs[2] || !creator) {
-    return <Custom404 />
+  if (!group) {
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:195]: Did you mean `group === undefined`? Below, a null group leads to the 404 page.

jahooma (approved):
Thanks Inga! This is a nice way to ship a portion of the private groups feature.

Overall strategy for rendering private groups looks good. Just had a couple comments.

web/hooks/use-group.ts
ingawei: Yeah I can move it out, it's mainly for refreshes on the group page, or else I get an error that's like "must be logged in to call API"

web/hooks/use-group.ts
ingawei: yeah I was getting an error because it doesn't recognize the return type of `getUserIsGroupMember` as a boolean

PR #1602, opened 2023-03-14
Ingawei/man 943 create private market functionality

ingawei:
Did static props separation for market page as well (very similar to groups PR from a few days ago)


web/pages/[username]/[contractSlug].tsx
```diff
@@ -165,64 +191,66 @@ export async function getStaticPaths() {
...
     totalPositions,
-    pointsString,
     creatorTwitter,
     relatedContracts,
-  } = props
-  const contract = useContract(props.contract?.id) ?? props.contract
+  } = contractParams
+  const contract =
+    useContract(contractParams.contract?.id) ??
+    (contractParams.contract as Contract) ??
```
Comment (jahooma) [inline web/pages/[username]/[contractSlug].tsx:245]: Why cast as `Contract`? Isn't it already `Contract | null`?

web/hooks/use-contracts.ts
```diff
@@ -168,3 +170,28 @@ export const useContracts = (
...
+  const [privateContract, setPrivateContract] = useState<
+    Contract<AnyContractType> | undefined | null
+  >(undefined)
+  const user = useUser()
+  useEffect(() => {
+    // if there is no user
+    if (user === null) {
+      setPrivateContract(null)
+    } else if (user) {
+      // need this timeout (1 sec works) or else get "must be signed in to make API calls" error
```
jahooma, line 184: Hmm, would be worth investigating at some point why this happens.

Oh, I think it could be because we load the user from localstorage optimistically. The actual auth state is set as a callback to `onIdTokenChanged` in `auth-context.ts`.

One solution would be to add another field to the AuthUser state like `authLoaded: boolean`. Then, only set it once the callback to onIdTokenChanged is called. And maybe have a hook `useAuthUser` that gives the whole object.

If you don't want to do this, I can, after you merge.

web/components/contract/private-contract.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+export function ContractParamsPageContent(props: {
+  contract: Contract<AnyContractType>
+}) {
+  const { contract } = props
+  const contractParams = useContractParams(contract)
+  return <ContractPageContent contractParams={contractParams} />
+}
+
+export function useContractParams(contract: Contract) {
```
jahooma, line 58: Hmm, this hook has a bunch of sub hooks that each make a new query. And each query sets state, triggering a ton of extra renders that slow things down.

An alternative would be to create an api endpoint that sends all the contract params data in a json blob, if you are authenticated and have permission.
I guess if RLS is working, it could just be a regular function that fetches all this data. And then you could call it in the `getStaticProps` for public markets as well.

jahooma (approved):
Nice work, Inga! In addition to paving the way for private markets, this also just migrates a lot of market page requests to supabase, which is good.

Had a few comments for you to address. I'll try to be more prompt with my reviews in the future.

web/pages/[username]/[contractSlug].tsx
Comment (ingawei) [reply web/pages/[username]/[contractSlug].tsx:245]: It's never null at that point since ContractPageContent doesn't show unless Contract is defined, I just don't know another way to redefine ContractParams

web/hooks/use-contracts.ts
ingawei: Please do this!

web/pages/[username]/[contractSlug].tsx
Comment (jahooma) [reply web/pages/[username]/[contractSlug].tsx:245]: Could do:
`export function ContractPageContent(props: { contractParams: ContractParams & { contract: Contract} }`

PR #1605, opened 2023-03-16
Change portfolio history to be native Supabase table

mqp:
The goal is to make the `user_portfolio_history` table have normal columns rather than a JSON blob and to migrate it completely away from Firestore. You can view this as a test run for future tables.

The plan is like this:

1. Alter the table to make the JSON data blob nullable and add new nullable real columns. Create any indexes we want with the native columns.
2. Create a `before update or insert` trigger on the table that, if it sees a JSON blob, populates the real columns with the contents of the blob. If it doesn't see a JSON blob, it does nothing.
3. **Now: all new rows will have their native columns populated.**
4. Run `update foo_table set id = id where some_real_column is null`, running the trigger a lot.
5. **Now: all rows will always have their native columns populated.**
6. Rewrite existing code to read from Supabase using the native columns.
7. Deploy this.
8. Rewrite existing code to write to Supabase using the native columns, writing a `null` JSON blob.
9. Deploy this.
10. **Now: we are fully migrated.**
11. Clean up -- remove the trigger, delete the `data` and `fs_updated_time` columns, turn off replication to the table, delete data from Firestore, add any constraints on the new native columns that we want. Repack table.

### DIFF
--- a/backend/functions/src/scheduled/backup-db.ts
+++ b/backend/functions/src/scheduled/backup-db.ts
@@ -51,7 +51,6 @@ export const backupDbCore = async (
     'latency',
     'views',
     'notifications',
-    'portfolioHistory',
     'folds',
   ]
   return await client.exportDocuments({ name, outputUriPrefix, collectionIds })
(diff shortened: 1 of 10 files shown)


mqp:
FYI @jahooma. I will plan on doing this migration tomorrow.

By the way, if we start migrating more tables away from Firestore that have real important data in them, we should figure out what is the backup story for Supabase.

jahooma:
Nice, the migration plan is clear. I like that we can keep the same table, just add new columns.
I glanced through the code and it looked reasonable.

PR #1609, opened 2023-03-17
Quests

IanPhilips:
Screenshot 2023-03-17 at 11 42 11 AM

Screenshot 2023-03-17 at 11 41 29 AM

Screenshot 2023-03-17 at 11 41 47 AM


backend/api/src/complete-quest.ts
```diff
@@ -0,0 +1,19 @@
...
+import { QUEST_TYPES } from 'common/quest'
+import { completeQuestInternal } from 'shared/quest'
+
+const bodySchema = z.object({
+  // must be of type QuestType
+  questType: z.enum(QUEST_TYPES),
+  userId: z.string(),
+})
+
+export const completequest = endpoint(async (req) => {
```
jahooma, line 13: It's easier to use `authEndpoint`, which would include the userId of the authenticated user.

web/public/mtg/app.js
```diff
@@ -96,7 +96,7 @@ function getKSamples() {
   shuffleArray(allCards)
   window.console.log(allCards)
   for (let j = 0; j < allCards.length; j++) {
-    key = allCards[j]
+    userKey = allCards[j]
```
jahooma, line 99: Did you mean to change this?

web/hooks/use-is-seen.ts
```diff
@@ -0,0 +1,33 @@
...
+      startMs = today.getTime()
+      endMs = startMs + DAY_MS
+    } else if (timePeriod === 'week') {
+      startMs = dayjs().startOf('week').add(1, 'day').valueOf()
+      endMs = startMs + DAY_MS * 7
+    }
+    getUserEventsCount(user.id, eventNames, startMs, endMs, db).then((count) =>
+      setSeen(count > 0)
+    )
+  }, [user])
```
jahooma, line 30: nit: This seems unexpected in a useSeen hook. Maybe rename to `useHasSeenEvent`? Or extract to a different hook.

jahooma (approved):
Nice work Ian! Cool to quickly get a quests reward system out.

From the product side, I dislike the weekly rewards, because it doesn't feel regular. Like, if I click share, it's hard to know that I did anything special except for the third time in a week.

We want people to eagerly get on each day and get their rewards. If they can expect to share something and get a reward, that is good. They will do it more. Same for market creation.

And weekly + daily rewards are more complex.

That said, happy to try this out and see how it goes!

backend/api/src/complete-quest.ts
IanPhilips: Then I'd have to get the user every time I want to call completeQuest, which is excessive imo. Anywhere you want to click a share link you have to add the user to the call.

web/public/mtg/app.js
IanPhilips: oh what, no

web/hooks/use-is-seen.ts
IanPhilips: It has this as it a required parameter:  `timePeriod: 'day' | 'week'`

IanPhilips: changed it to `useHasSeen`

backend/api/src/complete-quest.ts
jahooma: Not sure I understand. AuthEndpoint gets the user for you based on who is signed in (second param is auth, user id is `auth.uid`). You don't have to pass anything explicitly.

IanPhilips: Oh right, that was also based on not having the idtoken around but now I can wait for `useIsAuthorized()` so that works too

IanPhilips:
Daily quests definitely seem more effective but I personally don't feel prepared to share and create a market every day, so I didn't feel justified putting that pressure on our users. That could be my unique problem and others would enjoy doing those daily quests, however, so I've low confidence on what time period each quest should have.

jahooma:
It's kind of like the daily free market. You don't have to do it, but if you do, you get a reward.
I think the betting streak does have that pressure, however.

PR #1618, opened 2023-03-18
Ingawei/man 956 rewrite activity feed to use supabase

ingawei:


web/hooks/use-contract-supabase.ts
```diff
@@ -41,3 +59,43 @@ export const useContractParams = (contract: Contract) => {
...
+      'postgres_changes',
+      {
+        event: 'INSERT',
+        schema: 'public',
+        table: 'contracts',
+      },
+      (payload) => {
+        if (payload) {
+          const payloadContract = payload.new.data as Contract
+          setNewContract(payloadContract)
```
jahooma, line 87: Ditto.

web/lib/supabase/comments.ts
```diff
@@ -12,3 +12,14 @@ export async function getAllComments(contractId: string, maxCount: number) {
   )
   return comments.map((comment) => comment.data)
 }
+
+export async function getComments(limit: number) {
+  let q = selectJson(db, 'contract_comments')
+  q = q
+    .order('fs_updated_time', {
```
jahooma, line 19: I think this is the time that the row got added to supabase. But many comments were added initially with a script.

So you should actually be using data->>createdTime, I think.

jahooma (approved):
Nice! It's cool to switch this onto supabase. Might save us some money too.

jahooma:
Oh yeah, and improving the latency is a big win for everyone that loads the home page!

PR #1619, opened 2023-03-18
Improve efficiency of `get_recommended_contracts`

mqp:
I like doing this because 1) this was responsible for a decent chunk of our total load 2) it's really nice to have the stuff on the home page load snappily.

What is going on here is bad statistics on JSON columns biting us again. Basically, as it was, the planner was deciding:

- OK, so they want the top N scoring contracts that satisfy this big complicated criteria (can't be closing soon, binary, etc.)
- That looks like a big complicated criteria. I bet most of the contracts will not satisfy it. If I start going down the list of scores looking up contracts by ID, I will probably end up looking up a shitload of contracts before I have N of them that satisfy the criteria.
- So it would be faster to just scan the whole contracts table, figure out which ones satisfy the criteria, and then go down the scores list with that in mind.

But the planner is wrong; it happens to be that most of our contracts satisfy the criteria. It's wrong because it's so bad at looking inside JSON columns to compile stats about the distribution of values. If the fields weren't in JSON columns, it would know better and we wouldn't have this problem. But they are.

To work around this issue, I made a covering index for the whole filter that has all the criteria. Then the planner sees that it can scan the index to find the contracts that fit the criteria, instead of scanning the contracts table (way bigger.) So it ends up being much faster. It's possible there would have been other things that nudged the planner to do things differently, but this one seems pretty simple and foolproof.

Let me know if this makes sense @jahooma.

### DIFF
--- a/backend/supabase/functions.sql
+++ b/backend/supabase/functions.sql
@@ -110,9 +110,13 @@ returns table (data jsonb, score real)
 immutable parallel safe
 language sql
 as $$
-  with recommended_contracts as (
-    select data, score
+  with recommendation_scores as materialized (
+    select contract_id, score
     from get_recommended_contract_scores_unseen(uid)
+    order by score desc
+  ), recommended_contracts as not materialized (
+    select data, score
+    from recommendation_scores
     left join contracts
     on contracts.id = contract_id
     where is_valid_contract(data)
@@ -228,7 +232,7 @@ create or replace function is_valid_contract(data jsonb)
     stable parallel safe
 as $$
 select not (data->>'isResolved')::boolean
-       and (data->>'visibility') != 'unlisted'
+       and (data->>'visibility') = 'public'
        and (data->>'closeTime')::bigint > ts_to_millis(now() + interval '10 minutes')
 $$ language sql;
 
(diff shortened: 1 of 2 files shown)


backend/supabase/functions.sql
```diff
@@ -110,9 +110,13 @@ returns table (data jsonb, score real)
 immutable parallel safe
 language sql
 as $$
-  with recommended_contracts as (
-    select data, score
+  with recommendation_scores as materialized (
+    select contract_id, score
     from get_recommended_contract_scores_unseen(uid)
+    order by score desc
+  ), recommended_contracts as not materialized (
```
mqp, line 117: I am not sure if you are familiar with `materialized` and `not materialized` but it does what it sounds like -- if you write `as materialized` then the CTE is computed and written to disk (useful if you want it to not repeat its work), if you write `as not materialized` then the CTE is not written to disk and re-queried when it's used later (useful if the work is cheap or if you want the planner to "see through" the CTE and choose how to do the query based on how it's used.) If you write nothing then the planner decides.

In this case, I want the recommendation scores to be written to disk, so they don't get computed twice, but I definitely don't want the `recommended_contracts` to be written to disk because (since I know that most contracts are valid) I know that's going to be way more contracts than I actually need. I want the planner to wait and combine this query with the `trending_contracts` and `new_contracts` filters and use the index I made.

PR #1624, opened 2023-03-21
draw comments from supabase

ingawei:


web/components/comments/comments-list.tsx
```diff
@@ -27,24 +28,40 @@ function contractPath(slug: string) {
...
+  const [pageNum, setPageNum] = useState(0)
+  const numComments = useNumUserComments(user.id)
+  const [pageComments, setPageComments] = useState<
+    {
+      key: {
+        contractId: any
+        contractQuestion: any
+        contractSlug: any
+      }
+      items: unknown[]
```
jahooma, line 41: Why are they key fields `any` instead of `string`? Aren't items of type `ContractComment`s?

web/components/comments/comments-list.tsx
```diff
@@ -27,24 +28,40 @@ function contractPath(slug: string) {
...
-      contractId: c.contractId,
-      contractQuestion: c.contractQuestion,
-      contractSlug: c.contractSlug,
-    }
-  })
-
-  if (isLoading) {
-    return <LoadingIndicator className="mt-4" />
-  }
+  useEffect(() => {
```
jahooma, line 46: This works fine. However, it wouldn't keep any comments from previous pages in memory, so there will be latency going back and forth between pages.

One way to do that would be to keep a complete list of `items` and append to it for each new page loaded. Then, you would slice it based on the current page number.

If you figured this out, it would also be worth creating a hook for it so the code is reusable (like we had for firebase).

jahooma (approved):
Nice, cool to migrate more code over to Supabase!

web/components/comments/comments-list.tsx
ingawei: will figure out later

PR #1625, opened 2023-03-21
Auth to Supabase as Firebase user

mqp:
**WIP -- need to handle token refresh, subscriptions, and dev/prod security checks.**

To support Supabase RLS mostly, although it could also be useful for other things (RPCs can see your UID too.)


mqp:
OK, this doesn't handle refresh, but other than that I think it's good for now and I will keep poking at it.

PR #1626, opened 2023-03-21
use useIsAuthorized instead of useUser

ingawei:


jahooma:
can remove console.logz

PR #1636, opened 2023-03-25
Kill period unique bettors fields on contracts

mqp:
We don't need these anymore and we aren't likely to in the future.


jahooma (approved):
LGTM! This is a really nice thing to be rid of.

PR #1641, opened 2023-03-28
create waiting for supabase new groups/contracts

ingawei:
This will reroute to a loading page upon creation of a new market/group while it waits for supabase to have said group. 
image


jahooma:
The waiting for a market to be created is not a good experience. There must be another way. Could we just write to supabase at the same time as firebase in api/`createmarket`?

jahooma:
An alternative would be to load the contract using firebase admin, but that seems annoying.

Marshall suggests that the replicator should be pretty fast, but empirically it isn't. Maybe he can improve that speed. I don't really want to merge this unless it is faster.

jahooma:
Another way to do this would be to create a get-market api endpoint that reads from firebase (using admin library so no security rules issues!), and returns the market (if not private) and visibility. Then call that from `getStaticProps`

jahooma (approved):
Thanks Inga for finding a solution! Looks good.

PR #1647, opened 2023-03-29
Fix positions endpoint docs: sort -> order

kenakofer:
The api code uses the url parameter "order", not "sort", so this documentation did not match the functionality. Given that the rest of the api docs uses "order", "order" is probably the preferred choice here.


IanPhilips:
Good eye, thanks @kenakofer!

PR #1668, opened 2023-04-05
Ingawei/man 986 make supabase contract search

ingawei:
Here are some metrics on how well the function runs given different parameters:
-[ fuzzy search within a group, using last-updated sort ](https://explain.dalibo.com/plan/2226aeb8569bd33c)
-[ fuzzy search using last-updated sort ](https://explain.dalibo.com/plan/h0d4ea45515512gc)
-[ full text search using last-updated sort ](https://explain.dalibo.com/plan/811c9cd208770eba)
-[ full text search using popularity sort ](https://explain.dalibo.com/plan/f232e5955b4b3c9c)
-[ empty search using last-updated time ](https://explain.dalibo.com/plan/2g861ebha6abe7d3) <-- this one is a yikes at ~700 ms execution time and may benefit from moving the last-updated field into it's own column
-[ empty search using popularity](https://explain.dalibo.com/plan/a03261g3eec309c3)


web/components/search/omni-search.tsx
```diff
@@ -119,14 +119,32 @@ const Results = (props: { query: string }) => {
...
+        filter: 'all',
+        sort: 'most-popular',
+        offset: 0,
+        limit: marketHitLimit,
+      }),
+    ]).then(
+      ([
+        userHits,
+        groupHits,
+        { fuzzyOffset: _fuzzyOffset, data: marketHits },
```
jahooma, line 139: You could delete `fuzzyOffset: _fuzzyOffset,` if unused

web/components/supabase-search.tsx
```diff
@@ -0,0 +1,503 @@
...
+  sort: Sort
+  filter: filter
+  //   facetFilters: SearchOptions['facetFilters']
+}
+
+function getShowTime(sort: Sort) {
+  return sort === 'close-date' || sort === 'resolve-date' ? sort : null
+}
+
+export const INITIAL_STATE = {
```
jahooma, line 59: Could add type annotation so it stays in sync: 
`export const INITIAL_STATE: stateType`

web/components/supabase-search.tsx
```diff
@@ -0,0 +1,503 @@
...
+    fromGroupProps,
+    listViewDisabled,
+  } = props
+
+  const [state, setState] = usePersistentState<stateType>(
+    INITIAL_STATE,
+    !persistPrefix
+      ? undefined
+      : { key: `${persistPrefix}-supabase-search`, store: inMemoryStore() }
+  )
```
jahooma, line 137: Use `usePersistentInMemoryState` which is the new way.

Hmm, but I guess you're not persisting it without the persistPrefix? I guess you could conditionally use a plain `useState` if we don't want it to persist.

Otherwise, we should make the `persistPrefix` not optional. Optional properties it too easy to mess up if you really should be setting it.

web/components/supabase-search.tsx
```diff
@@ -0,0 +1,503 @@
...
+  const loadMoreContracts = () => {
+    performQuery(() => state)()
+  }
+
+  const searchParams = useRef<SupabaseSearchParameters | null>(null)
+  const searchParamsStore = inMemoryStore<SupabaseSearchParameters>()
+  const requestId = useRef(0)
+  const [asList, setAsList] = usePersistentState(!listViewDisabled, {
```
jahooma, line 145: `usePersistentInMemoryState` here.

web/components/supabase-search.tsx
```diff
@@ -0,0 +1,503 @@
...
+        setState({
+          fuzzyContractOffset: newFuzzyContractOffset,
+          contracts: freshContracts,
+          showTime: showTime,
+          shouldLoadMore: newContracts.length === 20,
+        })
+        if (freshQuery && isWholePage) window.scrollTo(0, 0)
+      }
+
+      // Rest of the function
```
jahooma, line 206: ?

web/pages/api/v0/search-markets.ts
```diff
@@ -12,6 +12,18 @@ export default async function handler(
...
 
-  const contracts = await searchContracts(keywords, 100)
-  res.status(200).json(contracts.map((c) => toFullMarket(c)))
+  const { fuzzyOffset: _fuzzyOffset, data } = await searchContract({
+    state: {
+      contracts: undefined,
+      fuzzyContractOffset: 0,
+      shouldLoadMore: false,
+      showTime: null,
+    },
```
jahooma, line 21: Should create this default state within `supabase/contract.ts`. You could even make it the default argument value of state, e.g. function searchContract(query: string, filter: string, ... state = defaultState) { ... }

web/lib/supabase/contracts.ts
```diff
@@ -117,17 +119,106 @@ export async function getContractVisibilityFromSlug(contractSlug: string) {
...
+  sort: Sort
+  offset: number
+  limit: number
+  group_id?: string
+  creator_id?: string
+}) {
+  const { state, query, filter, sort, offset, limit, group_id, creator_id } =
+    props
+
+  if (!query || query.length == 0) {
```
jahooma, line 135: Technically `''` is falsey, so the first condition is enough. But also, query has to be a string (based on the type), so the second condition would also be sufficient.

backend/supabase/contracts/empty-search-contracts-function.sql
```diff
@@ -0,0 +1,84 @@
+CREATE OR REPLACE FUNCTION empty_search_contracts(
```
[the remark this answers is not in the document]
jahooma: Also, I miss the new lines between sql functions.

backend/supabase/contracts/empty-search-contracts-function.sql
```diff
@@ -0,0 +1,84 @@
...
+    END DESC NULLS LAST,
+    CASE
+      %L
+      WHEN '' newest '' THEN created_time
+      WHEN '' resolve - date '' THEN resolution_time
+    END DESC NULLS LAST,
+    CASE
+      WHEN %L = '' close - date '' THEN close_time
+    END ASC NULLS LAST OFFSET %s
+  LIMIT %s ',
```
jahooma, line 73: Hmm, I think the client should be sending the field to sort on and the direction. What if we want to try a new sort? We'd have to update the sort here too. The backend should be more capable.

backend/supabase/contracts/empty-search-contracts-function.sql
```diff
@@ -0,0 +1,84 @@
...
+    limit_n INTEGER,
+    group_id TEXT DEFAULT NULL,
+    creator_id TEXT DEFAULT NULL
+  ) RETURNS TABLE (data jsonb) AS $$
+DECLARE base_query TEXT;
+where_clause TEXT;
+sql_query TEXT;
+BEGIN -- Common WHERE clause
+-- If fuzzy search is enabled and is group search
+IF group_id is not null then base_query := FORMAT(
```
jahooma, line 14: I'm not sure about generating the sql from the sql function. It's maybe ok if we're just doing a small amount of it. But it seems like we'd much prefer creating all the sort, where clause, and fuzzy vs full text search from a typescript api endpoint.

Any reason why not to do this?

jahooma (commented):
Thanks Inga! This is awesome to get our search off of algolia. Lots of detailed work to get this all running.

The only part I'm unsure about is how we generate the sql within the sql functions. We can talk about it.

backend/supabase/contracts/search-contracts-function.sql
```diff
@@ -0,0 +1,161 @@
...
+  contract_sort,
+  contract_sort,
+  creator_id,
+  creator_id
+);
+RETURN where_clause;
+END;
+$$ LANGUAGE plpgsql;
+-- generates latter half of search query
+CREATE OR REPLACE FUNCTION generate_sort_query(
```
jahooma, line 116: Should dedupe with above empty sort query function

backend/api/src/supabase-search-contract.ts
```diff
@@ -0,0 +1,188 @@
...
+    const pg = createSupabaseDirectClient()
+    const searchMarketSQL = getSearchContractSQL({
+      term: term,
+      filter: filter,
+      sort: sort,
+      offset: offset,
+      limit: limit,
+      fuzzy: fuzzy,
+      groupId: groupId,
+      creatorId: creatorId,
```
jahooma, line 30: nit: you can just write
`term,
filter,
sort,
...`

backend/api/src/supabase-search-contract.ts
```diff
@@ -0,0 +1,188 @@
...
+        ${whereSQL}
+        AND contractz.group_id = '${groupId}'`
+    }
+    // if fuzzy search within group
+    else if (fuzzy) {
+      query = `
+        SELECT contractz.data
+        FROM (
+            SELECT contracts.*,
+                similarity(contracts.question, '${term}') AS similarity_score,
```
jahooma, line 79: This approach looks very susceptible to SQL injection. We either have to get a library to clean all the parameters (term, filter sort, etc) or somehow compose the queries with pg-promise.

jahooma (commented):
Thanks for updating to use an api endpoint to compose the sql rather than a sql function!

Looks good except for the sql injection possibility.

PR #1672, opened 2023-04-07
Reverse order=shares for NO positions contract-metrics.ts

kenakofer:
In `getContractMetricsForContractId()`, switch order=shares NO holders to be sorted ascending, so that top NO holders can be fetched without parsing all positions, as it is already possible to do with YES holders.

This is a more intuitive, continuous ordering, starting with strong YES, then weak YES, then weak NO, then strong NO at the end.

### DIFF
--- a/common/src/supabase/contract-metrics.ts
+++ b/common/src/supabase/contract-metrics.ts
@@ -183,7 +183,7 @@ export async function getContractMetricsForContractId(
       .select('*')
       .eq('contract_id', contractId)
       .eq(`data->hasNoShares`, true)
-      .order(`data->totalShares->NO` as any, { ascending: false })
+      .order(`data->totalShares->NO` as any, { ascending: true })
 
     q = q
       .eq(`data->hasYesShares`, true)


IanPhilips:
Hey thanks for the PR! I use the NO shares sorted descending in the `user-positions-table` bc these positions are of the most interest to users. I see your point for the API though so if you feel very strongly about this and fix it just for the api then I'll merge your changes.

kenakofer:
Thanks for the response!

For my own edification, can you point me to where the `user-positions-table` is using NO shares sorted descending? Searching the repository for `getContractMetricsForContractId`, I only see (besides the API) this use, which is only for `'profit'` and would be unaffected by my change:
```ts
    useEffect(() => {
      if (sortBy === 'profit' && contractMetricsByProfit === undefined) {
        getContractMetricsForContractId(contractId, db, sortBy).then(
          setContractMetricsByProfit
        )
      }
    }, [contractId, contractMetricsByProfit, sortBy])
```

IanPhilips:
Oh right, I don't get the positions from supabase in the contract's static props _yet_ but I will as soon as we migrate the positions over from firebase. Right now they are queried via firebase in the same descending YES,descending NO orders via the `getCPMMContractUserContractMetrics` call.

kenakofer:
@IanPhilips OK, I pushed a solution that only modifies the `positions` endpoint. In the long run, it seems like it might be nicer to separate out YES from NO positions.

I don't know how to test my changes. I see the `Vercel - dev` deployment, but it looks like my github account can't access it? Or maybe I just don't know where to go.

IanPhilips:
I just approved the deployment to prod as well, but you actually have to click 'Visit Preview' on the first bot message to get the url/site to try out the changes.
Screenshot 2023-04-11 at 8 24 58 AM

kenakofer:
Ah, thanks. I need to fix a bug.

kenakofer:
@IanPhilips This pull request is ready for review.

PR #1688, opened 2023-04-17
cascade visibility changes

ingawei:
- I've denormalized visibility on bets and comments, this updates them when you change contract visibility. I might've done it in an inefficient way pls let me know


backend/functions/src/triggers/on-update-contract.ts
```diff
@@ -113,6 +119,46 @@ const getPropsThatTriggerRevalidation = (contract: Contract) => {
...
+async function processSubcollection(
+  subcollectionRef: FirebaseFirestore.CollectionReference,
+  newVisibility: 'public' | 'unlisted'
+) {
+  const batchSize = 500
+
+  let moreDocuments = true
+
+  while (moreDocuments) {
+    const snapshot = await subcollectionRef.limit(batchSize).get()
```
jahooma, line 146: Needs an offset or it would always keep fetching the same docs. See `processPaginated` in backend/shared/utils.ts

ingawei:
@jahooma I'm using processPaginated now, please check if this is right!

jahooma (approved):
Looks right, yeah!

PR #1690, opened 2023-04-17
Add .sql to prettier

sipec:
I couldn't find a nice postgres formatter that didn't do nonsense like always putting `select` on its own line
There's https://github.com/nene/prettier-plugin-sql-cst ... but it doesn't work for postgres syntax 😔

So I just went the path of least resistance and added the most used prettier sql plugin. (Which kinda doesn't make sense to put in `web/` but oh well).

I actually liked the most recent formatter that @ingawei was using more, but I don't feel like making everyone get the sqltools vscode extension

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -3,5 +3,16 @@
   "useTabs": false,
   "semi": false,
   "trailingComma": "es5",
-  "singleQuote": true
+  "singleQuote": true,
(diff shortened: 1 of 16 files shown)


ingawei (approved):
Thank you Sinclair! More white space is nicer

jahooma:
Seems ok and it's nice that it's prettier

PR #1698, opened 2023-04-19
Update a bunch of stuff to read contract metrics from Supabase

mqp:
Kind of WIP, I am planning to migrate the whole contract metrics table over, but I am waiting on some work Ian is doing to do the whole thing. First step is migrating over random reads.


mqp:
Obsolete, doing this work in a different order now.

PR #1700, opened 2023-04-19
Listen to contract metrics via supabase

IanPhilips:


jahooma:
What's the status of this PR? Wasn't this blocking @mqp?

web/lib/supabase/utils.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  useEffect(() => {
+    initChannelAndCallbacks()
+  }, [db])
+
+  const initChannelAndCallbacks = () => {
+    // very first channel opened
+    if (!mainChannel) subscribeToNewChannel()
+    // channel already exists, restart and resubscribe callbacks, adding my own
+    else restartChannel()
```
mqp, line 69: Why is it good to restart the channel?

web/lib/supabase/utils.ts
```diff
@@ -0,0 +1,256 @@
...
+let mainChannelResubscribeInterval: NodeJS.Timer | undefined = undefined
+
+export function useValuesFromSupabase<
+  T extends TableName | ViewName,
+  K extends keyof RowFor<T>
+>(
+  table: T,
+  rowGroupKey: K,
+  rowGroupValue: string,
+  uniqueRowDataKey: K,
```
mqp, line 39: This should always be the same for any table name, right? You could make a map from table name to primary keys, so that people don't need to pass it into this code.

mqp:
This is actually just way more complicated than I thought it would be -- I don't entirely understand this version yet, never mind how it should be made. I will have to look at it more tomorrow. If you like it, merge, but I think we will be due for a lot more work on this...

IanPhilips:
MQP and I had a 1:1 and decided on a few things to do before merging

web/lib/supabase/utils.ts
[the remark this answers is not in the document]
IanPhilips: good idea, done!

IanPhilips:
>    Given these, why do you need the resubscribe timer? In what case do you want to resubscribe but it's not being handled internally by one of these? Is it if the initial join fails? That seems like you could do it once rather than having a timer though.  

The resubscribe timer is needed bc it handles channel connection errors, and is on a timer bc the retry sometimes fails.

> Why do we restart the channel? This seems to add an incredible amount of complexity managing the callbacks and I don't understand why we need to do it.  

Unfortunately we can only hit subscribe once on a channel. If we want to add new events to subscribe to, we have to recreate the channel.

>    What is the plan for using this in cases where two different pieces of code are subscribing to an overlapping collection of stuff? (e.g. two places on the same page that both want the bets feed.) Is that possible for it to handle? It seems that this design will not work right, unless I am mistaken.  

This handles that case: the two hooks pass a unique callback invoked based on the same event.

>    It would seem more React-y to put this all in a React context, rather than have it at the top level of a random Javascript module. I'm not sure if there's any important reason to do that though.  

I'm not sure whether this is necessary either, @jahooma do you have an opinion? It'd be for the shared state (`manChannel`,`subscriptionStatus`,`callbacks`,`mainChannelResubscribeInterval`) at the top of `use-subscription.ts`

IanPhilips:
Just sent supabase the following message:
Hi,

We recently added a small number of realtime channel subscriptions to our app and are planning on adding them in more places. During testing I've been hitting the `error: too_many_connections`. Based on this I have a few questions:
1. Is the error I'm seeing related to the 500 max concurrent channels cap?  
2. What is the purpose of the 500 max concurrent channels?  Is this a technical limitation or just a safety rail?
3. Can you raise our concurrent channel cap? 
4. What is the pricing rate for higher concurrent channel caps?

IanPhilips:
Open question: how do we select the most up to date object if we don't use the firestore-dependent `fs_updated_time`?

web/hooks/use-subscription.ts
```diff
@@ -0,0 +1,253 @@
...
+      // During DELETE operation, oldValue will have the primary key values set and NO data
+      const oldValue = oldRecord as RowFor<T>
+      const index = prev.findIndex((m) => rowMatches(m, oldValue))
+      // New row with filled values
+      if (newRecord && primaryKeyColumns.every((k) => newRecord[k] !== null)) {
+        const newValue = newRecord as RowFor<T>
+        // We already have this row, so update it
+        if (oldRecord && index > -1) {
+          prev[index] = newValue
+          return prev
```
jahooma, line 117: Does it work to mutate the old array? I thought setState would not trigger a re-render if the reference is the same.

jahooma:
You could use React Context if you are trying to cache the data so that multiple hooks don't refetch. Or, you could just use a `usePersistentInMemoryState` — which is probably what I'd prefer.

That seems like a future upgrade to this hook since it's not 100% necessary yet. (You would have to store the data by unique table and filter).

jahooma:
> Open question: how do we select the most up to date object if we don't use the firestore-dependent `fs_updated_time`?

I don't see why we need this at all. In the realtime subscription, doesn't it give you the newer value? You then know that it is newer. If you want to know if it changed, then do a deep equality comparison with the old value, e.g. `isEqual` from lodash.

IanPhilips:
> > Open question: how do we select the most up to date object if we don't use the firestore-dependent `fs_updated_time`?
> 
> I don't see why we need this at all. In the realtime subscription, doesn't it give you the newer value? You then know that it is newer. If you want to know if it changed, then do a deep equality comparison with the old value, e.g. `isEqual` from lodash.

This is in reference to the state where you've subscribed, triggered fetching the initial state, gotten an update via subscription, then finished fetching the initial state.

web/hooks/use-subscription.ts
```diff
@@ -0,0 +1,253 @@
...
+
+//TODO: fs_updated_time relies on firebase, which we are moving off of. We
+// have no better solution for updated time atm.
+function hasFsUpdatedTime(obj: any): obj is { fs_updated_time: number } {
+  return 'fs_updated_time' in obj
+}
+type RowFilter<T extends TableName> = { k: keyof RowFor<T>; v: string }
+const CHANNEL_NAME = 'supabase-realtime-values'
+
+let mainChannel: undefined | RealtimeChannel = undefined
```
jahooma, line 21: Doesn't this imply there's only one channel? What if you have two hooks listening to different tables?

jahooma:
> This is in reference to the state where you've subscribed, triggered fetching the initial state, gotten an update via subscription, then finished fetching the initial state.

Minor race condition that mostly doesn't matter? You could also wait to start the subscription until after the initial state is loaded, as one solution.

jahooma:
I'm also skeptical about how this creates and restart channels. I've seen channels reconnect automatically. We've used them for the live feed and other cases and it hasn't been a problem just creating them once.

web/hooks/use-subscription.ts
IanPhilips: The channel has to restart and subscribe to updates to the two different tables. We should minimize channels per client due to realtime limitations:
Screenshot 2023-04-26 at 2 56 16 PM

IanPhilips:
> I'm also skeptical about how this creates and restart channels. I've seen channels reconnect automatically. We've used them for the live feed and other cases and it hasn't been a problem just creating them once.

I've tested this and if the client hits a channel_error you aren't resubscribed. Are you sure you've tested that case? Channels reconnect automatically if your wifi is toggled on/off, but not if there's a connection error.

web/hooks/use-subscription.ts
IanPhilips: Oh it looks like creating a new array is more robust, will do that instead

IanPhilips:
> I'm also skeptical about how this creates and restart channels. I've seen channels reconnect automatically. We've used them for the live feed and other cases and it hasn't been a problem just creating them once.

So, just checking up on this:  
`channel.subscribe(async (status) => {})`  
is the subscription code used throughout main, so it seems no one has actually tested an error on channel subscription.  

`  channel.subscribe(async (status, err) => {
      console.log('channel subscribe status', status, err)
// ...
})`  
This is how it is written in this PR

IanPhilips:
So I just added the subscription status and error checking on main into the `useRealtimeBets` hook and after a couple refreshes I got a channel error that the subscription did not resubscribe to.
Screenshot 2023-04-27 at 12 05 34 PM

jahooma:
I see, interesting! Thanks for finding this error. 500 channels seems extremely limited.

Did supabase get back to you yet?

IanPhilips:
No response yet

mqp:
BTW, after thinking about it for a bit I think we should just keep around an updated timestamp on all the DB tables we subscribe to -- it's easy to implement (you can just have it be a generated column in SQL) and then it means we can confidently order our updates without fear of the spooky race conditions Ian has mentioned. I basically see no downside.

IanPhilips:
Going to close this as mqp made his own pr for the same thing!

PR #1702, opened 2023-04-19
Change list view

ingawei:


jahooma:
What do you think about merging the % and Action columns so that it takes up less width? I.e. it's still labeled '%' but there's a Bet or Resolve button next to the percent.

jahooma:
Would it be possible to put the Bet button on mobile below the percent? Kind of wish you could bet from that screen.

web/components/answers/answer-resolve-panel.tsx
```diff
@@ -117,36 +120,73 @@ export function AnswerResolvePanel(props: {
...
+                Resolve{' '}
+                {resolveOption === 'CANCEL' ? (
+                  <>N/A</>
+                ) : resolveOption === 'CHOOSE' ? (
+                  chosenText
+                ) : (
+                  <>{answers.length} answers</>
+                )}
+              </>
+            </Button>
```
jahooma, line 188: This duplication of code / logic is not great. At the least, you should create variables for color, disabled, and button label defined above that you reuse.

web/components/answers/answer-resolve-panel.tsx
```diff
@@ -20,6 +20,7 @@ export function AnswerResolvePanel(props: {
     option: 'CHOOSE' | 'CHOOSE_MULTIPLE' | 'CANCEL' | undefined
   ) => void
   chosenAnswers: { [answerId: string]: number }
+  modalSetOpen?: (open: boolean) => void
```
jahooma, line 23: This isn't being used as a function in this component. Shouldn't is be a boolean instead? `isInsideModal`?

web/components/answers/answer-item.tsx
```diff
@@ -20,6 +20,7 @@ export function AnswerItem(props: {
   totalChosenProb?: number
   onChoose: (answerId: string, prob: number) => void
   onDeselect: (answerId: string) => void
+  modalSetOpen?: (open: boolean) => void
```
jahooma, line 23: This could also just be a boolean

web/components/answers/answers-panel.tsx
```diff
@@ -135,66 +136,87 @@ export function AnswersPanel(props: {
...
-
-      {resolution && answerItemComponents}
-
-      {!resolveOption && (
-        <Col className="gap-3">
-          {answerItems.map((item) => (
-            <OpenAnswer
-              key={item.id}
-              answer={item}
+  // if this is not in a modal
```
jahooma, line 143: If this is in a modal

web/components/contract/contracts-table.tsx
```diff
@@ -0,0 +1,284 @@
...
+          highlightContractIds?.includes(contract.id)
+            ? contractListEntryHighlightClass
+            : '',
+          (isClosed(contract) && contract.creatorId !== user?.id) ||
+            contract.isResolved
+            ? 'text-ink-500'
+            : '',
+          'hover:bg-primary-50 focus:bg-primary-50 group cursor-pointer'
+        )}
+        onClick={(e) => {
```
jahooma, line 191: This needs to be an actual link component. We've done this many times and users always complain. Without a native link, various features break like middle clicking or right clicking to open in a new tab.
I think you can just create the link within the table row: `<tr><Link></Link></tr>`.

web/components/resolution-panel.tsx
```diff
@@ -113,31 +121,60 @@ export function ResolutionPanel(props: {
...
+            color={
+              outcome === 'YES'
+                ? 'green'
+                : outcome === 'NO'
+                ? 'red'
+                : outcome === 'CANCEL'
+                ? 'yellow'
+                : outcome === 'MKT'
+                ? 'blue'
+                : 'indigo'
```
jahooma, line 161: ditto for code duplication

jahooma (approved):
Cool stuff! I'm sure users will greatly appreciate the added detail from this table view!

In fact, someone at the Solaris dinner yesterday said it would be great to see the number of traders from the list view and if they could bet directly from it as well. I guess they still wouldn't be able to on mobile... but would be great to make this possible.
I would even add a whole new row on mobile for each question if it showed the bet button and number of traders.

Had a few comments on the code before merging.

PR #1703, opened 2023-04-19
Kill algolia

sipec:
also simplifies `searchContract` to have reasonable defaults when not paginating

- [ ] cancel algolia account billing (after merge)


jahooma (approved):
Looks good to me! I think it's been long enough that we wouldn't need to revert to Algolia.

Ding-dong the witch is dead.

jahooma:
Ya, I can do that! There's actually only a pay-as-you-go option, so we will not be billed further if we don't use the api.

I also want to turn off the replication to Algolia via firebase extension.

PR #1706, opened 2023-04-24
Switch location of balance graph and portfolio value graph

bfinan:
Switching the location of these buttons so that:
- Profit and Portfolio Value are adjacent, since they are correlated
- Balance is adjacent to the "Get M$" button


jahooma:
Reasonable change, thanks!

PR #1708, opened 2023-04-25
Remove separate sharedholder stats querying on contract page

mqp:
This seems unnecessary because we are loading the whole list of positions right there, so we don't need to also get the count independently.


mqp:
Closing this because I realized we enforce a limit of 100 on how many of these we load so it isn't as easy as I thought.

PR #1717, opened 2023-04-27
Boosted markets

sipec:
commit order is spaghet
look at file diff instead

reward is still 5 mana and ads are 1 every 4. probably should change these


backend/supabase/functions.sql
```diff
@@ -29,6 +29,26 @@ where data->>'contentType' = 'contract'
...
+-- Get recomended ads
+-- TODO: finish
+create
+or replace function get_recommended_ads (uid text) returns table (data jsonb, score real) immutable parallel safe language sql as $$
+select 
+  data,
+  coalesce(dot(urf, crf) * crf.freshness_score, 0.0) as score
+from user_recommendation_features as urf
+  cross join contract_recommendation_features as crf
+  left join contracts on contracts.id = crf.contract_id
```
jahooma, line 41: Old method for recommendations. New: `user_embeddings.interest_embedding`

jahooma: Can delete

backend/supabase/seed.sql
```diff
@@ -845,6 +848,21 @@ create table if not exists
...
+    market_id text not null,
+    foreign key (market_id) references contracts (id),
+    funds numeric not null,
+    cost_per_view numeric not null,
+    created_at timestamp not null default now(),
+    embedding vector (1536) not null
+  );
+
+alter table market_ads
+add constraint market_ads_unique_market_id unique (market_id);
```
jahooma, line 864: I think this line will fail when run again. You should add the contraint to the bottom of the table definition:

create table if not exists
  market_ads (
    id text not null primary key default uuid_generate_v4 (),
    user_id text not null,
    market_id text not null,
    foreign key (market_id) references contracts (id),
    funds numeric not null,
    cost_per_view numeric not null,
    created_at timestamp not null default now(),
    embedding vector (1536) not null,
    unique (market_id)
  );

common/src/txn.ts
```diff
@@ -212,18 +214,30 @@ type QfDividend = {
...
+type PostAdRedeem = {
   category: 'AD_REDEEM'
   fromType: 'AD'
   toType: 'USER'
 }
 
+type MarketAdCreate = {
+  category: 'MARKET_BOOST_CREATE'
+  fromType: 'USER'
+  toTupe: 'AD'
```
jahooma, line 232: Typo

web/components/contract/creator-share-panel.tsx
```diff
@@ -4,32 +4,83 @@ import { formatMoney } from 'common/util/format'
...
+      <div className="text-ink-500 mb-2 text-base">
+        Bump up your market in the feed. We'll target it to users who like
+        questions like this one. All funds go to the viewers.
+      </div>
     </GradientContainer>
   )
 }
+
+// in mana. TODO: variable cost
+export const COST_PER_VIEW = 5
```
jahooma, line 52: Change to 1 or 2?

web/components/feed/feed-items.tsx
```diff
@@ -75,6 +103,20 @@ export const FeedItems = (props: {
   )
 }
 
+// every period items in A, insert an item from B
+function zipperMerge<A, B>(a: A[], b: B[], period: number): (A | B)[] {
```
jahooma, line 107: Wouldn't "zipperMerge" imply period 1? Maybe zipperPeriodMerge?

backend/api/src/place-bet.ts
```diff
@@ -49,7 +49,7 @@ const numericSchema = z.object({
 })
 
 export const placebet = authEndpoint(async (req, auth) => {
-  log(`Inside endpoint handler for ${auth.uid}.`)
+  log(`Inside endpoint handler for ${auth.uid}. LOL`)
```
jahooma, line 52: LOL

web/hooks/use-feed.ts
```diff
@@ -20,6 +27,12 @@ export const useFeed = (
   }
 ) => {
   const { topic } = options ?? {}
+
+  const [boosts, setBoosts] = useState<BoostsType>()
+  useEffect(() => {
+    if (user) getBoosts(user.id).then(setBoosts as any)
```
jahooma, line 33: Seems like we'll occasionally get errors when the ad runs out of funds (if someone else claims the last mana after you fetched the boost).
Any easy ideas to address this?

jahooma (approved):
Yay! This is great, Sinclair! Can't wait to see market boosts in action.

Just a couple comments.

jahooma:
Also, type error ```

./components/feed/feed-items.tsx:33:13
--
13:34:54.371 | Type error: Property 'map' does not exist on type '{}'.

```

File: backend/api/src/redeem-market-ad-reward.ts [hunk 8]
```diff
@@ -0,0 +1,74 @@
...
+
+  // find the advertisement
+  const data = await pg.one(
+    `select cost_per_view::numeric, funds::numeric from market_ads
+    where id = $1`,
+    [adId]
+  )
+
+  const reward = parseFloat(data.cost_per_view)
+  const funds = parseFloat(data.funds)
```
jahooma, line 45: This is the one place where we should rely on the txns as source of truth and not just rely on the ad's stored funds.

If we fetched funds from txns within the redeem transaction below, that would actually make it correct.

web/hooks/use-feed.ts
sipec: some options:
- require an extra 10 mana on boosting a new market as a "flat fee", and the query for ads is instead is `market_ads.funds > 10`. The fee usually will end up as "dust" in the ad, but occasionally it will be used to pay out both sides in a race
- pay out anyways so long as funds > -10, and we eat the cost
- subscribe to txns and hide dead ads (seems annoying for now)

sipec:
hmm static analysis doesn't consider supabase generated types

web/hooks/use-feed.ts
jahooma: Maybe if there's x mana of funds left, where x < 10, there's only an x% chance that it would show in your feed.

web/lib/firebase/api.ts
```diff
@@ -143,6 +143,23 @@ export function createPost(params: {
...
+}
+
+let nonce = 0
+export function redeemBoost(params: any) {
+  // TODO: rate limit on the backend instead?
+  const now = Date.now()
+  if (now - nonce < 15 * 1000) {
+    throw Error(
+      'Please wait 15 seconds after redeeming a boost before redeeming another.'
+    )
```
jahooma: Not a great user experience...  Ok to try it for now.

I think instead of the button we might want auto-claim with 5 seconds of viewing the ad on screen (with countdown 5-4-3-2-1).

PR #1719, opened 2023-04-27
Ingawei/make contracts page more efficient

ingawei:
Basically I've made getContractParams a backend api call, so that it is not bounded by client side RLS rules when fetching all comments/bets. It checks access to the contract once and then fetches everything, so should be more efficient for private market pages


backend/api/src/get-contract-params.ts
```diff
@@ -0,0 +1,180 @@
...
+    !contract.deleted &&
+    // can access if contract is not private
+    (contract.visibility != 'private' ||
+      // if contract is private, can't access if in static props
+      (!fromStaticProps &&
+        // otherwise, can access if user can access contract's group
+        auth &&
+        groupId &&
+        (await getUserIsMember(groupId, auth?.uid, db))))
+  if (canAccessContract) {
```
jahooma, line 58: Minor style point: if the `!canAccessContract` case is way shorter you might handle that within the if. Then the main body of the function can be the bulk of the code and un-indented.

backend/supabase/user_contract_metrics/triggers.sql
```diff
@@ -0,0 +1,38 @@
+alter table user_contract_metrics
```
jahooma, line 1: Converting `user_contract_metrics` to native columns should totally be a single PR. It's possible that Ian or Marshall might have opinions, but it's buried within other changes here.

backend/supabase/user_contract_metrics/triggers.sql
```diff
@@ -0,0 +1,38 @@
...
+add column has_yes_shares boolean,
+add column has_no_shares boolean,
+add column total_shares_yes numeric,
+add column total_shares_no numeric,
+add column profit numeric,
+add column has_shares boolean;
```
jahooma, line 7: We should come to agree on what to commit in sql. This approach is kind of a like a migration script of what you ran once.

I think the previous idea is that all the files could be run and produce the current state of the db. For example, when I add a column, I run the alter table command, but then only commit an edit to the initial creation of the table in `seed.sql` so that it represents the current schema.

For now this is fine, but I think we should pick a style and stick to it.

common/src/chart.ts
```diff
@@ -0,0 +1,36 @@
+import { AxisScale } from 'd3-axis'
```
jahooma, line 1: Refactoring chart types or moving around existing bet-fetching code should be a different PR.
It's good to separate refactors from changes in functionality.

web/hooks/use-group.ts
```diff
@@ -200,7 +200,7 @@ export function useIsGroupMember(groupSlug: string) {
   const isAuthorized = useIsAuthorized()
   useEffect(() => {
     // if there is no user
-    if (isAuthorized === null) {
+    if (isAuthorized === false) {
```
jahooma, line 203: You could change `useIsAuthorized` to return a boolean. I think I intended that but made the change quickly.

Also acceptable: make it return undefined if the user/no user is not loaded.
Returning null makes no sense because that is another case of not authorized.

jahooma (approved):
Thanks Inga! The approach of getting this data from an api endpoint makes a ton of sense to me!

What do you think about a simpler RLS policy that would never bring the database down:
Private groups / markets are never allowed to be fetched from the client. Public data is.
Then, all such private data would have to be fetched from an api endpoint (or from getStaticProps.

ingawei:
> Thanks Inga! The approach of getting this data from an api endpoint makes a ton of sense to me!
> 
> What do you think about a simpler RLS policy that would never bring the database down: Private groups / markets are never allowed to be fetched from the client. Public data is. Then, all such private data would have to be fetched from an api endpoint (or from getStaticProps.

I think yeah this might be eventually where we want to go, but for now it's nice to be able to fetch private stuff in the same call as other stuff for the user profile page for the time being

PR #1721, opened 2023-04-28
Adding arbitragebot to bots list

YSaxon:


akrolsmir (commented):
Hey @YSaxon, thanks for submitting this PR! There does already seem to be another bot by the name @ArbitrageBot -- yours is different via capitalization only, but could you consider changing your bot name to be a bit more unique? Thanks!

(Looks like "ArbBot" is taken, maybe "Arbitrader" or "ArbitraryBot"?

PR #1730, opened 2023-05-01
Improvements to leagues

SirSaltyy:
Added some stuff to leagues.

Also modified the tooltip component so that if it has children it won't show the ugly question mark and instead added a dotted line to the text that can be hovered/clicked.

### DIFF
--- a/web/components/widgets/info-tooltip.tsx
+++ b/web/components/widgets/info-tooltip.tsx
@@ -2,13 +2,28 @@ import { InformationCircleIcon } from '@heroicons/react/outline'
 import { Tooltip } from './tooltip'
 import clsx from 'clsx'
 
-export function InfoTooltip(props: { text: string; className?: string }) {
-  const { text, className } = props
+export function InfoTooltip(props: {
+  text: string
+  className?: string
+  children?: React.ReactNode
+}) {
+  const { text, className, children } = props
   return (
     <Tooltip className="inline-block" text={text}>
-      <InformationCircleIcon
-        className={clsx('text-ink-500 -mb-1 h-5 w-5', className)}
-      />
+      {children ? (
+        <span
+          className={clsx(
+            'cursor-help border-b border-dotted border-gray-600',
+            className
+          )}
+        >
+          {children}
+        </span>
+      ) : (
+        <InformationCircleIcon
+          className={clsx('text-ink-500 -mb-1 h-5 w-5', className)}
+        />
+      )}
     </Tooltip>
   )
 }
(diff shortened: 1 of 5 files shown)


web/components/widgets/info-tooltip.tsx
```diff
@@ -2,13 +2,28 @@ import { InformationCircleIcon } from '@heroicons/react/outline'
...
+  className?: string
+  children?: React.ReactNode
+}) {
+  const { text, className, children } = props
   return (
     <Tooltip className="inline-block" text={text}>
-      <InformationCircleIcon
-        className={clsx('text-ink-500 -mb-1 h-5 w-5', className)}
-      />
+      {children ? (
```
jahooma, line 13: Cool! Typically it is better to just have a different component rather than configure it with different modes based on options. Ok to ship this though.

web/pages/leagues.tsx
```diff
@@ -110,6 +123,79 @@ export default function Leagues(props: { rows: league_row[] }) {
             </Row>
           </Row>
 
+          <Row className="mb-4 items-center gap-3">
+            <text className="">
+              Compete for{' '}
+              <span
+                className="cursor-pointer border-b border-dotted border-blue-600 text-blue-600 hover:text-blue-800" // Add any styling you want for the clickable text
```
jahooma, line 130: The comment is unnecessary.

web/pages/leagues.tsx
```diff
@@ -191,16 +289,29 @@ const CohortTable = (props: {
                   isUser={currUserId === user.id}
                 />
               )}
-
-              {promotionCount > 0 && i + 1 === promotionCount && (
+              {doublePromotionCount > 0 && i + 1 === doublePromotionCount && (
```
jahooma, line 292: Shouldn't this line also have a tooltip explainer? I feel it's confusing right now.

jahooma (approved):
Nice work, David! Cool to see you & gpt coding.

Just a couple comments and you should feel free to merge after addressing them.

web/pages/leagues.tsx
```diff
@@ -158,14 +245,25 @@ const CohortTable = (props: {
...
   const users = useUsers(rows.map((row) => row.user_id))
   if (!users) return <LoadingIndicator />
 
   const division = rows[0].division
   const nextDivision = division + 1
+  const nextNextDivision = division + 2
   const nextDivisionName = DIVISION_NAMES[nextDivision] ?? SECRET_NEXT_DIVISION
+  const nextNextDivisionName =
+    DIVISION_NAMES[nextNextDivision] ?? SECRET_NEXT_DIVISION
+  DIVISION_NAMES[nextDivision] ?? SECRET_NEXT_DIVISION
```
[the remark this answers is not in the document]
SirSaltyy: why? we use these

PR #1732, opened 2023-05-02
Ingawei/add contract firebase rules

ingawei:


web/components/contract/contract-card.tsx
```diff
@@ -391,7 +391,7 @@ export function ContractCardNew(props: {
   const { className, promotedData } = props
   const user = useUser()
 
-  const contract = useContract(props.contract.id) ?? props.contract
+  const contract = useRealtimeContract(props.contract.id) ?? props.contract
```
jahooma, line 394: @IanPhilips If we start using more supabase subscription channels like this, would you expect that to work?

Seems like we are already getting errors with, e.g. trades tab not updating on a market after you place a bet. These are caused by too many channels, right?

web/hooks/use-contract-supabase.ts
```diff
@@ -26,6 +31,38 @@ export const usePublicContracts = (contractIds: string[] | undefined) => {
...
+    }
+    const length = changedContract.length
+    if (length > changesLength.current) {
+      setContract(changedContract[length - 1].new.data as Contract)
+      changesLength.current = length
+    }
+  }, [changedContract])
+
+  return contract
+}
```
mqp, line 64: This function looks surprisingly complicated. Why isn't it something like
```typescript
const { rows } = useSubscription('contracts', { k: 'id', v: contractId ?? '_' })
return rows != null && rows.length > 0 ? rows[0].data as Contract : undefined
```
It also is fishy that this function is getting called with `undefined` and implies that there's something wrong in the React component design that is calling this hook. You should consider factoring the design so that the components calling this hook don't render when there is no contract.

web/hooks/use-contract-supabase.ts
```diff
@@ -26,6 +31,38 @@ export const usePublicContracts = (contractIds: string[] | undefined) => {
...
+        .then((result) => setContract(result))
+        .catch((e) => console.log(e))
+    }
+  }, [contractId])
+  const changesLength = useRef(0)
+  const changedContract = useRealtimeChanges('UPDATE', 'contracts', {
+    k: 'id',
+    v: contractId ?? '',
+  })
+  console.log(changedContract)
```
mqp, line 50: This doesn't make sense to me, why would you want this function to maintain a list of changes to the contract over time? I don't think you understood the different realtime abstractions I wrote.

- `useRealtimeChannel` -- this is for when you want to get a callback whenever there is an insert/update/delete. It abstracts away the channel management stuff we were having a hard time with. It's useful if you want to do something quirky when you see the changes that isn't just "have a list of the changes" or "update the thing that changed."
- `useRealtimeChanges` -- this is a small wrapper around `useRealtimeChannel` that stores the list of changes in state.
- `useRealtimeRows` -- this is a convenience wrapper around `useRealtimeChanges` that only subscribes to inserts and returns the list of new rows. This tends to be useful for subscribing to ~append-only collections of things.
- `useSubscription` -- this is for when you want to subscribe to a subset of a table and have a live updating snapshot of what you subscribed to, like a Firestore subscription. It handles the "fetch existing and then apply updates to them as they come in" logic. This seems like the easiest thing for you to use here.

web/hooks/use-follows-supabase.ts
```diff
@@ -0,0 +1,59 @@
...
+  // For each element in deletedFollows, find its index in combinedFollows.
+  // If it exists (the index is not -1), remove it from combinedFollows.
+  deletedFollowIds.forEach((deletedFollowId) => {
+    let index = combinedFollows.indexOf(deletedFollowId ?? '')
+    if (index !== -1) {
+      combinedFollows.splice(index, 1)
+    }
+  })
+
+  return uniq(combinedFollows)
```
mqp, line 58: Unless I am missing something, I think this function could be replaced by
```typescript
const { rows } = useSubscription('contract_follows', { k: 'contract_id', v: contractId })
return rows?.map(r => r.follow_id)
```
As I wrote above, the fact that it accepts `undefined` is a code smell and indicates something is likely wrong with the React component design.

web/hooks/use-follows-supabase.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+  )
+  const insertedFollows: Insert<'contract_follows'>[] = useRealtimeChanges(
+    'INSERT',
+    'contract_follows',
+    {
+      k: 'contract_id',
+      v: contractId ?? '',
+    }
+  )
```
mqp, line 28: It seems strange to subscribe to these in two separate channels when you could have done `useRealtimeChanges('*', ...)` to pick up both inserts and deletes over one channel.

mqp:
Also, note that the realtime channel subscription tools at a lower level than `useSubscription` are unreliable, in the sense that they will not deliver all changes in cases where the channel gets disconnected (they have no way of getting the ones that came in while they were disconnected.) So don't use them in a context where it would be very bad for there to be outdated data. For example, it would be inappropriate to use them in a naive way to keep the contract page contract up-to-date, because if you left the tab and came back, it would just show an outdated contract forever until a new update came in.

ingawei:
@jahooma I've addressed Marshalls concerns, pls look again

web/hooks/use-contract-supabase.ts
```diff
@@ -26,6 +32,16 @@ export const usePublicContracts = (contractIds: string[] | undefined) => {
   return contracts
 }
 
+export function useRealtimeContract(contractId: string) {
```
jahooma, line 35: `contractId` type should include `undefined` or `null` if it could be that.

web/pages/challenges/[username]/[contractSlug]/[challengeSlug].tsx
```diff
@@ -68,8 +67,32 @@ export default function ChallengePage(props: {
...
+        user={user}
+        slug={slug}
+        bets={bets}
+        challenge={challenge}
+        challengeSlug={challengeSlug}
+      />
+    )
+}
+
+export function ChallengePageContent(props: {
```
Comment (jahooma) [inline web/pages/challenges/[username]/[contractSlug]/[challengeSlug].tsx:86]: Didn't realize this code still existed lol

jahooma (approved):
Sweet, this looks good Inga! Happy to get all these changes in. The march toward Supabase is inevitable.

jahooma (approved):
Sweet, this looks good Inga! Happy to get all these changes in. The march toward Supabase is inevitable.

PR #1733, opened 2023-05-02
Ancient markets page

NcyRocks:
Adds an ancient markets page at manifold.markets/ancient. Also adds links to most quests to make them easier to complete.

### DIFF
--- a/web/components/quests-or-streak.tsx
+++ b/web/components/quests-or-streak.tsx
@@ -26,6 +26,7 @@ import { getQuestScores } from 'common/supabase/set-scores'
 import { useQuestStatus } from 'web/hooks/use-quest-status'
 import { db } from 'web/lib/supabase/db'
 import { LoadingIndicator } from 'web/components/widgets/loading-indicator'
+import Link from 'next/link'
 
 const QUEST_STATS_CLICK_EVENT = 'click quest stats button'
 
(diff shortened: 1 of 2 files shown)


web/components/quests-or-streak.tsx
```diff
@@ -201,7 +207,7 @@ const QuestRow = (props: {
           </span>
           <Col>
             <span className={clsx('sm:text-xl')}>
-              {title}
+              {link ? <a href={url}>{title}</a> : title}
```
jahooma, line 210: We use NextJS's link components. You could do this like:

`<Link href={href}>{title}</Link>`

web/components/quests-or-streak.tsx
```diff
@@ -190,8 +194,10 @@ const QuestRow = (props: {
   status: string
   reward: number
   info?: string
+  link?: string
```
jahooma, line 197: Prefer the name `href` over `link` (it is not a full link)

jahooma (approved):
Awesome, seems pretty useful! Thanks Nathan!

web/components/quests-or-streak.tsx
NcyRocks: Makes sense, fixed

web/components/quests-or-streak.tsx
NcyRocks: Renamed, thanks!

PR #1738, opened 2023-05-04
Implement basic Supabase polling stuff

mqp:
Pretty self-explanatory, this provides `useSupabasePolling(q)` which re-runs the query given repeatedly and returns the results, and optionally takes deps that restart the polling (e.g. if the query is based on a user ID and they just logged in.)

I updated a couple places where we were using realtime for groups and just re-fetching everything every time to use this instead, as a proof of concept.


mqp:
@jahooma @ingawei thoughts?

jahooma:
Looks pretty simple. Seems like it could be useful in certain cases!

Not sure if I would call it real-time. A user of a real-time hook might be confused when it takes a while to update.

PR #1740, opened 2023-05-04
Ingawei/make private group invite links

ingawei:
@jahooma please review! This is the invite link PR I showed today. Nothing is on prod but everything should be deployed on dev

image

image


backend/supabase/group_invites/create.sql
```diff
@@ -0,0 +1,45 @@
+drop table group_invites;
+
+create table if not exists
+  group_invites (
+    id text not null primary key default random_alphanumeric (12),
```
jahooma, line 5: I think Sinclair did this for the id before:

```
create table if not exists
  market_ads (
    id text not null primary key default uuid_generate_v4 (),
 ```

File: backend/supabase/group_invites/create.sql [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+    group_id text not null,
+    foreign key (group_id) references groups (id),
```
jahooma, line 7: Nice!

backend/supabase/group_invites/create.sql
```diff
@@ -0,0 +1,45 @@
...
+    is_max_uses_reached boolean generated always as (
+      case
+        when max_uses is null then false
+        else uses >= max_uses
+      end
+    ) stored,
+    expire_time timestamptz
+  );
+
+  CREATE OR REPLACE FUNCTION set_expire_time()
```
jahooma, line 29: Doesn't auto-format change to lowercase? Would be nice to check in only consistent formats.

backend/supabase/group_invites/create.sql
```diff
@@ -0,0 +1,45 @@
...
+      or (duration is not null)
+    ),
+    uses numeric not null default 0,
+    max_uses numeric default null,
+    is_max_uses_reached boolean generated always as (
+      case
+        when max_uses is null then false
+        else uses >= max_uses
+      end
+    ) stored,
```
jahooma, line 25: Interesting.

backend/supabase/group_invites/create.sql
```diff
@@ -0,0 +1,45 @@
...
+    created_time timestamptz not null default now(),
+    duration interval default '1 week',
+    is_forever boolean not null default false,
+    check (
+      (
+        duration is null
+        and is_forever = true
+      )
+      or (duration is not null)
+    ),
```
jahooma, line 17: Could `is_forever` also be generated like `is_max_uses_reached`?

backend/supabase/group_invites/create.sql
jahooma: Seems like this could also be generated? Or is there a reason to prefer using a trigger?

backend/supabase/group_invites/functions.sql
```diff
@@ -0,0 +1,14 @@
+create
+or replace function get_last_week_long_link (this_group_id text) returns text as $$
+    SELECT id
+    FROM group_invites
+    WHERE (
+            group_id = this_group_id
+            -- just gets the last default link if it was generated less than 12 hours ago
+            and duration = '1 week'::interval
+            and max_uses is null
+            and created_time + duration > (now() + '6 days 12 hours'::interval)
```
jahooma, line 10: Seems like you could have just compared created_time without adding duration. Not a big deal though.

backend/supabase/group_invites/functions.sql
```diff
@@ -0,0 +1,14 @@
+create
+or replace function get_last_week_long_link (this_group_id text) returns text as $$
```
jahooma, line 2: This function could kind of be a view. You just filter the view by group id after. I might prefer that style for next time.

common/src/supabase/groups.ts
```diff
@@ -1,5 +1,15 @@
 import { SupabaseClient } from '@supabase/supabase-js'
 import { run } from './utils'
+import { Group } from 'common/group'
+
+export async function getGroup(groupId: string, db: SupabaseClient) {
```
jahooma, line 5: Prefer putting `db` as the first argument (I think that's the convention elsewhere)

web/pages/group-invite/[inviteid].tsx
```diff
@@ -0,0 +1,153 @@
...
+import clsx from 'clsx'
+import { useState } from 'react'
+import { useRouter } from 'next/router'
+
+export async function getStaticProps(props: { params: { inviteid: string } }) {
+  const { inviteid } = props.params
+  const adminDb = await initSupabaseAdmin()
+  const invite: GroupInvite = (await getInvite(inviteid, adminDb)) || null
+  const group = invite ? await getGroup(invite.group_id, adminDb) : null
+  return { props: { invite, groupName: group?.name, groupSlug: group?.slug } }
```
Comment (jahooma) [inline web/pages/group-invite/[inviteid].tsx:29]: Might want to add revalidate: 60s so that it more closely tracks is_max_uses_reached.

That would work except for the last guy, because it always sends even old data immediately, and revalidates for the next page load. So to be more correct, you could change to `getServerSideProps`

backend/supabase/group_invites/create.sql
```diff
@@ -0,0 +1,45 @@
+drop table group_invites;
```
jahooma, line 1: This would destroy all the current data in the table, correct? I don't think it makes sense to put that in the sql file

jahooma (approved):
Nice, looks great! Thanks Inga.

Had a few code comments before pushing.

backend/supabase/group_invites/create.sql
ingawei: Yeah but this ID is hella long, I think it would be ugly in a share link

backend/supabase/group_invites/create.sql
ingawei: yeah I was not able to make a generated column, because it adds something to the current time so it's "non-deterministic"

backend/supabase/group_invites/functions.sql
ingawei: I'm trying to avoid views where I don't think it's super useful because that's another thing we have to pay attention to managing permissions for. Right now this table is admin read only, and I'd have to keep track of that for any views generated off of it

PR #1743, opened 2023-05-09
Reduce ad reward to M3, add a fee of M2

sipec:
the fee is not stored on the ad, is just the same for every ad, so it'll apply to current ads


backend/shared/src/run-txn.ts
```diff
@@ -158,3 +159,22 @@ export async function runRedeemBoostTxn(
 
   return { status: 'success', txn }
 }
+
+export async function runRedeemBoostFeeTxn(
```
jahooma, line 163: Probably unnecessary helper

jahooma (approved):
Thanks Sinclair!

PR #1748, opened 2023-05-11
Add MiraBot to bots list.

Mira-public:
I will soon start automatically trading with this alt and would like the bot tag.


jahooma (approved):
Nice! Bring on the bots!

PR #1750, opened 2023-05-13
Write code for basic Supabase realtime functionality

mqp:
This introduces two hooks:

1. `useLiveStream` opens and manages a channel and returns a live-updated stream of inserts on the chosen table. Useful for live feed kind of stuff and maybe the contract page bets.
2. `useSubscription` opens and manages a channel and returns a live-updated snapshot of the chosen table subset by merging a fetched snapshot with changes that come in, like subscribing to a Firestore query. Useful for most of the random stuff we want to use subscriptions for.

Some relevant facts:

1. When you call `db.removeChannel`, the websocket will be disconnected iff it has no more channels. So if all active channels are configured (as in this PR) to remove themselves when the tab is hidden, that will suffice to disconnect the websocket and not count against our connection cap.
2. Unlike in #1700, we don't try to multiplex subscriptions over the same channel, because I don't believe there's any need for it -- it's not saving any important resource, shouldn't impact our billing, and I think the Supabase library wants us to use the multiple-channels paradigm.

This is still WIP -- I need to test more disconnection and error scenarios to get things just right -- but it should be roughly right.

### DIFF
--- a/common/src/leagues.ts
+++ b/common/src/leagues.ts
@@ -1,4 +1,4 @@
-import { PlainTablesAndViews } from './supabase/utils'
+import { Row } from './supabase/utils'
 
 export type season = typeof SEASONS[number]
 
@@ -38,7 +38,7 @@ export const getDemotionAndPromotionCount = (division: number) => {
   return { demotion: 5, promotion: 5, doublePromotion: 0 }
 }
 
-export type league_row = PlainTablesAndViews['leagues']
+export type league_row = Row<'leagues'>
 export type league_user_info = league_row & { rank: number }
 
 export const COHORT_SIZE = 25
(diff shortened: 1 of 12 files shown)


mqp:
CC @IanPhilips @jahooma. I found it easier to do this from scratch instead of branching from #1700.

jahooma (commented):
Looks pretty good!

Interesting to separate out listening to inserts vs. maintaining an updated list based on changes. Makes sense though.

I think the names of the hooks might be clearer. E.g. `useRealTimeInserts` and `useRealTime` instead of `useLiveStream` and `useSubscription`.

Would be happy to test these live to see what kinds of errors we run into.

jahooma:
Any updates on this? Inga still has a couple PR's blocked on this.

mqp:
OK, this seems like it's all working correctly to me.

PR #1751, opened 2023-05-13
Return a full Bet from /api/v0/market/marketId/sell, and not just status success

MayMeta:
Currently an API call `POST /v0/market/[marketId]/sell` returns a `{"status": "success"}`, while a call `POST /v0/bet` returns a full `Bet` object.

This Pull Request changes the return type of the `sell` handler to a also be a full `Bet` object, with an additional `"status": "success"` field for backwards compatibility.

It also updates the API documentation `/docs/docs/api.md` to reflect this change.

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -661,6 +661,8 @@ Parameters:
       depending on current unfilled limit bets and the AMM's liquidity. Any remaining
       portion of the bet not filled would remain to be matched against in the future.
   - An unfilled limit order bet can be cancelled using the cancel API.
+- `expiresAt`: Optional. A Unix timestamp (in milliseconds) at which the limit bet
+  should be automatically canceled.
 
 Example request:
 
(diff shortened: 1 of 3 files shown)


docs/docs/api.md
```diff
@@ -822,6 +826,9 @@ $ curl https://manifold.markets/api/v0/market/{marketId}/sell -X POST \
     --data-raw '{"outcome": "YES", "shares": 10}'
 ```
 
+Response type: A `Bet`, with an additional `"status": "success"` field, for backwards
```
akrolsmir, line 829: Hm, instead of mentioning "In previous versions it simply returned `{"status": "success"}`.", here, could you add the message to the changelog section? Thanks!

akrolsmir (approved):
This looks great, thanks for making this quick improvement!

docs/docs/api.md
MayMeta: Ok, no prob! I didn't spot that there was a changlelog section

PR #1752, opened 2023-05-14
If the user is trying to sell shares they don't own return an error 400 You don't have any shares to sell

MayMeta:
Currently if a user tries to sell shares they down own on a CPPM-1 market using an API, they will get the following result:
```
POST https://manifold.markets/api/v0/market/[marketId]/sell body='{"outcome": "NO", "shares": 1}'
--> 400 {"message": "You can only sell up to undefined shares."}
```

This Pull Request changes this error message to:
```
POST https://manifold.markets/api/v0/market/[marketId]/sell body='{"outcome": "NO", "shares": 1}'
--> 400 {"message": "You don't have any [outcome] shares to sell."}
```


jahooma (approved):
Sweet, thanks for the improvement!

PR #1753, opened 2023-05-15
Ingawei/add your groups to group page

ingawei:
@jahooma pls review

Separated groups page into two things, your groups and discover:
image
image

Also updated groups search to hopefully be more efficient and good (does the same thing as contract search, checks exact text and then does fuzzy search. The API calls are on dev if you'd like to try it out


web/lib/supabase/groups.ts
```diff
@@ -124,3 +203,37 @@ export async function getPublicGroups() {
...
+
+export async function listGroupsBySlug(groupSlugs: string[]) {
+  const { data } = await run(
+    db.from('groups').select('data').in('slug', groupSlugs)
+  )
+  if (data && data.length > 0) {
+    return data.map((group) => group.data as Group)
+  }
+  return []
+}
```
jahooma, line 239: I think @IanPhilips just added a few indexes for groups and related tables. Would be good to still check that these queries are covered! (Check seed.sql or see if your db tools shows what indexes were created, or run the query in your console with explain analyze and check it is < 40ms.)

web/pages/groups.tsx
```diff
@@ -89,29 +49,22 @@ export default function Groups(props: { groups: SearchGroupInfo[] }) {
...
+              className={'mb-4'}
+              tabs={[
+                {
+                  title: 'Your Groups',
+                  content: <YourGroups yourGroupIds={yourGroupIds} />,
+                },
+                {
+                  title: 'Discover',
+                  content: <DiscoverGroups yourGroupIds={yourGroupIds} />,
+                },
```
jahooma, line 63: For users that are not in any groups, what do you think about defaulting to the Discover groups tab?

jahooma (approved):
This is great, thanks Inga! Glad to see this overhaul of the groups page, including proper search!

Code looks good.

Will be excited to see more usage of groups. I think we should include them in the feed cards, (but maybe after we overhaul the feed to be more like a timeline, an idea we had today).

web/pages/groups.tsx
ingawei: ohh smart, will do

backend/api/src/supabase-search-groups.ts
```diff
@@ -0,0 +1,136 @@
...
+      query = `
+      select groupz.data
+      from (
+        select groups.*,
+            similarity(groups.name,$1) AS similarity_score
+        FROM groups 
+        join group_members 
+        on groups.id = group_members.group_id
+        where group_members.member_id = '${uid}'
+      ) AS groupz
```
IanPhilips, line 85: I sort of wish these `groupz` variables had better names, maybe like `similar_groups`?

backend/supabase/seed.sql
```diff
@@ -620,7 +620,9 @@ create table if not exists
     privacy_status text,
     slug text,
     name text,
+    name_fts tsvector generated always as (to_tsvector('english'::regconfig, name)) stored,
     creator_id text;
```
IanPhilips, line 624: should be a `,`, not `;`

backend/api/src/supabase-search-groups.ts
```diff
@@ -0,0 +1,136 @@
...
+      `
+    } else {
+      query = `
+      select groups.data
+        from groups 
+        join group_members 
+        on groups.id = group_members.group_id,
+        websearch_to_tsquery('english',  $1) as query
+      where group_members.member_id = '${uid}'
+      and groups.name_fts @@ query
```
IanPhilips, line 97: I don't see the `groups.name_fts` column in the prod db though it is in the seed.sql, you'll have to add it manually via 
`alter table groups
add column name_fts tsvector generated always as (to_tsvector(name)) stored;`

backend/supabase/groups/indexes.sql
```diff
@@ -0,0 +1,5 @@
+
+create index if not exists group_slug on public.groups (slug);
+create index if not exists group_name on public.groups (name);
+create index if not exists group_creator_id on public.groups (creator_id);
+create index if not exists total_members on public.groups (total_members desc);
```
IanPhilips, line 5: It looks like you're searching `name_fts` in non-fuzzy search, here's the query plan using the dev db:
Screenshot 2023-05-16 at 9 18 19 AM

backend/supabase/groups/indexes.sql
```diff
@@ -0,0 +1,5 @@
+
+create index if not exists group_slug on public.groups (slug);
+create index if not exists group_name on public.groups (name);
```
IanPhilips, line 3: Nice, looks like this index is working:
Screenshot 2023-05-16 at 9 21 10 AM

PR #1759, opened 2023-05-17
trying to add pretty border but the world hates me

ingawei:
image
adding borders
image
^ platinum vs silver, platinum is bluer and shinier
image
bronze  :(


backend/supabase/seed.sql
```diff
@@ -58,23 +59,6 @@ create policy "public read" on users for
 select
   using (true);
 
-create index if not exists users_data_gin on users using GIN (data);
```
jahooma, line 61: Aren't other indexes for tables here in the seed.sql file? Currently, I would expect these indexes here. We really need to decide how we're organizing our sql.

backend/supabase/posts/triggers.sql
```diff
@@ -29,6 +29,3 @@ or
 update on posts for each row
 execute function post_populate_cols ();
 
-update posts
-set
-  fs_updated_time = fs_updated_time;
```
jahooma, line 34: Wasn't there a purpose for this?

jahooma (approved):
Nice! Cool to show users' divisions this way.

The user avatars feel a little hard to read when they are at their smallest size now... maybe we should enlarge them? Alternately, the border could be thinner and darker?

Down to try this and iterate

backend/supabase/posts/triggers.sql
ingawei: yeah but Marshall was against having it here incase someone was just blindly running these

PR #1760, opened 2023-05-17
Delete dead old Numeric market code

sipec:
Prevents creating numeric markets via the API, and removes code for betting, resolving, payouts etc.


sipec:
hmm I still want to reconcile the weird array/union types for all contract types

jahooma:
Thanks for looking into deprecating these!

PR #1766, opened 2023-05-18
Check if username is a reserved path

MayMeta:
Currently anyone can create a new user with a username that is a reserved path, like `leagues`, `api`, `about`, `links`, etc. Existing users can also change their names to a reserved path. In both cases the personal page of the user becomes inaccessible, and the only way how the user can restore their access to Manifold is to write to discord, asking admins to change their name.

This Pull Requests restricts users ability to pick a reserved path as a username, by checking it against a list `RESERVED_PATHS` in `common/envs/constants.ts`.

Note that a some number of the paths that I propose to reserve are not currently in use, but they might be used by Manifold in the future. For example: `login`, `twitter`, and `ads`.


IanPhilips (approved):
This is great, thanks for writing this! We've been putting it off but this will save us time fixing users' little username adventures.

PR #1767, opened 2023-05-18
API POST endpoints should not accept extra parameters

MayMeta:
Currently any POST request to any of the auth endpoint can have extra parameters unspecified in the schema. This is prone to bugs where the caller expected that their parameter will have an effect on the result of the call, but it didn't.

Example: sending a POST request to the `/sell` endpoint and providing `amount` parameter by mistake (instead of `shares`), which results in all the available shares being sold, instead of the specified amount. The desired behavior in this scenario is to send a status code 400 with a body `{"message": 'Unknown parameter "amount"'}`.

This pull requests adds a `.strict()` call to all `zod` body schemas of all auth endpoints, that will reject any POST request that has extra parameters.

Note, that I'm not sure if is this change breaks any existing server logic, although it might help to reveal some bugs where the caller incorrectly assumed parameters in the endpoint schema.

I'm also unsure how the `.strict()` feature interacts with `z.union`, `z.intersection`, or `z.infer<typeof schema1> & (z.infer<typeof schema2> | {})`, therefore I didn't add any `.strict()` calls to the following files:
```
backend/api/src/create-chat-message.ts
backend/api/src/create-comment.ts
backend/api/src/create-market.ts
backend/api/src/create-post.ts
backend/api/src/resolve-market.ts
```
I request help from more experienced developers to assist or consult me on those cases.

### DIFF
--- a/backend/api/src/add-contract-to-group.ts
+++ b/backend/api/src/add-contract-to-group.ts
@@ -12,7 +12,7 @@ import { getUser } from 'shared/utils'
 const bodySchema = z.object({
   groupId: z.string(),
   contractId: z.string(),
-})
+}).strict()
 
 export const addcontracttogroup = authEndpoint(async (req, auth) => {
   const { groupId, contractId } = validate(bodySchema, req.body)
--- a/backend/api/src/add-group-member.ts
+++ b/backend/api/src/add-group-member.ts
@@ -13,7 +13,7 @@ const bodySchema = z.object({
   groupId: z.string(),
   userId: z.string(),
   role: z.string().optional(),
-})
+}).strict()
 
 export const addgroupmember = authEndpoint(async (req, auth) => {
   const { groupId, userId, role } = validate(bodySchema, req.body)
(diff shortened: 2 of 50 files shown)


IanPhilips:
This is cool! I support this

MayMeta:
I will wait for approval from another Core team member, before investigating `z.union`, `z.intersection`, and `z.infer<typeof schema1> & (z.infer<typeof schema2> | {})` further.

Do you have integration tests, or unit tests for those handlers? And if so, then where can I find them.

mqp:
> Do you have integration tests, or unit tests for those handlers? And if so, then where can I find them.

It's kind of embarrassing, but basically no. The way we typically test is by running the API locally and then testing against our local API, and potentially deploying to dev and testing against dev.

sipec:
this is a good idea that I think will make life easier for us as well. I'll try to get it merged ... sometime

sipec:
thanks @MayMeta! I kinda just did it myself in e94d2244be01fe3e09b1d84e4027831aca191255 because I didn't want to deal with the merge conflicts ... but this was definitely a good idea and I'm glad you made this PR!

PR #1768, opened 2023-05-18
Add MetaculusBot to botlist

bjubes:
To remove it from leagues


jahooma (approved):
Good idea! I still need to run a script to actually move it to the bots league.

PR #1769, opened 2023-05-18
Add more fields to LiteUser

MayMeta:
Currently API enpoints `/v0/me`, `/v0/user/[username]`, `/v0/user/by-id/[id]`, and `/v0/users` only return the following fields:
```
id, createdTime, name, username, url, avatarUrl?, bio?, website?, twitterHandle?, discordHandle?,
balance, totalDeposits, profitCached { daily, weekly, monthly, allTime }
```

This Pull requests adds 8 more fields to the returned `LiteUser` object:
```
 isBot, isCore, isChecked, isBannedFromPosting?, userDeleted?,
 followerCountCached, currentBettingStreak?, lastBetTime?
 ```
 Where `isBot`, `isCore`, `isChecked` are checked against the `BOT_USERNAMES`, `CORE_USERNAMES`, `CHECK_USERNAMES` constants from `common/envs/constants.ts` and the rest of the fields are taken from the `User` object.
 
 I'm open for discussion if `isCore` and `isChecked` is the best naming choice. Perhaps they're better be renamed to `isAdmin` and/or `isTrustworthy` in the `LiteUser` object, to avoid confusion.


IanPhilips:
Nice! This is hopefully helpful to the api users, can you change isCore to `isAdmin` and `isChecked` to `isTrustworthy` like you mention? I think that will be a better UX for the median api user.

IanPhilips (changes requested):
Nice! This is hopefully helpful to the api users, can you change `isCore` to `isAdmin` and `isChecked` to `isTrustworthy` like you mention? I think that will be a better UX for the median api user.

IanPhilips:
Also, has anyone asked for more fields on liteuser or is this just something you think would be useful?

MayMeta:
> Also, has anyone asked for more fields on liteuser or is this just something you think would be useful?

Primarily I've asked for this, and I also think it would be useful for others. 🙂

For example, some users create quite a number of markets regarding other user's streaks, like: [Will 5 or more people, with streaks of 150 or more, comment in this market?](https://manifold.markets/firstuserhere/will-5-or-more-people-with-streaks), and it's currently difficult if not impossible to verify a persons streak (the person can just photoshop their screenshot, and there's some `streakForgiveness` logic going on under the hood which would be difficult to reproduce on a local machine). A similar thing is happening with the followers count.

This PR should help in both of those use-cases.

Also `lastBetTime` should help those, who implement API scripts that rely on other users betting activity.

IanPhilips:
Okay this seems reasonable, but I also don't understand why we return only a lite user at the user/by-id/[id] endpoint. I'd figure we use lite-user when you could potentially get a lot of users, and a full user when you request one via id.

IanPhilips:
I just changed the api so that /user/by-id/[id] returns the full user object

MayMeta:
> I also don't understand why we return only a lite user at the user/by-id/[id] endpoint. I'd figure we use lite-user when you could potentially get a lot of users, and a full user when you request one via id.

I've checked, and `/api/v0/me` actually calls `getcurrentuser` which returns full `User`.

MayMeta:
> I just changed the api so that /user/by-id/[id] returns the full user object

Are you sure that it's ok to make `referredByUserId` public information?

IanPhilips:
yup, will only help people audit the referrals totals if anything.

PR #1772, opened 2023-05-18
Kill some dead Firestore-related code

mqp:
Since we are trying to migrate this kind of thing to Supabase, it's nice to keep it tidy so that we can actually see what is left that actually wants migrating.


web/pages/contract-search-firestore.tsx
mqp, line 1: It's hard for me to imagine why we would ever want this page again.

jahooma: Yeah, let's toast it

PR #1775, opened 2023-05-19
docs: add information about how to code market/client rate limits

kevinburke:


sipec:
ehh I'm not convinced this change to the docs is good, sorry

PR #1778, opened 2023-05-20
Update stats embed

wasabipesto:
Stats now pull data directly from the API so they're always up to date. I've got a couple basic charts in at the moment, hoping to add more soon.


jahooma (approved):
Wow, this is awesome!!

PR #1779, opened 2023-05-21
Read notifications from Supabase

mqp:
Also introduces some new features to the realtime abstractions to support this case, e.g. `usePersistentSubscription`.

Visible functionality should be unchanged.

### DIFF
--- a/web/pages/notifications.tsx
+++ b/web/pages/notifications.tsx
@@ -1,6 +1,6 @@
 import clsx from 'clsx'
 import { Notification, ReactionNotificationTypes } from 'common/notification'
-import { PrivateUser } from 'common/user'
+import { PrivateUser, User } from 'common/user'
 import { sortBy } from 'lodash'
 import { useRouter } from 'next/router'
 import React, { Fragment, ReactNode, useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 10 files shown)


web/pages/notifications.tsx
```diff
@@ -31,24 +31,30 @@ import {
...
   const [navigateToSection, setNavigateToSection] = useState<string>()
   const { isNative } = getNativePlatform()
-  useRedirectIfSignedOut()
+
   const { groupedNotifications, mostRecentNotification } =
-    useGroupedNonBalanceChangeNotifications(privateUser)
+    useGroupedNonBalanceChangeNotifications(privateUser.id)
   const balanceChangeGroupedNotifications =
-    useGroupedBalanceChangeNotifications(privateUser)
+    useGroupedBalanceChangeNotifications(privateUser.id)
```
mqp, line 57: @IanPhilips I see that you just changed all this code to handle some case where `privateUser` is not available, but I don't understand why you did any of that. Obviously it makes no sense to have a notifications page unless you are logged in. How can the private user not be available if the redirect didn't fire? Also, why was it doing a client-side redirect?

I changed it to a server-side redirect and removed all your "if there isn't a private user" logic. Let me know if I am missing something.

IanPhilips: I like the cleaner, top-level code that allows undefined as a private user on the `useNotifications` hook so that you can use it with `usePrivateUser()` without extra conditional code. I believe it was doing client side redirect because @jahooma and I don't like the delay that the use experiences when clicking the notifications page and nothing happening. Better to immediately navigate and while they wait, show a loading spinner.

mqp: I don't agree with you -- I have a highly confident general programming heuristic that says it's basically always less convenient to have a function that is shaped like
```typescript
f(x) {
  if (x == null) {
    return;
  } else {
    /* ... do work */
  }
}
```
vs.
```typescript
f(x) {
  /* ... do work */
}

// somewhere else
if (x != null) {
  f(x)
}
```
for a variety of reasons, but it's not very important. Since you have to live with this code I will change it back to be how you like.

mqp: Actually, I thought better of it. In this case rendering it initially with a fake user ID has a real cost, since it's really hitting the server and asking for nonsense (a subscription to a nonexistent set of notifications.) So I'm going to do it my way and if you want to go back and refactor the React tree so that you can do a client-side redirect without subscribing to nonsense, that should be relatively easy for you to do (you would just need to split out the "actual notification area" into a separate React component that is only rendered if you see a logged-in user.)

IanPhilips: Surely you're right as that is your competitive advantage as you've said yourself. However I suspect we will revert this to be client side in which case the react rule of hooks for conditional rendering will make me add another component if the 'useNotifications' hook doesn't accept a null private user, which I think is ugly.

IanPhilips: You follow your heart as you see fit and I'll see what I can live with after you merge. I don't have strong opinions here.

mqp: Just kidding! I changed it back to client-side with separate components because it seems like our `isAuthorized` stuff is kind of broken with SSR right now.

> in which case the react rule of hooks for conditional rendering will make me add another component if the 'useNotifications' hook doesn't accept a null private user, which I think is ugly.

I think this is typically anti-ugly -- my same aesthetic intuition above about pulling the conditional outward applies.

mqp:
I tested this a decent bit, it's pretty complicated so there might be some bugs but the easiest way to find them will be to put it live.

PR #1781, opened 2023-05-21
Clean up some stuff on the `user_seen_markets` table

mqp:
This is urgent because old `user_seen_markets` writes which are conflicting with the broken primary key stuff are currently wedging replication.

- Added autogenerated ID primary column -- the existing primary key didn't make sense for it to be unique.
- Dropped the GIN index. This table is relatively large and GIN indexes are beefy. We have no use for it.
- Dropped `fs_updated_time`.
- Turned off replication on `user_seen_markets` since we are not writing to Firestore anymore.

### DIFF
--- a/backend/scripts/supabase-import.ts
+++ b/backend/scripts/supabase-import.ts
@@ -213,14 +213,6 @@ async function importDatabase(
       timeChunk,
       startTime
     )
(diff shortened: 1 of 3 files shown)


mqp:
FYI @IanPhilips @jahooma

PR #1782, opened 2023-05-21
Add more reserved paths

MayMeta:
Long story short, I forgot to add "link" which is used for manalinks, and decided to add another 12 paths that might be used by Manifold in the future. If you're against adding some (or all) of the proposed 12 paths, I'm ok with leaving only the "link",

I've also thought about reserving "market", but some user already took that username: https://manifold.markets/market.

I will also use this opportunity to casually mention the existence of users with other weird usernames:
```
{"id": "u7XiGZiSPhMM6PJrY4b3XZCbABF2", "username": "", ...}
{"id": "xiFxCGuYlXUUJTbkkAvPsTwteGl2", "username": "home", ...}
{"id": "vlXewWdc1GhCfLn3GAIoN3dGDbE2", "username": "Daniel_Vogel@me.com", ...}
```


jahooma:
Thanks for the effort. I think PRs should only be submitted if they meet a certain bar of importance, otherwise even a slight attention cost is net-negative. Knowing the priority of changes is difficult for open source contributors, unfortunately.

PR #1787, opened 2023-05-23
Kill contract `uniqueBettorIds`

mqp:
Now it's gone and we will save lots of bandwidth sending these back and forth everywhere.


jahooma:
Hooray!

PR #1791, opened 2023-05-23
Don't subscribe to contracts in some places

mqp:
General philosophy: If the contract is not the focus of the page, and if users can't bet on it without clicking through, it's not really worth subscribing to changes on it. That's the case for related markets on the contract page, list mode search result contracts, and "today's changes" on the homepage.


jahooma:
Very mixed feelings on this.

For Today's updates, where it's front and center on the homepage, feels important to have the latest probability.
For the search list view, you can actually bet from that, so that also seems important to be correct.
Related markets is less important, but I don't see why we wouldn't want everything to be up-to-date.

I think we should see where the technology starts failing first, and only then make compromises on the product. So, I would punt on making a change like this.

mqp:
OK, I will leave it to you then.

PR #1792, opened 2023-05-24
Use dynamic imports for loot box animation

mqp:
This animation data is huge (several hundred kilobytes minified) and the animation library that runs it is huge (similar size) so we need to not load them with the main bundle.

This may cause a delay in showing the loot box on page load -- if we don't like that, then we should not ship the animation as a huge JSON blob that takes a big special-purpose library to render.


mqp:
FYI @ingawei @mantikoros

PR #1793, opened 2023-05-24
Improve efficiency of `updateContractMetrics`

mqp:
- Previously, this would get all the current probabilities for all contracts redundantly in each "historical snapshot" query for each time period, which was redundant.
- Now there is a covering index that has the probabilities in it, so the historical snapshot queries can do an index-only scan.

CC @IanPhilips since you were interested in this.

### DIFF
--- a/backend/functions/src/scheduled/update-contract-metrics.ts
+++ b/backend/functions/src/scheduled/update-contract-metrics.ts
@@ -34,17 +34,20 @@ export async function updateContractMetricsCore() {
   log(`Loaded ${contracts.length} contracts.`)
 
   const now = Date.now()
-  const yesterday = now - DAY_MS
+  const dayAgo = now - DAY_MS
   const weekAgo = now - 7 * DAY_MS
   const monthAgo = now - 30 * DAY_MS
 
+  log('Loading current contract probabilities...')
+  const contractProbs = await getCurrentProbs(pg)
+
   log('Loading historic contract probabilities...')
-  const [yesterdayProbs, weekAgoProbs, monthAgoProbs] = await Promise.all(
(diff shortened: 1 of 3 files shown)


backend/supabase/seed.sql
```diff
@@ -471,18 +471,17 @@ create index if not exists contract_bets_data_gin on contract_bets using GIN (da
...
-/* serving stats page, recent bets API */
-create index if not exists contract_bets_created_time_global on contract_bets (created_time desc);
-
 /* serving activity feed bets list */
 create index if not exists contract_bets_activity_feed on contract_bets (is_ante, is_redemption, created_time desc);
 
+/* serving update contract metrics */
+create index if not exists contract_bets_historical_probs
+  on contract_bets (created_time)
+  include (contract_id, prob_before, prob_after);
```
IanPhilips, line 480: How is this include thing different than just making a normal index like:
`contract_bets (created_time, contract_id, prob_before, prob_after)`?

backend/functions/src/scheduled/update-contract-metrics.ts
```diff
@@ -73,7 +80,7 @@ export async function updateContractMetricsCore() {
 
     if (hasChanges(contract, update)) {
       const contractDoc = firestore.collection('contracts').doc(contract.id)
-      writer.update(contractDoc, update)
+      //      writer.update(contractDoc, update)
```
IanPhilips, line 83: Intentionally commented out?

backend/supabase/seed.sql
IanPhilips: Oh I see you just posted about this in discord: https://discord.com/channels/915138780216823849/1108516565676200018/1111042390510272602

PR #1796, opened 2023-05-26
Migrate post reads to supabase

sipec:
still gotta test that the group about page works when edited (🙄 I should migrate the "about posts" into the group table)
also put back realtime in the right places


jahooma:
Cool! Looks good

PR #1805, opened 2023-06-01
notification thigns

ingawei:


web/hooks/use-group-supabase.ts
```diff
@@ -65,15 +79,17 @@ export function useRealtimeGroupContractIds(groupId: string) {
...
 
-  useEffect(() => {
-    if (contract.groupSlugs)
+  useEffectCheckEquality(() => {
+    if (contract && contract.groupSlugs)
       listGroupsBySlug(uniq(contract.groupSlugs)).then((groups) =>
         setGroups(filterDefined(groups))
       )
-  }, [contract.groupSlugs])
+  }, [contract])
```
jahooma, line 92: You want this to update whenever any property on contract changes? Seems unlikely. (You could change to `contract?.groupSlugs` if you want to update when contract becomes defined)

backend/api/src/create-market.ts
```diff
@@ -128,8 +131,18 @@ export async function createMarketHelper(body: schema, auth: AuthedUser) {
...
+  if (group && groupId) {
     await addGroupContract(groupId, group, contractRef, userId)
+    if (contract.visibility == 'private') {
+      const contractCreator = await getUser(contract.creatorId)
+      if (!contractCreator) throw new Error('Could not find contract creator')
+      await createNewContractInFromPrivateGroupNotification(
+        contractCreator,
+        contract,
+        group
+      )
```
jahooma, line 143: Sending emails and such is generally better to do in a trigger (i.e. onCreateContract), so it does not slow down the initial request to create the contract.

I would probably move this logic there (to `on-create-contract.ts`)

backend/shared/src/create-notification.ts
```diff
@@ -1105,6 +1110,68 @@ export const createNewContractNotification = async (
...
+        sourceSlug: contract.slug,
+        sourceTitle: contract.question,
+        sourceContractSlug: contract.slug,
+        sourceContractId: contract.id,
+        sourceContractTitle: contract.question,
+        sourceContractCreatorUsername: contract.creatorUsername,
+      }
+      await insertNotificationToSupabase(notification, pg)
+    }
+    // TODO: send private market email
```
jahooma, line 1154: TODO no longer relevant?

backend/shared/src/create-notification.ts
```diff
@@ -1105,6 +1110,68 @@ export const createNewContractNotification = async (
   }
 }
 
+export const createNewContractInFromPrivateGroupNotification = async (
```
jahooma, line 1113: Did you mean to include In and From in this name?

jahooma (approved):
Nice change, Inga! Should make private markets a much less lonely experience haha

backend/api/src/create-market.ts
ingawei: That's where I originally had it (this is what took me a long time to debug yesterday), but groups aren't populated when on-create-contract is triggered yet

PR #1811, opened 2023-06-06
Ingawei/feed tweaks

ingawei:
A couple of visual feed tweaks:
image
- avatar is larger, no more random reasons on top right side
- always shows when market was created under name
- brought back percentage bar thingy
- twitter like bottom bar with actions
- groups is overlaid on picture

image

- made yes and no bet panels singular

image

- explicitly state market for bets, likes and comments

image

- made boost button smaller and on top right

### DIFF
--- a/web/components/bet/bet-panel.tsx
+++ b/web/components/bet/bet-panel.tsx
@@ -57,6 +57,7 @@ export function BuyPanel(props: {
   hidden: boolean
   onBuySuccess?: () => void
   mobileView?: boolean
+  singularView?: 'YES' | 'NO'
   initialOutcome?: binaryOutcomes
   location?: string
   className?: string
@@ -67,6 +68,7 @@ export function BuyPanel(props: {
     hidden,
     onBuySuccess,
     mobileView,
+    singularView,
     initialOutcome,
     location = 'bet panel',
     className,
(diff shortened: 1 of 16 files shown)


ingawei:
@IanPhilips btw, feel free to leave feedback as well!

IanPhilips:
Thanks for working on this! I like the group overlay and twitter-like buttons on the bottom of the card. Does the user icon bring up the trades? Perhaps that should default to positions, as that's what I use more often. Ofc people may be confused about why there are fewer positions than traders but you could add a note to the modal about that.
A few notes:
- I thought we were going to axe the cover image unless uploaded by creator. However, we can't yet tell whether the creator uploaded the image or not, so the image should be axed by default.
- The yes/no buttons seem kind of small targets for mobile
- buttons should have the same style everywhere on our site, why revert the claim button to a pill shape?
- I think while the feed is in alpha and we're gathering feedback the reason should still show.  I can add that back in if it's too much trouble.
- Would love a timestamp of when the feed item was updated/created.

jahooma:
Looks very nice, thanks Inga! This is a big improvement, even to our current feed.

I like the stronger emphasis on probability with the bar, and and the bet panel changes — seems a lot simpler, and it opens the keyboard automatically on mobile to change the bet amount.

The payout and profit might be improved, they look kind of like another bar right now.
Screenshot 2023-06-06 at 3 50 43 PM


Otherwise, I do feel like this is more of a change for the current feed:
Many of the cards in the new feed are specialized to a specific purpose, whereas these are sort of "everything about a market".
For example, if the event is around a new comment on a market, we might want to make the comment more of a focus and have less info on the market.
Or if there's a change in probability, you might get that update because you already viewed the market. So you don't need as much info or context on the market, you just want to see the concise update, which was the change in probability.

I would take this component as a base, and then make a denser / more concise version for each of these cases.

As Ian said, I think showing the relevant timestamp is important, and possibly they should be in the same location across different card types. E.g. in the top right corner whether it's a new market or new comment.

PR #1813, opened 2023-06-06
Migrate post writes

sipec:
Successfully stores comments, post creation, post edits into supabase instead of firebase

- [x] `schema.ts` types are bOrk but not ready to run this migration on prod
- [x] I can't get realtime to work, requires refresh. Instead figure out our cache revalidate / optimistic update solution


web/hooks/use-comments.ts
```diff
@@ -20,12 +18,14 @@ export const useComments = (contractId: string) => {
...
-export const useCommentsOnPost = (postId: string | undefined) => {
-  const [comments, setComments] = useState<PostComment[] | undefined>()
 
-  useEffect(() => {
-    if (postId) return listenForCommentsOnPost(postId, setComments)
-  }, [postId])
-
-  return comments
+export const useNewCommentsOnPost = (postId: string) => {
+  return useRealtimeRows('post_comments', { k: 'post_id', v: postId }).map(
```
jahooma, line 23: useRealtimeRows misses updates when you are off the tab. Better to use `useSubscription` and pass in a fetch function to refetch data while you were away.

jahooma: Note that you need to fetch it the raw rows array. See use-comments-supabase.ts as an example (modified today):

```ts
export function useRealtimeComments(limit: number): Comment[] {
  const { rows } = useSubscription('contract_comments', undefined, () =>
    getCommentRows(limit)
  )
  return (rows ?? []).map((r) => r.data as Comment)
}
```

backend/shared/src/supabase/utils.ts
```diff
@@ -70,3 +70,15 @@ export async function bulkUpsert<
...
+// Replacement for firebase updateDoc. Updates just the data field (what firebase would've replicated to)
+export async function updateData<
+  T extends TableName,
+  K extends Record<string, any>
+>(db: SupabaseDirectClient, table: T, id: string, data: Partial<K>) {
+  await db.none(`
+  update ${table}
+  set data = data || '${JSON.stringify(data)}'
+  where id = '${id}'
+`)
```
jahooma, line 83: Hmm, is this actually doing a partial update? Seems like it would blow away the previous json keys if provided with only a subset to update.

backend/api/src/create-post.ts
```diff
@@ -78,89 +73,61 @@ export const createpost = authEndpoint(async (req, auth) => {
...
 
-export async function getPostFromSlug(slug: string) {
-  const firestore = admin.firestore()
-  const snap = await firestore
-    .collection('posts')
-    .where('slug', '==', slug)
-    .get()
+// TODO: migrate slug in new column with unique constraint
+export async function postExists(slug: string) {
+  const pg = createSupabaseDirectClient()
```
jahooma, line 126: Creating the pg connection here is an anti-pattern. It should usually be initialized at the top level of the api endpoint or script or whatever, and then passed in as a parameter.

jahooma (commented):
Nice! Cool to see this migrated. Should make it easier to migrate other stuff given this experience.

A couple comments. Realtime might work fine if you use `useSubscription`. I'll delete the other `useRealtimeRows` because I can't think of a case where we'd actually want that — usually it's just a bug.

backend/shared/src/supabase/utils.ts
sipec: the `data || '${JSON.stringify(newData)` part merges the old data and the new data

jahooma: cool

PR #1817, opened 2023-06-08
changing up groups page

ingawei:
Changing look of group page to prioritize private groups
image

Made join/leave buttons on mobile smol
image

also now have view group_role drawing from groups_rbac instead of groups for privacy


web/hooks/use-group-supabase.ts
```diff
@@ -277,3 +279,29 @@ export function useGroupsWhereUserHasRole(userId: string | undefined) {
...
+      getMemberPrivateGroups(userId).then((result) => {
+        setMemberPrivateGroups(result)
+      })
+    }
+  }, [userId])
+  return memberPrivateGroups
+}
+
+export function useYourNonPrivateNonModeratorGroups(userId: string) {
+  const [groups, setGroups] = useState<Group[] | undefined>(undefined)
```
jahooma, line 298: Should be able to use `usePersistentInMemoryState` instead of `useState` for better UX on returning to the page

jahooma (approved):
LGTM!

PR #1820, opened 2023-06-12
Migrate group members to supabase

sipec:
One issue with splitting up migration work by subcollection is that the firebase rules can't check for for membership (since membership data is no longer written to firebase) which means that until we also migrate the other parts of groups, private groups are less secure.
I suppose we could migrate `group_contracts` first instead. I originally didn't want to because the add/remove contract functions check your membership, and I wanted membership to be on supabase first to make migrating the rest of the function easier

EDIT: no longer think this is much of an issue


sipec:
another bug is that the denormalized members count doesn't decrement. it does increment though.
in practice this doesn't really matter, since it's primarily used for sorting rather than display. it can be fixed later.

PR #1822, opened 2023-06-13
Multi Binary contract!

jahooma:
https://www.notion.so/manifoldmarkets/Multi-set-of-binary-markets-8bd7ad1fde074e67b75bc1dd65f9a59a

https://www.notion.so/manifoldmarkets/MultiBinary-Product-Spec-fe7b741b84e049099d44cb826ae5f271


common/src/contract.ts
```diff
@@ -163,14 +165,34 @@ export type Uniswap2 = {
...
+export type CPMMMulti = {
+  mechanism: 'cpmm-multi-1'
+  outcomeType: 'MULTIPLE_CHOICE'
+  shouldAnswersSumToOne: boolean
+  totalLiquidity: number // for historical reasons, this the total subsidy amount added in Ṁ
+  subsidyPool: number // current value of subsidy pool in Ṁ
+
+  // Answers chosen on resolution, with the weights of each answer.
+  // Weights sum to 1 if shouldAnswersSumToOne is true. Otherwise, range from 0 to 1 for each answerId.
+  resolutions?: { [answerId: string]: number }
```
mantikoros, line 182: What about resolving N/A? If we want to handle conditional markets in the future, we'll need to be able to cancel some subset of the markets.

mantikoros (approved):
LGTM!

- I think the bet labels should probably be changed. Maybe "Buy" / "Short"
- contract.resolutions is not future proof for conditional markets
- answers are allowed to be exact string duplicates (and they shouldn't be!)
- Missing subsidize dialog.
- info stats table should show subsidy pool (that has not been added yet)
- Should hide old market types on /create: FR, numeric, etc.
- Duplicated multi does not copy over answers

jahooma:
Thanks for the review!

PR #1823, opened 2023-06-13
Ingawei/make minimal feed card

ingawei:
- made a smol market thing if things are attached to it
- consolidated some stuff between feed and feed-timeline
- made time consistently next to action, also made tiny
- rearranged some files
image
image
image
image


web/components/feed/feed-timeline-items.tsx
```diff
@@ -69,42 +65,45 @@ export const FeedTimelineItems = (props: {
   )
 
   return (
-    <Col>
+    <Col className="gap-3">
       {feedTimelineItems.map((item) => {
         // Boosted contract
-        if ('ad_id' in item) {
+        if (item.contract && ('ad_id' in item || 'contract' in item)) {
```
IanPhilips, line 71: Lol this is how I originally wrote it but I recently changed it bc I think it's slightly confusing to parse both ads and normal contracts in one place. I preferred to separate them as though this does introduce some duplication, it results in cleaner/easier to parse code IMO. It's fine as is, I don't feel strongly about it.

web/components/feed/feed-timeline-items.tsx
```diff
@@ -201,6 +146,25 @@ export const FeedTimelineItems = (props: {
   )
 }
 
+export function FeedRelatedItemFrame(props: {
+  children: React.ReactNode
+  href: string
+  className?: string
+}) {
+  const { children, href, className } = props
```
IanPhilips, line 154: This used to have a `useIsVisible` hook to market feed items as seen, which is/will be important for feed functionality. I can't find it moved anywhere else so this should be re-added here.

web/lib/util/shortenedFromNow.ts
```diff
@@ -0,0 +1,26 @@
+import dayjs from 'dayjs'
+import duration from 'dayjs/plugin/duration'
+import {} from 'dayjs/plugin/duration'
+
+dayjs.extend(duration)
+
+export function shortenedFromNow(timeInPast: number): string {
+  const diff = dayjs.duration(dayjs().diff(timeInPast))
+
+  const units: { [key: string]: number } = {
```
IanPhilips, line 10: lol I just made something like this for the profile! Yours is probably better though 😄

IanPhilips (commented):
This looks great, Inga! I'm super excited to start testing this on all our little guinea pigs. The new mostly image-less, minimal card designs will make the feed much more informative.

I left a few comments, thanks for working on this!

web/lib/util/shortenedFromNow.ts
ingawei: oh where is it maybe we can consolidate lol

web/components/feed/feed-timeline-items.tsx
ingawei: Yeah it was annoying to keep having to duplicate changes so I just consolidated them

web/components/feed/feed-timeline-items.tsx
ingawei: I think this a different thing? FeedItemFrame is still intact, I've just added FeedRelatedItemFrame
Screenshot 2023-06-14 at 3 49 04 PM

ingawei: @IanPhilips

IanPhilips: oh great, if you didn't change that then no worries!

PR #1831, opened 2023-06-16
adding to private group functionality

ingawei:
Allow adding to private group from create a market page
image
Changes once private group is selected
image
also transferred group search functionality over to search group function


backend/api/src/supabase-search-groups.ts
```diff
@@ -13,13 +13,20 @@ const bodySchema = z.object({
   limit: z.number().gt(0),
   fuzzy: z.boolean().optional(),
   yourGroups: z.boolean().optional(),
+  addingToContract: z.boolean().optional(),
+  newContract: z.boolean().optional(),
```
jahooma, line 17: In general, I think it's better to for the api to describe what it does, rather than current case you are using it for. Makes it easier to reuse, but also makes it easier to understand what is actually happening.

For example, instead of `addingToContract` it could be `hasModeratorOrAboveRole` and `newContract` could be `isNonPrivate`, etc.

Or maybe you could set filters to a list of visibilities: `['public', 'curated']`

jahooma (approved):
LGTM! Thanks for improving this case for people using private groups.

backend/api/src/supabase-search-groups.ts
ingawei: Yeah unfortunately hasModeratorOrAboveRole means something different if you're searching for groups to add to your contract vs just searching for groups in general.

PR #1834, opened 2023-06-20
Migrate group writes to supa

sipec:


backend/api/src/update-group.ts
```diff
@@ -0,0 +1,25 @@
...
+    )
+    .optional(),
+  aboutPostId: z.string().optional(),
+  bannerUrl: z.string().optional(),
+})
+
+export const updategroup = authEndpoint(async (req) => {
+  const { id, ...data } = validate(schema, req.body)
+  const db = createSupabaseDirectClient()
+  await updateData(db, 'posts', id, data)
```
jahooma, line 23: Update 'posts'? Shouldn't this update 'groups'?

jahooma: Also, don't you need to check if the authenticated user is a group creator / moderator to apply this change?

jahooma (approved):
Thanks Sinclair!! Code looks reasonable, except for the `updategroup` endpoint.

You still might want to test as I didn't look that closely.

Cool to reach another migration milestone!

PR #1835, opened 2023-06-20
added limit order new UI'

ingawei:
limit order is just a slider, no inputs, a lot less room for user error. Could use some help on the wording
image


web/components/bet/limit-slider.tsx
```diff
@@ -0,0 +1,143 @@
...
+    setLowLimitProb,
+    highLimitProb,
+    setHighLimitProb,
+    mobileView,
+  } = props
+
+  const mark = (value: number) => (
+    <span className="text-ink-400 text-xs">
+      <div className={'sm:h-0.5'} />
+      {value}%
```
jahooma, line 32: Have you checked if this works with the Pseudo-numeric type? (The kind Austin used to use for his Taco Tuesday markets).

Instead of percents, you would show numbers chosen within the lower and upper bound.

I'd be ok shipping with percents too as long as it still places limit orders correctly for pseudo numeric.

web/components/bet/limit-slider.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  const lowLimitPosition = sliderWidth * ((lowLimitProb ?? 0) / 100)
+  const highLimitPosition = sliderWidth * ((highLimitProb ?? 0) / 100)
+  const LIMIT_LABEL_MIN_DISTANCE = 130
+  return (
+    <Col className="mb-8">
+      <Row className="mt-1 mb-4 gap-4">
+        <Col className="w-full">
+          <div className="text-ink-800 text-sm">
+            Attempt to keep between probabilities:
```
jahooma, line 53: Not bad, though let me see what I can think of.

"Bet when the probability crosses a lower or upper bound"
"Bet if the probability leaves the range"
"Bet at a lower or upper bound"

web/components/bet/limit-slider.tsx
```diff
@@ -0,0 +1,143 @@
...
+          marks={{
+            '0': mark(0),
+            '50': mark(50),
+            '100': mark(100),
+          }}
+          value={[lowLimitProb, highLimitProb]}
+          onChange={(value) => {
+            if (value && Array.isArray(value) && value.length > 1) {
+              setLowLimitProb(value[0])
+              setHighLimitProb(value[1])
```
jahooma, line 122: Do you handle not allowing low and high to be the same value?

jahooma (approved):
Whoa, this is super cool!! Excited to see how users like this simpler UI.

PR #1841, opened 2023-06-22
things

ingawei:
added inputs back into limit order
image

prevents you from making the higher limit lower than the lower limit
image


web/components/widgets/number-input.tsx
```diff
@@ -58,3 +56,52 @@ export function NumberInput(props: {
     </Col>
   )
 }
+
+export function ControllableNumberInput(props: {
```
jahooma, line 60: Nice!

jahooma (approved):
Looks good! Thanks for iterating based on user feedback

PR #1847, opened 2023-06-23
changed to questions everywhere

ingawei:


common/src/user-notification-preferences.ts
```diff
@@ -7,44 +7,44 @@ import { PrivateUser } from './user'
 export type notification_destination_types = 'email' | 'browser' | 'mobile'
 export type notification_preference = keyof notification_preferences
 export type notification_preferences = {
-  // Watched Markets
-  all_comments_on_watched_markets: notification_destination_types[]
-  all_answers_on_watched_markets: notification_destination_types[]
+  // Watched Questions
+  all_comments_on_watched_questions: notification_destination_types[]
```
jahooma, line 11: Changing properties that are stored in the database would break things unless we also ran a script to change them. Probably not worth the effort.

Notification preferences fit this scenario, I think.

I wonder which other changes in the PR are like this?

backend/api/src/create-question-ad.ts
```diff
@@ -76,8 +76,8 @@ export const boostmarket = authEndpoint(async (req, auth) => {
     log('error adding txn! reversing funds to ad.')
 
     await pg.none(
-      `update market_ads
-      set funds = market_ads.funds - $1
+      `update question_ads
```
jahooma, line 79: Probably not planning to change this table name, right?

web/pages/api/v0/question/[id]/index.ts
```diff
@@ -17,6 +17,6 @@ export default async function handler(
     res.status(404).json({ error: 'Contract not found' })
     return
   }
-  res.setHeader('Cache-Control', marketCacheStrategy)
-  return res.status(200).json(toFullMarket(contract))
+  res.setHeader('Cache-Control', questionCacheStrategy)
+  return res.status(200).json(toFullQuestion(contract))
```
Comment (jahooma) [inline web/pages/api/v0/question/[id]/index.ts:21]: I guess this would break the api for everyone using `/market`. I'm not sure that's worth doing...
Can think about it

PR #1849, opened 2023-06-23
Add groupSlugs to FullMarket

ahalekelly:
I couldn't find any way to get the groups of a market in the API


IanPhilips:
Thanks for the PR!

PR #1858, opened 2023-06-28
Ingawei/add new contract type

ingawei:
Added a very simple type, lmk if I missed something or could implement it better, because this was like so much less than 500 files lol. I played with having a more basic contract type called NonBet to support things like these with no real betting (and hopefully in the future posts) but that seemed way more complicated.

### DIFF
--- a/common/src/contract.ts
+++ b/common/src/contract.ts
@@ -44,6 +44,7 @@ type AnyOutcomeType =
   | Cert
   | QuadraticFunding
   | Stonk
+  | BountiedQuestion
 
 type AnyContractType =
   | (CPMM & Binary)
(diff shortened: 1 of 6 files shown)


common/src/contract.ts
```diff
@@ -252,6 +259,13 @@ export type Stonk = {
   initialProbability: number
 }
 
+export type BountiedQuestion = {
+  outcomeType: 'BOUNTIED_QUESTION'
+  totalBounty: number
+  bountyPaid: number
+  bountyTxns: string[]
```
jahooma, line 266: Are these the id's of the txns? Should leave a comment at minimum

jahooma (approved):
Nice! Looks good!

It's cool that it seems you won't have to create any new tables for this feature. Saves trouble setting up the replicator.

common/src/contract.ts
ingawei: oh yeah that was the idea, will do

PR #1863, opened 2023-06-30
Migrate contract metrics reads

sipec:


web/hooks/use-contract-metrics.ts
```diff
@@ -1,107 +1,40 @@
...
+    const val = rows
+      .map((row) => row.data as ContractMetric)
+      .filter((metrics) => {
+        return metrics.totalShares[outcome] > 0.1
+      })
+      .sort((a, b) => b.totalShares[outcome] - a.totalShares[outcome])
 
-  return contractMetrics
-}
+    return [outcome, val]
```
jahooma, line 36: Casting this `as const` probably helps the types.

web/hooks/use-saved-contract-metrics.ts
```diff
@@ -1,33 +1,28 @@
...
-          ...savedMetrics,
-          ...contractMetrics,
-        } as ContractMetric)
-      : undefined
-
-  useEffectCheckEquality(() => {
-    if (metrics) setSavedMetrics(metrics)
-  }, [metrics, setSavedMetrics])
+  useEffect(() => {
+    getUserContractMetrics(user?.id ?? '_', contract.id, db).then((metrics) => {
```
jahooma, line 21: I think this is not live updating? Don't we need that for the case of making a bet from the market page and seeing your position update in the BetSummary?

web/components/contract/user-positions-table.tsx
```diff
@@ -79,28 +76,24 @@ export const BinaryUserPositionsTable = memo(
...
+        getContractMetricsOutcomeCount(contractId, 'yes', db),
+        getContractMetricsOutcomeCount(contractId, 'no', db),
+      ]).then(([yesCount, noCount]) => {
         setShareholderStats({
           yesShareholders: yesCount,
           noShareholders: noCount,
         })
-      )
+        setTotalPositions?.(yesCount + noCount)
+      })
```
jahooma, line 96: If the `useRealtimeContractMetrics` actually includes all user positions, then it might include these counts, so we could delete this.

web/hooks/use-saved-contract-metrics.ts
jahooma: We might be able to pass the current user's position down from the ContractPage if that has loaded all users' positions. And maybe other parts of the app don't need this to live update...

jahooma (approved):
Nice! Definitely simplifies things, as you can tell by -260 lines vs +76 lines of code.

A couple comments.

web/hooks/use-saved-contract-metrics.ts
sipec: Well, on the contract page, this is only used for the leaderboard, which only shows if it's resolved, meaning it doesn't have to live update and one fetch should suffice

PR #1883, opened 2023-07-12
Migrate contract metrics writes

sipec:
@IanPhilips fyi


backend/supabase/seed.sql
```diff
@@ -110,7 +110,7 @@ create table if not exists
     user_id text not null,
     contract_id text not null,
     data jsonb not null,
-    fs_updated_time timestamp not null,
+    fs_updated_time timestamp,
```
jahooma, line 113: Why change this to nullable?

jahooma (approved):
This is great! Will save us some firebase bill, and stop some replicator problems, perhaps.

backend/api/src/change-user-info.ts
```diff
@@ -129,26 +128,6 @@ export const changeUser = async (
   }
   log(`Updated ${betRows.length} bets.`)
 
-  log('Updating denormalized user data on contract metrics docs...')
-  const contractMetricsRows = await pg.manyOrNone(
```
IanPhilips, line 133: afaict this is still necessary, why remove it?

backend/shared/src/helpers/user-contract-metrics.ts
```diff
@@ -1,26 +1,38 @@
...
-    writer.update(
-      firestore.collection(`users/${userId}/contract-metrics`).doc(contract.id),
-      metrics
-    )
-  })
+  await bulkUpdateContractMetrics(metrics)
+}
 
-  await writer.flush()
+export async function bulkUpdateContractMetrics(metrics: ContractMetric[]) {
```
IanPhilips, line 23: Screenshot 2023-07-12 at 8 53 18 AM
It looks like the table has many more columns than you're updating here, why leave them out?

IanPhilips (changes requested):
Yay! Like James said, this is a huge proportion of our fb writes and the replicator tried to write the same 16 contract metrics like 600 times yesterday, with an average of 150s (not ms) per write.

backend/supabase/seed.sql
sipec: as it is not copied in via the replicator, it will not have an fs_updated_time (and actually the column can be dropped)

backend/api/src/change-user-info.ts
sipec: because it's annoying to re-write to do the same thing, it should just be a join, and the cache moving out of date doesn't impact much. as long as we do the user join / user display cache stuff quickly enough it after this, it should be ok

backend/shared/src/helpers/user-contract-metrics.ts
sipec: I'm leaving the on insert trigger that copies to those fields from `data`. which is how the data got there from the  firebase replicator.

This is the minimal change necessary for migrating at all. Moving read/writes to native columns can be easily done later

PR #1886, opened 2023-07-12
Ingawei/man 1143 sort by market type

ingawei:
image
Updated contract search to have market type sort, and also look of contract filter/sort

@jahooma pls review

also @mantikoros please take a look at predictionMarketSorts in `supabase-search.tsx` to see if there are some sorts there that would actually work for bountied questions/and or if theres something easy that can be done to make it work.

TODO: add sort by bounty option

### DIFF
--- a/backend/supabase/contracts/indexes.sql
+++ b/backend/supabase/contracts/indexes.sql
@@ -31,3 +31,11 @@ create index if not exists contracts_resolution_time on contracts (resolution_ti
 create index if not exists contracts_visibility_public on contracts (id)
 where
   visibility = 'public';
+
+create index contracts_outcome_type_binary on contracts (outcome_type)
+where
+  outcome_type = 'BINARY';
+
+create index contracts_outcome_type_not_binary on contracts (outcome_type)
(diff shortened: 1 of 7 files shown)


backend/supabase/contracts/indexes.sql
```diff
@@ -31,3 +31,11 @@ create index if not exists contracts_resolution_time on contracts (resolution_ti
...
 where
   visibility = 'public';
+
+create index contracts_outcome_type_binary on contracts (outcome_type)
+where
+  outcome_type = 'BINARY';
+
+create index contracts_outcome_type_not_binary on contracts (outcome_type)
+where
+  outcome_type <> 'BINARY';
```
jahooma, line 41: Does this index work for the other outcome_types?

jahooma (approved):
Looks good! Thanks Inga.

backend/supabase/contracts/indexes.sql
ingawei: I made another index below it that does that

PR #1896, opened 2023-07-14
Ingawei/man 1146 create page placeholder for question input should change

ingawei:
image
image
* made 2 step process for creating a contract of different types
* made example defaults for each type


ingawei:
@jahooma pls review

web/hooks/use-new-contract.ts
```diff
@@ -0,0 +1,285 @@
...
+import { extensions } from 'common/util/parse'
+import { useTextEditor } from 'web/components/widgets/editor'
+import { safeLocalStorage } from 'web/lib/util/local'
+import { createMarket } from 'web/lib/firebase/api'
+import { removeUndefinedProps } from 'common/util/object'
+import { track } from 'web/lib/service/analytics'
+
+const descriptionPlaceholder = 'Optional. Provide background info and details.'
+
+export const useNewContract = (
```
jahooma, line 27: I'm not sure about moving this to a separate file because it is so relevant to the first file.

Frankly, I would never have even split it into its own hook (stephen did that). I would put it directly into the component body like we do for most complex components.

If there are modular pieces that can be abstracted out, those could be hooks. But updating key state for a component seems like it should just be part of the component. And not at the bottom of the file lol, but in the component.

jahooma (approved):
Ooh, nice! Looks cool.

Code seems good too!

I'm not 100% sure about the 2-step process, but I'm down to try it. An alternative might be to show the type selection at the top of the page (you can scroll to see the rest of the form), but that when you choose one, it collapses.

Biggest UI gripe is that I think it's not super clear that the market type options are clickable at first. Maybe add a border by default? Or radio button or something.

Navigating back doesn't take you back to step one. I think you'd manage this with url params, e.g. "?step=1". This could be a follow-up though, I don't think I'd block on it. We probably want to get feedback on the overall idea.

web/hooks/use-new-contract.ts
ingawei: oh great yeah I didn't understand why that was a think I think i'm just gonna put it back in the function then lol

PR #1902, opened 2023-07-17
[API] Make markets fetchable by a list of ids

Will-Howard:
This is for use in https://github.com/Will-Howard/manifolio , which currently makes many independent requests to the `/market` endpoint to achieve the same thing.

Things for reviewers to focus on, because I don't have the context necessary:
 - Do the docs render correctly?
 - Does there need to be some kind of CORS handling now that the `/markets` endpoint accepts POST requests?


web/lib/supabase/contracts.ts
```diff
@@ -111,8 +111,15 @@ export const getPublicContracts = async (options: {
   limit: number
   beforeTime?: number
   order?: 'asc' | 'desc'
+  ids?: string[]
```
jahooma, line 114: I think this should be a separate function. The other params would generally not be used when getting all by id.

web/pages/api/v0/markets.ts
```diff
@@ -18,6 +18,7 @@ const queryParams = z
       .or(z.string().regex(/\d+/).transform(Number))
       .refine((n) => n >= 0 && n <= 1000, 'Limit must be between 0 and 1000'),
     before: z.string().optional(),
+    ids: z.array(z.string()).optional(),
```
jahooma, line 21: Would add a max length of 1000 for ids.

docs/docs/api.md
```diff
@@ -101,9 +101,10 @@ Gets a group's markets by its unique ID.
 Requires no authorization.
 Note: group is singular in the URL.
 
-### `GET /v0/markets`
+### `GET /v0/markets or POST /v0/markets`
```
jahooma, line 104: I prefer not allowing post

jahooma:
While looking up a list of markets serves api users, it can't be cached as easily, so it could get expensive.

I prefer we wait until we implement usage limits or mana-cost per request before we add such a feature.

Because of that, I'm closing this PR for now.

PR #1908, opened 2023-07-18
adding poll creation option

ingawei:
- added the ability to create a poll contract in dev
image
- for a poll, default close time (when you can stop voting and a majority is automatically chosen) is never (this never option is only for polls and is actually just 1000 years in the future)
image
- pretty short PR but I'd like to get your feedback on implementation so far and also a second set of eyes because I'm feeling a the kind of sick where you're losing your mind so this code might be gross


backend/api/src/create-market.ts
```diff
@@ -441,7 +442,8 @@ async function generateAntes(
     outcomeType === 'BINARY' ||
     outcomeType === 'PSEUDO_NUMERIC' ||
     outcomeType === 'STONK' ||
-    outcomeType === 'MULTIPLE_CHOICE'
+    outcomeType === 'MULTIPLE_CHOICE' ||
+    outcomeType === 'POLL'
```
jahooma, line 446: Generating a liquidity doc for polls? We can skip all this for polls.

web/components/new-contract/contract-params-form.tsx
```diff
@@ -154,6 +155,19 @@ export function ContractParamsForm(props: {
...
     : undefined
 
+  const closeDateMap: { [key: string]: number | string } = {
+    'A day': 1,
+    'A week': 7,
+    '30 days': 30,
+    'This year': daysLeftInTheYear,
+  }
+
+  const NEVER_IN_DAYS = 30 * 12 * 1000 // ~1000 years
```
jahooma, line 165: I think Twitter polls always close (seemingly too soon — I feel like I often still want to vote in them).

I'm ok trying out this Never option. We'll see if users want it.

web/components/new-contract/contract-params-form.tsx
```diff
@@ -154,6 +155,19 @@ export function ContractParamsForm(props: {
...
+  if (outcomeType == 'POLL') {
+    closeDateMap['Never'] = NEVER_IN_DAYS
+  }
+  const [neverCloses, setNeverCloses] = useState(false)
```
jahooma, line 169: Does this need to be a separate useState? Isn't it just:

`const neverCloses = closeDateInDays === NEVER_IN_DAYS`

jahooma (approved):
Looks pretty good, Inga! I think the data schema is right. I guess there's one more table to be created for recording user votes in the poll.

web/components/new-contract/contract-params-form.tsx
ingawei: closeDateInDays doesn't exist, it just adds that amount to now

PR #1913, opened 2023-07-21
Ingawei/poll market page

ingawei:
added basic backend functionality:
pre-vote:
image

post-vote:
image

also changed up multi choice bar on mobile
before:
image

now:
image

probably could be adjusted more but the version before looks almost broken to me

TODO:
- notifications
- cards
- clean up poll page more
- handle closing. Right now I'm thinking it's just a UI change that highlights the most voted answer but that's just because idk how to automatically schedule a backend function to select a resolution upon close, lmk if you have thoughts on that


backend/api/src/cast-poll-vote.ts
```diff
@@ -0,0 +1,59 @@
...
+const schema = z.object({
+  contractId: z.string(),
+  voteId: z.string(),
+})
+
+export const castpollvote = authEndpoint(async (req, auth) => {
+  const { contractId, voteId } = validate(schema, req.body)
+  const pg = createSupabaseDirectClient()
+
+  const voters = await pg.manyOrNone(
```
jahooma, line 18: We could use a pg promise transaction:

(example)
```
db.tx(async t => {
    await t.none('UPDATE users SET active = $1 WHERE id = $2', [true, 123]);
    await t.one('INSERT INTO audit(entity, id) VALUES($1, $2) RETURNING id', ['users', 123]);
})
``` 

If we both 1. read the voters and 2. insert a voter, within a transaction, that could solve some double-vote race conditions.

https://github.com/vitaly-t/pg-promise#transactions

backend/api/src/cast-poll-vote.ts
```diff
@@ -0,0 +1,59 @@
...
+  // Find the option to update
+  let optionToUpdate = options.find((o) => o.id === voteId)
+
+  // Update the votes field
+  if (optionToUpdate) {
+    optionToUpdate.votes = voters.length + 1
+  }
+
+  // Write the updated options back to the document
+  contractRef.update({ options: options })
```
jahooma, line 46: await this

jahooma: I also think it would be ok to put this within or outside the pg.tx().

The important thing is to conceptualize the source of truth for vote count as the supabase table of voters, and this is merely a cache that should try to be mostly accurate.

web/components/poll/poll-panel.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { firebaseLogin } from 'web/lib/firebase/users'
+import { getHasVoted } from 'web/lib/supabase/polls'
+import { AnswerBar } from '../answers/answer-item'
+import { Button } from '../buttons/button'
+import { Col } from '../layout/col'
+
+export function PollPanel(props: { contract: PollContract }) {
+  const { contract } = props
+  const { options, closeTime } = contract
+  const totalVotes = options.reduce((acc, option) => acc + option.votes, 0)
```
jahooma, line 15: Could be: sumBy(options, option => option.votes)

jahooma (approved):
Sweet! Looks good!

A few minor comments.

Vote back end, vote UI, and multi choice changes seem good.

For close time, you could create a scheduled function that runs every minute that just queries to see if any polls are past their close time. Then it could set the resolution, correct? And isResolved: true, I guess.

jahooma:
Another idea that could be cool:
We have a positions tab on markets. The equivalent here would be to show every user that voted, grouped by which option they chose. It's kind of a spicy thing to show, but has the same reasoning as showing people's bets in that it is interesting content.
I think this could be a followup after shipping polls though.

jahooma:
Notifications
- for the creator when a vote option is chosen
- for the user when the vote is resolved? I think that would be cool

ingawei:
> Another idea that could be cool: We have a positions tab on markets. The equivalent here would be to show every user that voted, grouped by which option they chose. It's kind of a spicy thing to show, but has the same reasoning as showing people's bets in that it is interesting content. I think this could be a followup after shipping polls though.

Yeah I was thinking of doing this but instead having it as a popup when you click on the number of votes

PR #1917, opened 2023-07-21
things

ingawei:
smol PR but mainly want to get a second look on if changing the way tabs are done in groups will break anything

### DIFF
--- a/common/src/group.ts
+++ b/common/src/group.ts
@@ -45,17 +45,8 @@ export type GroupLink = {
   userId?: string
 }
 
-export function groupPath(
-  groupSlug: string,
-  subpath?:
-    | 'edit'
-    | 'markets'
-    | 'about'
-    | typeof GROUP_CHAT_SLUG
-    | 'leaderboards'
-    | 'posts'
-) {
-  return `/group/${groupSlug}${subpath ? `/${subpath}` : ''}`
+export function groupPath(groupSlug: string) {
+  return `/group/${groupSlug}`
 }
 
 export const GroupsByTopic = {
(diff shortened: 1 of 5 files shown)


common/src/group.ts
```diff
@@ -45,17 +45,8 @@ export type GroupLink = {
...
-    | 'markets'
-    | 'about'
-    | typeof GROUP_CHAT_SLUG
-    | 'leaderboards'
-    | 'posts'
-) {
-  return `/group/${groupSlug}${subpath ? `/${subpath}` : ''}`
+export function groupPath(groupSlug: string) {
+  return `/group/${groupSlug}`
 }
```
jahooma, line 50: Don't we still support linking to subpages this way? Seems like we should keep this.

jahooma: I guess that right now this would only be for leaderboards. It's ok to break this, but what if we have other tabs in the future for groups? I'd only do this if it was interfering with the other url params, which I don't think that it would be?

web/pages/group/[...slugs]/index.tsx
```diff
@@ -150,9 +154,14 @@ export function GroupPageContent(props: { groupParams?: GroupParams }) {
   const { groupParams } = props
   const router = useRouter()
   const { slugs } = router.query as { slugs: string[] }
-  const page = slugs?.[1] as typeof groupSubpages[number]
+  // const page = slugs?.[1] as typeof groupSubpages[number]
+
+  const [tab, setTab] = usePersistentState('questions', {
```
Comment (jahooma) [inline web/pages/group/[...slugs]/index.tsx:159]: Not sure why we can't just include the second slug as the start state, so you can begin on a tab.

jahooma (approved):
Looks pretty good!

common/src/group.ts
jahooma: Also, kind of minor, but I don't like the pattern of using tab=blah instead of adding a path /blah. From the user's perspective, the latter is way cleaner and easier to parse.
I think this is partly on NextJS for making this hard.

PR #1921, opened 2023-07-25
Add name column to user

sipec:
- also reorganizes user sql stuff from seed to supabase/users

### migration
```sql
alter table users
add column name text;

create
or replace function users_populate_cols () returns trigger language plpgsql as $$ begin
    if new.data is not null then 
      new.name := (new.data)->>'name';
      new.username := (new.data)->>'username';
    end if;
    return new;
end $$;

update users
set fs_updated_time = fs_updated_time;

alter table users
alter column name set not null,
alter column username set not null;
create index if not exists users_name_idx on users (name);

create index if not exists users_username_idx on users (username);

drop index users_name_gin;

drop index users_username_gin;

drop index user_profit_cached_all_time_idx;

drop index users_instance_id_email_idx;

drop index users_instance_id_idx;

-- drop index users_data_gin;
```

- then recreate views `group_role` and `user_groups`
- then recreate functions `search_users` and `engaged_users`

### DIFF
--- a/backend/supabase/users/create.sql
+++ b/backend/supabase/users/create.sql
@@ -1,0 +1,12 @@
+create table if not exists
+  users (
+    id text not null primary key,
+    data jsonb not null,
+    fs_updated_time timestamp not null,
+    name text not null,
+    username text not null,
+    name_username_vector tsvector generated always as (to_tsvector(name || ' ' || username)) stored
+  );
+
+alter table users
+cluster on users_pkey;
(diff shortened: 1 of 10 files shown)


backend/supabase/users/create.sql
```diff
@@ -0,0 +1,12 @@
+create table if not exists
```
jahooma, line 1: I'm not sure it's better to have a separate file for each of the table creation, indexes, rls, and triggers.

Seems better to put those all in the same file. I guess the advantage here is you can see if, say, rls exists by looking at the filenames? Not sure that outweighs colocality.

jahooma (approved):
Looks good, thanks Sinclair!

Another step in the migration, saving us from some gin indexes

PR #1923, opened 2023-07-25
Migrate follow reads

sipec:
following button state is now manually updated when the follow/unfollow async call finishes, instead of listening.

the follower panel does use realtime though


jahooma (approved):
Looks good. Thanks Sinclair!

PR #1924, opened 2023-07-25
Migrate follow writes

sipec:
- [x] merge the reads
- [x] test migration process on dev

```sql
alter table user_follows
add column created_time timestamptz;

update user_follows
set created_time = millis_to_ts((data->>'timestamp')::bigint);

alter table user_follows
alter column created_time set not null,
alter column created_time set default now();

-- then go check the table before continuing

alter table user_follows
drop column fs_updated_time,
drop column data;
```
then
```bash
firebase use prod
firebase deploy --only functions:logSubcollections,backupDb
```


jahooma (approved):
Seems reasonable. Good luck with this migration!

sipec:
ah right, I gotta rewrite this as api function, since firebase uid is no more and client code isn't how we write to supabase anymore

PR #1925, opened 2023-07-25
adding onepercentbot to botlist

firstuserhere:
adding onepercentbot to list of bots. Testing phase over, now deploying


sipec:
I think you don't have prettier configured right

sipec:
ok I just did it here instead: f84fdb24691f75419804b63ebe0d68b38dfdc38c

PR #1933, opened 2023-07-27
Changed bet-summary profit tooltip to clarify that profit on question is sumed over all bets

spagovir:


jahooma:
Seems reasonable.

PR #1934, opened 2023-07-28
adding resolution

ingawei:
autoresolve polls and send notifications
image
image


ingawei:
@jahooma would appreciate a sanity check to make sure I don't break anything with these schedules functions

jahooma (approved):
Looks good to me!

PR #1947, opened 2023-08-03
made navigating functinoal

ingawei:
URL params now work in group and user contracts page


jahooma (approved):
Thanks Inga!! I know this one is tricky. Looks good.

sipec:
@ingawei @jahooma the purpose of the previous code is to set the url params WITHOUT triggering the nextjs router. this is obvious if you read the comment on line 69, that links to https://github.com/vercel/next.js/discussions/18072

The problem with doing it this way is that the search pages will now trigger extra searches (because it the router store sets the data whenever the Next Router changes). This also means if you try to set state of two different things in the url params it fails. That is why @IanPhilips was driven to create the JSON in url params monstrocity, so that they'd update at the same time.

Anyways, curious what the original thing being solved by this change was

ingawei:
> @ingawei @jahooma the purpose of the previous code is to set the url params WITHOUT triggering the nextjs router. this is obvious if you read the comment on line 69, that links to [vercel/next.js#18072](https://github.com/vercel/next.js/discussions/18072)
> 
> The problem with doing it this way is that the search pages will now trigger extra searches (because it the router store sets the data whenever the Next Router changes). This also means if you try to set state of two different things in the url params it fails. That is why @IanPhilips was driven to create the JSON in url params monstrosity, so that they'd update at the same time.
> 
> Anyways, curious what the original thing being solved by this change was

The original thing being fixed was just adding url param store to pages that weren't the search page (user contracts, group contracts, etc.). I picked this way cause it just seemed cleaner.

jahooma:
Without this change, navigating back was broken. It didn't save scroll position or even the specific sorts on on user / group page. That's a giant, giant problem from the UX perspective because navigating back to click into more markets is like *the* main flow.

The old code looks like a hack that can be fixed other ways. We should be able to keep something in memory that prevents duplicate searches.

sipec:
hmm tested it on the user page and yes it does seem like the old way is super broken (though not exactly in the way you describe)

anyways portaling over to #internal-dev for continued conversation on search

PR #1965, opened 2023-08-10
Fix the path to the list of secret keys in the functions' documentation

maksugr:
Hi folks,
The path to secret keys in functions' documentation is incorrect (the file is absent).
I'm trying to fix that.


IanPhilips:
Good eye, thanks!

PR #1968, opened 2023-08-11
Group breadcrumbs

IanPhilips:
Screenshot 2023-08-11 at 12 06 25 PM

Can edit:
Screenshot 2023-08-11 at 12 06 10 PM


sipec:
these are soo pretty aaaa

sipec:
oh, but it should be  [ + Category ] rather than just a lone pencil icon if there's no groups. or something else easy to parse
image

PR #1983, opened 2023-08-21
Fix mana-printing bug

ProgramCrafter:
Fix for https://manifold.markets/derikk?tab=managrams, where managrams with multiple destinations can leave the user with negative mana balance:
изображение


IanPhilips:
Thanks for the quick fix!

PR #1984, opened 2023-08-23
Update constants.ts

Distortinator:
added a bot


jahooma:
Could you update the manifold account's description to say it is a bot?

Also, there are merge conflicts after I merged the last PR adding a bot.

Distortinator:
Updated the description and the conflict should be resolved.

PR #1989, opened 2023-08-24
Android mobile improvements

IanPhilips:
- Opens in-app android browser for links with target ='_blank'
- Changes internal facing links to target = '_self'. This includes a hack for iframe embeds of markets on native.
- Adds workarounds for severed native<->client communication post android OOM: https://github.com/react-native-webview/react-native-webview/issues/2680
- Removes Sentry, the remote mobile logging monitor as it was causing android build problems 😢


native/App.tsx
```diff
@@ -94,19 +92,17 @@ const App = () => {
...
-      setTimeout(() => {
-        communicateWithWebview('nativeFbUser', fbUser)
-      }, 100)
-      // My older android phone needs a bit longer
-      setTimeout(() => {
-        communicateWithWebview('nativeFbUser', fbUser)
-      }, 250)
-    }
-  }, [listeningToNative, fbUser])
+  const sendWebviewAuthInfo = (user: FirebaseUser) => {
```
IanPhilips, line 95: Now that this is called directly, if the ` auth.onAuthStateChanged` is actually used for anything, we won't get updates from the useEffect. If people are left signed out we'll probably want to add this back in.

native/App.tsx
```diff
@@ -122,15 +118,21 @@ const App = () => {
   const [theme, setTheme] = useState<'dark' | 'light'>('light')
 
   const setEndpointWithNativeQuery = (endpoint?: string) => {
-    const newUrl = baseUri + (endpoint ?? 'home') + nativeQuery
+    const newUrl =
+      baseUri +
+      (endpoint ?? 'home') +
+      nativeQuery +
+      `&rand=${Math.random().toString()}`
```
IanPhilips, line 125: Adding a random string ensures the `<WebView/>` `src` parameter will treat a new url from our `setUrlToLoad` as a new state even if it was previously set to the same url.

PR #1995, opened 2023-08-26
Revalidate props on stats page after DB update

NGTmeaty:
This should allow for more timely updates to the stats page after they are written to Postgres.


jahooma (approved):
Thanks, that's actually a good idea!

PR #1996, opened 2023-08-29
Ingawei/adjust dashboard look

ingawei:


web/components/feed/card-reason.tsx
```diff
@@ -20,7 +21,34 @@ export function CardReason(props: {
   )
   const positiveChange = probChange && probChange > 0
```
ingawei, line 23: added timestamp to cards that don't have an explicit reason

jahooma (approved):
Thanks Inga! Code looks good.

The single column items seem better, probably. The side panel is helpful. I like the probability change in the last day!

LGTM. I think as soon as we get users building these, they'll have their own ideas and requests for what they want their dashboard to look like.

PR #2001, opened 2023-08-30
Ingawei/add graph to card

ingawei:
Basically if change today is more than 5%, card shows graph of today

image
image


jahooma (approved):
Looks good! Thanks Inga. The graphs make the home feed a lot better IMO

PR #2002, opened 2023-08-31
expand description if have images/embed

ingawei:
Just an idea I think we could try out, basically if a description has images/embed (visually stimulating content), we add the description to the feed card and make it really big so that it can read more like a post:

examples:
image
image


jahooma (approved):
Great idea! Looks good.

PR #2025, opened 2023-09-07
Ingawei/review no show

ingawei:
Changes:

Rating is now in contracts tab to be more reflection of user's contract and not user
image

Reviews visual tweaks (now goes in order of most recent, de-emphasizes market and emphasizes review/reviewer):
image

notification button:
image

notification modal:
image

notification after rating:
image

market after rating:
image

also made it so stars change upon hover cause it's cute


mantikoros (commented):
Moving the rating and reviews into the user's questions section makes a lot of sense, but I still think showing each user's rating as a numerical score is bad UX versus a qualitative label.  1. Users will be confused when they realize that their score isn't the average of their ratings (esp when they've received all 5 star reviews but their score is lower). 2. What really matters is just highlighting if a market creator is bad or unreliable; we don't really want people fussing about improving their score slightly.

ingawei:
> Moving the rating and reviews into the user's questions section makes a lot of sense, but I still think showing each user's rating as a numerical score is bad UX versus a qualitative label. 1. Users will be confused when they realize that their score isn't the average of their ratings (esp when they've received all 5 star reviews but their score is lower). 2. What really matters is just highlighting if a market creator is bad or unreliable; we don't really want people fussing about improving their score slightly.

@mantikoros Oh why are we not just averaging? I just think it's confusing to use a qualitative label when we review resolutions quantitatively. If we really want to keep it not qualitative, we should rate on a qualitative scale (at least in the front end)

web/components/notifications/notification-types.tsx
```diff
@@ -664,7 +673,28 @@ export function MarketResolvedNotification(props: {
...
+            >
+              <Row className="gap-1">
+                <StarIcon className="h-4 w-4" />
+                Rate {notification.sourceUserName}'s resolution
+              </Row>
+            </Button>
+          )}
+          {!showReviewButton && (
+            <Row className="text-ink-500 items-center gap-0.5 text-sm italic">
+              You rated this resolution {justNowReview ?? userReview?.rating}{' '}
```
sipec, line 694: nit: maybe just "You rated this {..." instead?

web/components/reviews/review.tsx
```diff
@@ -22,28 +24,33 @@ export const Review = (props: {
 
   if (!user || !contract) return null
 
+  const isEmpty = text ? JSONEmpty(text) : true
```
sipec, line 27:
```suggestion
  const isEmpty = !text || JSONEmpty(text)
```

sipec (approved):
lgtm

mantikoros:
> > Moving the rating and reviews into the user's questions section makes a lot of sense, but I still think showing each user's rating as a numerical score is bad UX versus a qualitative label. 1. Users will be confused when they realize that their score isn't the average of their ratings (esp when they've received all 5 star reviews but their score is lower). 2. What really matters is just highlighting if a market creator is bad or unreliable; we don't really want people fussing about improving their score slightly.
> 
> @mantikoros Oh why are we not just averaging? I just think it's confusing to use a qualitative label when we review resolutions quantitatively. If we really want to keep it not qualitative, we should rate on a qualitative scale (at least in the front end)

Simple average doesn't work: https://www.evanmiller.org/how-not-to-sort-by-average-rating.html

I think a thumb up or down would also be a reasonable alternative UI. But I also disagree with your point: 5 star score + qualitative summary is probably fine.

ingawei:
> 

Would you mind if I tried pushing this and seeing user feedback? I think our current solution is too big brain and provides no transparency. I think averaging works if people also have the context of how many reviews there are, which I've added.

PR #2028, opened 2023-09-08
Re-implement follow user notification in api call

sipec:
### migration
```sql
drop index users_follower_count_cached;

create or replace view user_groups as (...
```
```sh
./deploy.sh dev
```
```sh
firebase deploy --only functions:onCreateBet,functions:onFollowUser,functions:onUnfollowUser
```


backend/functions/src/triggers/on-create-bet.ts
```diff
@@ -153,7 +153,7 @@ const addBetToFollowersFeeds = async (
   contract: Contract,
   bet: Bet
 ) => {
-  if (bettor.followerCountCached <= 0 || contract.mechanism === 'dpm-2') return
```
IanPhilips, line 156: This was handy for exiting early, would be nice to have this in so we don't waste time querying peoples' bets who don't have followers.

PR #2041, opened 2023-09-16
Add confetti to FAQ

Bohaska:
It's fun!


sipec:
easter eggs must remain buried u_u

PR #2049, opened 2023-09-21
Fix topic links in new questions interface

NGTmeaty:
Fixes share link in the new questions interface.


akrolsmir:
Oh wow, thanks for the quick fix!

PR #2059, opened 2023-10-06
Add `after` filter to `GET /managrams`

jkerkhoff:
Most of the time I will only be interested in new managrams, so would be nice to be able to query that directly.

(Sorry for commit spam. Don't see an option to squash the commits in the github interface.)


IanPhilips (commented):
Looks good!

PR #2060, opened 2023-10-08
Add new bot user

NGTmeaty:
Adding my bot user. I also have seen a few others without the bot tag, but not sure on the exact procedure there.

https://manifold.markets/ithaca
https://manifold.markets/BottieMcBotface


akrolsmir:
Thanks for adding this!

PR #2063, opened 2023-10-10
Add bots: ithaca, GigaGaussian, BottieMcBotface

NiciusB:


akrolsmir:
Awesome, thanks for doing this @NiciusB!

PR #2065, opened 2023-10-11
Fix remove group dialog

Bohaska:
Successfully removed group to non-predictive group -> Succesfully removed group from non-predictive group


IanPhilips:
Shoot, it looks like Salty change this already, sorry we took too long!

IanPhilips:
Well, he changed it to new dialogs at least

PR #2067, opened 2023-10-12
Allow bot accounts to use managrams immediately

jkerkhoff:
Since getting the bot tag requires admin intervention I think the waiting period is not necessary here. If someone (me) wants to build something cool using managrams, they should not be blocked by this!

Also removed the Manifest ticket exception, which doesn't do anything anymore.


akrolsmir:
LGTM, thanks for fighting anti-bot discrimination!

PR #2069, opened 2023-10-13
Update constants.ts

Mira-public:
I would prefer not to have this anymore.


akrolsmir:
Thanks for being trustworthy for the past year or so! Mind saying why you'd like to decline this now?

PR #2074, opened 2023-10-17
Change non-predictive toggle

SirSaltyy:
It's not working on my local host (the toggles get stuck and aren't reflecting the correct state). I think it's because the properties "isRanked" and "isSubsidised" isn't being written to firebase for some reason, and the default undefined state is "true".

I tried fixing it for 1 hour and I'm pretty sure it should work and it just doesn't....

Is it because it only works when live because the writes aren't actually happening in Firebase or something since I'm creating a new boolean property for the contract?


backend/shared/src/update-group-contracts-internal.ts
```diff
@@ -55,12 +58,21 @@ export async function addGroupToContract(
...
-      nonPredictive: true,
+      isRanked: false,
     })
     if (recordEdit) {
-      await recordContractEdit(contract, recordEdit.userId, ['nonPredictive'])
+      await recordContractEdit(contract, recordEdit.userId, ['isRanked'])
+    }
+  }
+
+  if (group.id === UNSUBSIDIZED_GROUP_ID && !contract.isSubsidised) {
```
SirSaltyy, line 70: Or is it the fact I am just putting both these IF statements for unsubsidized and unranked in the same function and need to create 2 functions?

SirSaltyy: hmm that doesn't seem to be it either tho

backend/shared/src/update-group-contracts-internal.ts
```diff
@@ -103,11 +115,17 @@ export async function removeGroupFromContract(
...
+    })
+    await recordContractEdit(contract, userId, ['isRanked'])
+  }
+  if (group.id === UNSUBSIDIZED_GROUP_ID && contract.isSubsidised) {
     await firestore.collection('contracts').doc(contract.id).update({
-      nonPredictive: false,
+      isSubsidised: true,
     })
-    await recordContractEdit(contract, userId, ['nonPredictive'])
+    await recordContractEdit(contract, userId, ['isRanked'])
```
IanPhilips, line 128: this should be ['isSubsidised'], correct?

backend/shared/src/update-group-contracts-internal.ts
```diff
@@ -103,11 +115,17 @@ export async function removeGroupFromContract(
...
+  if (group.id === UNRANKED_GROUP_ID && contract.isRanked) {
+    await firestore.collection('contracts').doc(contract.id).update({
+      isRanked: true,
+    })
+    await recordContractEdit(contract, userId, ['isRanked'])
+  }
+  if (group.id === UNSUBSIDIZED_GROUP_ID && contract.isSubsidised) {
     await firestore.collection('contracts').doc(contract.id).update({
-      nonPredictive: false,
+      isSubsidised: true,
```
IanPhilips, line 126: should be false?

backend/shared/src/update-group-contracts-internal.ts
```diff
@@ -103,11 +115,17 @@ export async function removeGroupFromContract(
       groupLinks: newLinks,
     })
 
-  if (group.id === NON_PREDICTIVE_GROUP_ID && contract.nonPredictive) {
+  if (group.id === UNRANKED_GROUP_ID && contract.isRanked) {
+    await firestore.collection('contracts').doc(contract.id).update({
+      isRanked: true,
```
IanPhilips, line 120: false here as well?

backend/shared/src/update-group-contracts-internal.ts
```diff
@@ -55,12 +58,21 @@ export async function addGroupToContract(
       })
   }
 
-  if (group.id === NON_PREDICTIVE_GROUP_ID && !contract.nonPredictive) {
+  if (group.id === UNRANKED_GROUP_ID && !contract.isRanked) {
     await firestore.collection('contracts').doc(contract.id).update({
-      nonPredictive: true,
+      isRanked: false,
```
IanPhilips, line 63: should be true

backend/shared/src/update-group-contracts-internal.ts
```diff
@@ -55,12 +58,21 @@ export async function addGroupToContract(
...
     })
     if (recordEdit) {
-      await recordContractEdit(contract, recordEdit.userId, ['nonPredictive'])
+      await recordContractEdit(contract, recordEdit.userId, ['isRanked'])
+    }
+  }
+
+  if (group.id === UNSUBSIDIZED_GROUP_ID && !contract.isSubsidised) {
+    await firestore.collection('contracts').doc(contract.id).update({
+      isSubsidised: false,
```
IanPhilips, line 72: should be true

common/src/contract.ts
```diff
@@ -111,7 +111,8 @@ export type Contract<T extends AnyContractType = AnyContractType> = {
   isTwitchContract?: boolean
 
   coverImageUrl?: string
-  nonPredictive?: boolean // If true, don't include profits in leagues, etc.
+  isRanked?: boolean
+  isSubsidised?: boolean
```
IanPhilips, line 115: should be spelled `isSubsidized` - and everywhere else it's misspelled

backend/shared/src/update-group-contracts-internal.ts
IanPhilips: I think you just have your cases backwards.

IanPhilips (dismissed):
I don't understand the difference between unsubsidized and unranked but I assume you know what you're doing!

common/src/envs/constants.ts
```diff
@@ -246,7 +246,7 @@ export const GROUP_SLUGS_TO_IGNORE_IN_MARKETS_EMAIL = [
   'manifold-features-25bad7c7792e',
   'bugs',
   'manifold-leagues',
-  'nonpredictive',
```
IanPhilips, line 249: This is wrong, you're still using the nonpredictive slug here: https://manifold.markets/browse?topic=nonpredictive&t=&q=

common/src/envs/constants.ts
```diff
@@ -322,7 +322,7 @@ export const GROUP_SLUGS_TO_HIDE_FROM_WELCOME_FLOW = [
   '-sircryptomind-crypto-stock',
   'selfresolving',
   'fun',
-  'nonpredictive',
```
IanPhilips, line 325: https://manifold.markets/browse?topic=nonpredictive&t=&q=

common/src/envs/constants.ts
IanPhilips: If you wanted to switch the group slugs you'd have to switch it on all the denormalized data on contracts, which would not be trivial.

web/components/leagues/mana-earned-breakdown.tsx
```diff
@@ -61,9 +61,7 @@ export const ManaEarnedBreakdown = (props: {
   const contractIds = loadingBets
     ? uniq(loadingBets.map((b) => b.contractId))
     : undefined
-  const contracts = usePublicContracts(contractIds)?.filter(
-    (c) => !c.nonPredictive
-  )
+  const contracts = usePublicContracts(contractIds)?.filter((c) => !c.isRanked)
```
IanPhilips, line 64: this should be `c.isRanked`, not the other way around

common/src/envs/constants.ts
SirSaltyy: Yeah u right! Forgot I had made this change. Was planning to change the slug but then realised was easier not to and forgot to change this back. Good catch

PR #2083, opened 2023-10-20
Replicate `privateUsers` -> `private_users` table

mqp:
Self-explanatory. Uses RLS for reads. I plan to get reads migrated over and then probably instead of using RLS for writes, I will make API endpoints for writing to `private_users` from the client.


jahooma:
Yay! First PR in a long while.

PR #2091, opened 2023-10-23
love: Format height

derikk:
Convert into feet and inches for display


jahooma (approved):
Looks good, thanks! I was wanting this change.

PR #2102, opened 2023-10-25
love: filter potential lovers by those looking for matches

derikk:


jahooma (approved):
Thanks! Was on our list to fix haha

PR #2109, opened 2023-10-26
location search preliminary

ingawei:
Location search yayy!

image
image


love/components/search-location.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const dropdownRef = useRef<HTMLUListElement>(null)
+
+  useEffect(() => {
+    if (query.length < 2) {
+      setCities([])
+    }
+  }, [query])
+
+  useEffect(() => {
+    const fetchData = async () => {
```
jahooma, line 48: Could keep a `useRef` with the query count, and only set state if it's the latest query count.

love/components/search-location.tsx
```diff
@@ -0,0 +1,154 @@
...
+  return {
+    city: lover.city,
+    regionCode: lover.region_code,
+    country: lover.country,
+    latitude: lover.latitude,
+    longitude: lover.longitude,
+  } as City
+}
+
+export default function CitySearchBox(props: {
```
jahooma, line 30: We usually don't export default unless it's a page.

jahooma (approved):
Looks good! Thanks Inga!

PR #2112, opened 2023-10-26
Charity page now loads txns from Supabase, in static props

mqp:
Pretty self-explanatory, now the page doesn't subscribe to the transaction list, it just loads them up front.


jahooma:
If someone donates, would it update with their donation?

mqp:
> If someone donates, would it update with their donation?

It doesn't subscribe to other people's donations, if that's what you mean. It only updates if you make a donation on the page, or if you refresh.

PR #2131, opened 2023-10-29
Make filters more sensible for multiple values in "relationship style"

ProgramCrafter:
> In the search filter, the "relationship style" options are all unchecked by default and if you check them, they're ANDed together - only users that have checked all of those options will be displayed. That is counterintuitive. If I check "monogamous" and "polyamorous", I'm expecting users who checked either, not the much smaller number of users that checked both. They should be ORed together, and all checked by default.
> 
> chrisjbillington, https://discord.com/channels/915138780216823849/1166835928518631454/1168154982676181073

### DIFF
--- a/love/components/filters.tsx
+++ b/love/components/filters.tsx
@@ -240,7 +240,7 @@ export const Filters = (props: {
                 />
               </Col>
               <Col className={clsx(rowClassName)}>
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Nice! Thanks for the change. Seems like an improvment

PR #2133, opened 2023-10-30
Add index to notifications table to serve loan payout code

mqp:
This index is for looking up who already got loan payouts.


mqp:
@jahooma it looks to me by my testing like this will make this part of the update loans code fast for now.

PR #2135, opened 2023-10-30
Trying to rearrange ad queries

ProgramCrafter:
May fix boost clicks showing incorrectly.


web/components/contract/boost-button.tsx
```diff
@@ -351,25 +351,34 @@ function FeedAnalytics(props: { contractId: string }) {
         .eq('contract_id', contractId)
   )
 
+  if (adQuery.error || viewQuery.error) {
```
jahooma, line 354: You can't early return if there's a react hook being used below: https://stackoverflow.com/a/56976469

I think just consolidate the error case like before and we should be good.

PR #2139, opened 2023-10-31
Ingawei/get near you places

ingawei:
Yayy can now search by proximity
image

can set radius in 50 mile increments
image


love/components/filters.tsx
```diff
@@ -31,12 +36,33 @@ const initialFilters = {
...
     Partial<rowFor<'lovers'> & User>
   >(initialFilters, 'profile-filters')
 
+  const [nearbyOriginLocation, setNearbyOriginLocation] = useState<
+    string | null | undefined
+  >(undefined)
+
+  const [radius, setRadius] = useState<number>(100)
+  const [debouncedRadius, setDebouncedRadius] = useState(radius)
+  const debouncedSetRadius = debounce(setDebouncedRadius, 200)
```
jahooma, line 52: I think this should be:

`const [debouncedSetRadius] = useState(() => debounce(setDebouncedRadius, 200))`

So that the function is not recreated on each render

jahooma (approved):
Looks good. Thanks Inga! This is a pretty sweet feature for a dating app that is only a couple weeks old.

PR #2142, opened 2023-10-31
Make embed QR green + bottom left + helper text

akrolsmir:
New: 
Old: 
image

I think the green's an improvement over black but it's not a perfect match


IanPhilips:
I think the team voted against this idea, but thanks for the taking a shot at it!

PR #2146, opened 2023-10-31
Add Back Button to Profile on Mobile

casesandberg:
**Problem:** The UX on mobile when clicking into a profile is confusing because there is no way to get back out of the profile. Most modern phones will have auto-hid the browser bar thus making it very tricky to go back to the page you were just on. 

To solve for this I have added a sticky header to the top of user profile on mobile with a back button and relocated the More Options menu here as well. As you scroll down the page it shows the username in the header as well. 

**Before:**

Screenshot 2023-10-31 at 3 39 31 PM

**After:**

Screenshot 2023-10-31 at 4 14 07 PM

(Page on first load)

Screenshot 2023-10-31 at 4 14 15 PM

(Scrolled shows username and the bumped down sticky table headers)

### DIFF
--- a/web/components/bet/user-bets-table.tsx
+++ b/web/components/bet/user-bets-table.tsx
@@ -477,9 +477,10 @@ function BetsTable(props: {
     <Col className="mb-4 flex-1 gap-4">
       <Col className={'w-full'}>
         <Row
-          className={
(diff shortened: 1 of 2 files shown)


jahooma (approved):
Wow, thanks! This looks like an improvement.

Congrats on your first contribution!

PR #2148, opened 2023-11-01
Add `Order by` to Profiles View

casesandberg:
This PR adds a select next to the filter button to allow users to switch ordering between `Recently Active` and `Newly Registered`. I also included a change to the display of filters to make them 2-column on desktop for easier scanning. 

New Order By select:

Screenshot 2023-10-31 at 6 09 17 PM

Two-column filters on desktop:

Screenshot 2023-10-31 at 6 09 33 PM

One-column filters on mobile:

Screenshot 2023-10-31 at 6 09 46 PM

### DIFF
--- a/love/components/filters.tsx
+++ b/love/components/filters.tsx
@@ -140,14 +140,14 @@ export const Filters = (props: {
     console.log(filteredLovers)
   }
   const cities: { [key: string]: string } = {
+    All: '',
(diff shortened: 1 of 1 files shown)


jahooma (approved):
Awesome, looks like another good change. I wanted to add this sort myself.

I think the display name of the sorts could be shorter?

PR #2150, opened 2023-11-01
Red arrow go down

Di10n:
When users reply to a comment with NO bets, it should be indicated by a red arrow going down, not up.


IanPhilips:
I thought of this and expect it may make uninformed viewers think people are downvoting a comment when it's attracted many no bets, which is not true: it may be an insightful comment in support of a no resolution. Additionally, you're raising the NO prob by betting no, so it really is accurate how it is, imo. This is the second time someone's suggest this tho, so we can try it and see if people misconceive it.

PR #2151, opened 2023-11-01
Preserve newlines in Linkify component while linkifying text

casesandberg:
This PR modifies the Linkify component to preserve newlines in the text while also converting mentions and URLs into links.

Screenshot 2023-10-31 at 9 18 03 PM

### DIFF
--- a/web/components/widgets/linkify.tsx
+++ b/web/components/widgets/linkify.tsx
@@ -38,13 +38,7 @@ export function Linkify(props: { text: string; className?: string }) {
     <span className={clsx(className, 'break-anywhere')}>
       {text.split(regex).map((part, i) => (
         <Fragment key={i}>
-          {/* Ensure newlines are respected */}
-          {part.split('\n').map((line, j) => (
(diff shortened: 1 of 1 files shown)


web/components/widgets/linkify.tsx
```diff
@@ -38,7 +38,13 @@ export function Linkify(props: { text: string; className?: string }) {
     <span className={clsx(className, 'break-anywhere')}>
       {text.split(regex).map((part, i) => (
         <Fragment key={i}>
-          {part}
+          {/* Ensure newlines are respected */}
+          {part.split('\n').map((line, j) => (
```
jahooma, line 42: I think there's a css-only solution with whitespace-preline or something. Will take a look tomorrow

casesandberg: You are right! Just pushed up a css-only version.

jahooma:
Thanks, didn't want to change this component since it is used many places, but I did it otherwise.

PR #2154, opened 2023-11-01
Make user embeddings fetch optimize better

mqp:
For some reason I don't fully understand, the optimizer can't see through this `u.id = any(array[...]::text)` clause, so it does a scan on the USM table every time because it thinks there's basically no filter.

Changing it to use a (more normal) `u.id in (...)` clause works fine.

FYI @IanPhilips


IanPhilips:
Interesting, thanks!

PR #2162, opened 2023-11-02
Overhaul watch into bookmarks

sipec:
- No bookmark on bet, comment. only on create, boost, answer
- Move bookmark button into top bar. Hide likes on mobile

We'd also need to run the following migration:
```sql
delete from contract_follows
```

Resolves MAN-1296


sipec:
image
image
image

web/components/contract/like-button.tsx
```diff
@@ -132,26 +132,18 @@ export const LikeButton = memo(function LikeButton(props: {
...
-          <div className="relative">
-            <HeartIcon
-              className={clsx(
-                'h-6 w-6',
-                liked &&
-                  'fill-scarlet-200 stroke-scarlet-300 dark:stroke-scarlet-600'
-              )}
-            />
-          </div>
+          <HeartIcon
```
IanPhilips, line 138: Change to a bookmark sign?

backend/functions/src/triggers/on-create-comment-on-contract.ts
```diff
@@ -105,8 +104,6 @@ export const onCreateCommentOnContract = functions
     const commentCreator = await getUser(comment.userId)
     if (!commentCreator) throw new Error('Could not find comment creator')
 
-    await addUserToContractFollowers(contract.id, commentCreator.id)
```
IanPhilips, line 108: I think we should send users who commented on the contract, a resolution notification.

backend/functions/src/triggers/on-create-bet.ts
```diff
@@ -145,11 +144,6 @@ export const onCreateBet = functions
     await giveUniqueBettorAndLiquidityBonus(contract, eventId, bettor, bet)
 
     await Promise.all([
-      // They may be selling out of a position completely, so only add them if they're buying
```
IanPhilips, line 148: I think bettors will still receive a resolution notification bc they'll have a contractMetric associated with the contract, so that's good.

IanPhilips (commented):
This seems like a good change. I think the commentors & subsidizers should still be alerted when the contract resolves, though. 

Ways to extend the follow that @jahooma and I talked about:  
Mobile users don't have anywhere to see their daily movers. You can surface the bookmarked markets with the daily movers with a toggleable component on the feed. First thing they see everyday is the component expanded, and toggling it hides it that day. It would show their bookmarked markets at top, followed by biggest movers underneath, a la Apple Stocks.

image

backend/functions/src/triggers/on-create-bet.ts
IanPhilips: The sooner you merge this the sooner we can stop sending traders emails for every unlinked binary market resolution!

IanPhilips:
I think two things should happen:
- when emails get sent out for independent multi binary markets, only those users who bet on the answer get an email. 
- You should build your contract bookmark feature as separate from the contract follows and leave them as they are. 
  - if you really think people who comment on markets shouldn't follow contracts, just undo that.

sipec:
i think this should be started over as a new collection, or just turning contract likes into bookmarks

PR #2163, opened 2023-11-03
Nicer MVP badge tooltip text

Bohaska:
Change tooltip text for MVP badge to "Thanks for helping Manifold!"


sipec:
I don't think this is any clearer. Also this badge is short lived anyways

PR #2173, opened 2023-11-05
Claiming that mana is not property

ProgramCrafter:
> unrelated but still related: I wish it was crystal clear to any random person who looks that mana is not a crypto thing and also not real money
> 
> Eliza (Trustworthy-ish), https://discord.com/channels/915138780216823849/1169517299196252200/1170203384805072917

### DIFF
--- a/web/components/explainer-panel.tsx
+++ b/web/components/explainer-panel.tsx
@@ -59,6 +59,16 @@ export const WhatIsMana = () => (
     <br />
     Most users never have to buy mana!
     <br />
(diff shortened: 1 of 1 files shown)


sipec:
I think this makes the explanation overly technical which could turn people away. The purpose of these sections is be marketing copy on a landing page. The phrasing should be enticing and simple, pandering even

And like, we still want people to find mana valuable. Emphasizing the ways it gets taken away is bad

PR #2175, opened 2023-11-05
API: Get Comments and Markets by `userId`

casesandberg:


web/pages/api/v0/comments.ts
```diff
@@ -51,7 +55,16 @@ export default async function handler(
...
+  const { userId, limit } = params
+
   try {
+    if (userId) {
+      const comments = await getUserComments(userId, limit, 0)
+      res.setHeader('Cache-Control', 'max-age=15, public')
+      return res.status(200).json(comments)
+    }
+
     const contractId = await getContractId(params)
```
jahooma, line 68: Seems like if you specify both userId and contractId, it should get that user's comments from that contract

casesandberg: Yep, missed that. Just pushed up a fix.

jahooma (approved):
Cool, thanks Case!

PR #2176, opened 2023-11-06
API: Add market/[id]/update route

### LINKED ISSUE #2043: Update market descriptions via the API
I have a bunch of markets that I batch created via the API and I would like to batch update the descriptions using the API (I accidentally used "Description" instead of "DescriptionMarkdown" to upload the text).

Unfortunately as far as I can tell this is not possible, it would be great if it was possible.

eppsilon:
Hi - I've attempted to add the ability to update a market question or description via the API. It's a feature I could use and it was a nice reason to explore the Manifold source code. Is this contribution something that could be accepted/merged if I also update the API docs and figure out how to test it?

---

- Move `getDescriptionJson()` from `backend/api/src/create-market.ts` to `common/src/contract.ts` so it can be used for creating or updating
- Modify `backend/api/src/update-market.ts` to accept `question` and `description` (optional) and validate/save them
- Add `web/pages/api/v0/market/[id]/update.ts` to pass requests to the updatemarket() function.

Resolves #2043.

### DIFF
--- a/backend/api/src/update-market.ts
+++ b/backend/api/src/update-market.ts
@@ -1,76 +1,114 @@
-import { APIError, authEndpoint, validate } from 'api/helpers'
+import { APIError, typedEndpoint, validate } from 'api/helpers'
 import { getContractSupabase } from 'shared/utils'
 import * as admin from 'firebase-admin'
-import { z } from 'zod'
 import { trackPublicEvent } from 'shared/analytics'
 import { throwErrorIfNotMod } from 'shared/helpers/auth'
 import { removeUndefinedProps } from 'common/util/object'
 import { recordContractEdit } from 'shared/record-contract-edit'
+import { getDescriptionJson } from 'common/contract'
+import { JSONContent } from '@tiptap/core'
+import { updateMarketProps } from 'common/api/market-types'
 
-const bodySchema = z
-  .object({
-    contractId: z.string(),
-    visibility: z.enum(['unlisted', 'public']).optional(),
-    closeTime: z.number().optional(),
-    addAnswersMode: z.enum(['ONLY_CREATOR', 'ANYONE']).optional(),
-    sort: z.string().optional(),
-  })
-  .strict()
-
-export const updatemarket = authEndpoint(async (req, auth, log) => {
-  const { contractId, visibility, addAnswersMode, closeTime, sort } = validate(
-    bodySchema,
-    req.body
-  )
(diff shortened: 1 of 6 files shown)


backend/api/src/update-market.ts
```diff
@@ -1,26 +1,53 @@
...
+    !question &&
+    !description &&
+    !descriptionHtml &&
+    !descriptionMarkdown &&
+    !descriptionJson
+  )
+    throw new APIError(
+      400,
+      'Must provide visibility, closeTime, question, description, descriptionHtml, descriptionMarkdown, or descriptionJson'
+    )
```
sipec, line 50: this API error is extraneous - `validate` actually already throws an error if the api params aren't right

backend/api/src/update-market.ts
```diff
@@ -49,10 +81,39 @@ export const updatemarket = authEndpoint(async (req, auth) => {
...
   await recordContractEdit(
     contract,
     auth.uid,
-    buildArray([visibility && 'visibility', closeTime && 'closeTime'])
+    buildArray([
+      visibility && 'visibility',
+      closeTime && 'closeTime',
+      question && 'question',
+      resolvedDescription && 'description',
+    ])
```
sipec, line 116: Currently the contract edits (what shows up in edit history) for descriptions and questions titles are handled via the `on-update-contract.ts` firebase trigger. In the frontend, we're writing to firebase directly to update those fields. The code here would record the title and description edits twice I think.

To solve, you can either remove the insert to `contract_edits` in the firebase trigger and change `saveDescription()` in the frontend to call the api. Or you can not update the question and description here

sipec (changes requested):
I like where this pr is going, it just needs a few fixes. @eppsilon do you think you'll have time to take a look? (ok if not)

backend/api/src/update-market.ts
jkerkhoff: Doesn't that only do validation on the individual fields though? Seems like this is checking if any field is set, since otherwise it'd be a no-op.

backend/api/src/update-market.ts
eppsilon: OK, I have removed the recordContractEdit() calls for the question and description.

eppsilon:
Hi - I've rebased this PR and made a couple tweaks.

sipec:
superseded by 0fae39b5f7f442cfdad677a82dadfb651980d622
thanks tho!

PR #2179, opened 2023-11-06
Allow Streamlit embeds

JulHeg:
Hey! This should make it possible to embed Streamlit apps in Manifold comments and descriptions, so that for example people can play with new ways to visualize markets.

### DIFF
--- a/web/components/editor/embed-modal.tsx
+++ b/web/components/editor/embed-modal.tsx
@@ -84,10 +84,7 @@ const embedPatterns: EmbedPattern[] = [
 ]
 const allowed_domains: string[] = [
   "streamlit.app",
-  "kalshi.com",
-  "polymarket.com",
-  "predictit.org",
-  "gjopen.com",
+  "wikipedia.org",
 ];
 
 function isAllowedDomain(url: string) {
--- a/web/components/editor/embed-modal.tsx
+++ b/web/components/editor/embed-modal.tsx
@@ -82,7 +82,7 @@ const embedPatterns: EmbedPattern[] = [
       `<iframe src="https://www.tiktok.com/embed/v2/${id}"></iframe>`,
   },
 ]
-const allowed_domains: string[] = [
+const allowedDomains: string[] = [
   "streamlit.app",
   "wikipedia.org",
 ];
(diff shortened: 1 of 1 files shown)


akrolsmir:
Thanks for , but a couple of thoughts:
1. This is already possible if you just embed a link with `<iframe src="..."></iframe>`. We disallowed this from the ui https://github.com/manifoldmarkets/manifold/commit/29f874c33337ec1c334dcdb8e3ab20497e8dc38e but adding the 
2. The code in embed-modal.tsx was supposed to rewrite embeds for the purpose of custom embed links that are different than the default link. Streamlit embeds are supposed to have append a `?embed=true` parameter which will result in slightly different behavior
3. It's not clear to me that the user experience of using an embed for the Manifold 2x2 viewer is better than just copying over the generated plot...

JulHeg:
Good points @akrolsmir! The first one definitely does make this PR more of a nice-to-have, but I guess we’re already here. So about the other two:
2. I made another code change so there is now a list of domains that are allowlisted but don’t require any changes to the URL. I guess that makes sense to have anyway? 
3. The upside over copying just the image is that it updates with the current market price. I added some custom CSS to resize the mosaic plot so it fits the iframe without having to scroll. It still takes a few seconds to load and is generally less nice than something built into the site, but much improved if I say so myself: <https://manifold.markets/BoltonBailey/will-a-democrat-win-the-2024-us-pre-f7bdb63135a1#ng14xgo4JHbTNx99xpKT>

web/components/editor/embed-modal.tsx
```diff
@@ -81,18 +81,41 @@ const embedPatterns: EmbedPattern[] = [
...
     const match = text.match(pattern.regex)
     if (match) {
       return pattern.rewrite(match[1])
     }
+    
+  }
+
+  const urlPattern = /^(https?:\/\/.*)/;
+  const match = text.match(urlPattern);
+  if (match) {
```
akrolsmir, line 112: Would simplify this to
```
// If it's a URL, check against the allowed domains
if (match && isAllowedDomain(match[1])) {
  return ...
}
```
and exclude the extra return null, because that's already caught by the next case below

web/components/editor/embed-modal.tsx
```diff
@@ -81,18 +81,41 @@ const embedPatterns: EmbedPattern[] = [
     rewrite: (id) =>
       `<iframe src="https://www.tiktok.com/embed/v2/${id}"></iframe>`,
   },
-  // {
-  //   regex: /^(https?:\/\/.*)/,
-  //   rewrite: (url) => `<iframe src="${url}"></iframe>`,
-  // },
 ]
+const allowed_domains: string[] = [
```
akrolsmir, line 85: nit: our convention here is to use camelCase so `allowedDomains`

akrolsmir (approved):
Looks great, thanks!

JulHeg:
Yay!

PR #2181, opened 2023-11-07
Clean up header on profile pages

Di10n:
On profile pages, there is a random gap that appears to the left of the header which I've removed. I also replaced the stacked usernames in the top bar with just the username + badge because having all of it appear there is redundant.

### DIFF
--- a/web/pages/[username]/index.tsx
+++ b/web/pages/[username]/index.tsx
@@ -217,11 +217,11 @@ function UserProfile(props: {
       )}
       {showConfetti && <FullscreenConfetti />}
 
-      <Col className="mx-4 mt-1">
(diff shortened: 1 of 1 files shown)


jahooma:
Thanks! In the future for visual changes like this, it's good to put a screenshot (ideally a before / after). But nice!

Di10n:
> Thanks! In the future for visual changes like this, it's good to put a screenshot (ideally a before / after). But nice!

Sure thing! Before:

Screenshot 2023-11-06 at 10 35 50 PM

Screenshot 2023-11-06 at 10 35 56 PM

and after:

Screenshot 2023-11-06 at 10 44 25 PM

PR #2187, opened 2023-11-08
Update delete account

SirSaltyy:
-Now deletes *all personal info* from firebase.
-Improved copy to be more transparent.

### DIFF
--- a/web/pages/profile.tsx
+++ b/web/pages/profile.tsx
@@ -30,6 +30,7 @@ import { useUser } from 'web/hooks/use-user'
 import ShortToggle from 'web/components/widgets/short-toggle'
 import { InfoTooltip } from 'web/components/widgets/info-tooltip'
 import { useEditableUserInfo } from 'web/hooks/use-editable-user-info'
+import { Tooltip } from 'web/components/widgets/tooltip'
 
 export const getServerSideProps = redirectIfLoggedOut('/', async (_, creds) => {
   return { props: { auth: await getUserAndPrivateUser(creds.uid) } }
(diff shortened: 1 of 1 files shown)


SirSaltyy:
FYI @mantikoros I wasn't able to test the backend functionality, as my dev.sh environment is still not working.

As far as I can tell I don't need to redeploy anything once merged for these changes to come into effect as I'm not changing any backend functions, but let me know if that isn't the case. When I try to delete in prod local host I get a "failed to delete account" though, not sure if it's due to changes I made or just because it's in the wrong environment.

mantikoros (approved):
LGTM, but don't merge this in until after you've tested this on dev.

web/pages/profile.tsx
```diff
@@ -112,6 +113,12 @@ export default function ProfilePage(props: {
       //eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore
       twitchInfo: deleteField(),
+      //eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      initialIpAddress: deleteField(),
```
IanPhilips, line 118: I'm not sure I think this is a good idea. What about trolls that come onto the site, wreak havoc with spam, then go and delete their account - how can we ban their ip address going forward?

web/pages/profile.tsx
```diff
@@ -112,6 +113,12 @@ export default function ProfilePage(props: {
...
+      //eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      initialDeviceToken: deleteField(),
```
IanPhilips, line 121: This isn't really personalized, just basically a cookie on their browser

IanPhilips:
I think we should hold off on this until we have a good solution for handling banned users' ip addresses.

PR #2188, opened 2023-11-09
Ingawei/questions add

ingawei:
image
image
image
image
image

I refresh the answers after you submit in a pretty jank way, please check if it is acceptable lol


love/components/questions-form.tsx
```diff
@@ -151,9 +161,80 @@ const QuestionRow = (props: { row: rowFor<'love_questions'>; user: User }) => {
...
     </Col>
   )
 }
+
+export const IndividualQuestionRow = (props: {
+  row: rowFor<'love_questions'>
+  initialAnswer?: rowFor<'love_answers'>
+  user: User
+  onCancel: () => void
+  onSubmit?: (form: loveAnswerState) => void
```
IanPhilips, line 176: doesn't look like anyone uses the parameter of the callback, maybe leave it out?

love/components/answers/lover-answers.tsx
```diff
@@ -6,21 +6,40 @@ import { partition } from 'lodash'
...
+  const questions = allQuestions.filter((question) =>
+    answerQuestionIds.has(question.id)
+  )
 
   const [multiChoiceAnswers, otherAnswers] = partition(
     answers,
     (a) => a.multiple_choice != null
   )
 
+  function refreshAnswers() {
```
IanPhilips, line 39: This is not the right way to do this. The useUserAnswersAndQuestions should return a `refreshResults` callback in addition to the `questions` and `answers` it returns now. Then anyone can call the refresh callback on demand. With your current setup, anyone who wants to refresh their results has to create this extra scaffolding of a refresh state and an incrementing index.

love/hooks/use-questions.ts
```diff
@@ -29,6 +40,7 @@ export const useUserAnswersAndQuestions = (userId: string | undefined) => {
         setQuestions(questions)
       })
     }
-  }, [userId])
+  }, [userId, refresh])
```
IanPhilips, line 43: Move the logic currently within the useEffect into a function within the hook so that you can return a callback to it. Then call the function in the useEffect.

love/lib/supabase/answers.ts
```diff
@@ -0,0 +1,15 @@
+import { Row as rowFor, run } from 'common/supabase/utils'
+import { db } from 'web/lib/supabase/db'
+
+export const deleteAnswer = async (
```
IanPhilips, line 4: Have you added this ability in the `love_answers.sql`? I believe you need:
` drop policy if exists "self delete" on love_answers;
create policy  "self delete" on love_answers for delete
    using (creator_id = firebase_uid());`

love/components/questions-form.tsx
```diff
@@ -59,75 +60,84 @@ export const QuestionsForm = (props: { questionType: QuestionType }) => {
...
+  ) => void
+) => {
+  const res = await run(
+    db
+      .from('love_answers')
+      .select('*')
+      .eq('question_id', id)
+      .eq('creator_id', userId)
+  )
+  if (res.data.length) {
```
IanPhilips, line 81: Hmmm this is weird to pass in a setState as a parameter to a db fetching function like this, why not just set the state from the parent with the results of the function?

love/components/questions-form.tsx
```diff
@@ -59,75 +60,84 @@ export const QuestionsForm = (props: { questionType: QuestionType }) => {
...
-        .eq('question_id', id)
-        .eq('creator_id', user.id)
-    )
-    if (res.data.length) {
-      setForm(res.data[0])
-    }
-  }
   useEffect(() => {
-    fetchPrevious()
+    fetchPrevious(id, user.id, setForm)
```
IanPhilips, line 128: e.g. this should be `fetchPrevious(id, user.id).then(setForm)`

IanPhilips (commented):
This looks great, Inga! This will be much more approachable than having 100 open text boxes. I made a few comment on ze code.

PR #2191, opened 2023-11-09
Run spellcheck

calvinballing:
Ran cspell locally and fixed a number of typos.  Very roughly the more important changes are in earlier commits.

Be sure to check that none of these will have any unintended ill effects, such as the changes to `users_popuate`, or the `eventTrackingName="copy lover referrral"`.  If so we can pull those out and merge the rest.


calvinballing:
Could I get a workflow approval on this?

backend/supabase/users/triggers.sql
```diff
@@ -7,7 +7,7 @@ create or replace function users_populate_cols () returns trigger language plpgs
     return new;
 end $$;
 
-create trigger users_popuate before insert
+create trigger users_populate before insert
```
jahooma, line 10: Undo this one. I think it would create a different trigger if we change it

PR #2193, opened 2023-11-11
Make a GCE scheduler service to run jobs

mqp:
Pretty self-explanatory. I want to convert the cloud function scheduled jobs to this. Benefits:

1. They only run once without any fiddly locking. (This is the main driver.)
2. Runs of the same job won't run "on top of each other" if we misunderestimate how long they will take.
3. We can easily see what's running and what's scheduled in one place.
4. They don't recycle instances every run (useful for jobs that like to cache stuff in memory.)
5. Firebase functions just kind of suck.
6. It's a nice prototype of how to deploy a containerized service to GCE that we can use in the future (e.g. for the API.)


mqp:
It feels pretty fishy to use a scheduler inside Node, but the alternatives are:

- Use an external cloud scheduler application. (Google Cloud Scheduler refuses to give us an only-once guarantee, which is why we are in this mess in the first place.)
- Use `cron` on a box. Requires a box that actually has software running outside of a container, terrifying my coworkers. Also lacks "don't run again if it's taking a long time" functionality.
- Use `systemd` on a box. Requires a more complicated box, terrifying my coworkers more.
- Use Google Kubernetes Engine, which has a cron built in. Requires learning Kubernetes.

So I hope this solution can tide us over for now and avoid having to do something more sophisticated.

PR #2194, opened 2023-11-11
Fixing conditional probablity render in the case of resolved dating markets

tbnichols:
https://discord.com/channels/915138780216823849/1166835928518631454/1172784262047617025

as noted here, there are issues with showing conditional probability for dates when one of the previous markets has already resolved

replicable on second date market between Annie L and https://manifold.love/emmy here

Fix looks at markets that have resolved yes using a ternary statement and sets their probability to 1 instead of whatever the answer was before resolution

### DIFF
--- a/love/lib/util/relationship-market.ts
+++ b/love/lib/util/relationship-market.ts
@@ -7,6 +7,6 @@ export const getCumulativeRelationshipProb = (
   const { answers } = contract
   return answers
     .slice(0, index + 1)
-    .map((a) => a.prob)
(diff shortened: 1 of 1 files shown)


tbnichols:
image
Fix shows on local 
image
compared to prod

jahooma:
Thanks for this! The issue is mostly a backend problem that should be setting the answer.prob to 1. I've fixed it for the case you cited. Happy to fix other cases you find, but hopefully it is fixed going forward.

PR #2196, opened 2023-11-13
Add @Sayaka as a bot account

xzi:
I am the owner of the account and am running a small scale bot off it. I would like the bot tag on it if that is ok. Thanks and have a good one


akrolsmir:
Thanks for listing your bot!

PR #2199, opened 2023-11-14
update calibration page

SirSaltyy:


mantikoros (changes requested):
I made a few changes to the formatting and content, but it mostly looks good!

The hydration error was a result of putting elements like <div> inside <p>, which you could have discovered by more thoroughly reading the error messages in the console.

The calibration graph is still messed up but I'm not sure what the problem is. That needs to be fixed before pushing. I'll have @sinclairity (the original creator of the graph) take a look.

mantikoros:
@SirSaltyy, in the future don't use <p>! Just use <div>. That way you'll avoid the weird hydration issues.

sipec (approved):

PR #2206, opened 2023-11-17
Logging with severity, endpoint, and trace id

IanPhilips:
Just a sample of what logging could look like. If this looks good I will phase out the old `log = (...args: unknown[])` style and move log out of the scheduler into `/shared`. I'm not sure how to distinguish between DEBUG and INFO, I guess debug is more verbose? Not sure we'll really use that fine-grained of logging. Also will want to pass the`APIError` a `log.error` to use


IanPhilips:
Looks like it works on dev! Copying the trace from the cloud run request worked as well.
Screenshot 2023-11-16 at 5 16 39 PM

jahooma:
Nice!

I would be ok with `log` being a function that always is of status DEBUG. I don't think I've ever wanted to use the others?

PR #2211, opened 2023-11-20
Await Mobile Sign-in, Typed Native Bridge Messages

casesandberg:
Typescript was having a real tough time typing a mix of parametrized functions and object arg functions, so I have converged all of the messages to pass around an object for ease of types. 

This PR introduces `postMessageToNativeAwaitResponse` which is a promisify wrapper around the window messages between native and the web view. In this case, in order for `firebaseLogin` to behave similarly on mobile and desktop (returning the user object), we need to await the response of a successful (or aborted) login on native. 

Let me know if the typed changes are too drastic, I can remove those commits from the PR to only include the main feature change. 

### TODO
- [ ] Verify `NativeLoginAbortMessage` functions correctly in ios/android


jahooma:
Nice!

IanPhilips:
Oh this looks cool, thanks for submitting it! I'm not entirely sure what user problem this solves, is there a bug this is fixing or UX this improves? I do see that it seems to improve the developer UX.

casesandberg:
I was working on a change to make it so all `firebaseLogin` calls do not short-circuit the action if they are logged out. Currently, if a user initiates an action logged out they get intercepted to log in or create an account, and then they have to re-initiate the action they wanted to do. If we are able to await login then we can continue the action as soon as they are logged in. 

It ended up being a big can of worms (this included) so I decided to split it up into two parts to make it easier to review.

native/components/auth-page.tsx
```diff
@@ -81,10 +109,7 @@ export const AuthPage = (props: {
       if (data?.displayName && !user.displayName) {
         await updateProfile(user, { displayName: data.displayName })
       }
-      const fbUser = user.toJSON()
```
IanPhilips, line 84: This is a necessary bit, the user isn't stringifiable without it

native/components/auth-page.tsx
```diff
@@ -46,23 +47,50 @@ export const AuthPage = (props: {
...
+  }
+
   // We can't just log in to google within the webview: see https://developers.googleblog.com/2021/06/upcoming-security-changes-to-googles-oauth-2.0-authorization-endpoint.html#instructions-ios
   useEffect(() => {
     try {
       if (response?.type === 'success') {
         const { id_token } = response.params
         const credential = GoogleAuthProvider.credential(id_token)
         signInWithCredential(auth, credential).then((result) => {
-          const fbUser = result.user.toJSON()
```
IanPhilips, line 56: Necessary bit

native/App.tsx
```diff
@@ -344,7 +353,7 @@ const App = () => {
     // Receiving cached firebase user from webview cache
     else if (type === 'users') {
       try {
-        const fbUserAndPrivateUser = JSON.parse(payload)
```
IanPhilips, line 347: the payload is stringified, right?

native/App.tsx
```diff
@@ -332,10 +337,14 @@ const App = () => {
     // User needs to enable push notifications
     else if (type === 'promptEnablePushNotifications') {
       registerForPushNotificationsAsync().then((token) => {
-        if (token)
-          communicateWithWebview('pushToken', {
-            token,
-            userId: fbUser?.uid,
+        if (token && fbUser)
```
IanPhilips, line 340: Not clear to me that we need both the fbuser on our side just to pass the token

native/App.tsx
```diff
@@ -315,15 +312,23 @@ const App = () => {
         if (status === 'granted') {
           const token = await getPushToken()
           if (!webview.current) return
-          if (token)
-            communicateWithWebview('pushToken', {
-              token,
-              userId: fbUser?.uid,
+          if (token && fbUser)
```
IanPhilips, line 315: Not sure about why the check for `fbUser`

casesandberg:
Thanks @IanPhilips! I ended up just removing all of the typescript changes because it was hard to tell what was going on and was likely introducing bugs (like the `toJson`). This should be easier to review.

PR #2215, opened 2023-11-21
[WEB] Fix Bet Intent Tracking

casesandberg:
This PR fixes the bet intent tracking I introduced last week. As I explore the codebase more, I found that the initial intent tracking does not actually cover most cases. This will now track intent on all "bet" buttons, including ones that are shown inside modals.


mantikoros (approved):
LGTM! nice change

PR #2228, opened 2023-11-27
Remove CORS requirement from all API endpoints

akrolsmir:
This allows other websites (like manifund.org) make client-side requests to Manifold's API endpoints without any additional work. Note that before this PR, about half of our 32 APIs restricted CORS access to Manifold's website and localhost; the other half had no restrictions. AFAICT there was no underlying reason for this, so I'm getting rid of all of it.

This also fixes the Manifund issue where it is unable to pull up your account and balance via the /v0/me API:

image

### DIFF
--- a/web/lib/api/cors.ts
+++ b/web/lib/api/cors.ts
@@ -4,11 +4,11 @@ import { NextApiRequest, NextApiResponse } from 'next'
 export function applyCorsHeaders(
   req: NextApiRequest,
   res: NextApiResponse,
-  params: Cors.CorsOptions
+  params?: Cors.CorsOptions
 ) {
   // This cors module is made as express.js middleware, so it's easier to promisify it for ourselves.
   return new Promise((resolve, reject) => {
-    Cors(params)(req, res, (result) => {
+    Cors(params ?? CORS_UNRESTRICTED)(req, res, (result) => {
       if (result instanceof Error) {
         return reject(result)
       }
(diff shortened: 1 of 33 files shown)


akrolsmir:
@mqp Does anything bad happen when we get rid of CORS requirements?

jahooma:
I guess I didn't get either why we needed any CORS restrictions. Seems like a nice feature for other sites to use our API

mqp:
Is there a reason why you didn't just add Manifund to the CORS whitelist for the stuff it wanted to use?

If you look up CORS you will quickly understand what it's for. It's trying to make it so that some third party can't make a button on their website that says "free cake" and when you click it then the button sends all your mana to them. Mostly this is only a problem for people who use cookies for authentication, since without CORS the button will use your cookies. So I can't think of anything that would happen to go wrong with unrestricted CORS on our API right now, although it's possible I didn't think of something. I would probably still leave it on for defense-in-depth reasons unless there is a reason to turn it off.

If you toast it, I will laugh at you when later you add some API functionality that reads your auth cookie and someone exploits it to fuck everyone over. But other than that it's your call.

akrolsmir:
Cool, thanks for the color!

> Is there a reason why you didn't just add Manifund to the CORS whitelist for the stuff it wanted to use?

Yeah: 1) in case other people want to create integrations (eg Beeminder) for which the CORS stuff is annoying, and 2) because of the inconsistency in the APIs in the first place where CORS was only required for a random half of them.

PR #2229, opened 2023-11-28
Specify engine in Dream API Call

wasabipesto:
Dreamstudio [deprecated some engines](https://platform.stability.ai/docs/features/api-parameters#engine) and so a new one must be specified.

### DIFF
--- a/backend/shared/src/dream-utils.ts
+++ b/backend/shared/src/dream-utils.ts
@@ -1,12 +1,15 @@
 import { DOMAIN } from 'common/envs/constants'
 const MODIFIERS =
   '8k, beautiful, illustration, trending on art station, picture of the day, epic composition'
+// Select engine from https://platform.stability.ai/docs/features/api-parameters#engine
+const ENGINE = 'stable-diffusion-xl-1024-v0-9'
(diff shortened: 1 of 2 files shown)


IanPhilips:
Hmm this doesn't seem to work on the dev preview, we can try prod

IanPhilips:
Seems like it doesn't return an image

Screenshot 2023-11-28 at 10 14 23 AM

IanPhilips:
Let me try again, I think I made a spelling error

IanPhilips:
Thanks @wasabipesto!

wasabipesto:
I'm not sure this is 100% fixed - I changed to the SDXL v0.9 model since that was the first one that worked from my testing, but it seems like the vercel function might be timing out while it waits for the image response. The function times out after 15s but the image generation can take 20-40s.

IanPhilips:
Oh I see, we can fix that if it's a real problem. I haven't experienced it yet

wasabipesto:
Well it's working fine on prod now, maybe dev has different timeout limits.

One last thing, I would keep an eye on usage costs. I don't know how much more the SDXL API costs but there was a note that it would be higher than the old engine.

PR #2230, opened 2023-11-28
Add Auth Check Handler to Continue User Actions

casesandberg:
Right now when a logged out user goes to place a bet, they are forced to log in (or create an account) and then have to re-attempt placing the bet. This PR introduces an `authCheckHandler` to replace basic firebaseLogin calls and continue the user action automatically once they are successfully logged in.

### DIFF
--- a/web/hooks/use-auth-check-handler.ts
+++ b/web/hooks/use-auth-check-handler.ts
@@ -1,0 +1,34 @@
+import { useUser } from './use-user'
+import { useCallback } from 'react'
+import { firebaseLogin } from 'web/lib/firebase/users'
+
+export function useAuthCheckHandler() {
+  const user = useUser()
+  const isAuthenticated = !!user
+
+  return {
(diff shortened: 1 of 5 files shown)


jahooma:
Would users expect the bet to go through? They might want to reconsider based on what balance they get on signing up?

If new users do want this, happy to merge after fixing merge conflicts.

web/hooks/use-auth-check-handler.ts
```diff
@@ -0,0 +1,34 @@
+import { useUser } from './use-user'
+import { useCallback } from 'react'
+import { firebaseLogin } from 'web/lib/firebase/users'
+
+export function useAuthCheckHandler() {
+  const user = useUser()
+  const isAuthenticated = !!user
+
+  return {
+    authCheckHandler: useCallback(
```
jahooma, line 10: Why pass this within an object if it's the only return value?

PR #2231, opened 2023-11-29
Ingawei/compatibility answers

ingawei:
TODO:
Editing/deleting answers
Calculating your compatibility score with others (maybe a good task for @mantikoros?)
Truncated view as more people add questions


backend/supabase/love/love-compatibility-answers.sql
```diff
@@ -0,0 +1,41 @@
+create table if not exists
+  love_compatibility_answers (
+    id bigint generated always as identity primary key,
+    question_id bigint not null,
+    creator_id text not null,
```
jahooma, line 5: This is the user who answered it right, not the creator of the question? I guess it's fine, but might have been better as user_id?

jahooma (approved):
This looks fantastic, Inga! Seems great as a working initial version.

I like how easy it is to get the zone answering questions.

- Would be nice to truly skip an answer — it seems like right now it prevents you from going to the next question? E.g. I could imagine someone not wanting to answer the question about sex, but then being unable to answer more questions.
- Once skipping works, you'd want a view to browse through all the questions, including ones you haven't answered (tabs for answered vs. not answered?), and maybe show how many people have answered each. And, you'd want to be able to create new questions, without having answered them all.
- Would be nice to show a compatibility score with someone else. And, as you said, show how your answers are or are not compatible when browsing someone else's profile.
- Not a super fan of the big buttons on your profile. Already looked a little cluttered before we added more.

I think the code looks good! Cool that we could reuse the love_questions table.

PR #2234, opened 2023-11-30
[API] Expose More Endpoints

casesandberg:
For @jkerkhoff's [Mirrorbot Community Fund project](https://manifund.org/projects/mirrorbot), I have exposed more endpoints so that they can add support for Multiple Choice markets.


sipec:
hmm this is probably going to conflict with #2176 and the soon-coming api refactor I'm working on.
I think exposing more endpoints is a good idea.
idk about having multiple http methods per endpoint - we don't really follow that convention

sipec:
> I can remove the update market logic altogether

ok if you do that I think I can just merge it

casesandberg:
Wahoo! Removed!

PR #2243, opened 2023-12-04
handle multiple choice markets in daily profit chart

bgila:
bug report: my lil bro reported an error from tapping "$ invested" on `/home` and I repro'd

probable cause: multiple choice markets don't have a `probChanges` value defined

fix: check whether probChanges is defined, and if it isn't, leave `propChange` as undefined, which `ContractMention` can handle, rather than crashing.

bug:
bug on localhost

confirmed fix locally:
fixed

### DIFF
--- a/web/components/daily-profit.tsx
+++ b/web/components/daily-profit.tsx
@@ -234,10 +234,13 @@ const MarketCell = (props: {
   from: 'day' | 'week' | 'month'
 }) => {
   const c = props.contract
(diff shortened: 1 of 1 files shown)


jahooma:
Thanks for the quick fix! Good to have quick turnaround on this type of client-side crash.

I think @IanPhilips might want to follow up — the backend should be filling in probChanges in all cases for multi markets.

PR #2244, opened 2023-12-05
Ingawei/manifold wrapped

ingawei:
Doesn't work for Acceleration or V, but seems to work for everyone else. Would love some help figuring out if we can make it possible to make it work for those 2 tomorrow haha


backend/supabase/user_contract_metrics.sql
```diff
@@ -1,44 +1,46 @@
...
+    contract_id text not null,
+    data jsonb not null,
+    fs_updated_time timestamp,
+    has_yes_shares boolean,
+    has_no_shares boolean,
+    total_shares_yes numeric,
+    total_shares_no numeric,
+    profit numeric,
+    has_shares boolean,
+    answer_id text
```
jahooma, line 14: It's generally bad to have these whitespace changes in the diff, because it obscures what was actually changed.

backend/supabase/wrapped-functions.sql
```diff
@@ -0,0 +1,42 @@
...
+
+CREATE OR REPLACE FUNCTION calculate_user_profit_for_2023(user_id_input TEXT)
+RETURNS NUMERIC AS $$
+DECLARE
+    total_profit NUMERIC;
+BEGIN
+    SELECT SUM(profit) INTO total_profit
+    FROM public.user_contract_metrics
+    WHERE user_id = user_id_input
+    AND fs_updated_time >= '2023-01-01'::TIMESTAMPTZ
```
jahooma, line 37: Hmm, doesn't this include profit from markets you bet on last year, but got resolved this year?

The most correct version would be this but excluding profit that accrued on those bets last year. There's no simple way to calculate this.

But I think you can just check the saved portfolio metrics? It should have recorded the user's all time profit on Jan 1st (or 0 if they signed up this year). Then you just subtract that from their current all time profit.

backend/supabase/wrapped-functions.sql
```diff
@@ -0,0 +1,42 @@
...
+  month TIMESTAMPTZ,
+  bet_count bigint,
+  total_amount numeric
+) as $$
+BEGIN
+    RETURN QUERY
+    SELECT
+        date_trunc('month', created_time) AS month,
+        COUNT(bet_id) AS bet_count,
+        SUM(amount) AS total_amount  -- Add this line to calculate the sum of amounts
```
jahooma, line 13: This would discount negative bet amounts, mainly sales. If you only had a large sale at a profit, you could even have a negative total bet amount, which doesn't really make sense lol.

I would add an absolute value: SUM(ABS(amount)). But then you need to filter out redemptions below.

backend/supabase/user_contract_metrics.sql
```diff
@@ -1,44 +1,46 @@
...
+where
+  total_shares_yes is not null;
 
-create index contract_metrics_answer_id
-    on user_contract_metrics (contract_id,answer_id);
+create index contract_metrics_answer_id on user_contract_metrics (contract_id, answer_id);
+
+create index idx_user_id_fs_updated_time on public.user_contract_metrics (user_id, fs_updated_time);
+
+CREATE INDEX idx_profit ON public.user_contract_metrics (profit);
```
jahooma, line 46: Kind of weird to mix upper and lower case. Here and in the next sql file.

backend/supabase/wrapped-functions.sql
```diff
@@ -0,0 +1,42 @@
...
+    FROM
+        contract_bets
+    WHERE
+        user_id = user_id_input AND
+        created_time >= '2023-01-01' AND
+        created_time <= '2024-01-01' AND
+        is_ante = false
```
jahooma, line 20: I'd add `is_redemption = false`

jahooma (commented):
Cool stuff, Inga! Maybe we can even build off of this next year haha.

Left a couple comments.

PR #2246, opened 2023-12-05
add manifold politics id to core user list

bgila:


IanPhilips (approved):
You'll also need to add this id into the sql and firebase admins definition

PR #2254, opened 2023-12-06
initial update sort behavior

bgila:
https://linear.app/manifoldmarkets/issue/MAN-1376/market-creator-should-be-able-to-set-default-sort-default-value-for

discussed with ian

### DIFF
--- a/web/components/contract/contract-overview.tsx
+++ b/web/components/contract/contract-overview.tsx
@@ -62,6 +62,7 @@ import { FaArrowTrendDown, FaArrowTrendUp } from 'react-icons/fa6'
 import { formatPercent } from 'common/util/format'
 import { updateContract } from 'web/lib/firebase/contracts'
 import { isAdminId, isModId } from 'common/envs/constants'
+import { updateMarket } from 'web/lib/firebase/api'
 
 export const ContractOverview = memo(
   (props: {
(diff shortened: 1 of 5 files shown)


web/components/contract/contract-overview.tsx
```diff
@@ -560,6 +565,24 @@ const ChoiceOverview = (props: {
...
   )
 
+  useEffect(() => {
+    if (
+      currentUserId &&
+      (isModId(currentUserId) ||
+        isAdminId(currentUserId) ||
+        contract.creatorId === currentUserId)
+    ) {
+      updateContract(contract.id, { sort }).then(
```
IanPhilips, line 582: I think you can wrap this in a `toast.promise`

web/components/contract/contract-overview.tsx
```diff
@@ -560,6 +565,24 @@ const ChoiceOverview = (props: {
     [answers, resolutions, shouldAnswersSumToOne, sort]
   )
 
+  useEffect(() => {
+    if (
+      currentUserId &&
```
IanPhilips, line 577: I think this logic is cursed, should be `currentUserId && (case1 || case2 || case3)`

IanPhilips (changes requested):
Almost there! Last thing you need to do is add the logic in the `/backend/api/src/update-market.ts`

bgila:
addressed, ty

PR #2258, opened 2023-12-07
Probability  -> Value

NcyRocks:
The limit order screen reads "Value" instead of "Probability" on numeric markets


jahooma:
I think that's deliberate b/c pseudo numeric show a numeric value rather than a percent

jahooma (approved):

jahooma:
Oh sure! Sorry, read this from my phone. Looks good.

PR #2263, opened 2023-12-08
migrate vercel APIs pt 2

sipec:
migrates basically all the GETs

what's left: dalle & dream, and search, contract tagging, link preview. oh and revalidate - but we need to keep that I think

and of course our internal private apis and updating the docs ... but that can wait

### DIFF
--- a/backend/api/src/app.ts
+++ b/backend/api/src/app.ts
@@ -123,6 +123,7 @@ import { getUser } from './get-user'
 import { getUsers } from './get-users'
 import { getMarket } from './get-market'
 import { getGroup } from './get-group'
+import { getPositions } from './get-positions'
 
 const allowCorsUnrestricted: RequestHandler = cors({})
 const allowCorsManifold: RequestHandler = cors({
@@ -198,6 +199,7 @@ const handlers: { [k in APIPath]: RequestHandler } = {
   markets: markets,
   'send-mana': sendMana,
   managrams: getManagrams,
+  positions: getPositions,
   me: getCurrentUser,
   user: getUser,
   users: getUsers,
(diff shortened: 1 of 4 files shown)


backend/api/src/markets.ts
```diff
@@ -35,37 +35,47 @@ const getPublicContracts = async (
     beforeTime?: string
     order?: 'asc' | 'desc'
     userId?: string
+    groupId?: string
   }
 ) => {
-  let q = selectJson(db, 'public_contracts')
-  q = q.order('created_time', {
+  const q = selectJson(db, 'public_contracts')
+  q.order('created_time', {
```
jahooma, line 42: I thought these queries were immutable and returned new instances? But here it looks like it is mutating the query.

web/pages/api/v0/bets.ts
```diff
@@ -1,133 +1,4 @@
...
-      : await getPublicBets(opts)
-    res.setHeader('Cache-Control', 'max-age=15, public')
-    return res.status(200).json(bets)
-  } catch (e) {
-    console.error(`Error while fetching bets: ${e}`)
-    return res.status(500).json({ error: 'Error while fetching bets: ' + e })
-  }
-}
+const handler = nextHandler('bets')
+export default handler
```
jahooma, line 4: If we remove the vercel cache and forward to our api, I think that can actually cost us a lot: vercel will treat each call as non-cached and charge us for each. (Whereas right now it doesn't charge us if it hits the vercel cache.)

I think we need to keep all the same caches as before for all the vercel v0 api.

jahooma (changes requested):
Approved except for change requested from my last comment (add back vercel cache)

PR #2264, opened 2023-12-08
[API] Add Leagues

casesandberg:


jahooma (approved):
LGTM! Thanks Case!

jahooma:
I'll let @sinclairity also review it

sipec:
I'm going to change this to not require the userId

sipec:
also idk about including the `rankSnapshot` and `Primary Key` in the results

casesandberg:
It would be good to know what place you got in the league (I believe thats rank snapshot right?). `Primary key` is not needed.

sipec:
nvm I am dum. I just thought it was always null (but I guess that just means unranked?)

PR #2265, opened 2023-12-08
[API] Expose Leagues Endpoint

casesandberg:
I forgot to expose the api as an endpoint...

### DIFF
--- a/web/pages/api/v0/leagues.ts
+++ b/web/pages/api/v0/leagues.ts
@@ -1,0 +1,4 @@
+import { nextHandler } from 'web/lib/api/handler'
+
+const handler = nextHandler('leagues')
+export default handler


jahooma:
I think it's ok to only expose on the new api URL. Everyone should start using that instead
@sinclairity

jahooma:
My understanding is that this should URL work once we redeploy (doing that now...):

https://api.manifold.markets/leagues?userId=5LZ4LgYuySdL1huCWe7bti02ghx2

casesandberg:
OH! Got it, thats easy. Should I build all my stuff using this new endpoint then?

jahooma:
I guess not everything is migrated over yet, but most of the getters are, like /bets and /markets, which you could start using, yes!

PR #2274, opened 2023-12-12
[API] Allow for Unrestricted Cors On Undocumented Apis

casesandberg:
I was trying to use the newly exposed `search-users` endpoint, however it looks like because its considered undocumented its getting blocked with cors.


sipec:
good idea but I just did it like this 14d19f292a9b instead because its more elegant

PR #2281, opened 2023-12-15
Use vercel edge middleware for apis

sipec:
This allows us to let people stay on manifold.markets/api/v0.

The edge middleware handles all of the fetching from our backend. And edge functions are billed only for calculation time and not time spent waiting for data https://vercel.com/docs/functions/edge-functions/usage-and-pricing. so I thiiiink this should basically not cost us anything

Of course, while this still requires us to move back the the backend routes to match the old routes. And in this PR I'd also need to undelete api/og and some other things, but just this is sufficient to show that it works!


web/lib/api/proxy.ts
```diff
@@ -27,7 +35,7 @@ function getProxiedRequestHeaders(req: NextApiRequest, whitelist: string[]) {
   }
   // mqp: the backend uses this in the cloud armor rules to bypass GCP throttling
   result.append('X-Vercel-Proxy-Secret', process.env.VERCEL_PROXY_SECRET ?? '')
-  result.append('X-Forwarded-For', req.socket.remoteAddress || '')
+  result.append('X-Forwarded-For', '')
```
mqp, line 38: If you deploy something like this please be careful not to leave this broken, this would break the GCP throttling rules.

sipec:
closing now that #2285 which includes this is merged

PR #2285, opened 2023-12-16
Add back old api paths

sipec:
meta: not a fan of how the commit order looks on github. I added back the old paths, then merged the edge api middleware branch into this one (to delete all the `pages/api/v0` files) then pushed more commits on top of that.

`save-twitch` is still on the new path, but the rest have been returned


mqp:
Very general criticism that isn't specific to this particular change. Your API schema definition is not to my taste.

- The `visibility` is used to infer the path, but that doesn't make sense. `v0` is for versioning. To the extent that we care about versioning at all, calling code should specify what version of the API they are using. How would you infer this if we created a `v1`?
- Why is `authed` in the schema? You should call all our APIs with a credential if you have it, and then the API can do what it likes with that. We don't appear to even be using this parameter so I don't understand why it's there. Having it be a boolean is also kind of suspicious because there are many knobs that we may wish to turn with auth in our API, e.g. allowing only JWT auth for a particular endpoint, allowing only keys with particular permissions, etc.
- Having a single `props` which is deciding to check either the query params or the request body depending on the method is too magical and also too restrictive. There should be both a `queryParamsSchema` and `bodySchema`. (Why is it called `props`?)
- `returns` being specialized to JSON I am actually OK with, since it's reasonable for our API to 100% return JSON (barring random HTTP endpoints on our backend that are like webhooks and stuff and not in any sense "part of the API".)
- `cache` is ridiculous. Endpoints will not typically always respond with the same cache headers, they will want different caching headers depending on the code path. The fact that many of our endpoints always respond with the same one is laziness, not something you should enshrine into an abstraction; for example it would be normal to return more lenient cache durations (or no cache headers at all) when no results come back from a query to the API, under the thesis that the fact that someone is asking suggests that probably some results might appear there soon. Cache headers should be clearly written in the API code, they definitely should not be tucked away in this file.

In general, it looks like you thought that a good property for the API endpoints is that they shouldn't be handling raw HTTP `Request` and `Response` objects, which I think is misguided -- why should they not? Those are already a carefully designed abstraction that is well suited for this purpose. So I would undo a lot of the decisions that you made to facilitate that.

You can feel free to merge this but I care enough about this that I will spend a couple of hours later changing everything to be more flexible.

common/src/api/schema.ts
```diff
@@ -99,14 +99,15 @@ export const API = (_apiTypeCheck = {
       })
       .strict(),
   },
-  'cancel-bet': {
+  'bet/cancel/:betId': {
```
mqp, line 102: It's OK to use the Express named parameter syntax for this, I would probably bother to typecheck it though, because the whole thing is getting abstract enough that it might start getting painful to debug any errors.

mqp:
I guess there are not currently any paths with multiple methods, so we didn't have to fix the code to handle that yet. That's fine.

web/middleware.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  if (pathsToSkip.includes(path)) {
+    return NextResponse.next()
+  }
+
+  try {
+    return await fetchBackend(req, path)
+  } catch (err) {
+    console.error('Error talking to cloud function: ', err)
+    return new Response('Error communicating with backend', { status: 500 })
```
mqp, line 15: This should probably be `NextResponse.json({ success: false, error: 'Error communicating with backend' }, { status: 500 })` to be consistent with our normal responses.

mqp (approved):
Cool, I didn't check all the individual paths to make sure they were correct, presumably it was a relatively mechanical process. In general it makes sense. I will just come in later and produce some kind of counter-proposal for how the endpoints should be defined.

PR #2288, opened 2023-12-18
[API] Expose ResolverId on Contract, Add to Answer

casesandberg:
Not sure if `resolverId` was being filtered out intentionally or not. This also saves the resolverId on an answer similar to a contract.


jahooma (approved):
Looks good, thanks!

PR #2289, opened 2023-12-18
Fix bugs with double formatting in some searches

mqp:
If you call pg-promise's formatter multiple times on SQL strings (either directly or implicitly by passing parameters to e.g. `db.map`), you may end up with bugs where the first format introduces formatting language syntax to the second format. For example, in the cases I fixed here, we were doing something like:

```typescript
const where = pg.as.format('where foo = $1', [query]);
const sql = 'select * from bars ' + where + ' order by $1';
const result = await db.many(sql, [sort]);
```
This will throw a template syntax exception given the query string `i have a $5 bill`.

Note that `db.many(sql, [])` is different from `db.many(sql)` or `db.many(sql, null)`. The first invokes the template formatting engine; the second does not.

The best way to handle this is to never write code that runs the template formatting on the same SQL snippets multiple times.

### DIFF
--- a/backend/api/src/supabase-search-contract.ts
+++ b/backend/api/src/supabase-search-contract.ts
@@ -108,7 +108,7 @@ const search = async (
           searchType,
         })
         return pg
-          .map(searchSQL, [], (r) => ({
+          .map(searchSQL, null, (r) => ({
             data: convertContract(r),
             searchType,
           }))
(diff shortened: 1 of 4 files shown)


mqp:
FYI @jahooma because I think you wrote the "SQL builder" stuff and you might want to be aware of this "gotcha."

PR #2294, opened 2023-12-19
Move remaining Vercel API endpoints to edge runtime

mqp:
I think the revalidate one might actually be costing us some money for how long it spends sitting around waiting for contract pages to revalidate and so on? So I guess this way it will likely cost us less.


mqp:
Never mind -- I don't think it's actually possible to make the revalidation endpoint work like this.

PR #2300, opened 2023-12-20
Make daily/weekly/etc stock/crypto price markets non-predictive

chrisjbillington:
e.g "Will BTC close higher on blah date than blah date"

Very good [coverage of the relevant markets](https://manifold.markets/browse?q=%22close%20higher%22&s=score&f=all&ct=ALL) and few (if any) false positives matching on "close higher"

As suggested by Ian on discord

(I have not tested if this actually works, just copying the "daily-coinflip" code)

image


IanPhilips:
Thanks Chris! This makes sense to me so I merged it. Will see what the rest of the team says at standup.

PR #2302, opened 2023-12-20
fix: prefer png over webp for social images

### LINKED ISSUE #2301: Twitter cards don't render the og:image
I think it's because the current meta tag setup is providing webp images. png would be more compatible; it looks like simply subbing `.webp` in the URL with `.png` should do it.

quantizor:
Fixes #2301


mqp:
Based on a cursory Google this seems like a plausible hypothesis to me, so I am inclined to merge it after the holidays.

mqp:
(Only, if this is true, we should modify `buildOgUrl` instead of doing this replace, so it applies to other call sites.)

PR #2305, opened 2023-12-24
Migrate comment creation trigger code to Supabase

mqp:
Just random stuff that's reading comments and bets.


sipec (approved):
the code certainly looks fine

though a firebase trigger that does things in supabase feels odd.
we should just call it in the create contract endpoint

mqp:
> though a firebase trigger that does things in supabase feels odd.

I don't really agree that there's anything odd about it, but I agree that we should migrate it away from being a trigger.

IanPhilips:
I have a weak preference for making the create-comment have minimal extra logic in it, and the db trigger to have any other random logic that may take longer, like a feed addition, or status attatchment, etc. With that said, I think it makes sense to turn this into a supabase trigger so that it's easy to test. You can host the api locally, use ngrok to open a tunnel, copy the url provided and set the supabase trigger to trigger your local url in 30s - 1minute. You can add arbitrary urls to the supabase trigger so that this will work even when live.

sipec:
Ok I tested and with a few fixes it does seem to add the notifications correctly

PR #2307, opened 2023-12-26
Make `lastUpdatedTime` on contracts more useful, improve API

mqp:
`lastUpdatedTime` used to mean `max(lastBetTime, lastCommentTime)`. Now it also covers other user-salient edits to contract properties.

In addition, the new `/markets` parameters will allow API users to easily get "fresh" markets from the API without having to constantly poll everything in the universe.

### DIFF
--- a/backend/api/src/markets.ts
+++ b/backend/api/src/markets.ts
@@ -1,83 +1,69 @@
 import { SupabaseClient, createSupabaseClient } from 'shared/supabase/init'
-import { run, selectJson } from 'common/supabase/utils'
+import { Column, Row, run, selectJson } from 'common/supabase/utils'
 import { toLiteMarket } from 'common/api/market-types'
 import { APIError, type APIHandler } from './helpers'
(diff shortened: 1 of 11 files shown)


KingSupernova31:
Copying my comment from Discord: A few other things that I think it should cover, in decreasing order of importance:

* The addition of market liquidity via subsidy.
* The addition of an answer in a multiple-choice market.
* The editing of an answer in a multiple-choice market.
* The "is ranked" toggle being changed.
* The bounty in a bounty market being added to.
* A bounty being paid out in a bounty market.
* The "is subsidized" toggle being changed.
* The "is public" toggle being changed.
* Someone voting in a poll.
* A multiple choice market being toggled between "creator only" and "anyone can add answers".
* The addition or editing of a market graph annotation.

Only the first four would be useful for me at present, but I mention the others since in theory I'd expect "last updated time" to include any change at all to the market's data.

PR #2316, opened 2024-01-03
fix duplication of other when it is bad

bgila:
tested - http://localhost:3000/ManifoldPolitics/which-party-will-win-the-us-preside-26a1eb1b8ce6


bgila:
tested on prod https://prod-13ljxcq4d-mantic.vercel.app/ManifoldPolitics/which-party-will-win-the-us-preside-98654274ab42

jahooma:
Thanks Barak! Looks good

PR #2332, opened 2024-01-05
Minor fixes to the API docs

KingSupernova31:

PR #2339, opened 2024-01-06
Update api.md

KingSupernova31:
Cleaned up grammar, fixed outdated rate limit explanation, added  explanations of some non-obvious parameters and a mention of the internal API.


mqp:
If you fix the merge conflicts between this and your prior change #2332 I will merge it.

KingSupernova31:
Whoops, sorry. I think I've fixed it. (I'm not very familiar with Github, apologies.)

docs/docs/api.md
```diff
@@ -46,7 +48,7 @@ GET requests with parameters should have the parameters in the query string. POS
 API responses should always either have a body with a JSON result object (if the response was a 200)
 or with a JSON object representing an error (if the response was a 4xx or 5xx).
 
-All times are UNIX timestamps in milliseconds since epoch.
+All times are UNIX timestamps in milliseconds since epoch. (Javascript timestamps.)
```
mqp, line 51: There's no such thing as a "Javascript timestamp."

docs/docs/api.md
```diff
@@ -1042,6 +1041,18 @@ Parameters:
...
+Manifold has some internal API endpoints that are not part of the official API. These are largely undocumented, but a few are mentioned here for third-party use until a more permanent solution is implimented. These endpoints are not preceeded by `/v0` and are even more subject to sudden changes than the official API endpoints.
+
+### `POST /unresolve`
+
+Unresolves a market. Requires auth.
+
+Parameters:
+
+- `contractId`: The ID of the market to unresolve.
+
```
mqp, line 1055: Not really jazzed about including this. There's a 0% chance anyone will ever update this if the `unresolve` endpoint changes, and if moderators want this endpoint, we should put it in the `v0` namespace, not tell them to use this one.

mqp:
I like the rest, so I will merge the parts I like and leave the remainder. Thanks for the improvements.

docs/docs/api.md
KingSupernova31: I don't think the terminology is particularly important, so if you don't want that in there, I won't object. But "Javascript timestamp" is meaningful, since it's distinguishable from Unix time by the fact that it's denoted in miliseconds. Unix time is in seconds.

docs/docs/api.md
```diff
@@ -902,25 +901,25 @@ Parameters:
     get bets 11 through 20.
 - `after`: Optional. Include only bets created after the bet with this ID.
   - For example, if you request the 10 most recent bets and then perform a second query with `after=[the id of the 1st bet]`, you will receive up to 10 new bets, if available.
-- `kinds`: Optional. Specifies subsets of bets to return. Possible kinds: `open-limit` (open limit orders.)
+- `kinds`: Optional. Specifies subsets of bets to return. Possible kinds: `open-limit` (open limit orders, including ones on closed and reolved markets).
```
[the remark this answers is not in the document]
KingSupernova31: if you're fixing the bug right now, then sure, no objections. But if your plan is to leave the bug in for now and leave it undocumented, seems bad. If the bug is going to exist anyway, it's strictly better for it to be documented.

docs/docs/api.md
KingSupernova31: Wanna bet against me at 0%? :) Or even 10%? The fact that I've added it to the documentation is strong evidence against your claim.

In general, yes, of course it would be better for this to be a more official, reliable endpoint. Feel free to add a v0 version. But right now there is no such official unresolve endpoint, so it's better for people to be aware of the one that does exist. "Either we document this or we make it better" is a false dichotomy; what's likely to actually happen is that it will be neither fixed nor documented. For the time being while it's not fixed, being documented seems better than not being documented.

mqp:
OK, you know what, I am too lazy to argue about it so I will just merge your version and change it later if I feel like it.

PR #2355, opened 2024-01-09
[WEB] Skip Market Type on Duplicate Market

casesandberg:
URL params are coming in after the page has rendered and thus required selecting the market type on duplicate market, even though its already in the params. This useEffect keeps local state in sync with the params coming in, allowing users to skip selecting market type if its provided.


jahooma (approved):
LGTM. Thanks!

PR #2358, opened 2024-01-10
Add rewrite rule to redirect to new api

jahooma:


mqp:
Please explain - what is the motivation? Is this different than the existing approach? Does it pass through an `X-Forwarded-For` IP correctly?

jahooma:
I think we don't get charged for execution time of the middleware. That's all.

> Does it pass through an X-Forwarded-For IP correctly?
I have no idea 😂

mqp:
Then please don't merge this until you have verified that the Google Cloud rate limiting will work with requests passed through this system. If all Vercel users were collectively limited to 1000 req/min because GCP wasn't able to distinguish between them, that would break people's existing integrations.

jahooma:
This might work, but would need to managed by someone who understands the google cloud rate limiting rules.

PR #2363, opened 2024-01-11
Fix line breaks

Calvin-LL:
Example market link: https://manifold.markets/CodeandSolder/will-any-new-states-implement-allma-adb38b008e05

## Before

Screenshot 2024-01-11 at 15 17 21

## After

Screenshot 2024-01-11 at 15 17 46

### DIFF
--- a/love/tailwind.config.js
+++ b/love/tailwind.config.js
@@ -408,6 +408,7 @@ module.exports = {
           },
         },
         '.break-anywhere': {
+          'line-break': 'anywhere',
           'overflow-wrap': 'anywhere',
           'word-break': 'break-word', // for Safari
         },
--- a/politics/tailwind.config.js
+++ b/politics/tailwind.config.js
@@ -437,6 +437,7 @@ module.exports = {
           },
         },
         '.break-anywhere': {
+          'line-break': 'anywhere',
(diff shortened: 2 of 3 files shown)


sipec:
Hmm this class is used in a lot of places and I'm wary of breaking mid-word in full generality. But thanks for point out this particular css bug.

Calvin-LL:
> Hmm this class is used in a lot of places and I think breaking mid wo
> rd is often bad. But thanks for point out this particular css bug.

The existing styles already break words. See MDN page for (line-break)[https://developer.mozilla.org/en-US/docs/Web/CSS/line-break] and (word-break)[https://developer.mozilla.org/en-US/docs/Web/CSS/word-break]

sipec:
ok I've fixed this particular bug (roughly) the way you have suggested, by adding word break opportunities at the start of the dots.

and you're right about what the class is supposed to do 😅 
I'm still skeptical after reading mdn though. the default behavior of `overflow-wrap: anywhere` to try to wrap at whitespace first is actually good. Less importantly, breaking at `&nbsp;` is pretty unintuitive and breaking `&zwj;` can split zwj sequence emojis like ❤️‍🔥 (maybe? not sure).

PR #2364, opened 2024-01-12
Fix dark theme

Calvin-LL:
There are currently two problems with dark theme. This PR fixes both of them.

## Problem 1

When a user is using dark theme on their device and opens the website for the first time, the website show light theme despite showing "Auto" in the side bar.

You can reproduce by deleting site data and reloading the site.

The device is set to dark theme in this recording:

https://github.com/manifoldmarkets/manifold/assets/8357970/f84e1467-a9da-4a2e-9609-86619146512d

You can see in this recording, only after clicking the theme toggle button 3 times and cycling back to "Auto" does the website show dark theme.

This is caused by the `useTheme` hook setting the initial value of `localStorage.getItem('theme')` to `"loading"` but `init-theme.js` not recognizing that value. This is fixed by checking for that value and treating it as `"auto"`, just like in the side bar:

https://github.com/manifoldmarkets/manifold/blob/e2e6048bc2d0c9f4f68c32cf1c7e3bf5474e28fe/web/components/nav/sidebar.tsx#L187

## Problem 2

When the theme is set to "Auto", the device is set to dark theme, and the network is slow, the site flashes light theme briefly. This has been annoying me when I use the site in bed at night.

You can reproduce by going to the Chrome DevTools's Network tab and switch "No throttling" to "Fast 3G".

In this recording you can see the site briefly flashing light theme before switching to dark theme.

https://github.com/manifoldmarkets/manifold/assets/8357970/e629186d-c3f2-4007-b313-0b6c606f395a

This is solved by adding a `no-js` class to the html element and removing it once the `init-theme.js` runs. A styling rule is added to `globals.css` to style the site in dark theme if the device is in dark theme and the `no-js` class is present. Once the `init-theme.js` runs, the `no-js` class is removed and the theme is set according to user setting.

### DIFF
--- a/love/pages/_document.tsx
+++ b/love/pages/_document.tsx
@@ -4,7 +4,7 @@ import Script from 'next/script'
 
 export default function Document() {
   return (
-    <Html lang="en">
+    <Html lang="en" className="no-js">
       <Head>
         <link rel="icon" href={ENV_CONFIG.faviconPath} />
         <Script src="/init-theme.js" strategy="beforeInteractive" />
--- a/love/public/init-theme.js
+++ b/love/public/init-theme.js
@@ -4,7 +4,12 @@
   const localTheme = localStorage.getItem('theme')
   const theme = localTheme ? JSON.parse(localTheme) : 'auto'
 
-  if (theme === 'dark' || (theme === 'auto' && autoDark)) {
+  document.documentElement.classList.remove('no-js')
+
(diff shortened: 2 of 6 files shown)


sipec:
thanks for looking into this. I'll try to review this today

jahooma:
Seems reasonable, merging. Thanks for the fix!

PR #2366, opened 2024-01-13
Label bet buttons with new amounts instead of ++, +, -, --

jahooma:
This is the craziest thing. It's cool.


jahooma:
It was not intuitive haha

PR #2369, opened 2024-01-15
Migrate reactions to supabase

sipec:
- adds columns for everything important about reactions
  - no longer denormalize name, username, avatar. on client, fetch on hover 
- move notification-specific stuff into helper function that's used to create the notification
- remove "reaction type". later, we could add a column `type` and set every row to "like" - this would be a trivial.

I've basically done the entire migration in this PR; but we could split it up into add columns & replicator, move reads, and move writes. I'm inclined not to because downtime / stateful bugs of likes don't matter much


sipec:
I haven't tested the new index, and not sure if we really need it

web/components/profile/user-liked-contracts-button.tsx
```diff
@@ -1,103 +0,0 @@
-import { User } from 'common/user'
```
IanPhilips, line 1: I predict users will complain if you remove this

IanPhilips (commented):
Looks great! Thanks for migrating these. The only thing I have to say is that I tried to remove the user likes button on the profile page in the past and many users complained. So I wouldn't recommend removing it.

PR #2374, opened 2024-01-16
Migrate manalinks to supabase

sipec:
Need to test


backend/api/src/claim-manalink.ts
```diff
@@ -1,43 +1,54 @@
...
     }
-    const manalink = manalinkSnap.data() as Manalink
 
-    const { amount, fromId, claimedUserIds } = manalink
+    const { amount, creator_id, expires_time, max_uses } = manalink
 
-    if (amount <= 0 || isNaN(amount) || !isFinite(amount))
-      throw new APIError(500, 'Invalid amount')
+    const claimedUserIds = await pg.map(
+      `select txn.toId from manalink_claims
```
mqp, line 38: `txns` doesn't have native columns yet.

firestore.rules
```diff
@@ -163,14 +163,6 @@ service cloud.firestore {
...
     }
 
-    // Note: `resource` = existing doc, `request.resource` = incoming doc
-    match /manalinks/{slug} {
-      // Only you can create a manalink with your fromId
-      allow create: if request.auth.uid == request.resource.data.fromId;
-      // Only you can list and change your own manalinks
-      allow list, update: if request.auth.uid == resource.data.fromId;
-    }
-
```
mqp, line 173: Seems a little fishy to allow listing the old manalinks that used to be private which are still hanging around, unless you are also deleting all the old Firestore data (which would be reasonable, but seems a bit aggressive from a "what if I fucked up some part of the migration" perspective.)

sipec: deleting the rule lets anyone access it?

sipec: okay I removed the "cleanup" stuff from this

PR #2381, opened 2024-01-17
Migrate weekly updates to supabase

sipec:
need to test
we don't need to backfill the old weekly updates. do we?

Resolves MAN-1408


sipec:
wait what am I talking about. I should just merge it and see if it breaks

IanPhilips (approved):
Looks good to me! It's been a while since I worked on this, so if there's any bugs we can just fix them as they come up. It's a simple job, theoretically.

PR #2382, opened 2024-01-17
Fix Idaho market

Bohaska:
The slug was wrong and set to Indiana


Bohaska:
Manifold already fixed this

PR #2395, opened 2024-01-20
Switch backend code to connect to Supavisor pooler

mqp:
Per Supabase we need to switch to this in the next week because the old pgbouncer URL (`db.instance.supabase.co`) is getting shut off.

See https://supabase.com/blog/supavisor-postgres-connection-pooler for some marketing material.

Should be no big deal.


sipec (approved):

jahooma:
Don't we need this by tomorrow? Should merge soon. @mqp

PR #2406, opened 2024-01-24
Show number of followers when adding a topic

Bohaska:
mattyb said in discord:

> when there’s 4 “Apple”s, limiting to 5 feels way too low
> if tags were clearly organized, and lacking dupes, this would be fine (imo)
> but i can’t tell which Apple I’m using, and thus need them all.
> 
> If there was “Apple (30 Markets)” and “Apple (350 markets)” this would also be better

We don't have market stats, but we do have follower stats.


jahooma (approved):
Nice, seems like a straight improvement!

PR #2408, opened 2024-01-24
New user flow with corrected stats

IanPhilips:
@jahooma would you take a look at the new `update-stats` code? Well, at least glance at it. I tested it and it look like it works but lmk if I misunderstood something about the nd1, etc.

Screenshot 2024-01-24 at 7 57 43 PM
Screenshot 2024-01-24 at 7 57 54 PM


backend/api/src/get-singup-bonus.ts
```diff
@@ -0,0 +1,25 @@
+import { APIError, APIHandler } from 'api/helpers/endpoint'
+import { getUser } from 'shared/utils'
+import { MARKET_VISIT_BONUS_TOTAL } from 'common/economy'
+import { sendOnboardingMarketVisitBonus } from 'shared/onboarding-helpers'
+import * as admin from 'firebase-admin'
+
+export const getSingupBonus: APIHandler<'get-signup-bonus'> = async (
```
jahooma, line 7: Typo, plus I don't think it should start with 'get' if we are going to be using endpoints for just fetching data. Maybe 'claim-signup-bonus'?

backend/scheduler/src/jobs/update-stats.ts
```diff
@@ -407,8 +430,14 @@ export const updateStatsCore = async () => {
     const end = i + 1
     return average(dailyActivationRate.slice(start, end))
   })
+  const newUserBetAverage = dailyNewRealUsers.map((newUsers) => {
```
jahooma, line 433: Seems right

jahooma:
- Maybe "+M100" for visiting a new question (with "+")?
- The Earn up to 500 could be right below the Related questions header? Doesn't it sound a little bit salesy to be earn up to? Instead of earn 100 per market visited.

IanPhilips:
> * The Earn up to 500 could be right below the Related questions header? Doesn't it sound a little bit salesy to be earn up to? Instead of earn 100 per market visited.

There's a reason they do that!

PR #2422, opened 2024-01-27
New `user_contract_views` table to track contract views efficiently

mqp:
Our existing `user_seen_markets` table is very large, because it records a new row for every view ever. (That's kind of larger than you would think because we have big contract and user IDs.) With indexes, it's currently 60 GB. As a result it's challenging to query it efficiently. But querying it is important for high-throughput feed generation work, because all the time we want to answer the question "has the guy already seen this thing we are trying to put in their feed."

We can make a way smaller and easier to query version that fulfills all of our needs by recording "last seen time" and "count per view type" per contract and user. `user_contract_views` will be the good version.

There's also a small additional point which is that the old table has a design mistake, which is that we don't put logged-out views into it, but we care about logged-out views. (Instead we end up querying `user_events` for them which sucks.) So we should put those into the new table with `user_id` `null`.

We can migrate like this:

1. This PR adds the new table and an API endpoint for recording views that puts data both into the new and old table. The frontend now calls that API to record views.
2. Run a query to copy the old `user_seen_markets` data prior to this PR into the new table. (Maybe we will be slightly inaccurate here but it's basically fine.)
3. Change the places querying `user_seen_markets` to query the new table instead.
4. Toast the old table.


mqp:
There's a kind of weird bug which already exists, which double-counts views when you hit market pages, because first it calls the tracking code as a logged out user before calling it as a logged-in user. I want to fix this before merging.

sipec:
resolves MAN-1412

jahooma:
Any updates on this @mqp?

mqp:
OK, I am going to get this out the door now.

jahooma:
This is great.

Note: We also are using views to populate the "Recent" section on your portfolio page now. (Sql function get_your_recent_contracts, which queries user_seen_markets)

PR #2426, opened 2024-01-30
[WEB] Track users last seen time

casesandberg:
The mods have been requesting better visibility into inactive users (see [Good User Suggestions](https://manifoldmarkets.notion.site/Good-user-suggestions-86f377ac982a411bb84067d0971bf31a?p=de637cf182714a1ab69b4f7ba86e26ad&pm=s)). This tracks users last seen time so that it can be exposed in user hover cards, profiles, etc...

I am getting a firebase error `Missing or insufficient permissions` so putting up a draft PR to attempt to debug in deployed dev env.

### DIFF
--- a/common/src/api/user-types.ts
+++ b/common/src/api/user-types.ts
@@ -25,6 +25,7 @@ export type LiteUser = {
   balance: number
   totalDeposits: number
   lastBetTime?: number
+  lastSeenTime?: number
   currentBettingStreak?: number
   profitCached: {
     daily: number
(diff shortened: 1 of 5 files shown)


casesandberg:
Ideally the call would happen [alongside the user being identified](https://github.com/manifoldmarkets/manifold/blob/bb2e7389f881afe17127066ee5f877729de416eb/web/components/auth-context.tsx#L165), however getting that same permissions error there as well.

PR #2430, opened 2024-01-30
Add documentation for User schema

Bohaska:
/me returns a User object, not a LiteUser object, which may be confusing for people who try to use the same parsers to parse responses from both /me and /user. This clarifies in the docs that they may return different stuff.


sipec:
obviated by changing it to return LiteUser in 8ba10dbda60bd4c37206358fc9db8e5aa695886d

PR #2432, opened 2024-01-30
[WEB] Show Limit Order Volume on Selected Answer

casesandberg:
Screenshot 2024-01-30 at 1 09 20 PM

Shows limit order volume on selected answers with a button for the Order Book. I chose to use the scale icon but happy to switch it to another if you would prefer!


web/components/answers/answers-panel.tsx
```diff
@@ -497,6 +522,27 @@ function Answer(props: {
                   Edit
                 </Button>
               )}
+
+            {unfilledBets?.length && limitOrderVolume ? (
```
jahooma, line 526: If length is 0, won't it render '0' instead of nothing? Classic React.

web/components/answers/answers-panel.tsx
```diff
@@ -383,6 +394,20 @@ function Answer(props: {
...
   const isMobile = useIsMobile()
 
+  const limitOrderVolume = useMemo(
+    () =>
+      unfilledBets?.reduce((acc, bet) => {
+        acc += bet.orderAmount
+
+        bet.fills.map((fill) => {
+          acc -= fill.amount
+        })
```
jahooma, line 404: I think the sum of the fills is always `bet.amount`. So you could just subtract that.

jahooma (approved):
Nice!! Something we want to do this quarter is improve the limit order UX, so showing the total open amount is great, and reducing the clicks to get to limit order UI is also great.

web/components/answers/answers-panel.tsx
casesandberg: It shouldn't because its a ternary right? If its 0 it will fallback to `null`

web/components/answers/answers-panel.tsx
casesandberg: I was trying to account for partial fills but maybe thats not necessary? I was just trying to get the amount of mana left tied up in limit orders but maybe I am calculating it wrong.

web/components/answers/answers-panel.tsx
jahooma: Oh my bad, you're right, didn't see the ternary

web/components/answers/answers-panel.tsx
jahooma: Yeah, this is right. It's just `bet.orderAmount - bet.amount`

casesandberg: Right on! Just pushed a fix.

jahooma:
Cool. Merging this.

PR #2439, opened 2024-02-01
Stop people from following themselves

Bohaska:
You can follow yourself through your own feed on your own market -> three dot button.

This PR should now return an error if you try to do that.


sipec:
we should also remove it as an option from the menu on the frontend on your own profile!

sipec:
addressed by 477e5d8 so closing

PR #2442, opened 2024-02-02
Fix generateStaticParams for politics pages

IanPhilips:


IanPhilips:
@akrolsmir why don't you use `generateStaticParams` ([documentation](https://nextjs.org/docs/app/building-your-application/caching#generatestaticparams)) in manifund? 
It sounds like it is equivalent to getStaticPaths:

> For [dynamic segments](https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes) (e.g. app/blog/[slug]/page.js), paths provided by generateStaticParams are cached in the Full Route Cache at build time. At request time, Next.js will also cache paths that weren't known at build time the first time they're visited.

But I can't get it to work with a super simple route, see `/example-ssr/[username]/[contractSlug]`

This is the error I get:
```
[Error: An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details. A digest property is included on this error instance which may provide additional details about the nature of the error.] {
  digest: 'DYNAMIC_SERVER_USAGE',
  page: '/example-ssr/ian/stonker'
}
[Error: An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details. A digest property is included on this error instance which may provide additional details about the nature of the error.] {
  digest: 'DYNAMIC_SERVER_USAGE',
  page: '/example-ssr/ian/stonker'
}
Error: Runtime exited with error: exit status 1
Runtime.ExitError
```

PR #2445, opened 2024-02-02
[WEB] Answer Limit Order Bar Charts

casesandberg:
Screenshot 2024-02-01 at 7 47 39 PM

I decided to make these max out at 50% height to reduce amount that can end up behind text. Happy to tweak the weights of the scale as well.

### DIFF
--- a/web/components/answers/answers-panel.tsx
+++ b/web/components/answers/answers-panel.tsx
@@ -49,6 +49,10 @@ import { OrderBookButton } from '../bet/order-book'
 import { useUnfilledBets } from 'web/hooks/use-bets'
 import { Tooltip } from '../widgets/tooltip'
 import { formatMoney, shortFormatNumber } from 'common/util/format'
+import { safeLocalStorage } from 'web/lib/util/local'
+import { useIsClient } from 'web/hooks/use-is-client'
+
+const SHOW_LIMIT_ORDER_CHARTS_KEY = 'SHOW_LIMIT_ORDER_CHARTS_KEY'
 
 // full resorting, hover, clickiness, search and add
 export function AnswersPanel(props: {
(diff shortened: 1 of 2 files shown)


web/components/answers/answers-panel.tsx
```diff
@@ -101,6 +105,20 @@ export function AnswersPanel(props: {
...
+  const handleToggleLimitOrderCharts = () => {
+    const newValue = !shouldShowLimitOrderChart
+
+    setShouldShowLimitOrderChart(newValue)
+    safeLocalStorage?.setItem(
+      SHOW_LIMIT_ORDER_CHARTS_KEY,
+      newValue ? 'true' : 'false'
+    )
+  }
+
```
jahooma, line 121: This looks fine, but could you try using `usePersistentLocalState`:

Example:

```
const [lover, setLover] = usePersistentLocalState<
    Row<'lovers'> | undefined | null
  >(undefined, `lover-${user?.id}`)
  ```

Comment (casesandberg) [reply web/components/answers/answers-panel.tsx]: Right on, I figured I was missing a hook that did that. Updated!

File: web/components/answers/answers-panel.tsx [hunk 2]
```diff
@@ -556,3 +604,49 @@ function Answer(props: {
...
 }
+
+function LimitOrderBarChart({
+  limitOrders,
+  prob,
+  activeColor,
+}: {
+  limitOrders?: Array<LimitBet>
+  prob: number
+  activeColor: string
```
jahooma, line 615: I realize this is a more concise way of declaring and unpacking props. Maybe we should do it this way everywhere

web/components/answers/answers-panel.tsx
```diff
@@ -556,3 +604,49 @@ function Answer(props: {
...
+
+  return (
+    <div className="pointer-events-none absolute inset-0 overflow-hidden">
+      {Object.entries(limitOrdersByProb).map(([limitProb, bets]) => {
+        const numericLimitProb = Number(limitProb)
+        const volume = sumBy(bets, (bet) => bet.orderAmount - bet.amount)
+
+        const logVolume = Math.log(volume)
+        const logMaxOfVolume = Math.log(40000)
+        const scaledResult = (logVolume / logMaxOfVolume) * 50 // Only fill up max 50% of the height
```
jahooma, line 630: Seems reasonable!

jahooma (approved):
Nice work, Case! Let's see how the users like it.

PR #2446, opened 2024-02-02
Migrate txns to supabase

sipec:
need to test

Resolves MAN-1414


backend/shared/src/txn/run-txn.ts
```diff
@@ -1,149 +1,195 @@
...
+        txn.amount,
+        txn.fromId,
+        txn.toId,
+        txn.fromType,
+        txn.toType,
+        txn.category,
+      ]
+    )
+    .catch((e) => {
+      logFailedToRecordTxn(txn)
```
jahooma, line 158: From discussion: Would be good to trigger these in dev and search the logs for them

backend/shared/src/onboarding-helpers.ts
```diff
@@ -105,17 +108,22 @@ const sendNextDayManaBonus = async (
...
     }
-    const manaBonusTxn = await runTxnFromBank(transaction, signupBonusTxn)
-    if (manaBonusTxn.status != 'error' && manaBonusTxn.txn) {
-      transaction.update(toDoc, {
-        manaBonusSent: true,
-        weeklyTrendingEmailSent: true, // not yet, but about to!
+
+    const manaBonusTxn = await pg
+      .tx((tx) => {
+        return runTxnFromBank(tx, signupBonusTxn)
```
jahooma, line 114: Seems like this nests the firestore transactions. That would break?

jahooma: Ok, this does a firestore write not a txn.

jahooma: [c] Should test creating a user to make sure this works [J] Yes

backend/api/src/boost-market.ts
```diff
@@ -41,57 +39,39 @@ export const boostmarket = authEndpoint(async (req, auth) => {
...
-  } catch (e) {
-    log('error adding txn! reversing funds to ad.')
 
-    await pg.none(
-      `update market_ads
-      set funds = market_ads.funds - $1
-      where id = $2`,
-      [totalCost, id]
-    )
-    log(`done subtracting ${totalCost} from ad ${id}`)
```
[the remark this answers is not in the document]
jahooma: [c] this is to put it back if there was an error. But now it's in a tx

jahooma (approved):
Looks good! Thanks Sinclair.

Monitoring for errors is very important after this.

Would be good to keep tabs on any errors. We should have a plan for what exactly to query for.
[c] Could create a dashboard from these logs.
- If it fails in firebase transaction, will roll back correctly
- If a supabase part fails, it will throw an error
- If firestore succeeds and supabase fails, we do log if the txn insert failed. [c] Can check if there are other postgres things after firestore

One idea: Wrap pg.tx in our runTxn code. Allow the caller to pass a pre-query (function that takes tx) that runs before txn is created.

[c] This code might not be in this state for that long. Once we migrate users then balance can be in the pg.tx transaction.

backend/shared/src/onboarding-helpers.ts
sipec: it did break but now it's all un-nested

sipec:
TXN Migration plan
- Merge and wait for frontend to deploy
- Alter table id default, created_time default not null, fs_updated time drop not null
- Recreate txn_populate_cols (don’t populate created time)
- Deploy API
- Redeploy logCollections, onUpdatePrivateUser
- Wait for firebase writes to finish
- Alter table add not nulls for other columns

PR #2447, opened 2024-02-02
Ingawei/add graph

ingawei:
Added this graph but probably did it in a gross way:
image

Separated grabbing elections page props sharing between `web` and `politics`, in `get-home-props`

Made `getChartParams` a separate thing because I originally wanted to make a server side chart component that could grab it's own points, but that didn't work and I couldn't figure out why

example from [next js site](https://nextjs.org/docs/app/building-your-application/data-fetching/patterns):
```
async function Playlists({ artistID }: { artistID: string }) {
  // Wait for the playlists
  const playlists = await getArtistPlaylists(artistID)
 
  return (
    <ul>
      {playlists.map((playlist) => (
        <li key={playlist.id}>{playlist.name}</li>
      ))}
    </ul>
  )
}
 
export default async function Page({
  params: { username },
}: {
  params: { username: string }
}) {
  // Wait for the artist
  const artist = await getArtist(username)
 
  return (
    <>
      <h1>{artist.name}</h1>
      <Suspense fallback={<div>Loading...</div>}>
        <Playlists artistID={artist.id} />
      </Suspense>
    </>
  )
}
```
Also did something gross in `politics-card.tsx` where I take the graph out of the ClickFrame because hover and clicking wasn't responsive within it

Also ported over a bunch of graph objects from web so I can start styling those over time


politics/lib/get-home-props.ts
```diff
@@ -0,0 +1,167 @@
...
+  ChartAnnotation,
+  getChartAnnotations,
+} from 'common/supabase/chart-annotations'
+import { getContractFromSlug } from 'common/supabase/contracts'
+import { SupabaseClient } from 'common/supabase/utils'
+import { unstable_cache } from 'next/cache'
+import { initSupabaseAdmin } from 'web/lib/supabase/admin-db'
+export const REVALIDATE_CONTRACTS_SECONDS = 60
+
+export async function getHomeProps(useUnstableCache: boolean) {
```
IanPhilips, line 23: Only the politics site uses `useUnstableCache`, so you can eliminate that option from both where you got this code, and where it lives now.

politics/lib/get-home-props.ts
```diff
@@ -0,0 +1,167 @@
...
+    newHampshireContract,
+    republicanVPContract,
+    democraticVPContract,
+  ] = await Promise.all(contractsPromises)
+
+  const linkPreviews = await fetchLinkPreviews([NH_LINK])
+
+  let historyDataProps = undefined
+  let chartAnnotationsProps: ChartAnnotation[] = []
+  if (!!electionPartyContract) {
```
IanPhilips, line 71: If we don't have the `electionPartyContract` we should just throw an error, right? No need to use mutable variables here

IanPhilips (commented):
Looks good to me! I'm probably missing the context here, but it seems(?) like you could duplicate fewer components by passing some classes/colors down, but it's not a big deal.

PR #2449, opened 2024-02-05
Add request-loan to internal API docs

blackdiamand:


sipec:
no longer necessary now that getting new loans is no longer supported

PR #2456, opened 2024-02-07
Move all writes of private user to the backend

sipec:
precursor to migrating private users to supa

note this breaks admin/journeys


jahooma:
Cool, looks pretty good. We are relegating private user to the server so that we can remove firebase from the client?

Or is it easier now to completely migrate to supabase and use RLS on the private-user table. Feel free to chat with me to explain.

sipec:
ok I'm slashing scope of this PR to take out the auth context stuff, because it's complicated. I'm certain what remains works.

PR #2463, opened 2024-02-12
Show fractional shares for advanced traders

Bohaska:
This shows fractional shares to 2 decimal places if advanced trader mode is on.

Inspired by Eliza's discord suggestion "Displaying fractional mana/shares can teach people that small bets at 5% or 95% are still meaningful"


Bohaska:
wait a second, that's not right...
image

PR #2464, opened 2024-02-13
Cache user display info on client

sipec:
not all usages of Avatar are refactored to use the id, but most are

resolves MAN-1407


web/hooks/use-user.ts
```diff
@@ -34,6 +34,102 @@ export const usePrefetchUsers = (userIds: string[]) => {
...
+export type DisplayUser = {
+  id: string
+  name: string
+  username: string
+  avatarUrl?: string
+  createdTime?: number
+  isBannedFromPosting?: boolean
+}
+
+const cache: {
```
casesandberg, line 46: I might consider using react's state management (like context) for this since global variables don't inherently trigger re-renders and can't leverage modern concurrency features like automatic batching.

jahooma: That sounds like a good suggestion!

How would you get it to only re-render components that use the users that changed? E.g. if a component uses `useDisplayUser(userId)` it should only re-render if that user object is updated in the cache.

I'm not super familiar with react context.

casesandberg: Similar to how the cache is narrowed down here, you can create a hook that does the same:

```
function useDisplayUser(userId: string) {
  const { cache, setCache } = useContext(UserCacheContext);

  ...

  return cache[userId]
}
```

jahooma: I see. Maybe there's something I'm missing, but it seems like calling `setCache` would trigger a re-render in every component that uses `useDisplayUser`, regardless of the userId. Even memoizing the return value would only help if passed as props through children components that are memoized.

sipec: hmm I think I fundamentally disagree with react context for this because the user information is async server state (comes from database) rather than ui state (changes based on user input). it all has to pass through a `useEffect` anyways so it's conceptually simpler if the cache lives in vanilla rather than in react

on the other hand, we already have these nice hooks for storing stuff in localstorage. hmm...

sipec:
resolved by 1c52cba1d6d2342d025820c75f2744c5f1e17f1f

PR #2465, opened 2024-02-13
Change calibration calculation to exclude implicit redemption sales in addition to explicit sales

jacyanthis:


jacyanthis:
This PR reflects what seems to be consensus on how calibration should be calculated in [the Discord thread](https://discord.com/channels/915138780216823849/1192249627790028945).

akrolsmir:
Requesting @mantikoros to take a look, as the person who did the calibration calculations!

web/pages/[username]/calibration.tsx
```diff
@@ -205,8 +205,27 @@ const getCalibrationPoints = (betsData: [Contract, LimitBet[]][]) => {
     if (resolution !== 'YES' && resolution !== 'NO') continue
     const resolvedYES = resolution === 'YES'
 
-    for (const bet of bets as Bet[]) {
-      if (bet.amount < 0) continue // skip sales
+    let currentPosition = 0
+
+    // bets is reversed in place to track user current position
+    for (const bet of bets.reverse() as Bet[]) {
```
Comment (mantikoros) [inline web/pages/[username]/calibration.tsx:211]: .reverse() is a mutable method, which could lead to subtle errors in the future. Prefer .slice().reverse()

web/pages/[username]/calibration.tsx
```diff
@@ -205,8 +205,27 @@ const getCalibrationPoints = (betsData: [Contract, LimitBet[]][]) => {
...
+      const betSign = bet.outcome === 'YES' ? 1 : -1
+      const nextPosition = currentPosition + bet.amount * betSign
```
Comment (mantikoros) [inline web/pages/[username]/calibration.tsx:213]: Position calculation is wrong. It should use shares, not mana amounts (which is what bet.amount is)

web/pages/[username]/calibration.tsx
```diff
@@ -205,8 +205,27 @@ const getCalibrationPoints = (betsData: [Contract, LimitBet[]][]) => {
...
+      if (bet.amount < 0 || ((Math.sign(currentPosition) !== betSign) && (Math.abs(currentPosition) >= Math.abs(bet.amount)))) {
+        currentPosition = nextPosition  // update position before continuing
+        continue
+      }
+
+      // set bet weight, adjusted for any partial implicit sale
+      if (Math.sign(currentPosition) !== betSign && Math.abs(currentPosition) < Math.abs(bet.amount)) {
+        const w = nextPosition
+      } else {
+        const w = bet.amount
```
Comment (mantikoros) [inline web/pages/[username]/calibration.tsx:225]: should be consistent and use # of shares

mantikoros (changes requested):
The position calculation should use shares, not mana amounts.

jacyanthis:
Thanks, @mantikoros! I replaced `bet.reverse()` with out-of-place `bet.slice().reverse()` and `bet.amount` with `bet.shares` (assuming that means what I think it means; excluding the `bet.amount < 0` preexisting explicit sale check) as well as an unrelated fix to the incorrect scope of `w`. Of course it should still be tested by someone who can do that.

web/pages/[username]/calibration.tsx
```diff
@@ -205,8 +205,26 @@ const getCalibrationPoints = (betsData: [Contract, LimitBet[]][]) => {
...
+      // skip explicit and exclusive implicit sales
+      if (bet.amount < 0 || ((Math.sign(currentPosition) !== betSign) && (Math.abs(currentPosition) >= Math.abs(bet.shares)))) {
+        currentPosition = nextPosition  // update position before continuing
+        continue
+      }
+
+      // set bet weight, adjusted for any partial implicit sale
+      let w = bet.shares
+      if (Math.sign(currentPosition) !== betSign && Math.abs(currentPosition) < Math.abs(bet.shares)) {
+        w = nextPosition
```
Comment (mantikoros) [inline web/pages/[username]/calibration.tsx:224]: This should be the abs value of next pos; the weight should always be positive.

Other than that, it looks good. I'll make this fix and merge it in...soon...

mantikoros (approved):
Thanks for the contribution!

PR #2467, opened 2024-02-15
Fixed backend writes

sipec:
fixed version of #2456 
let's deploy to dev and test first 😅


jahooma:
Reported in this thread (https://discord.com/channels/915138780216823849/1208291219483004960): users are having trouble signing up. Seems like it could be related to this change, so I'm going to revert it to be cautious

PR #2469, opened 2024-02-17
[WEB] Hovercards Everywhere

casesandberg:
This PR adds hovercards to all places where Avatars, UserLinks or Avatar + UserLink combos exist. It also fixes a bug where the hovering the avatar would sometimes not trigger the hovercard, and adds the date joined.

### DIFF
--- a/web/components/user/user-hovercard.tsx
+++ b/web/components/user/user-hovercard.tsx
@@ -44,65 +44,61 @@ function FetchUserHovercardContent({ userId }: { userId: string }) {
   const followerIds = useFollowers(userId)
   const isMod = useAdminOrTrusted()
 
-  return (
-    user && (
-      <HoverCard.Content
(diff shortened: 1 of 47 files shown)


web/components/user/user-hovercard.tsx
jahooma, line 1: Only thing I'm worried about is that the HoverCard fetches a lot of data without anyone actually doing the hover: (useFollows, useFollowers, getFullUserById). This could be bad for pages which have a lot of these components (feed, comments, browse)

Could you make the data fetching (and excess divs) only happen if the hover is triggered?

casesandberg: It does that already! `FetchUserHovercardContent` only renders after the hover is triggered, which then kicks off all the data fetching.

jahooma: Oh gotcha! Nice

sipec:
I think it's not good to wrap avatar and username with an extra component. would have preferred the hover card just be added to the avatar and username components themselves, with a prop to disable hover effect.

PR #2471, opened 2024-02-17
[WEB] Custom answer colors

casesandberg:
This PR adds a color picker for custom answer colors. It also adjusts the existing minibar to account for color overrides so they match (currently it was out of sync with even existing answer colors).

Screenshot 2024-02-16 at 6 06 19 PM


YairRand:
Might make sense to make the colors in the selector match the default set? (`colors` param in react-colors' `CirclePicker`.) Some of the colors might seem out of place otherwise.

sipec (approved):

casesandberg:
Sounds good! I believe these are just the base material design colors so I trust google design, although I believe 4 of them are a bit close to poor contrast for my liking. 

This should be good to merge and I will follow up with an additional PR adding API editing of colors and will include a contrast check on the color to invert the text color if need be.

PR #2473, opened 2024-02-20
Greatly improve backend logging infrastructure

mqp:
Previously, we had four different kinds of logging in the backend:

- One was raw `console.log` statements -- usually in common code, if present.
- One was code using the `log` function in `shared/utils`. This would just call `console.log` with a prepended timestamp.
- One was code using the `gLog` object's methods in `shared/utils`. These would log structured JSON with a severity level, formatted the way GCP likes it. In addition, this output could be piped through a special script Sinclair wrote to format it nicer for a TTY.
- Finally, some code would pass down a decorated `gLog` method that included context data inside its JSON, like an endpoint path and trace ID, and its call tree would use that.

Now, all of the last three are taken care of via the `log` object in `shared/log` (also reexported from `shared/utils`). In all those cases, the code in `shared/log` will do all of the useful parts of each. It will produce a nice console-formatted log if you are in local dev, and a GCP structured log in production. You can specify a severity, or not. (`console.log` still just does `console.log`.)

Global context for the `log` object can be provided using the `withLogContext` function, which uses Node's [`AsyncLocalStorage`](https://nodejs.org/api/async_context.html#asynchronous-context-tracking) data structure to maintain state down a call tree, including async code. This makes it suitable for e.g. storing request metadata over the lifetime of an Express request handler, or storing job metadata over the lifetime of a scheduler job, which we now do.

Since this is a large change, I did the minimum necessary to get it implemented, and didn't remove all of the `log` and `logError` objects that we're manually passing around (they all just end up getting `log.info` and `log.error`.) Once it's live for a little bit and we can see that the `AsyncLocalStorage` context implementation is working well in production, then we can go clean everything up.

Incidentally, we could easily adapt this so that we're using it in `common` code too, we would just have to be careful to make a non-Node-specific code path.


mqp:
FYI @IanPhilips since you have thought about this logging stuff a bunch.

jahooma:
Nice!

PR #2474, opened 2024-02-20
Migrate backend user reads

sipec:


backend/shared/src/create-notification.ts
```diff
@@ -981,17 +982,16 @@ export const createUniqueBettorBonusNotification = async (
...
-        )
-      )
+
+  const lastBettorIds = uniqueBettorsExcludingCreator.slice(
+    uniqueBettorsExcludingCreator.length - TOTAL_NEW_BETTORS_TO_REPORT,
+    uniqueBettorsExcludingCreator.length
+  )
+
+  const mostRecentUniqueBettors = await pg.map(
+    `select * from users where id in $1`,
```
IanPhilips, line 992: does that syntax work? I thought it had to be of the type:
`id = ANY ($1)`

backend/shared/src/helpers/generate-and-update-avatar-urls.ts
```diff
@@ -1,27 +1,8 @@
...
-export const generateAndUpdateAvatarUrls = async (users: User[]) => {
-  const storage = getStorage()
-  const bucket = storage.bucket()
-  console.log('bucket name', bucket.name)
-
-  await Promise.all(
-    users.map(async (user) => {
-      const userDoc = firestore.collection('users').doc(user.id)
-      console.log('backfilling user avatar:', user.id)
-      const avatarUrl = await generateAvatarUrl(user.id, user.name, bucket)
```
IanPhilips, line 19: Why remove this? It's nice to have an auto-generated avatar url. And remove an optional field

IanPhilips (commented):
Looks good! Had 2 notes

backend/shared/src/create-notification.ts
mqp: https://www.postgresql.org/docs/current/functions-comparisons.html#FUNCTIONS-COMPARISONS-IN-SCALAR

`in` and `any` are similar. `any` is specifically for Postgres arrays and `in` is for lists of arbitrary expressions.

backend/shared/src/helpers/generate-and-update-avatar-urls.ts
sipec: this particular code isn't used anywhere. maybe was part of a script?
anyways it reads from the users collection and I don't really feel like migrating or testing it

PR #2477, opened 2024-02-22
Work on pagination stuff, implement pagination of Supabase-sourced LP docs in contract page

mqp:
***WIP -- want to hook the LP docs up to Supabase realtime before merging (right now it just loads them once.)***

This PR has two parts:

1. The `usePagination` hook was revamped to be more flexible. Specifically, you can now specify a `prefix` of items that can change after initial load, and they will just keep piling up at the front of the pagination. This means that now it's much more suitable for cases where we are doing a realtime subscription to some stuff that will appear at the front of the list (e.g. contract bets and LPs.)

2. The contract page now uses Supabase to load LP docs and uses the new pagination thingy to get the bets and LP docs into the trades tab.

Incidentally improves feed perf due to not loading all bets eagerly for all feed items to populate the traders modal.


mqp:
I'm working on the realtime stuff, which is related to this contract page work, so I am actually going to merge the pagination changes and leave the rest for a bit later.

mqp:
See https://github.com/manifoldmarkets/manifold/pull/2492.

PR #2478, opened 2024-02-23
Charge no fees when bot bets result in added liquidity

john-h-k:
This was (briefly) discussed on the discord and I thought it was worth making a PR to discuss it further.
The idea of this change is to incentivise market-making and generally make it easier for bots to provide liquidity for markets.

Alternative approaches:
* Instead of no fee, introduce a reduced fee such as 1 mana
* Keep the .25 mana fee but _raise_ the fee for market-taking bets

### DIFF
--- a/backend/functions/src/triggers/on-create-bet.ts
+++ b/backend/functions/src/triggers/on-create-bet.ts
@@ -104,8 +104,9 @@ export const onCreateBet = functions
     if (bet.replyToCommentId)
       await handleBetReplyToComment(bet, contract, bettor, pg)
 
+    const betRemovesLiquidity = bet.isFilled ?? true
     const isApiOrBot = bet.isApi || BOT_USERNAMES.includes(bettor.username)
-    if (isApiOrBot) {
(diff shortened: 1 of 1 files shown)


benmanns:
I would appreciate this! I'm working on a bot and would be happy to provide a lot of liquidity, but the fees force me to predict both P(market resolution) as well as P(order gets filled) in order to justify paying a M0.25 fee for each order.

I also understand wanting to avoid bots making lots and lots of orders, as Firestore is pay per write and even with Supabase, a bunch of writes will require scaling up and paying more. If it were possible to estimate within an order of magnitude what a non-executed limit order would cost (number of writes * cost of writes + estimate of future reads * cost of future reads) and charge that, it would mean I could add a lot of liquidity and hopefully pay a lot less. I'm also happy to pay to cancel orders (same deal, hopefully a low cost based on infrastructure costs) and keep the fee on execution, since it's easy to model profitability on a successfully executed order with fees.

jahooma:
Bot fee was reduced to 0.1 mana. Normal bets have fees now, but not limit order bets

PR #2479, opened 2024-02-23
Numeric multi

IanPhilips:


IanPhilips:
I'm going to merge this early with numeric creation disabled to keep from having to remerge main into it every day

jahooma:
Good idea! Feature flags were a common thing at Google -- you'd ship a feature with a separate way to enable or disable it

PR #2491, opened 2024-02-28
Speed up "prob at time" query 60000 ms -> 400 ms

sipec:
This is one of our heavier queries, [supabase's Query Performance tab](https://supabase.com/dashboard/project/pxidrgkatumlvfqaxcll/reports/query-performance?preset=time) says it takes 59754.82ms on average and taking 5% of total time consumed. it's run 3 times whenever we recalculate contract metrics.

To find the prob for a binary contract or answer at time `t`, I argue that we don't need to get the probAfter of the point right before `t`, just the probBefore of the point after `t`, because we fall back to the current pool probability anyways, which should be correct in the case where there are no bets after `t`

### DIFF
--- a/backend/shared/src/update-contract-metrics-core.ts
+++ b/backend/shared/src/update-contract-metrics-core.ts
@@ -184,34 +184,20 @@ const getCurrentProbs = async (pg: SupabaseDirectClient) => {
   )
 }
 
+// Does not get answers that have had no bets since the given time.
+// You must use the current pool probs for those instead.
 const getBetProbsAt = async (pg: SupabaseDirectClient, when: number) => {
   return Object.fromEntries(
     await pg.map(
-      `with probs_before as (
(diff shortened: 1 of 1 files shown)


jahooma:
Nice!

jahooma:
Seems right to me

mqp:
I don't really want to think really hard about this right now, but one thing I advise if you're not 100% sure about your change, is that you run the metrics update using both the old and new calculation, logging any differences in the resulting metrics, so that you can test empirically that it's correct. That's what I do.

Does the new code handle correctly the case where the contract resolved at some point between T and now? That's another tricky case. For example, if the probability as of the last bet a month ago was 60%, but a week ago the contract resolved to 80%, or something like that.

jahooma:
Bump

sipec:
tested it and it's not quite the same

PR #2497, opened 2024-02-29
Clean up contracts table stuff

mqp:
Just glancing at it briefly I found these handful of contracts table indexes that are either no longer used, unnecessary, or redundant with other indexes on the table.

- Popularity score is no longer used.
- The index on visibility is pointless given that there is also an index with only visibility-public IDs and is of questionable utility anyway.
- The index on unique bettors serves only the partner progress page, which is easy to serve with the existing index on created time instead.
- The index on close time etc. isn't adding value over our existing index on just close time.
- The index on importance score and resolution time isn't adding value over the existing index on importance score.
- The calibration page index isn't relevant to the calibration page (whose query is mostly about getting bets).

### DIFF
--- a/backend/supabase/contracts.sql
+++ b/backend/supabase/contracts.sql
@@ -55,8 +55,6 @@ create index if not exists contracts_unique_bettors on contracts (((data->>'uniq
 
 create index if not exists contracts_close_time on contracts (close_time desc);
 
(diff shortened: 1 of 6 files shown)


jahooma:
Great! Thanks for this cleanup

PR #2500, opened 2024-03-01
Compute partner progress for ≥20 trader markets correctly

syvb:
`getContractsCreatedProgress` checks for markets where `data->>uniqueBettorCount` is ≥20, but it should be checking `data->uniqueBettorCount`.

`->>` casts the count to a string, which results in the unique bettor count being lexicographically compared to "20", when it should be numerically compared to 20. As a result, the query includes all markets where the first digit of the bettor count is 2 to 9. (e.g. bettor counts of 5 and 50 are included, but 1 and 100 aren't).

cc @jahooma


jahooma:
Ahh, wow, good catch!!

PR #2501, opened 2024-03-01
Migrate all reads to `user_seen_markets` to use `user_contract_views`

mqp:
See https://github.com/manifoldmarkets/manifold/pull/2422.

I tested most of these queries and as expected the new version typically is a big perf improvement.

### DIFF
--- a/backend/shared/src/update-contract-views.ts
+++ b/backend/shared/src/update-contract-views.ts
@@ -7,7 +7,7 @@ import { JobContext } from 'shared/utils'
 import { uniq } from 'lodash'
 import { SafeBulkWriter } from 'shared/safe-bulk-writer'
 
-export async function updateContractViews({ log, lastEndTime }: JobContext) {
+export async function updateContractViews({ log }: JobContext) {
   const firestore = admin.firestore()
   const pg = createSupabaseDirectClient()
   log('Loading contract data...')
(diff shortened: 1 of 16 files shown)


mqp:
There's a kind of small deficiency in the new table, which is that it's no longer able to answer the question "find markets where people have done a page view since time T." It can only answer questions like "find markets where people have done a page view, and seen them at all since time T (perhaps as a card.)"

It's my judgment that the queries using that information will still be doing a good job with the new version, but if this bugs us, we could introduce more columns in the `user_contract_views` table for tracking "last card view time", "last page view time", and "last promoted view time" separately instead of just the single "last view time" column.

I think it's OK how it is though.

backend/shared/src/update-contract-views.ts
```diff
@@ -7,7 +7,7 @@ import { JobContext } from 'shared/utils'
 import { uniq } from 'lodash'
 import { SafeBulkWriter } from 'shared/safe-bulk-writer'
 
-export async function updateContractViews({ log, lastEndTime }: JobContext) {
+export async function updateContractViews({ log }: JobContext) {
```
mqp, line 10: I don't think caching this `views` data will be important at all anymore using the new faster table, so in a future PR I will probably remove this job and query it on demand.

mqp:
> There's a kind of small deficiency in the new table, which is that it's no longer able to answer the question "find markets where people have done a page view since time T." It can only answer questions like "find markets where people have done a page view, and also seen them in any way since time T (perhaps as a card.)"
> 
> It's my judgment that the queries using that information will still be doing a good job with the new version, but if this bugs us, we could introduce more columns in the `user_contract_views` table for tracking "last card view time", "last page view time", and "last promoted view time" separately instead of just the single "last view time" column.
> 
> I think it's OK how it is though.

Hmm... I'm actually having second thoughts about this in the context of the new portfolio page. I think it does a disservice to that page.

mqp:
OK, now `user_contract_views` has per-type timestamps, and we can maintain all the old behavior perfectly.

jahooma:
Cool! Glad you found a solution

PR #2509, opened 2024-03-03
Clean up unused stuff from user feed table

mqp:
I just randomly noticed there's a bunch of unnecessary fields here that we don't use anymore. Since `user_feed` population and serving is the single most intense thing in our software let's not make it inefficient for no reason.

The `reason` was always the `reasons[0]`, so there's no need to store it separately.

The others were all just getting `null` and never populated anymore.


mqp:
Seems the deploy strategy for this is

1. Deploy the web stuff that doesn't use the reason anymore by merging this PR.
2. Remove the non-null constraint on reason.
3. 3. Deploy the API, scheduler, and functions to not populate the deleted fields.
4. Alter the table to delete the fields.

PR #2511, opened 2024-03-04
Revert "Revert "Fix dark theme (#2364)""

Calvin-LL:
Not sure why #2364 was reverted. The both dark theme bugs came back after it was reverted.

# Problem 1

Almost exactly as stated in #2364 but was modified a little by https://github.com/manifoldmarkets/manifold/commit/1fcc593c20770353ded81097a9094fbe83d099f9

This video is recorded on an Android emulator.

https://github.com/manifoldmarkets/manifold/assets/8357970/09a6575b-7e45-4d35-a648-047a96cfae99

This is caused by the initial state of `theme` in local storage being set to `"loading"` in

https://github.com/manifoldmarkets/manifold/blob/744253f7f2c7d0c060b5769c713887420f76a5c3/web/hooks/use-theme.ts#L11

which is neither `"dark"` or `"auto"` thus not being set to dark theme in

https://github.com/manifoldmarkets/manifold/blob/744253f7f2c7d0c060b5769c713887420f76a5c3/web/public/init-theme.js#L7

# Problem 2

Exactly as stated in #2364

Even on my gigabit network, on both my Pixel 8 phone and my Android emulator (connected via ethernet), when the site is set to either `"auto"` or `"dark"`, it flashes the light theme on load.

https://github.com/manifoldmarkets/manifold/assets/8357970/a46f8b2b-a86f-4009-8b1d-dbad5719029a

### DIFF
--- a/love/pages/_document.tsx
+++ b/love/pages/_document.tsx
@@ -4,7 +4,7 @@ import Script from 'next/script'
 
 export default function Document() {
   return (
-    <Html lang="en">
+    <Html lang="en" className="no-js">
       <Head>
         <link rel="icon" href={ENV_CONFIG.faviconPath} />
         <Script src="/init-theme.js" strategy="beforeInteractive" />
--- a/love/public/init-theme.js
+++ b/love/public/init-theme.js
@@ -4,7 +4,12 @@
   const localTheme = localStorage.getItem('theme')
   const theme = localTheme ? JSON.parse(localTheme) : 'auto'
 
-  if (theme === 'dark' || (theme === 'auto' && autoDark)) {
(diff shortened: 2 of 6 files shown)


sipec:
I believe James reverted your change to test if it would fix an issue with dark mode (I think the issue was dark theme was always set on by default, but I don't recall exactly.) This did not actually solve the issue.

It should now be fixed by b8031bf9. I fixed problem 1 by defaulting to 'auto' rather than 'loading', and problem 2 via a style tag that just sets background and text color based on media query. I like this better because it is less verbose and works in the case where the global stylesheet arrives later than the html document - very noticable in development at least.

Thanks for looking into this for us!

PR #2516, opened 2024-03-05
Tidy up user comments list code

mqp:
I was interested in doing this because Ian mentioned that he tried to use the `usePagination` stuff here but had a hard time.

I modified `usePagination` to be easier to use for `limit`/`offset` style queries like the comments list API, and fixed some small bugs I noticed (e.g. if you went back and forth from the end of the list it would fire off a duplicate query for the next page.)

The end result is a less buggy comment list (the old one did stuff like flashing "No comments yet" before first load, and wiping out the list in between loads) and a better `usePagination` for adapting to later stuff elsewhere.

By the way, I note that this used to check for being authorized for no obvious reason? You don't need to be authorized to call this API and get comments.

FYI @IanPhilips

### DIFF
--- a/web/components/comments/comments-list.tsx
+++ b/web/components/comments/comments-list.tsx
@@ -1,7 +1,7 @@
+import { useCallback } from 'react'
 import { ContractComment } from 'common/comment'
 import { User } from 'common/user'
 import { groupConsecutive } from 'common/util/array'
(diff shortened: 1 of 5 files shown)


mqp:
Merging this because I want to use the pagination changes in my work.

PR #2521, opened 2024-03-07
Add ability to delete groups with no markets

sipec:
I haven't really thought about _who_ should be able to delete groups. should be same "mod level" as renaming groups probably, but for some reason that is manifold admin / group admin only right now


jahooma:
Cool!

PR #2532, opened 2024-03-15
Added units tests for common/src/supabase/contracts.ts

aidevbot:
### Summary
This pull request introduces unit tests for common/src/supabase/contracts.ts.

### Changes
Added unit tests in common/src/supabase/contracts.ts covering:
- getContracts: Tests for empty input, correct contract retrieval, chunking over 300 IDs, and error handling.
- getContractsByUsers: Verifies correct grouping of contracts by user IDs.
- getAnswersForContracts: Ensures answers are properly grouped by contract ID.
- getContractFromSlug: Confirms null return for non-existent contract slugs.

### Coverage report

| File          | % Stmts | % Branch | % Funcs | % Lines |
|---------------|---------|----------|---------|---------|
| **All files**     | **72.18** | **56.52** | **47.36** | **70.9**  |
| calculate     | 78.57   | 66.66    | 40      | 76.92   |
| uniswap2.ts   | 78.57   | 66.66    | 40      | 76.92   |
| **supabase**     | **71.42** | **55**    | **48.48** | **70.1**  |
| contracts.ts  | 77.94   | 50       | 55.55   | 75      |
| utils.ts      | 62.74   | 58.33    | 40      | 64.44   |

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,6 @@ jobs:
       working-directory: ./common
     - name: Run lint
       run: yarn lint
-      working-directory: ./common
     - name: Run unit tests
       run: yarn test
       working-directory: ./common
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,27 @@
+name: Unit tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


jahooma:
Interesting, thanks for this PR!

Seems like the unit tests lock in the behavior of current code, which could be useful, and potentially catch a regression.

But I'd guess we wouldn't change these functions at all, and if we did, that the tests would be an extra thing to fix after the fact to match the new behavior.

Merging as an experiment though!

PR #2533, opened 2024-03-16
[WEB] Basic Apple Login

casesandberg:
**Note: This change requires enabling Login with Apple for the web app within the Firebase Console before it will work, I don't have access. Hopefully its just a toggle since there is already login with apple for mobile.** 

This PR is intended to be a workaround for users that have created accounts via apple on mobile and are unable to sign in on web. It's a separate page that nothing else links to for now.

### DIFF
--- a/web/lib/firebase/users.ts
+++ b/web/lib/firebase/users.ts
@@ -7,7 +7,12 @@ import {
 } from 'common/user'
 import dayjs from 'dayjs'
 import utc from 'dayjs/plugin/utc'
-import { GoogleAuthProvider, getAuth, signInWithPopup } from 'firebase/auth'
+import {
+  GoogleAuthProvider,
+  OAuthProvider,
+  getAuth,
+  signInWithPopup,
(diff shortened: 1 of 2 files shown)


jahooma:
I'm not sure what the next step is to get Apple login working for web. I thought it might work out of the box.

Screenshot 2024-03-20 at 12 18 35 PM

I don't know what our strategy is for Apple sign ins. I don't think we want to support these on the web, because we prefer one sign in method. It's not obviously good to allow a work-around, because then they will expect Apple sign in to always work on web.

Unless this takes 2 minutes for me, or another team member wants to take this on, I'd be tempted to close the PR.

casesandberg:
@jahooma Yeah I agree. Just trying to help out the poor users who created accounts via Apple on mobile and thus can never use the web app. Maybe a better solution is to just put a disclaimer under the "Sign in with Apple" option on mobile.

jahooma:
Ok, I will give this a shot. Help out those poor apple users

PR #2534, opened 2024-03-17
Show spent mana in positions tab

Bohaska:
This shows the `invested` amount of each user under their shares/profit in the positions tab

### DIFF
--- a/web/components/contract/user-positions-table.tsx
+++ b/web/components/contract/user-positions-table.tsx
@@ -564,7 +564,7 @@ const PositionRow = memo(function PositionRow(props: {
       </UserHovercard>
       <Col>
         <span className={clsx(colorClassName, 'shrink-0', 'text-right')}>{numberToShow}</span>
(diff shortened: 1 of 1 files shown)


Bohaska:
here's what it looks like

image

Bohaska:
looks like this at phone dimensions image

jahooma:
This is great information to show.

However, it looks pretty busy, especially on mobile. I'm not sure if it makes sense to be in the same column as Max payout.

This was a difficult screen for users to figure out already.

Maybe remove it on mobile (`hidden sm:flex`) and I'll merge it to get feedback.

Bohaska:
> This is great information to show.
> 
> However, it looks pretty busy, especially on mobile. I'm not sure if it makes sense to be in the same column as Max payout.
> 
> This was a difficult screen for users to figure out already.
> 
> Maybe remove it on mobile (`hidden sm:flex`) and I'll merge it to get feedback.

Done!

PR #2551, opened 2024-03-23
Set resolution for independent MC to CANCEL if it's cancelled

Bohaska:
Helps fix the bug where you can't delete independent MC markets because their resolution type is MKT even if they are all resolved

### DIFF
--- a/backend/shared/src/resolve-market-helpers.ts
+++ b/backend/shared/src/resolve-market-helpers.ts
@@ -98,6 +98,7 @@ export const resolveMarketHelper = async (
       outcome === 'YES'
     const marketCancelled = unresolvedContract.answers
       .every((a) => a.resolution === 'CANCEL')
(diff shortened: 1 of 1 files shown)


backend/shared/src/resolve-market-helpers.ts
```diff
@@ -96,6 +96,9 @@ export const resolveMarketHelper = async (
     const hasAnswerResolvedYes =
       unresolvedContract.answers.some((a) => a.resolution === 'YES') ||
       outcome === 'YES'
+    const marketCancelled = unresolvedContract.answers
+      .every((a) => a.resolution === 'CANCEL')
+    const finalResolution = marketCancelled ? 'CANCEL' : 'MKT'
```
jahooma, line 101: I would move this inside the if statement. But it doesn't matter that much so I'll merge it

PR #2553, opened 2024-03-25
Added tests for contract-params.ts

aidevbot:
### Summary
This pull request introduces unit tests for src/contract-params.ts.

### Changes
Added the following test cases for src/contract-params.ts:
- should return not authed state for unauthorized access
- should return not found state for a deleted contract with unauthorized access
- should handle errors gracefully when external API calls fail
- should correctly fetch and process CPMM contract parameters
- should correctly fetch and process NUMBER outcome type contract parameters
- should correctly handle cases with 'cpmm-multi-1' mechanism and 'MULTIPLE_CHOICE' outcome type

### Coverage report

| File          | % Stmts | % Branch | % Funcs | % Lines |
|---------------|---------|----------|---------|---------|
| contract-params.ts      | 85.71 (up from 0)   | 70.9 (up from 0)    | 61.53 (up from 0)     | 88.23 (up from 0)   |


common/src/contract-params.test.ts
```diff
@@ -0,0 +1,474 @@
...
+    await expect(
+      getContractParams(mockContract, mockDb, true, undefined)
+    ).resolves.toMatchObject({
+      state: expect.any(String),
+    })
+  })
+
+  it('should correctly fetch and process CPMM contract parameters', async () => {
+    const mockDb = createClient('instanceId', 'key')
+    const mockContract = {
```
jahooma, line 169: Should there be a common place for mock contracts so they can be reused between tests? (And if we need to update it, because, for example, we added a new field, we would not have to update the mock contracts in a bunch of places.)

jahooma:
Cool! Thanks for setting up these tests. The getContractParams function is indeed a more critical piece of code.

However, my prediction is that after merging this, the tests will start failing because we updated something about the contracts. And probably no one will bother to fix it, except maybe to delete the failing unit tests (or all of them haha). I'm not sure unit tests are part of our dev culture yet.

But secondly, these tests don't necessarily enforce any higher logic or invariants that we want to preserve. They just preserve the exact current cases of our code. Which doesn't really seem that useful to me?

Merging for the experimental value though!

PR #2560, opened 2024-04-02
Fix mobile notifications for profits rounding to 0

benmanns:
If you make small bets or bets at 1% or 99%, the bet may end up profitable, while `Math.round(profit) === 0`. In that circumstance, the mobile notification will say "Resolved: {YOUR BET}. You **lost** M0, outperforming N others!" when it should say something like "Resolved: {YOUR BET}. You **made** M0 (+0.5%), outperforming N others!"

This changes `sendNotificationsIfSettingsPermit` to round profit only at display using `formatMoney`, rather than on computation, so the `profit > 0` check is correct even for profit in (M0, M0.5).

8FE14002-FC05-439F-9292-EFCD96C7EE24_4_5005_c

FF6013F9-5EB9-4865-BE6C-A75F32D2E50F_1_201_a

F295785D-FA19-48CD-9BFE-9B8F867D9EEA_4_5005_c

And an additional note—this is my first time contributing, and I don't have everything set up, so I haven't tested the mobile notification code path. I would appreciate if one of the devs could run this before committing. I think it's all correct, but wouldn't want to break anything!

### DIFF
--- a/backend/shared/src/create-notification.ts
+++ b/backend/shared/src/create-notification.ts
@@ -1266,12 +1266,12 @@ export const createContractResolvedNotifications = async (
         profitRank && totalShareholders && betterThan > 0
           ? `, outperforming ${betterThan} other${betterThan > 1 ? 's' : ''}!`
           : '.'
-      const profit = Math.round(userPayout - userInvestment)
+      const profit = userPayout - userInvestment
       const profitPercent = Math.round((profit / userInvestment) * 100)
-      const profitString = ` You made M${getMoneyNumber(
+      const profitString = ` You made ${formatMoney(
         profit
       )} (+${profitPercent}%)`
-      const lossString = ` You lost M${getMoneyNumber(-profit)}`
+      const lossString = ` You lost ${formatMoney(-profit)}`
       await createPushNotification(
         notification,
         privateUser,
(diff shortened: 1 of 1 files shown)


backend/shared/src/create-notification.ts
```diff
@@ -1266,12 +1266,12 @@ export const createContractResolvedNotifications = async (
...
           : '.'
-      const profit = Math.round(userPayout - userInvestment)
+      const profit = userPayout - userInvestment
       const profitPercent = Math.round((profit / userInvestment) * 100)
-      const profitString = ` You made M${getMoneyNumber(
+      const profitString = ` You made ${formatMoney(
         profit
       )} (+${profitPercent}%)`
-      const lossString = ` You lost M${getMoneyNumber(-profit)}`
+      const lossString = ` You lost ${formatMoney(-profit)}`
```
jahooma, line 1274: `formatMoney` uses our special M character that won't render in a notification

jahooma, line 1274: Seems like you could do profit >= 0 instead of profit > 0 on line 1284 to achieve what you want.

benmanns: The trouble is small losses (-0.5,0.0) and small wins (0.0, 0.5) both get treated the same if we use the rounded profit for the comparison with 0.

sipec (approved):

benmanns:
Thanks @sipec! I think it's working as expected.

image showing "Resolved: Summer Solstice 2024: NO. You made MO (+1%), outperforming 2 others!"

PR #2575, opened 2024-04-12
Fix up some Supabase realtime-related code

mqp:
.Not clear what the future of this is, but getting this merged into production, since I already did it and this stuff will be necessary work if we do decide to use Supabase broadcast stuff in the future.

Now there are 4 realtime-related hooks:

- `useChannel`. This manages a Realtime channel, logging errors and turning it on and off if you background the page.
- `useBroadcast`. This opens a channel with `useChannel` and subscribes it to a broadcast channel, and lets you get the messages that come in.
- `usePostgresChanges`. This opens a channel with `useChannel` and subscribes the channel to Postgres changes of your choosing. Renamed from the old ambiguous `useRealtime`.
- `useSubscription`. This is the one that does an initial query and merges it with changes from `usePostgresChanges` so that it's like you have a consistent snapshot of the state.

### DIFF
--- a/backend/api/src/on-create-bet.ts
+++ b/backend/api/src/on-create-bet.ts
@@ -78,21 +78,6 @@ export const onCreateBets = async (
   ordersToCancel: LimitBet[] | undefined,
   makers: maker[] | undefined
 ) => {
-  // Temporarily disable joining contract channel.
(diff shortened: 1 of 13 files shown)


mqp:
I used to have JWT auth stuff in this PR but now Supabase has its own auth stuff so I pried all of that out, since we would use theirs.

PR #2576, opened 2024-04-16
The spice must flow

sipec:
I've added a basic flow and tested that resolve, redeem, and unresolve works right. It's all behind a feature flag so I think this is a good point to check it in


jahooma:
Great work, Sinclair! This looks good and complete.

Let the spice flow!! (i.e. merge it!!)

jahooma (approved):

PR #2577, opened 2024-04-16
Say "for M̶250" not "for $M̶250"

syvb:
There shouldn't be a dollar sign here:

Refer a friend for for $M̶250 each!

The code does `${formatMoney(REFERRAL_AMOUNT)}` as if it was in a format string, but it's actually in a JSX markup expression which doesn't require a `$` there.


jahooma:
Thanks!

PR #2582, opened 2024-04-18
Add code for recording custom GCP metrics in backend

mqp:
Now we can make charts of things that aren't already on charts, like revalidations, or Postgres queries, or Postgres connections, or bets performed, or notifications made, or any other weird things we want to log, or whatever.


mqp:
Current status: Everything is working and I added a handful of metrics, I will continue to add more and make some more useful dashboards with our new metrics.

PR #2584, opened 2024-04-18
Reintroduce JWT signing of broadcast payloads

mqp:
The [broadcast authorization provided by Supabase](https://github.com/orgs/supabase/discussions/22484) seems overly complex for our needs. It's designed for people who want to authorize all kinds of different users into Phoenix channels, by looking up facts about them in Postgres based on their JWT. We don't care about anything so complicated right now; all we want is for nobody but our server to be able to send messages.

I think it's better to just implement what we want, which is simple enough, and not turn on their new authorization stuff until it's been fully baked.

We can try out their authorization later if we have an important reason why we want to use broadcasts on public channels for non-public info (e.g. contracts with special access rules, maybe private messages.)

### DIFF
--- a/backend/shared/src/supabase/channel.ts
+++ b/backend/shared/src/supabase/channel.ts
@@ -1,4 +1,5 @@
 import { getInstanceId, getInstanceHostname } from './init'
+import { sign } from 'jsonwebtoken'
 
 type BroadcastMessage = {
   channel: string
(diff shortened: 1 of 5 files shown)


sipec:
sounds reasonable. I've updated towards us just using polling over channels though.

mqp:
Closing this since we are just doing our own thing.

PR #2589, opened 2024-04-19
Say "for M̶250" not "for $M̶250", again

syvb:
This PR is identical to #2577, which was (accidentally?) reverted by @jahooma in d3483dfba19f601a34bd1cfadfe446e48e529f80.


jahooma:
Oops, yeah my bad, was a merge error.

PR #2594, opened 2024-04-24
stacked graph

ingawei:


web/components/portfolio/portfolio-value-graph.tsx
```diff
@@ -27,37 +34,138 @@ export const PortfolioTooltip = (props: { date: Date }) => {
...
+  setPortfolioFocus: (mode: PortfolioMode) => void
+  portfolioHoveredGraph: PortfolioHoveredGraphType
+  setPortfolioHoveredGraph: (hovered: PortfolioHoveredGraphType) => void
 }) => {
   const {
     mode,
     duration,
-    points,
-    onMouseOver,
+    // points,
```
jahooma, line 55: remove?

web/components/portfolio/portfolio-value-graph.tsx
```diff
@@ -27,37 +34,138 @@ export const PortfolioTooltip = (props: { date: Date }) => {
...
+          obj: p,
+        })
+        networthPoints.push({
+          x: p.timestamp,
+          y: p.balance + p.investmentValue,
+          obj: p,
+        })
+      })
+
+      return { profitPoints, investmentPoints, balancePoints, networthPoints }
```
jahooma, line 111: This might be nice to move into `common` as a utility function. Or else at the bottom of the file.

web/components/portfolio/portfolio-chart.tsx
```diff
@@ -0,0 +1,282 @@
...
+  y0: number // Lower boundary of the area
+  y1: number // Upper boundary of the area
+}
+
+// hacky solution
+const BALANCE_IDX = 0
+const INVESTED_IDX = 1
+
+// multi line chart
+export const PortfolioChart = <P extends HistoryPoint>(props: {
```
jahooma, line 39: This looks good. It's a shame there's not a good way to build a simpler generic version of this stacked line graph, and then configure it for the portfolio case? It would certainly be cleaner to read if so.

jahooma (approved):
Great, looks good Inga! I'm sure users will appreciate the combined balance + investment chart.

Charts work is always harder than it seems... haha.

PR #2597, opened 2024-04-30
Log custom metrics for HTTP requests

mqp:
Now we can see a latency breakdown graph by endpoint.


jahooma:
This will be very cool

PR #2605, opened 2024-05-07
Ingawei/negative pattern portfolio

ingawei:
Changed graph so that when invested is negative, the fill is diagonal lines:
image
image

Would like a sanity check because I feel like maybe I did it in a dumb way


jahooma (approved):
Cool, this is tricky, but seems like a reasonable solution!

PR #2606, opened 2024-05-07
Prototype API websocket server for broadcasts

mqp:
***WIP - I literally didn't run this code yet, I just want to get it out there to give any interested parties an idea of what's going on.***

Implements:

- A websocket server on the API server that listens to connections on the path `/ws`.
- Code to manage a list of topic subscriptions per connection, which can then receive broadcast messages on those topics.
- Client code for interfacing with the above.

### DIFF
--- a/backend/api/src/websockets/server.ts
+++ b/backend/api/src/websockets/server.ts
@@ -1,0 +1,119 @@
+import { Server as HttpServer } from 'node:http'
+import { Server as WebSocketServer, RawData, WebSocket } from 'ws'
+import { isError } from 'lodash'
+import { log } from 'shared/utils'
+import { Switchboard } from './switchboard'
+import {
+  ClientMessage,
+  ServerMessage,
+  CLIENT_MESSAGE_SCHEMA,
+} from 'common/api/websockets'
+
+const SWITCHBOARD = new Switchboard()
+
+export class MessageParseError extends Error {
+  details?: unknown
+  constructor(message: string, details?: unknown) {
+    super(message)
+    this.name = 'MessageParseError'
+    this.details = details
+  }
+}
+
+function serializeError(err: unknown) {
+  return isError(err) ? err.message : 'Unexpected error.'
+}
+
+function parseMessage(data: RawData): ClientMessage {
+  let messageObj: any
+  try {
+    messageObj = JSON.parse(data.toString())
+  } catch (err) {
+    log.error(err)
+    throw new MessageParseError('Message was not valid UTF-8 encoded JSON.')
+  }
+  const result = CLIENT_MESSAGE_SCHEMA.safeParse(messageObj)
+  if (!result.success) {
+    const issues = result.error.issues.map((i) => {
+      return {
+        field: i.path.join('.') || null,
+        error: i.message,
+      }
+    })
+    throw new MessageParseError('Error parsing message.', issues)
+  } else {
+    return result.data
+  }
+}
+
+function processMessage(ws: WebSocket, data: RawData): ServerMessage<'ack'> {
+  try {
+    const msg = parseMessage(data)
+    const { type, txid } = msg
+    try {
+      switch (type) {
+        case 'identify': {
+          SWITCHBOARD.identify(ws, msg.uid)
+          break
+        }
+        case 'subscribe': {
+          SWITCHBOARD.subscribe(ws, ...msg.topics)
+          break
+        }
+        case 'unsubscribe': {
+          SWITCHBOARD.unsubscribe(ws, ...msg.topics)
+          break
+        }
+        case 'ping': {
+          SWITCHBOARD.markSeen(ws)
+          break
+        }
+        default:
+          throw new Error("Unknown message type; shouldn't be possible here.")
+      }
+    } catch (err) {
+      log.error(err)
+      return { type: 'ack', txid, success: false, error: serializeError(err) }
+    }
+    return { type: 'ack', txid, success: true }
+  } catch (err) {
+    log.error(err)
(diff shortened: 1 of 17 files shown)


backend/api/src/websockets/switchboard.ts
```diff
@@ -0,0 +1,65 @@
+import { WebSocket } from 'ws'
+
+export type ClientState = {
+  uid?: string
+  lastSeen: number
+  subscriptions: Set<string>
+}
+
+/** Tracks the relationship of clients to websockets and subscription lists. */
+export class Switchboard {
```
jahooma, line 10: Classes? Hmm... I guess it's fine. Manifold employees might not be familiar with all the nuances of 'this' in classes.

backend/api/src/websockets/server.ts
```diff
@@ -0,0 +1,119 @@
...
+      return { type: 'ack', txid, success: false, error: serializeError(err) }
+    }
+    return { type: 'ack', txid, success: true }
+  } catch (err) {
+    log.error(err)
+    return { type: 'ack', success: false, error: serializeError(err) }
+  }
+}
+
+export function broadcast(topic: string, data: Record<string, unknown>) {
```
jahooma, line 85: Would we want some kind of event type system? I think supabase channels worked this way with multiple event types per channel/topic.

Probably can just add it later though if we need it.

backend/api/src/websockets/server.ts
```diff
@@ -0,0 +1,119 @@
...
+  return isError(err) ? err.message : 'Unexpected error.'
+}
+
+function parseMessage(data: RawData): ClientMessage {
+  let messageObj: any
+  try {
+    messageObj = JSON.parse(data.toString())
+  } catch (err) {
+    log.error(err)
+    throw new MessageParseError('Message was not valid UTF-8 encoded JSON.')
```
jahooma, line 33: Thanks for this error

web/lib/api/ws.ts
```diff
@@ -0,0 +1,138 @@
...
+} from 'common/api/websockets'
+import { getApiUrl } from 'common/api/utils'
+
+// mqp: useful for debugging
+const VERBOSE_LOGGING = true
+
+export type BroadcastHandler = (msg: ServerMessage<'broadcast'>) => void
+
+export class APIRealtimeClient {
+  ws!: WebSocket
```
jahooma, line 14: Wow, I guess WebSocket is a browser-defined type.

jahooma (approved):
This is great! Looks pretty understandable to me.

How do we plan to use topics? Probably like we did for channels: one topic per contract, or maybe one topic for getting all the bets for the live view.

Very excited to use this websocket technology!

Happy to merge this, but we should follow up by trying to use it for bets, I guess

backend/api/src/websockets/switchboard.ts
mqp: I guess on reflection "Manifold house style" is to just make some global variables that aren't in classes?

In Javascript if I want a bag of mutable state that I need to operate on repeatedly then I typically make a class and instantiate the class.

It's funny that you mention `this`, since probably the main reason Javascript introduced classes at all, was because it was harder to understand the implementation of `this` in the old "explicitly add functions to a prototype" or "use a closure like a class" styles that people would use in circa-2010 JS.

backend/api/src/websockets/server.ts
mqp: We'll see if we need it. FYI, in the Supabase implementation the events are just a client abstraction anyway. You couldn't actually ask for only some events from the server. The topics were the real thing.

web/lib/api/ws.ts
mqp: Browser web sockets are like 10+ years old. The browser WebSocket client implementation is also provided with Node in the newest Node (22), but there's no WebSocket server in Node, so we will still need to use a library for that for the foreseeable future.

mqp:
> How do we plan to use topics? Probably like we did for channels: one topic per contract, or maybe one topic for getting all the bets for the live view.

Basically I would break up the topics so that the stuff on our website can more or less maximally efficiently subscribe to what it needs and no more. Since we are free from arbitrary Supabase restrictions about "number of channels" or channel overhead or authorization stuff or so on, we can basically do whatever we want and it will work fine. It doesn't seem to matter much if there are lots of topics, or a few topics, or a medium number of topics.

backend/api/src/websockets/switchboard.ts
jahooma: One style is `initX()` which gives you the state, and then a bunch of functions that take x as their first parameter and mutate it.

Another style which is cool would be to have a function `createX()` which returns an object with all the methods. This has the downside of recreating the methods for every instantiation, whereas the class creates them once on the prototype object. (But I don't think that matters much here.)

E.g. with the latter style
```
function createSwitchboard() {
  const state = { ... }
  
  return {
    method1: () => {},
    method2: () => {}
  }
}
```

backend/api/src/websockets/server.ts
jahooma: Sounds good!

jahooma:
> Basically I would break up the topics so that the stuff on our website can more or less maximally efficiently subscribe to what it needs and no more. Since we are free from arbitrary Supabase restrictions about "number of channels" or channel overhead or authorization stuff or so on, we can basically do whatever we want and it will work fine. It doesn't seem to matter much if there are lots of topics, or a few topics, or a medium number of topics.

Awesome. Can't wait to use this to make our app good

mqp:
I am not going to say that I am 100% sure this is all working, but I tested all the code paths and poked at it in some different ways, and I think it's pretty good.

PR #2616, opened 2024-05-11
Migrate users

sipec:


jahooma (commented):
This looks great, Sinclair! It's useful for me to at least follow along with how the user migration is being implemented.

I haven't examined the PR in complete detail, but it seems reasonable. A dual transaction for firebase and supabase seems fine.

Good luck testing and merging!

PR #2625, opened 2024-05-16
Adjust profits for resolved markets

IanPhilips:
A number of changes:
1. 10x ranked profits on markets and answers resolved before/on may 15th
2. Any market that is unranked or unlisted has its profits nullified on the leaderboard 

Dev notes: 
1. The user has their `resolvedProfitAdjustment` cached so we don't have to recalculate this on every metric update
2. The user has any unresolved market/answer `profit_adjustments` calculated on every metric update on their user_contract_metrics
3. the new `profit` column in user_portfolio_history is their profit for the leaderboard


web/components/portfolio/portfolio-value-graph.tsx
IanPhilips, line 1: @sinclairity are we supposed to be adding in the p.spiceBalance to profit calcs now?

web/components/portfolio/portfolio-value-section.tsx
IanPhilips, line 1: Add in `spiceBalance`?

backend/shared/src/get-user-portfolio-internal.ts
IanPhilips, line 1: Add in `spiceBalance`?

backend/shared/src/update-user-metrics-core.ts
```diff
@@ -223,12 +216,18 @@ export async function updateUserMetricsCore() {
       contractsById,
       unresolvedBetsOnly
     )
+    const { balance, investmentValue, totalDeposits } = newPortfolio
+    const profit =
```
IanPhilips, line 220: add in `spiceBalance`?

PR #2629, opened 2024-05-17
New mod flow

SirSaltyy:


web/pages/reports.tsx
```diff
@@ -0,0 +1,297 @@
...
+          },
+          {} as { [key: number]: ReportStatus }
+        )
+        const initialNotes = sortedReports.reduce(
+          (acc: any, report: Report) => {
+            acc[report.report_id] = report.mod_note
+            return acc
+          },
+          {} as { [key: number]: string | undefined }
+        )
```
jahooma, line 74: Do this instead:

```
const reportsById = keyBy(newReports, 'report_id')
const initialStatuses = mapValues(reportsById, r => r.status)
const initialNotes = mapValues(reportsById, r => r.mod_note)
```

web/pages/reports.tsx
```diff
@@ -0,0 +1,297 @@
...
+              new Date(b.created_time ?? 0).getTime() -
+              new Date(a.created_time ?? 0).getTime()
+          )
+
+        setReports(sortedReports)
+        const initialStatuses = sortedReports.reduce(
+          (acc: any, report: Report) => {
+            acc[report.report_id] = report.status
+            return acc
+          },
```
jahooma, line 65: Move this and below lines into a hook. Create a new file, like `use-mod-reports.ts`.

web/pages/reports.tsx
```diff
@@ -0,0 +1,297 @@
...
+    }))
+
+    await updateReport(reportId, { mod_note: newNote })
+  }
+
+  useEffect(() => {
+    const fetchOwner = async () => {
+      const ownerData: { [key: string]: any } = {}
+      for (const report of reports) {
+        const user = await getUserById(report.user_id)
```
jahooma, line 123: If you need the User object, you should return that in the endpoint instead of the user_id. Does that make sense?

I.e. query for the user within the endpoint, not on the client.

web/pages/reports.tsx
```diff
@@ -0,0 +1,297 @@
...
+  const [reports, setReports] = useState<Report[]>([])
+  const [initialLoading, setInitialLoading] = useState(true)
+  const [owner, setOwner] = useState<{ [key: string]: any }>({})
+  const [reportStatuses, setReportStatuses] = useState<{
+    [key: number]: ReportStatus
+  }>({})
+  const [modNotes, setModNotes] = useState<{
+    [key: number]: string | undefined
+  }>({})
+  const initialLoadRef = useRef(true)
```
jahooma, line 44: Don't use refs for this. Mutating a ref does not trigger a re-render.

Most likely `reports` should optionally be `undefined`, and default to undefined, and then you know when it's loaded when it becomes defined. Then you don't need another piece of state.

web/pages/reports.tsx
```diff
@@ -0,0 +1,297 @@
...
+                  Resolved: 'resolved',
+                  'Needs Admin': 'needs admin',
+                }}
+                setChoice={(val) =>
+                  handleStatusChange(report.report_id, val as ReportStatus)
+                }
+              />
+            </Row>
+            <Row className="mt-2 items-center">
+              Mod note: &nbsp;
```
jahooma, line 243: instead of &nbsp;, use {' '}

web/pages/reports.tsx
```diff
@@ -0,0 +1,297 @@
...
+                        size="sm"
+                      />
+                      <UserLink
+                        user={owner[report.user_id]}
+                        className="text-ink-800 ml-2"
+                      />
+                      {owner[report.user_id]?.isBannedFromPosting && (
+                        <BannedBadge />
+                      )}
+                    </div>
```
jahooma: I would make a component for `ReportItem` or something which has all of this and below.

jahooma (commented):
Nice work Salty! Looks like a good feature. Just a few dozen comments for you to fix hahaha.

Let me know if you have any questions. And once you finish addressing the comments, I'll take another look.

backend/supabase/mod_reports.sql
```diff
@@ -0,0 +1,33 @@
...
+    report_id SERIAL primary key,
+    created_time timestamptz not null default now(),
+    comment_id text not null,
+    contract_id text not null,
+    user_id text not null,
+    status status_type not null default 'new',
+    contract_slug text not null,
+    contract_question text not null,
+    content jsonb not null,
+    creator_username text not null,
```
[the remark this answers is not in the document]
SirSaltyy: I understand what you want me to do and the reasoning for it.

Struggling with the execution. GPT first suggested using .leftjoin which didn't work so then started using raw SQL. But there was a lot of errors and the approach started requiring a lot of new lines of code including creating a new custom postgress function. Maybe this is correct but I wanted to check if there is a method you suggest to join the queries.

Also:
user_id
comment_id
contract_id
created_time

are all duplicated from contract_comments. Should I delete all of these too? Intuitively I would guess it is fine to delete all of them except comment_id because it would still be good to keep one identifier for which comment as a property on the report itself?

common/src/api/schema.ts
```diff
@@ -1154,6 +1154,32 @@ export const API = (_apiTypeCheck = {
...
+    method: 'POST',
+    visibility: 'public',
+    authed: true,
+    props: z
+      .object({
+        reportId: z.number(),
+        updates: z
+          .object({
+            status: z
+              .enum(['new', 'under review', 'resolved', 'needs admin'])
```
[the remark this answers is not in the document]
SirSaltyy: I don't know what this means. I tried changing the name from status: z to reportStatus: z which broke things. I then tried modifying update-reports to accommodate the change but it didn't help.

common/src/api/schema.ts
```diff
@@ -1154,6 +1154,32 @@ export const API = (_apiTypeCheck = {
...
+              .optional(),
+            mod_note: z.string(),
+          })
+          .partial(),
+      })
+      .strict(),
+    returns: {} as { status: string; data: any },
```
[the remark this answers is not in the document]
SirSaltyy: I having to write like 20 more lines to accommodate this? Is this correct lol?

web/hooks/use-reports.ts
```diff
@@ -0,0 +1,38 @@
...
+import { initSupabaseClient } from 'web/lib/supabase/db'
+
+type Report = {
+  report_id: number
+  created_time: string
+  comment_id: string
+  contract_id: string
+  user_id: string
+  status: 'new' | 'under review' | 'resolved' | 'needs admin'
+}
```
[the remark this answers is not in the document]
SirSaltyy: Actually common/report.ts is from Ian's old report system and has nothing to do with my new reports. Is there somewhere else I should define it?

I also have report-types.ts so maybe I actually should just delete where it is being defined in this file (I think I wrote this first and then realised I needed to define it in common later anyways)

backend/supabase/mod_reports.sql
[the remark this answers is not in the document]
jahooma: ID's are more often fine to copy around, but you could remove them except comment_id if you want. I'd probably keep them for convenience.

I think it's nice to have a created_time for this specific row in mod_reports, so I would leave that.

How do you join in the endpoint to get the comment data? Something like this:

```
const fullModReports = await pg.many(`
  select mr.*, cc.content, contracts.question, contracts.slug, users.username
  from mod_reports mr
  join contract_comments cc on cc.comment_id = mr.comment_id
  join contracts on contracts.id = mr.contract_id
  join users on users.id = contracts.creator_id
`)
```

common/src/api/schema.ts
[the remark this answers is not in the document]
jahooma: You already defined ReportStatus to be this enum, so you should use it:

```
z.object({
  status: ReportStatus,
  mode_note z.string(),
}).partial(),

File: web/hooks/use-reports.ts (hunk 10 above)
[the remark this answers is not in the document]
Comment (jahooma) [reply web/hooks/use-reports.ts:11]: Oh yeah, It should be called `ModReport` and live in `common/mod-report.ts`. I guess `report-types.ts` is ok too.

Try to define the types once in common and then reuse the types everywhere.

File: common/src/api/schema.ts (hunk 9 above)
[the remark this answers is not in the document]
Comment (jahooma) [reply common/src/api/schema.ts:1174]: No haha. Your endpoint update returns a field `data`, probably the type is `ModReport`?

File: backend/api/src/get-mod-reports.ts [hunk 11]
```diff
@@ -0,0 +1,26 @@
...
+    SELECT 
+      mr.*, 
+      cc.data->>'content' AS comment_content, 
+      c.question AS contract_question, 
+      c.slug AS contract_slug, 
+      creator.username AS creator_username,
+      owner.data->>'username' AS owner_username,
+      owner.data->>'avatarUrl' AS owner_avatarUrl, 
+      (owner.data->>'isBannedFromPosting')::boolean AS owner_isBannedFromPosting
+    FROM mod_reports mr
```
jahooma, line 18: Don't use capital letters in sql. It will come out as lowercase. I think that might be your bug

jahooma: I.e. `owner_avatarUrl => owner_avatar_url`
`owner_isBannedFromPosting => owner_is_banned_from_posting`

web/components/mod-report-item.tsx
```diff
@@ -110,7 +132,7 @@ const ModReportItem: React.FC<ReportItemProps> = ({
           />
         </Row>
         <Row className="mt-2 items-center">
-          Mod note: {''}
+          Mod note:&nbsp;
```
SirSaltyy, line 135: for some reason {''} wasn't doing anything. I tried wrapping it in a span and that also didn't work.

jahooma: You need a space in between the quotes: `{' '}`

backend/api/src/get-mod-reports.ts
```diff
@@ -6,20 +6,21 @@ export const getModReports: APIHandler<'get-mod-reports'> = async () => {
...
-      creator.username AS creator_username,
-      owner.data->>'username' AS owner_username,
-      owner.data->>'avatarUrl' AS owner_avatarUrl, 
-      (owner.data->>'isBannedFromPosting')::boolean AS owner_isBannedFromPosting
-    FROM mod_reports mr
-    JOIN contract_comments cc ON cc.comment_id = mr.comment_id
-    JOIN contracts c ON c.id = mr.contract_id
-    JOIN users creator ON creator.id = c.creator_id
-    JOIN users owner ON owner.id = mr.user_id
+      cc.data->>'content' as comment_content,
```
jahooma, line 11: Oh, I think you might have wanted only one '>':  cc.data->'content'. That would return the jsonb data instead of a string. I'm not 100% sure, but it could have solved your issue.

backend/api/src/update-mod-report.ts
```diff
@@ -0,0 +1,39 @@
...
+    status?: 'new' | 'under review' | 'resolved' | 'needs admin'
+    mod_note?: string
+  } = {}
+
+  if (updates.status) {
+    updateData.status = updates.status
+  }
+  if (updates.mod_note !== undefined) {
+    updateData.mod_note = updates.mod_note
+  }
```
jahooma, line 21: I'm pretty sure `updateData` is the same as `removeUndefinedProps(updates)`

common/src/api/zod-types.ts
```diff
@@ -51,3 +51,6 @@ export const DashboardItemSchema = z.union([
   DashboardLinkItemSchema,
   DashboardTextItemSchema,
 ])
+
+
+
```
jahooma, line 56: Can remove these lines

web/components/mod-report-item.tsx
```diff
@@ -0,0 +1,145 @@
...
+    return text.trim()
+  }
+
+  const parseCommentContent = (commentContent: any) => {
+    if (typeof commentContent === 'string') {
+      try {
+        return JSON.parse(commentContent)
+      } catch (error) {
+        console.error('Failed to parse comment content', error)
+        return null
```
jahooma, line 66: This is definitely not the right way to do this. I would try to use the rich text library to parse the content. Maybe you can ask @sinclairity?

I would be ok with you fixing this in a follow-up PR though if you add a TODO.

web/hooks/use-mod-reports.ts
```diff
@@ -0,0 +1,63 @@
...
+      const response = await api('get-mod-reports', {})
+      if (response && response.status === 'success') {
+        const newReports = response.reports
+
+        const sortedReports = newReports
+          .filter((report: ModReport) => report.created_time)
+          .sort(
+            (a: ModReport, b: ModReport) =>
+              new Date(b.created_time ?? 0).getTime() -
+              new Date(a.created_time ?? 0).getTime()
```
jahooma, line 26: I thought all reports would have a `created_time` making the filter unnecessary (and `?? 0`)

jahooma (approved):
Looks great! Thanks for implementing all the requested changes, Salty.

This is a pretty complex feature you've implemented. You're almost programming at the intern level! 😛

But seriously, you've learned quite a lot. And I'm sure our moderators will appreciate this feature.

web/components/mod-report-item.tsx
jahooma: `{richTextToString(comment.content)}`

backend/api/src/get-mod-reports.ts
SirSaltyy: Nice this allowed me to use our existing function parseJsonContentToText, good catch

web/components/mod-report-item.tsx
SirSaltyy: Tried for far too long (10-minutes) to get {' '} to work but couldn't so leaving it as is. Why is &nbsp bad to use anyways? Alternatively could wrap the text in a div and add padding but that just seems worse.

jahooma: Ah, yeah, it doesn't matter that much, mostly it's for learning so you know how to do it better in the future.

You're right, Row, which has display `flex` is typically for containers of other containers, not for text.

I think I'd wrap mod note in a div like this `<div>Mode note</div>` and add `gap-2` to the Row parent.

PR #2631, opened 2024-05-21
Ingawei/add prize market filter

ingawei:


common/src/api/market-search-types.ts
```diff
@@ -51,5 +51,11 @@ export const searchProps = z
     topicSlug: z.string().regex(FIRESTORE_DOC_REF_ID_REGEX).optional(),
     creatorId: z.string().regex(FIRESTORE_DOC_REF_ID_REGEX).optional(),
     isPolitics: z.coerce.boolean().optional(),
+    isPrizeMarket: z
+      .union([
+        z.literal('true'),
+        z.literal('false'),
+      ])
+      .default('false'),
```
jahooma, line 59: It's fine to use strings for parameters in the url, but I think it's better to use a boolean here. People will be calling our search API and not expect true and false to be strings.

jahooma (approved):
I think this is right! Just one small recommendation for you.

Thanks Inga!

common/src/api/market-search-types.ts
ingawei: Yeah that was to fix another bug, for some reason when I did a boolean it always registered as true

PR #2639, opened 2024-05-23
More websocket broadcast productionizing, test code

mqp:
Now things broadcast to contract-specific topics, broadcasts contain data about the broadcasted item, there are more broadcasts, broadcasts respect contract visibility, and I added a useful test script.


jahooma:
Nice!

PR #2643, opened 2024-05-24
Bulk write bets

sipec:
unclear if this actually improves much - all the inserts were already part of the same postgres transaction


jahooma:
Sorry, I forgot about this PR! The basic idea is good, especially when I think about linked multiple choice with 100 options. It's worth finishing at some point, though now there are some conflicts.

PR #2646, opened 2024-05-24
Migrate liquidity to supabase

sipec:
This no longer reads/writes contract transactionally, so the recorded pool (and p?) may be off if someone bets at the same time. That doesn't matter since we don't use those fields.
Somewhat more concerning - I think the _subsidyPool_ could be off if someone adds a subsidy the same time that the house tries to add a subsidy. We could avoid this by making the pg.tx `SERIAL` but idc

### DIFF
--- a/backend/api/src/add-subsidy.ts
+++ b/backend/api/src/add-subsidy.ts
@@ -10,6 +10,7 @@ import { broadcastNewSubsidy } from 'shared/websockets/helpers'
 import { onCreateLiquidityProvision } from './on-update-liquidity-provision'
 import { insertLiquidity } from 'shared/supabase/liquidity'
 import { convertLiquidity } from 'common/supabase/liquidity'
+import { FieldValue } from 'firebase-admin/firestore'
 
 export const addLiquidity: APIHandler<
   'market/:contractId/add-liquidity'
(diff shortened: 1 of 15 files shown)


backend/api/src/add-subsidy.ts
```diff
@@ -51,33 +53,20 @@ export const addContractLiquidity = async (
...
 
-    const liquidity = await firestore.runTransaction(async (transaction) => {
-      const contractDoc = firestore.doc(`contracts/${contractId}`)
+    const subsidyAmount = (1 - SUBSIDY_FEE) * amount
 
-      const newLiquidityProvisionDoc = firestore
-        .collection(`contracts/${contractId}/liquidity`)
-        .doc()
+    const { newLiquidityProvision, newTotalLiquidity, newSubsidyPool } =
+      getNewLiquidityProvision(userId, subsidyAmount, contract)
```
jahooma, line 59: I think `newSubsidyPool` can be changed to `subsidyPoolIncrease` or something, which then can use firestore's increment() to update the contract subsidyPool field that would not cause a problem with multiple subsidies being added at the same time.

jahooma: Same for `newTotalLiquidity`

jahooma (commented):
Cool!

Seems like liquidity docs are read in `resolve-market-helper.ts`, wouldn't you need to change the query to read from supabase instead of firebase?

Also the liquidity collection is written to in `create-market.ts` and `create-answer-cpmm.ts`

PR #2653, opened 2024-05-31
Migrate answers to supabase

sipec:
notable changes: I made createAnswerCpmm not use evil transaction anymore since it only needs to update the contract liquidity. I also changed drizzle liquidity significantly cuz the previous implementation made no sense to me


backend/functions/src/scheduled/drizzle-liquidity.ts
```diff
@@ -62,31 +70,30 @@ const drizzleMarket = async (contractId: string) => {
     const amount = subsidyPool <= 1 ? subsidyPool : r * v * subsidyPool
 
     if (contract.mechanism === 'cpmm-multi-1') {
-      const answersSnap = await trans.get(
-        firestore.collection(`contracts/${contractId}/answersCpmm`)
-      )
-      const answers = answersSnap.docs.map((doc) => doc.data() as Answer)
+      if (!contract.shouldAnswersSumToOne) return // will be drizzled by drizzleAnswer
```
jahooma, line 73: Not true. If you subsidize independent multi from the boost button, it does it on the whole market rather than specific answers. Then this code is supposed to drizzle it evenly among answers

backend/shared/src/supabase/answers.ts
```diff
@@ -1,6 +1,16 @@
...
+import { SupabaseDirectClient, pgp } from 'shared/supabase/init'
 import { convertAnswer } from 'common/supabase/contracts'
 import { groupBy } from 'lodash'
+import { Answer } from 'common/answer'
+import { bulkInsert, updateData, insert as upsert } from './utils'
+import { randomString } from 'common/util/random'
+import { removeUndefinedProps } from 'common/util/object'
+import { millisToTs } from 'common/supabase/utils'
+
+export const getAnswer = async (pg: SupabaseDirectClient, id: string) => {
```
jahooma, line 10: Thanks, this file of helper functions is great. We should do it for every table.

jahooma (commented):
Thanks Sinclair! Looks good overall. It's cool how close we are to eliminating firebase entirely in our bet transactions.

The one thing to fix was the comment left before about supporting subsidizing independent multi via one subsidy for the whole contract that then gets split up to each answer.

jahooma (approved):
Looks good! I don't see anything obviously wrong. Can't believe we're almost done with the migration!

PR #2656, opened 2024-06-03
Broadcast answers over websocket

sipec:


backend/api/src/on-create-bet.ts
```diff
@@ -77,10 +80,31 @@ export const onCreateBets = async (
...
+        const answer = answerSnap.data() as Answer
+        broadcastUpdatedAnswer(contract, answer)
+      }
+    } else {
+      const answersCol = firestore.collection(
+        `contracts/${contract.id}/answersCpmm`
+      )
+      const answersSnap = await answersCol.get()
+      const answers = answersSnap.docs.map((doc) => doc.data() as Answer)
+      answers.forEach((answer) => broadcastUpdatedAnswer(contract, answer))
```
jahooma, line 104: I'm a little worried about how this scales. For a 100 answer sums-to-one multi, then every bet will cause 100 broadcasts.

I think it's actually not that different from 1 broadcast of the 100 answers, except the front end code might want to debounce the updates, since browsers could lock up with that many rerenders.

PR #2661, opened 2024-06-05
Ingawei/add option to upgrade after creation

ingawei:
image
image

backwards calculates current tier for now, but hoping to get rid of that once we have a script that adds a tier to all markets

### DIFF
--- a/web/components/contract/boost-button.tsx
+++ b/web/components/contract/boost-button.tsx
@@ -58,182 +58,36 @@ export function BoostDialog(props: {
   const { contract, isOpen, setOpen } = props
 
   const [index, setIndex] = useState(0)
-  const [showTabs, setShowTabs] = useState(false)
 
   const [amount, setAmount] = useState<number | undefined>(5000)
-
-  const subsidyDisabled =
-    contract.isResolved ||
-    (contract.closeTime ?? Infinity) < Date.now() ||
-    (contract.mechanism !== 'cpmm-1' && contract.mechanism !== 'cpmm-multi-1')
 
   return (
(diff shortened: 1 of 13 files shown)


web/components/contract/upgrade-tier-button.tsx
```diff
@@ -0,0 +1,266 @@
...
+import { useUser } from 'web/hooks/use-user'
+import { ENV_CONFIG } from 'common/envs/constants'
+import { AddFundsModal } from '../add-funds-modal'
+import { track } from 'web/lib/service/analytics'
+import toast from 'react-hot-toast'
+
+export type ContractWithLiquidity =
+  | CPMMContract
+  | CPMMMultiContract
+  | CPMMNumericContract
```
jahooma, line 32: I think I created a MarketContract type which is the same!

web/components/contract/boost-button.tsx
```diff
@@ -58,185 +58,39 @@ export function BoostDialog(props: {
...
         <Title className="!mb-2">🚀 Boost this question</Title>
-
-        {!showTabs && (
-          <SimpleBoostRow
-            contract={contract}
-            subsidyDisabled={subsidyDisabled}
-            amount={amount}
-            setAmount={setAmount}
-          />
-        )}
```
jahooma, line 82: Nice, I thought the simple boost was a little confusing anyway, good to simplify that to just feed promotion

jahooma (approved):
LGTM! Looks good. This is a great feature. You should add the tier to the contract page header always!

PR #2662, opened 2024-06-05
Get rid of tier col

ingawei:
I got rid of tier in the backend, so old markets have tiers now. Decided to keep the answer cost because I wasn't sure what effect it would have if creating however many answers cost 0 mana, but then adding an additional answer after the fact cost mana. Basically my solution was to cap the number of answers to 89 because then you couldn't possibly get to the higher tier amount with just the additional cost of answers alone, but idk if this solution is gross. Lmk!

### DIFF
--- a/common/src/answer.ts
+++ b/common/src/answer.ts
@@ -36,8 +36,8 @@ export type Answer = {
 
 export const MAX_ANSWER_LENGTH = 240
 
-export const MAX_ANSWERS = 100
-export const MAX_INDEPENDENT_ANSWERS = 200
+export const MAX_ANSWERS = 89
+export const MAX_INDEPENDENT_ANSWERS = 89
 
 export const getMaximumAnswers = (shouldAnswersSumToOne: boolean) =>
   shouldAnswersSumToOne ? MAX_ANSWERS : MAX_INDEPENDENT_ANSWERS
--- a/common/src/answer.ts
+++ b/common/src/answer.ts
@@ -36,8 +36,8 @@ export type Answer = {
 
 export const MAX_ANSWER_LENGTH = 240
 
-export const MAX_ANSWERS = 89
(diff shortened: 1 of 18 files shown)


common/src/answer.ts
```diff
@@ -36,8 +36,8 @@ export type Answer = {
 
 export const MAX_ANSWER_LENGTH = 240
 
-export const MAX_ANSWERS = 100
-export const MAX_INDEPENDENT_ANSWERS = 200
+export const MAX_ANSWERS = 89
+export const MAX_INDEPENDENT_ANSWERS = 89
```
jahooma, line 40: I wouldn't change this. Right now, people are submitting new answers to e.g. the manifest prob bets market that has > 100 answers.

One idea would be to judge the market tier based on liquidity *per answer* so that adding more answers wouldn't change the tier.

I'm not that confident here. It's a new idea to scale answer cost by the existing liquidity, so I don't know if users want that or if they prefer a fixed cost per answer.

ingawei: Yeah but if that's the case then I should just add tiers to the backend then right? This was my fix if we could figure out a simple solution so that tiers could be sorter/filtered in an easy way, but it seems like because of the variable answer cost that there really isn't a straightforward frontend solution?

jahooma: Hmm, yes, it's not as simple, but you could calculate the tier from the contract, assuming the `answers` field is up-to-date?

jahooma: Try the last idea of finding the answer with most `totalLiquidity` and using that. (Possibly with a multiplier?)

ingawei: Ehh I just feel like it's kind of gross, particularly for the use case of searching through a lot of markets, because like some you would just look at total liquidity but for multi choice it's this weird metric? Just feels gross lol

PR #2669, opened 2024-06-11
added dropdownz

ingawei:
image
Can now filter by any number of tierz. Used binary to determine tier so as to not making the url querying too complicated, not sure if that's too hacky


backend/shared/src/supabase/search-contracts.ts
```diff
@@ -24,24 +24,40 @@ import {
...
+  filter: string
+  contractType: string
+  limit: number
+  offset: number
+  sort: 'score' | 'freshness-score'
+  isPrizeMarket: boolean
+  marketTier: TierParamsType
+  privateUser?: PrivateUser
+  threshold?: number
+}) {
```
jahooma, line 46: Thanks for making this a params object!

jahooma (approved):
Awesome, looks great! I still think we should add the other two tiers as filter options, but it's not blocking

web/components/search/contract-filters.tsx
```diff
@@ -90,6 +93,15 @@ export function ContractFilters(props: {
     })
   }
 
+  const toggleTier = (tier: MarketTierType) => {
+    const tierIndex = tiers.indexOf(tier)
+    if (tierIndex >= 0 && tierIndex < currentTiers.length) {
+      const tiersArray = currentTiers.split('')
+      tiersArray[tierIndex] = tiersArray[tierIndex] === '0' ? '1' : '0'
+      updateParams({ mt: tiersArray.join('') as TierParamsType })
```
jahooma, line 101: This binary string is a reasonable solution

PR #2675, opened 2024-06-14
Move private user reads to backend apis

sipec:
this is step 1 of migrating it
definitely take a look at the auth stuff
also I need to figure out how we're gonna keep the private user up to date. websockets wouldn't be secure (anyone can subscribe) so I guess we can poll just like for user :/

adapted from #2467

### DIFF
--- a/web/components/auth-context.tsx
+++ b/web/components/auth-context.tsx
@@ -23,7 +23,7 @@ import { safeLocalStorage } from 'web/lib/util/local'
 import { getSavedContractVisitsLocally } from 'web/hooks/use-save-visits'
 import { getSupabaseToken } from 'web/lib/firebase/api'
 import { updateSupabaseAuth } from 'web/lib/supabase/db'
-import { usePollUser } from 'web/hooks/use-user'
+import { usePollUser, useWebsocketPrivateUser } from 'web/hooks/use-user'
 import { useEffectCheckEquality } from 'web/hooks/use-effect-check-equality'
 import { getPrivateUserSafe, getUserSafe } from 'web/lib/supabase/users'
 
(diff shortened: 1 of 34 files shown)


web/hooks/use-users.ts
```diff
@@ -32,12 +31,9 @@ export const useRecentlyActiveUsersAndPrivateUsers = (limit: number) => {
   >()
   const loadUsers = async () => {
     const users = await getRecentlyActiveUsers(limit)
-    const privateUsers = await Promise.all(
-      users.map(async (u) => getPrivateUser(u.id))
-    )
```
jahooma, line 37: Why not load private users here with api here?

web/components/auth-context.tsx
```diff
@@ -211,17 +206,6 @@ export function AuthProvider(props: {
...
-    if (authLoaded && uid) {
-      const privateUserListener = listenForPrivateUser(uid, (privateUser) => {
-        setPrivateUser(privateUser ?? undefined)
-      })
-      return () => {
-        privateUserListener()
-      }
-    }
-  }, [authLoaded, uid])
-
```
jahooma, line 224: Uh, shouldn't we replace this with polling in the same PR?

I think we could also figure out auth'd websockets by sending encrypted messages with asymmetric key encryption. Which apparently @mqp did for supabase channels before we switched to websockets.

jahooma (commented):
Mostly looks good!

web/hooks/use-users.ts
sipec: i am not convinced the api will be authed when called in getServerSide props

web/components/auth-context.tsx
sipec: or we could just not send the info at all, and expect the client to refetch 🤔

web/hooks/use-users.ts
jahooma: I mean useEffects are not run on the server so yes.

But this whole hook doesn't make sense. It can't be called something to get private users and then always return undefined lol. Then just remove the private users part.

PR #2676, opened 2024-06-14
N/A all contracts on independent multi

SirSaltyy:


web/components/answers/answer-resolve-panel.tsx
```diff
@@ -419,12 +422,49 @@ export const IndependentAnswersResolvePanel = (props: {
...
+            console.error(e)
+            setError('Error resolving question')
+          }
+        }
+      }
+      console.log('All contracts canceled')
+    } catch (e) {
+      console.error(e)
+      setError('Error canceling all contracts')
+    }
```
jahooma, line 460: You don't need the outer try catch if you have the inner one around each api call.

web/components/answers/answer-resolve-panel.tsx
```diff
@@ -419,12 +422,49 @@ export const IndependentAnswersResolvePanel = (props: {
...
+  const answers = useAnswersCpmm(contract.id) ?? contract.answers
+  const isAdminorMod = useAdminOrMod()
+
+  const [isSubmitting, setIsSubmitting] = useState(false)
+  const [error, setError] = useState<string | undefined>(undefined)
+
+  const onCancelAll = async () => {
+    setIsSubmitting(true)
+    try {
+      for (const answer of answers) {
```
jahooma, line 434: Ask chat gpt to turn this for loop into a Promise.all. This will make it only a bit faster because the server will still resolve each answer sequentially.

web/components/answers/answer-resolve-panel.tsx
```diff
@@ -419,12 +422,49 @@ export const IndependentAnswersResolvePanel = (props: {
 
   const isAdmin = useAdmin()
   const user = useUser()
+  const answers = useAnswersCpmm(contract.id) ?? contract.answers
+  const isAdminorMod = useAdminOrMod()
```
jahooma, line 426: name the variable `isAdminOrMod`

web/components/answers/answer-resolve-panel.tsx
```diff
@@ -419,12 +422,49 @@ export const IndependentAnswersResolvePanel = (props: {
...
+    }
+    setIsSubmitting(false)
+  }
 
-  const { answers, addAnswersMode } = contract
   const sortedAnswers = sortBy(
     answers,
     (a) => (a.resolution ? -a.subsidyPool : -Infinity),
-    (a) => (addAnswersMode === 'ANYONE' ? -1 * a.prob : a.index)
+    (a) => a.index
```
jahooma, line 467: Are you sure we want this? Actually, we should try to use the same sort as the answers panel.

Use the `sortAnswers` function from common/answer instead of all this.

jahooma (approved):
Nice! Approved, pending a few comments.

web/components/answers/answer-resolve-panel.tsx
ProgramCrafter: Won't that hit by-IP rate limit?

jahooma: Good question! I don't know our exact limits, but I think a hundred requests in a short period would be ok.

SirSaltyy: I did this (see recent commit), but I think it has introduced an API error.

APIError: High volume of requests. Please try again later.

It managed to get through about 3-4 N/As each time before it then failed to resolve the rest.

PR #2677, opened 2024-06-14
Migrate private user writes to Supabase

sipec:
need to test and write migration script


jahooma (approved):
LGTM! Code seems reasonable. Good luck with the merge.

I like the solution for the websockets to ping when the private user has changed, but not leak any info about it and have the client refetch.

PR #2683, opened 2024-06-25
Delete all private groups

sipec:
I've already N/A'ed all remaining multi, canceled all private bounties, and closed all private polls. Also, everyone who asked to keep their private markets / groups has already had their markets turned unlisted and their group turned curated.
so running this script should just take care of private groups once and for all

- comments are deleted
- remaining markets get N/A'ed
- contracts and answers get redacted
- private groups are literally deleted

### DIFF
--- a/backend/scripts/delete-private-groups.ts
+++ b/backend/scripts/delete-private-groups.ts
@@ -31,7 +31,7 @@ runScript(async ({ pg }) => {
       continue
     }
     await resolveMarketHelper(contract, sinclair, creator, {
-      outcome: 'CANCELLED',
+      outcome: 'CANCEL',
     })
     await setAdjustProfitFromResolvedMarkets(contract.id)
   }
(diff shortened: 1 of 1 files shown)


backend/scripts/delete-private-groups.ts
```diff
@@ -0,0 +1,93 @@
...
+  )
+
+  for (const contract of unresolvedContracts) {
+    const creator = await getUser(contract.creatorId)
+    if (!creator) {
+      console.error(`Creator not found for contract ${contract.id}`)
+      continue
+    }
+    await resolveMarketHelper(contract, sinclair, creator, {
+      outcome: 'CANCELLED',
```
jahooma, line 34: Wasn't it "CANCEL"?

backend/scripts/delete-private-groups.ts
```diff
@@ -0,0 +1,93 @@
...
+    || jsonb_build_object('slug', 'private-auto-deleted-' || random_alphanumeric(8))
+    || jsonb_build_object('lastUpdatedTime', ts_to_millis(now()))
+    where visibility = 'private'
+    and (deleted != true or resolution is null)`
+  )
+  // note that answers will get overwritten by the denormalizer, but we set to empty to prevent data leakage.
+
+  console.log('deleting all private groups')
+  // copied from scripts/delete-group.ts
+  const groups = pg.many(
```
jahooma, line 72: nit: I would name this variable groupIds

backend/scripts/delete-private-groups.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  console.log(`overwriting and marking deleted all private contracts`)
+  await pg.none(
+    `update contracts
+    set data = data || '{
+        "wasPrivate": true,
+        "visibility": "unlisted",
+        "isRanked": false,
+        "deleted": true
+        "question": "Private Question (deleted)",
```
jahooma, line 57: I think we should ovewrite the slug as well

jahooma: Nvm I see you did

jahooma (approved):
Thanks Sinclair, looks good!

Was there anything in the notifications that we wanted to scrub? Might be worth removing the contract question if that was in plain text in a notification.

backend/scripts/delete-private-groups.ts
sipec: it is overwritten on line 63

PR #2684, opened 2024-06-26
Delete replicator

sipec:
need to also delete from gcloud and am not sure how


mqp:
Don't forget to clean up the SQL stuff and the Firestore write triggers. 

For deleting the gcloud infrastructure, it should be easy because the replicator is just a single instance and doesn't have anything fancy. If you just click into the VM instances list and terminate the instance, that should be basically it. There's also a Cloud Scheduler cronjob to replay failed writes every minute that you should delete.

sipec:
looks like someone already stopped the replicator vm

PR #2685, opened 2024-06-26
Make reset betting streak much faster/lighter

sipec:
no longer needs 4gb and runs in 5 seconds

image


sipec:
pushed to dev to test

jahooma (approved):
LGTM!

PR #2698, opened 2024-07-02
Transfer ownership script

SirSaltyy:
Created new script that uses Supabase.

Have not been able to run it yet due to RLS.

### DIFF
--- a/backend/scripts/transfer-contract-ownership.ts
+++ b/backend/scripts/transfer-contract-ownership.ts
@@ -1,36 +1,134 @@
-import { SafeBulkWriter } from 'shared/safe-bulk-writer'
+import {
+  createSupabaseClient,
+  createSupabaseDirectClient,
+} from 'shared/supabase/init'
+
 import { runScript } from './run-script'
-import { convertUser } from 'common/supabase/users'
+import yargs from 'yargs'
+import { hideBin } from 'yargs/helpers'
+
+// Define types for command-line arguments
+interface Args {
+  slugs: string[]
+  toUserId: string
+}
+
+// Define types for the fetched data
+interface User {
+  id: string
+  username: string
+  name: string
+  created_time: string
+}
+
+interface Contract {
+  id: string
+  data: any
+}
+
+// Initialize Supabase client
+const supabaseClient = createSupabaseClient()
+
+// Parse command-line arguments
+const argv = yargs(hideBin(process.argv))
+  .option('slugs', {
+    alias: 's',
+    type: 'array',
+    description: 'List of slugs to transfer',
+    demandOption: true,
+  })
+  .option('toUserId', {
+    alias: 'u',
+    type: 'string',
+    description: 'User ID to transfer contracts to',
+    demandOption: true,
+  })
+  .parseSync() as Args
+
+const BATCH_SIZE = 10 // Adjust batch size as needed
 
 if (require.main === module) {
-  runScript(async ({ pg, firestore }) => {
-    const fromUserId = 'zgCIqq8AmRUYVu6AdQ9vVEJN8On1'
-    const contractIdsToTransfer = await pg.map(
-      `select id from contracts where creator_id = $1`,
-      [fromUserId],
-      (row) => row.id
-    )
-    const toUserId = '4juQfJkFnwX9nws3dFOpz4gc1mi2'
-    const toUser = await pg.one(
-      `select * from users where id = $1`,
-      [toUserId],
-      (row) => convertUser(row)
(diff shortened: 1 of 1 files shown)


backend/scripts/transfer-contract-ownership.ts
```diff
@@ -1,37 +1,117 @@
...
+  name: string
+  created_time: string
+}
+
+interface Contract {
+  id: string
+  data: any
+}
+
+const supabaseDirectClient = createSupabaseDirectClient()
```
jahooma, line 24: This initializes before the secret keys are loaded from runScript. You are supposed to use the pg passed from runscript:
`runScript(async ({ pg }) => {`

backend/scripts/transfer-contract-ownership.ts
```diff
@@ -1,37 +1,117 @@
...
+interface User {
+  id: string
+  username: string
+  name: string
+  created_time: string
+}
+
+interface Contract {
+  id: string
+  data: any
```
IanPhilips, line 21: Using `any` isn't great code style. You should use follow the pattern of `getContracts` found elsewhere in the codebase.

backend/scripts/transfer-contract-ownership.ts
IanPhilips: +1

backend/scripts/transfer-contract-ownership.ts
```diff
@@ -1,37 +1,117 @@
...
-        creatorName: toUser.name,
-        creatorCreatedTime: toUser.createdTime,
-      })
-    }
-    await writer.close()
+  runScript(async () => {
+    const { slugs: slugsToTransfer, toUserId } = argv
+
+    try {
+      const toUserData = await supabaseDirectClient.oneOrNone<User>(
```
IanPhilips, line 48: Why not just use `getUser`  in backend/shared/src/utils.ts

backend/scripts/transfer-contract-ownership.ts
```diff
@@ -1,37 +1,117 @@
...
+
+      for (let i = 0; i < contractsToTransfer.length; i += BATCH_SIZE) {
+        const batch = contractsToTransfer.slice(i, i + BATCH_SIZE)
+
+        const updates = batch.map(({ id, data }) => {
+          const description = data.description || ''
+          const creatorUsername = data.creatorUsername || 'Unknown creator'
+          const creatorName = data.creatorName || 'Unknown name'
+
+          const updatedDescription = `${
```
IanPhilips, line 84: This is not going to work with tiptap. If you really want to do this you have to inspect a tiptap json object and add a new object into the data that is of the type and value you want, i.e. paragraph.

backend/scripts/transfer-contract-ownership.ts
```diff
@@ -1,37 +1,117 @@
...
+              ...data,
+              description: updatedDescription,
+              creatorUsername: toUser.username,
+              creatorName: toUser.name,
+              creatorAvatarUrl: data.creatorAvatarUrl,
+            },
+          }
+        })
+
+        const upsertQuery =
```
IanPhilips, line 102: I would just use the simpler `pg.none( update contracts set ...)` and update each row individually as you're already batching the contracts.

backend/scripts/transfer-contract-ownership.ts
```diff
@@ -1,37 +1,117 @@
...
+          const creatorName = data.creatorName || 'Unknown name'
+
+          const updatedDescription = `${
+            description ? description : ''
+          }\n\nThis market was created by ${creatorUsername} (${creatorName}) and has changed ownership.`
+
+          return {
+            id,
+            creator_id: toUser.id,
+            created_time: toUser.created_time,
```
IanPhilips, line 91: Why are you setting the contract created time to the user's created time?

backend/scripts/transfer-contract-ownership.ts
```diff
@@ -1,37 +1,117 @@
...
+            data: {
+              ...data,
```
IanPhilips, line 93: you should only update the fields you want to update, and not rewrite the entire data column. Someone could bet in between when you read and write and now you've overwrriten their change to the contract with old data.

IanPhilips (changes requested):
Nice, this is getting closer! Just a few adjustments and you'll be done

SirSaltyy:
@IanPhilips should all be good now. I tested it and it worked well.

IanPhilips (approved):
Looks good!

PR #2700, opened 2024-07-02
ig

ingawei:
Put first 8 topics in a dropdown on home page
image

Please check my work I am feeling under the weather so my brain is no good at the moment

### DIFF
--- a/web/components/supabase-search.tsx
+++ b/web/components/supabase-search.tsx
@@ -140,6 +140,7 @@ export type SearchParams = {
   [PRIZE_MARKET_KEY]: BinaryDigit
   [FOR_YOU_KEY]: BinaryDigit
   [MARKET_TIER_KEY]: TierParamsType
+  [TOPIC_FILTER_KEY]: string | undefined
 }
 
 export const QUERY_KEY = 'q'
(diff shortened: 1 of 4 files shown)


web/components/supabase-search.tsx
```diff
@@ -567,6 +582,7 @@ const useContractSearch = (
       p: isPrizeMarketString,
       fy: forYou,
       mt: marketTier,
+      tf: topicFilter,
```
jahooma, line 585: You're adding tf as a parameter, but wouldn't you need to modify the backend to accept it?

Also, there already exists the `topicSlug` parameter. It would be good to reuse that if possible

jahooma (commented):
Nice, this is great!

A couple notes:
- Lots of users want to be able to choose all, i.e. not for you and not a topic. Could we add that as an option?
- Is there a way to look through more topics? E.g. AI and other possibly important ones are missing
- It might be weird if the top topics change over time (based on importance score) and users can't select a topic anymore

Otherwise, see my other comment about needing to modify the backend to support `tf`

web/components/supabase-search.tsx
ingawei: No, see the code further below (line 615)
```
   const newContracts = await searchContracts({
        term: query,
        filter,
        sort,
        contractType,
        offset: freshQuery ? 0 : state.contracts?.length ?? 0,
        limit: CONTRACTS_PER_SEARCH_PAGE,
        topicSlug:
          topicSlug !== ''
            ? topicSlug
            : topicFilter !== ''
            ? topicFilter
            : undefined,
        creatorId: additionalFilter?.creatorId,
        isPolitics: additionalFilter?.isPolitics,
        isPrizeMarket: isPrizeMarketString,
        marketTier,
        forYou,
      })
      ```
      
      I didn't wanna use topicSlug because i didn't want to mess with the topic page usecase too much

jahooma: Ah, ok, I guess this is fine!

PR #2706, opened 2024-07-06
Add method for calling Stability API for image generation

wearsshoes:


wearsshoes:
Notes on this:
- Needs a stability API key in env.
- SDXL runs cost about 0.6¢ per run. Sd-ultra runs cost way more, 8¢ per run. As written, default sd-ultra generation is enabled only for market tiers plus and up; SDXL for cheap markets. For comparison Dall-E 3 is 6¢ per run, and does fall between the two in terms of quality.  
- Might be nice to eventually add methods for uploading or regenerating bad banner images by paying mana later.

wearsshoes:
Potential upgrade: add animated workflows via some serverless comfyui backend (possibly https://comfy.icu for ease of config, but it has a minimum allocation of $10/mo.) Cost per generation would probably be in the range of 50¢ per market, so probably reserved for premium and crystal markets.

wearsshoes:
Don’t merge, just close, as it’s almost certainly broken by now. Shouldn’t have put in a pull for something connecting to a API which is still under development, things you learn

PR #2716, opened 2024-07-11
Ingawei/answer adjustment

ingawei:
Changed metric to be profit
image

Moved metrics to 3 dot menu
image


IanPhilips (approved):
I like the new dropdown! It's nice to not have the UI elements change the page layout when you tap them 😅. I'm unsure about the lack of spent/shares indication in favor of profit, though. I'd personally like to see how many shares I have, and people on discord are asking for their spent figures back. 

I do like the more prominent, indigo text tho!

PR #2718, opened 2024-07-12
Move private messages to api

IanPhilips:
Loads messages which seems to be broken w/ the supabase client

### DIFF
--- a/web/hooks/use-persistent-supabase-polling.ts
+++ b/web/hooks/use-persistent-supabase-polling.ts
@@ -106,6 +106,92 @@ export function usePersistentSupabasePolling<T extends TableName>(
   return results
 }
 
+export function usePersistentApiPolling<T extends TableName>(
+  table: T,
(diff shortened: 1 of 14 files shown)


web/hooks/use-persistent-supabase-polling.ts
```diff
@@ -106,6 +106,92 @@ export function usePersistentSupabasePolling<T extends TableName>(
   return results
 }
 
+export function usePersistentApiPolling<T extends TableName>(
```
jahooma, line 109: Would be good to one day switch to websockets.

For private data, I know you don't want to send the messages over it, but you can still send a ping that there's some update to your private messages, and then query from the client (just like when you poll and it fetches the current status).
This is how we do updates for private user, btw.

IanPhilips: ah thanks for letting me know! Good idea. One day.

PR #2725, opened 2024-07-17
Ingawei/make pages swipeable

ingawei:
https://www.loom.com/share/20f69a3064ec44ec9555cc1f6d30640d?sid=9facc4d9-d03a-4074-be61-26aa839952fd

^ Added the paginated answers to a carousel bc ppl were complaining that the pages were jumping around. Also allows for scrolling left and right with scroll/finger instead of button

### DIFF
--- a/web/components/answers/answers-panel.tsx
+++ b/web/components/answers/answers-panel.tsx
@@ -250,7 +250,6 @@ export function AnswersPanel(props: {
       const itemWidth = carouselRef.offsetWidth + 16
       carouselRef.scrollTo({
         left: index * itemWidth,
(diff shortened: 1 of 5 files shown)


IanPhilips:
That's cool! I do think maybe the paginated experience isn't ideal, though. Users like to quickly see either the top options, or ALL of the options, not page 2,3,4,5, etc. of the options. It does suck to have to scroll through all of the options on mobile, though. Maybe you could add an expand all button and then a floating/sticky collapse all button that would let people skim them and then collapse them when done?

IanPhilips:
Closing fur now

PR #2737, opened 2024-07-24
Do all of create market within a transaction

sipec:
The most important thing is that running the create market txn and create liquidity antes happen in the same transaction

I was considering *just* putting the finance stuff in the same tx, and leaving create group etc out of it, but i think that if the client shouldn't get an error if the create market succeeds. we could put things in a continuation i suppose, but it seems unnecessary performance wise


jahooma (approved):
Thanks Sinclair!

PR #2739, opened 2024-07-24
Backfill missing liquidity and reimburse resolved markets affect by the bug

sipec:
https://www.notion.so/manifoldmarkets/Market-creation-outage-d6c0ed5a001a4057bcb89f82787a5b84

### DIFF
--- a/backend/api/src/create-market.ts
+++ b/backend/api/src/create-market.ts
@@ -567,7 +567,7 @@ async function getLoveAnswerUserIds(answers: string[]) {
   )
 }
 
-async function generateAntes(
+export async function generateAntes(
   pg: SupabaseDirectClient,
   providerId: string,
   contract: Contract,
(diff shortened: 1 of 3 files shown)


backend/api/src/create-market.ts
```diff
@@ -615,7 +615,7 @@ async function generateAntes(
     drizzledAmount > 0 &&
     (contract.mechanism === 'cpmm-1' || contract.mechanism === 'cpmm-multi-1')
   ) {
-    return await pg.tx(async (tx) => {
+    return await pg.txIf(async (tx) => {
```
jahooma, line 618: Ooh, don't know what this does

jahooma (approved):
Looks good!

backend/api/src/create-market.ts
sipec: it creates a new transaction iff it is not called from within another transaction

PR #2741, opened 2024-07-25
Adding an idempotency key for contract creation.

tarehart:
This is related to the bug discussed at https://manifoldmarkets.notion.site/Market-creation-outage-d6c0ed5a001a4057bcb89f82787a5b84

I'm intending to mitigate the effect of any future bugs by preventing duplicate contract creation.

This change is NOT properly tested on the backend because I don't have a suitable dev environment set up (is this possible?) but maybe the magic Vercel PR validation environment will help.

### DIFF
--- a/backend/api/src/create-market.ts
+++ b/backend/api/src/create-market.ts
@@ -20,6 +20,7 @@ import {
   NO_CLOSE_TIME_TYPES,
   OutcomeType,
   add_answers_mode,
+  contractUrl,
 } from 'common/contract'
 import { getAnte, getTieredCost } from 'common/economy'
 import { MAX_GROUPS_PER_MARKET } from 'common/group'
@@ -157,8 +158,9 @@ export async function createMarketHelper(body: Body, auth: AuthedUser) {
     : unmodifiedAnte
   const ante = Math.min(unmodifiedAnte, totalMarketCost)
 
-  if (await isDuplicateSubmission(idempotencyKey, pg)) {
-    throw new APIError(400, 'Contract has already been created')
+  const duplicateSubmissionUrl = await getDuplicateSubmissionUrl(idempotencyKey, pg)
+  if (duplicateSubmissionUrl) {
+    throw new APIError(400, 'Contract has already been created at ' + duplicateSubmissionUrl)
   }
 
   return await pg.tx(async (tx) => {
@@ -300,13 +302,16 @@ const runCreateMarketTxn = async (args: {
   }
 }
 
-async function isDuplicateSubmission(
+async function getDuplicateSubmissionUrl(
   idempotencyKey: string | undefined,
   pg: SupabaseDirectClient
-) {
-  if (!idempotencyKey) return false
+): Promise<string | undefined> {
+  if (!idempotencyKey) return undefined;
   const contracts = await getContractsDirect([idempotencyKey], pg)
-  return contracts.length > 0
+  if (contracts.length > 0) {
+    return contractUrl(contracts[0])
+  }
+  return undefined;
 }
 
 async function getCloseTimestamp(
--- a/backend/api/src/create-market.ts
+++ b/backend/api/src/create-market.ts
@@ -33,6 +33,7 @@ import { slugify } from 'common/util/slugify'
 import { getCloseDate } from 'shared/helpers/openai-utils'
 import {
   generateContractEmbeddings,
+  getContractsDirect,
   updateContract,
 } from 'shared/supabase/contracts'
 import {
@@ -103,6 +104,7 @@ export async function createMarketHelper(body: Body, auth: AuthedUser) {
     visibility,
     specialLiquidityPerAnswer,
     marketTier,
+    idempotencyKey,
   } = validateMarketBody(body)
 
   if (outcomeType === 'BOUNTIED_QUESTION') {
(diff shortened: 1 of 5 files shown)


akrolsmir:
Thanks for the PR, tagging @sinclairity for review! Quick uninformed flyby thoughts are that this makes sense & seems like a best practice though I'm not sure this issue warrants the extra complexity, handling this case might be a bit like fighting the last battle.

Testing against the Vercel dev build is pretty reasonable, though imo we should just hand out our GCP dev keys to any aspiring open source developers who ask -- any downsides to this @sinclairity?

akrolsmir:
Also just found this Discord thread with more info: https://discord.com/channels/915138780216823849/1265525070223118468

backend/api/src/create-market.ts
```diff
@@ -175,7 +181,8 @@ export async function createMarketHelper(body: Body, auth: AuthedUser) {
 
     const contract = getNewContract(
       removeUndefinedProps({
-        id: randomString(),
+        // Risk: idempotencyKey comes from the client, mischievous users could pass unexpected strings.
+        id: idempotencyKey ?? randomString(),
```
IanPhilips, line 185: This allows too much mischief for me! You could just have an in-memory map on the api so we don't have to enter any random strings into the db. It would be reset on api redeploys but this doesn't happen that often, and unlikely to happen between market submissions. Alternatively maybe the zod library can do more checks on the string to make sure it's safe, but I'm not sure.

IanPhilips (commented):
Overall though, I'm not sure this is worth fixing. Afaik we only run into this problem when we're experiencing database problems, which means these duplicate markets might be serving as a helpful sign something is wrong.

tarehart:
> Overall though, I'm not sure this is worth fixing. Afaik we only run into this problem when we're experiencing database problems, which means these duplicate markets might be serving as a helpful sign something is wrong.

Do you have any alarms on 5xx errors? That might be a more elegant way to get a signal, since there's a decent amount of pain associated with bugs like this
- User frustration / confusion
- Users' mana getting depleted
- Burden on mods to clean up duplicate markets or any other debris
- Burden on users to mitigate the effect of their duplicate markets, e.g. renaming and making them private
- Lingering mess, e.g. I still have duplicates I haven't complained about till now https://manifold.markets/tarehart?tab=questions

backend/api/src/create-market.ts
tarehart: I like the idea of using zod, maybe to do a simple regex validation. I can amend the PR with that if you're interested.

backend/api/src/create-market.ts
```diff
@@ -155,6 +157,10 @@ export async function createMarketHelper(body: Body, auth: AuthedUser) {
     : unmodifiedAnte
   const ante = Math.min(unmodifiedAnte, totalMarketCost)
 
+  if (await isDuplicateSubmission(idempotencyKey, pg)) {
+    throw new APIError(400, 'Contract has already been created')
```
IanPhilips, line 161: Would be great to return the url for them to visit, but I can add this later if that's too much trouble.

common/src/util/random.ts
```diff
@@ -1,9 +1,13 @@
 // max 10 length string. For longer, concat multiple
+// Often used as a unique identifier.
 export const randomString = (length = 10) =>
   Math.random()
     .toString(36)
     .substring(2, length + 2)
 
+// Matches the output of the randomString function, for validation purposes.
+export const randomStringRegex = /[0-9a-z]+/
```
IanPhilips, line 9: Nice, looks like it's just a bit off, it should be:
`export const randomStringRegex = /^[0-9a-z]+$/`

web/components/new-contract/new-contract-panel.tsx
```diff
@@ -50,6 +51,10 @@ export function NewContractPanel(props: {
     params?.outcomeType ? 'filling contract params' : 'choosing contract'
   )
 
+  // Don't use the usePersistentLocalState hook for this, because there's too high a risk that it will survive in local storage
+  // longer than it should under a trivial paramsKey like '', and improperly prevent users from creating any new contracts.
+  const [idempotencyKey] = useState(randomString())
```
IanPhilips, line 56: I think it'd be fine to just add this key to the `ContractParamsForm` component rather than pass the prop down.

IanPhilips (commented):
Thanks for pushing through, this looks almost done!

backend/api/src/create-market.ts
tarehart: Will do!

common/src/util/random.ts
tarehart: Good catch, will fix.

web/components/new-contract/new-contract-panel.tsx
tarehart: Sure, I'll do that. 

My purpose for defining it here is to make the idempotencyKey stable even if the user goes back and re-chooses the contract type, which they may likely do when banging around to try and make an error go away. If you feel that's a compelling reason, I'll put it back.

IanPhilips: Oh I see, as long as we provide them with the url of their newly minted contract I bet they won't go banging around.

tarehart: Yeah fair point! I did add it to the ContractParamsForm component rather than pass the prop down, per your suggestion. Does it look good at this point?

IanPhilips (approved):
Looks great, now! will merge this in. Thanks @tarehart!

PR #2750, opened 2024-07-28
api.md: Update /v0/comments limit parameter default

michaelskyba:
Official instance shows
```
{"message":"Error validating request.","details":[{"field":"limit","error":"Number must be less than or equal to 1000"}]}
```


IanPhilips:
Thanks!

PR #2755, opened 2024-07-31
Write to answer native columns and remove trigger

sipec:
to migrate, these commits should be deployed/run serially. (1st write native column and data, 2nd drop the populate trigger, 3rd stop writing to data).

we don't actually add house liquidity to answers ever so i was a bit lazy with that code.


jahooma (approved):
Looks good, thanks Sinclair!

sipec:
rebase merged

PR #2756, opened 2024-07-31
Have client calculate dependencies and pass to API

IanPhilips:
This has the client calculate its deps, letting us avoid the case where the API bricks bc it can't get a pg connection from the pool bc of too many concurrent website users. This doesn't handle the case where the API server is being hammered from bots/API users and runs out of connections trying to simulate the deps. We could use the cache from the bet-cache branch if we think this case is worth handling now with the extra complexity.

Screenshot 2024-07-31 at 3 37 33 PM


jahooma (approved):
LGTM! This is a nice idea

PR #2757, opened 2024-08-01
Redeem maker shares in continuation via queue

IanPhilips:
We may want to add a new `system` level timeout to the bet queue so that these redemption transactions wait as long as they need to in order to run. We probably also want to run the bettor's `redeemShares` in the continuation so their bet returns faster and the lock on the contract frees sooner.


jahooma:
@devloai -- please review this PR

PR #2760, opened 2024-08-02
Start limit order column migration

IanPhilips:


common/src/supabase/bets.ts
```diff
@@ -8,4 +8,6 @@ export const convertBet = (row: Row<'contract_bets'>) =>
     updated_time: tsToMillis,
     created_time: tsToMillis,
     answer_id: (a) => (a != null ? a : undefined),
+    is_filled: false,
+    is_cancelled: false,
```
IanPhilips, line 12: will revert these when the migration is done

PR #2764, opened 2024-08-04
Search users by username first

benmanns:
I ran into an issue where users with short usernames sometimes don't get returned from the search-users API in the first few rows. This is a problem, as the @ dialog in the comment box uses search-users with a limit of 6, so there is sometimes no way to mention a user, even with an exact username match. This change adds a query to search for an exact match username and present that as the first result, if present.

Note: I've not gotten manifold running locally, so I haven't been able to test this.

### DIFF
--- a/backend/api/src/supabase-search-users.ts
+++ b/backend/api/src/supabase-search-users.ts
@@ -21,14 +21,16 @@ export const searchUsers: APIHandler<'search-users'> = async (props, auth) => {
 
   const offset = page * limit
   const userId = auth?.uid
+  const searchExactSQL = getSearchExactUserSQL({ term })
   const searchFollowersSQL = getSearchUserSQL({ term, offset, limit, userId })
   const searchAllSQL = getSearchUserSQL({ term, offset, limit })
-  const [followers, all] = await Promise.all([
+  const [exact, followers, all] = await Promise.all([
+    pg.map(searchExactSQL, null, convertUser),
     pg.map(searchFollowersSQL, null, convertUser),
     pg.map(searchAllSQL, null, convertUser),
   ])
 
-  return uniqBy([...followers, ...all], 'id')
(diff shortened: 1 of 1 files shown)


IanPhilips:
Hey thanks for submitting this! Could you share a username this works for that our current search doesn't work for?

benmanns:
I was trying to tag [Vo username \@d](https://manifold.markets/d) on [this market](https://manifold.markets/market/will-djokovic-get-olympic-gold-in-2).

image

image

IanPhilips:
It's a bit more complicated because if I type zvi it should match ZviMowshowitz and not 'zvi', so not sure how to handle that yet

jahooma:
I support the exact match use case, even with the downside that it might not surface the most popular person.

One solution would be to put the exact match after the first other result!

benmanns:
Yeah, the problem I was trying to fix was mostly that it's currently impossible to tag `@d`, even with an exact name match, and there's not really anything to do to get more specific. I do think there could be more work put into the scoring of results, but as someone who has worked on search scoring, you can spend a decade on it and still not feel done.

For the ZviMowshowitz example, even if someone registered Z, Zv, Zvi, ZviM, ZviMo, ZviMow, ... you could still tag the full username if you typed the whole thing out. It's an interesting problem, though.

Some other example sites, on X if you start typing `@e`, the first result is `@elonmusk`, even though `@e` exists. Same for the rest of the prefixes (`@e` through `@elonmus`).

image

However, I would also note that if I just left it as-is, the tweet would mention `@e`, which is different from Manifold where you have to select from the dropdown to trigger a mention. Perhaps it would be a better solution to tag the exact match if it exists on typing @..., even without selecting from the suggestions.

IanPhilips:
I like the suggestion of making it auto-mention the exact match if no other is selected!

PR #2776, opened 2024-08-11
New watermark on SVG graphs.

tarehart:
Updating the watermark:
- In the background, behind the graph data
- Larger, so you can still read it if the graph data is in front of it
- Included the logo

Related suggestions in discord:
- [The watermark blocks some of the content](https://discord.com/channels/915138780216823849/1269095533956563014/1269095533956563014)
- [I appreciate the addition of a watermark but obscuring the chart with branding is bad](https://discord.com/channels/915138780216823849/1263557326326206554/1268663914669932675)

image

image

There's another file, probability-needle.tsx, that has the old watermark; I haven't touched it because I think it looks fine on the probability needle.

### DIFF
--- a/web/components/charts/helpers.tsx
+++ b/web/components/charts/helpers.tsx
@@ -477,6 +477,29 @@ export const SVGChart = <X, TT extends { x: number; y: number }>(props: {
           </clipPath>
         </defs>
 
+        {!noWatermark && (
+          <svg x={10} y={h - 35} opacity={0.15}>
+            <g transform='scale(1.5)'>
+              <path
+                d="M5.24854 17.0952L18.7175 6.80301L14.3444 20M5.24854 17.0952L9.79649 18.5476M5.24854 17.0952L4.27398 6.52755M14.3444 20L9.79649 18.5476M14.3444 20L22 12.638L16.3935 13.8147M9.79649 18.5476L12.3953 15.0668M4.27398 6.52755L10.0714 13.389M4.27398 6.52755L2 9.0818L4.47389 8.85643M12.9451 11.1603L10.971 5L8.65369 11.6611"
+                strokeLinecap="round"
+                strokeLinejoin="round"
+                fill='none'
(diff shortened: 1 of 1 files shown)


tarehart:
Hard for me to guess why dev failed deployment, could someone please provide logs?

jahooma:
It was a flaky error for our /stats page, which is unrelated!

ingawei (commented):
Hello! I see the vision behind this change, but am hesitant to approve because there are just too many instances where this kind of watermark will be entirely illegible, kind of defeating the purpose:
image

tarehart:
Good call, that example scribbled it out worse than I imagined. Philosophically I think the data should take priority, but I guess we could explore putting the watermark above the graph so they never conflict, does that sound good?

tarehart:
It'll take me a while to experiment with a watermark near the time range controls since there are a bunch of code paths to change and test, and I'm a little less confident about what would look good. Might be a week before I make time.

Should we consider merging this in the meantime as an incremental improvement?

tarehart:
Side note, as an outside contributor with no GOOGLE_APPLICATION_CREDENTIALS_DEV, it's hard to test the contract page. I've resorted to temporarily using a regular db here rather than an adminDb: https://github.com/manifoldmarkets/manifold/blob/8d62a25541e135b20dc20559d62548c927be8cc6/web/pages/%5Busername%5D/%5BcontractSlug%5D.tsx#L96-L97

What's the reason for this distinction, and is adminDb actually needed for the contract page, which appears to be read only?

tarehart:
Looking at the mobile dimensions again, I'm skeptical that I can make the alternative placement look good in the available space. 

I think this change as it stands is a clear improvement, and I suggest merging it. The people suffer https://discord.com/channels/915138780216823849/1269095533956563014/1280347805101326429

IanPhilips:
It looks good on the binary graphs but is basically invisible on the MC graph
Screenshot 2024-09-25 at 10 30 40 AM

PR #2781, opened 2024-08-14
prize cash backend architecture

sipec:


backend/shared/src/create-cash-contract.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const contract = getNewContract({
+      id: randomString(),
+      ante: subsidyAmount,
+      token: 'CASH',
+      description: htmlToRichText('<p></p>'),
+      initialProb: manaContract.prob,
+
+      creator,
+      slug: manaContract.slug,
```
jahooma, line 58: Probably good to make this unique as other code assumes so. Can add as suffix '-cash'

PR #2782, opened 2024-08-15
Ingawei/move messages to notifs

ingawei:


IanPhilips:
Heyo! I'm trying out the preview and when I send myself a message from another account I don't see a bubble pop up

PR #2789, opened 2024-08-20
Add ability to filter positions by answer through the API

### LINKED ISSUE #2785: Add `AnswerId` parameter to `positions` API
As described in [this bounty](https://manifold.markets/case/handle-multiple-choice-answers-in-c), currently there is no way to use the API to get the list of positions on a specific answer in a multi-answer market. The simplest way to resolve this would probably be to add an `AnswerId` parameter to the `positions` endpoint and pass it through from the `getPositions` API handler to `getOrderedContractMetricRowsForContractId`, which already supports filtering by `AnswerId`.

carrutstick:
Resolves #2785.

Note: I have not yet been able to test this commit.


IanPhilips:
Seems like it should work! Let's try it

PR #2824, opened 2024-09-03
Allow banned users to trade

SirSaltyy:
Also added native popup message when deleted/banned from trading users are automatically logged out to let them know why.


SirSaltyy:
hmm it was fine, I fully tested it. Not sure why such simple changes are failing deployment...

Did I do something wrong with the merge?

mantikoros (approved):
LGTM

We should implement bannedFromTrading as a user attr soon tho

PR #2828, opened 2024-09-04
buttonz

ingawei:
Made sweeps redeem button always visible, but if you cannot redeem for any reason a modal pops up that either shows you why or gives you actionable items like registering
image
image

I'm not sure if my interpretation of the messages and action items were accurate, so I would appreciate a second look please!


IanPhilips (approved):
LGTM! I just removed the location error bc it requires a private user to know and didn't feel like adding that everywhere. We can add that later

PR #2831, opened 2024-09-04
Give kyc rewards according to frozen net worth

sipec:
Resolves MAN-1499


backend/shared/src/distribute-kyc-bonus.ts
```diff
@@ -0,0 +1,48 @@
...
+    }>(
+      'select reward_amount, claimed from kyc_bonus_rewards where user_id = $1',
+      [userId]
+    )
+
+    if (data?.claimed) {
+      return
+    }
+
+    const reward = Math.max(data?.reward_amount ?? 0, MIN_KYC_REWARD)
```
mantikoros, line 21: MIN_KYC_REWARD should be SWEEPS_INITIAL_BONUS from economy.ts instead

Fetching from kyc_bonus_rewards is the special case; not the other way around

backend/supabase/kyc_bonus_rewards.sql
```diff
@@ -0,0 +1,17 @@
+-- This file is autogenerated from regen-schema.ts
+create table if not exists
+  kyc_bonus_rewards (
+    user_id text not null,
+    reward_amount numeric not null,
+    created_time timestamp with time zone default now(),
+    claimed boolean default false
```
mantikoros, line 7: we probably want to store claimed_time as well

mantikoros (commented):
LGTM, after requested changes

PR #2853, opened 2024-09-13
Trading ban

SirSaltyy:


SirSaltyy:
@mantikoros if u think someone else should review let me know. Also, I might need help resolving the conflict as I think the conflict editor is being dumb (or I am) and I don't want to accidentally undo someone's changes.

twitch-bot/common/types/manifold-api-types.ts
SirSaltyy, line 1: Don't mind this lol, doesn't matter.

SirSaltyy:
Oh, I just did a trading ban (rather than sweepcash and mana-specific), because I couldn't think of a single scenario where we'd want to ban someone from one and not just both.

Admin accounts not being allowed to trade sweepcash is covered by other logic.

mantikoros:
I can think of several cases where we'd only want to ban sweeps trading but not mana trading: Being an admin account is an overly restrictive condition. Future Manifold employees will not be admins. Non-employees who are connected to Manifold may also need to be banned. Users may also voluntarily request to be banned from trading as part of our responsible gaming  program.

mantikoros (changes requested):
Still think we should have separate bans for sweeps + mana, given the cases I described in a comment above. It's much easier to make this change now than when everything is in production and we have a more generic ban in use...

backend/api/src/ban-user-from-posting.ts
```diff
@@ -14,7 +14,7 @@ const bodySchema = z
   })
   .strict()
 
-export const banuser = authEndpoint(async (req, auth) => {
+export const banUserFromPosting = authEndpoint(async (req, auth) => {
   const { userId, unban } = validate(bodySchema, req.body)
   const db = createSupabaseDirectClient()
```
mantikoros, line 19: i believe our style is to call this "pg" instead of "db"

backend/api/src/ban-user-from-posting.ts
```diff
@@ -14,15 +14,15 @@ const bodySchema = z
   })
   .strict()
 
-export const banuser = authEndpoint(async (req, auth) => {
+export const banUserFromPosting = authEndpoint(async (req, auth) => {
   const { userId, unban } = validate(bodySchema, req.body)
```
IanPhilips, line 18: best to name properties in the positive valence, like `ban` rather than `unban`

backend/api/src/ban-user-from-mana.ts
```diff
@@ -0,0 +1,30 @@
...
+export const banUserFromMana = authEndpoint(async (req, auth) => {
+  const { userId, unban } = validate(bodySchema, req.body)
+  const pg = createSupabaseDirectClient()
+  await throwErrorIfNotMod(auth.uid)
+  if (isAdminId(userId)) throw new APIError(403, 'Cannot ban admin')
+  await trackPublicEvent(auth.uid, 'ban user from trading mana', {
+    userId,
+  })
+  await updateUser(pg, userId, {
+    isBannedFromMana: !unban,
```
IanPhilips, line 26: would prefer `isBannedFromManaTrading`

backend/api/src/ban-user-from-sweepcash.ts
```diff
@@ -0,0 +1,30 @@
...
+export const banUserFromSweepcash = authEndpoint(async (req, auth) => {
+  const { userId, unban } = validate(bodySchema, req.body)
+  const pg = createSupabaseDirectClient()
+  await throwErrorIfNotMod(auth.uid)
+  if (isAdminId(userId)) throw new APIError(403, 'Cannot ban admin')
+  await trackPublicEvent(auth.uid, 'ban user from trading sweepcash', {
+    userId,
+  })
+  await updateUser(pg, userId, {
+    isBannedFromSweepcash: !unban,
```
IanPhilips, line 26: `isBannedFromSweeipcashTrading`

backend/api/src/place-bet.ts
```diff
@@ -16,11 +16,15 @@ import { Answer } from 'common/answer'
 import { CpmmState, getCpmmProbability } from 'common/calculate-cpmm'
 import { ValidatedAPIParams } from 'common/api/schema'
 import { onCreateBets } from 'api/on-create-bet'
+<<<<<<< trading-ban
```
IanPhilips, line 19: fix these merge errors

backend/api/src/place-bet.ts
```diff
@@ -307,11 +319,16 @@ export const fetchContractBetDataAndValidate = async (
       'You must be kyc verified to trade on sweepstakes markets.'
     )
   }
+<<<<<<< trading-ban
```
IanPhilips, line 322: more merge indicators to remove

backend/api/src/place-bet.ts
```diff
@@ -307,11 +319,16 @@ export const fetchContractBetDataAndValidate = async (
...
+  if (user.userDeleted) {
+    throw new APIError(403, 'You are banned or deleted.')
```
IanPhilips, line 324: just deleted now

backend/api/src/place-bet.ts
```diff
@@ -322,7 +339,12 @@ export const fetchContractBetDataAndValidate = async (
   log(
     `Loaded user ${user.username} with id ${user.id} betting on slug ${contract.slug} with contract id: ${contract.id}.`
   )
-
+  if (user.isBannedFromMana && contract.token !== 'CASH') {
```
IanPhilips, line 342: might as well check for the mana token

web/lib/api/api.ts
```diff
@@ -312,15 +312,29 @@ export const updateMarket = curriedAPI('market/:contractId/update')
 
 export const updateUser = curriedAPI('me/update')
 
-export function banUser(params: { userId: string; unban?: boolean }) {
-  return call(getApiUrl('ban-user'), 'POST', params)
+export function banUserFromPosting(params: {
```
IanPhilips, line 315: This isn't how we create api endpoints anymore, but I guess this is okay. In the future you should use the `schema.ts`

web/lib/supabase/users.ts
```diff
@@ -50,10 +50,14 @@ export async function searchUsers(prompt: string, limit: number) {
...
   const { data } = await run(
-    selectFrom(db, 'users', ...defaultFields, 'isBannedFromPosting').in(
-      'id',
-      userIds
-    )
+    selectFrom(
+      db,
+      'users',
+      ...defaultFields,
+      'isBannedFromPosting',
```
IanPhilips, line 57: just add these to the default fields, right?

IanPhilips (changes requested):
Nice! Almost done

PR #2888, opened 2024-09-19
conditional orderbook tab and button

SirSaltyy:
One thing to note is that logged out mobile users still can't see the orderbook. This was an intentional choice as there wasn't a good spot to put it and I don't think a logged out mobile user cares that much.

Signed out desktop users, non-eligible sweeps signed-in desktop users, and non-eligible sweeps signed-in mobile users can now see the orderbook (although this is limited to binary markets rn - which should be fine since those are the only sweeps markets).

### DIFF
--- a/web/components/contract/contract-tabs.tsx
+++ b/web/components/contract/contract-tabs.tsx
@@ -10,7 +10,14 @@ import {
 } from 'common/antes'
 import { Bet } from 'common/bet'
 import { ContractComment } from 'common/comment'
-import { BinaryContract, Contract, CPMMNumericContract } from 'common/contract'
+import {
+  BinaryContract,
+  Contract,
+  CPMMMultiContract,
+  CPMMNumericContract,
+  PseudoNumericContract,
+  StonkContract,
+} from 'common/contract'
 import { buildArray } from 'common/util/array'
 import { shortFormatNumber, maybePluralize } from 'common/util/format'
 import { MINUTE_MS } from 'common/util/time'
@@ -43,6 +50,8 @@ import generateFilterDropdownItems from '../search/search-dropdown-helpers'
 import { useAPIGetter } from 'web/hooks/use-api-getter'
 import { api } from 'web/lib/api/api'
 import { TRADE_TERM } from 'common/envs/constants'
+import { useUnfilledBets } from 'web/hooks/use-bets'
+import { OrderBookPanel } from '../bet/order-book'
 
 export function ContractTabs(props: {
   mainContract: Contract
@@ -94,6 +103,8 @@ export function ContractTabs(props: {
     (totalPositions > 0 ? `${shortFormatNumber(totalPositions)} ` : '') +
     maybePluralize('Holder', totalPositions)
 
+  const unfilledBets = useUnfilledBets(liveContract.id) ?? []
+
   return (
     <ControlledTabs
       className="mb-4"
@@ -108,6 +119,8 @@ export function ContractTabs(props: {
               ? 'trades'
               : title === positionsTitle
               ? 'positions'
+              : title === 'Order book'
+              ? 'orderbook'
               : 'contract'
           } tab`
         )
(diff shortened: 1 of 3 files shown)


IanPhilips:
Could you post a screeny?

web/components/contract/contract-tabs.tsx
```diff
@@ -94,6 +103,8 @@ export function ContractTabs(props: {
     (totalPositions > 0 ? `${shortFormatNumber(totalPositions)} ` : '') +
     maybePluralize('Holder', totalPositions)
 
+  const unfilledBets = useUnfilledBets(liveContract.id) ?? []
```
IanPhilips, line 106: It'd be best to only load these bets if someone has toggled the order book, not on every page load.

IanPhilips (changes requested):

web/components/contract/twomba-contract-page.tsx
```diff
@@ -442,13 +453,45 @@ export function TwombaContractPageContent(props: ContractParams) {
...
             {!!user && (
-              <ContractSharePanel
-                isClosed={isClosed}
-                isCreator={isCreator}
-                showResolver={showResolver}
-                // TODO: upgrade tier
-                contract={props.contract}
-              />
+              <Col>
+                <Row className="items-center gap-2">
```
mantikoros, line 457: this seems like an awkward place to put the order book, but i'm not sure exactly where it should go

mantikoros (commented):
seems reasonable, but uncertain about design / layout

web/components/contract/twomba-contract-page.tsx
SirSaltyy: agreed but all the other places were worse, it's only there on mobile at least

web/components/contract/twomba-contract-page.tsx
```diff
@@ -442,13 +454,49 @@ export function TwombaContractPageContent(props: ContractParams) {
...
+                  {totalBets > 0 && (
+                    <div className="flex sm:hidden">
+                      <Button
+                        color="indigo-outline"
+                        size="lg"
+                        onClick={() => {
+                          setShowOrderBook((prev) => !prev)
+                          if (!showOrderBook) {
```
IanPhilips, line 466: what's this clause for?

web/components/contract/contract-tabs.tsx
```diff
@@ -94,6 +103,92 @@ export function ContractTabs(props: {
...
+    },
+    totalBets > 0 &&
+      liveContract.mechanism === 'cpmm-1' && {
+        title: 'Order book',
+        className: 'hidden md:block',
+        content: null,
+      }
+  )
+
+  const filteredTabs = tabs.filter(Boolean)
```
IanPhilips, line 165: I don't think you need this filter

web/components/contract/contract-tabs.tsx
```diff
@@ -94,6 +103,92 @@ export function ContractTabs(props: {
...
+    (tab) => tab.title === 'Order book'
+  )
+
+  const isOrderBookTabActive =
+    orderBookTabIndex >= 0 && activeIndex === orderBookTabIndex
+
+  const unfilledBets =
+    useUnfilledBets(liveContract.id, { enabled: isOrderBookTabActive }) ?? []
+
+  if (orderBookTabIndex >= 0) {
```
IanPhilips, line 177: what is this? Why isn't the order book simply in the tabs array as the `content` object?

IanPhilips (commented):
Almost done! still some odd bits in there

web/components/contract/contract-tabs.tsx
SirSaltyy: hmm I'm confused I thought this as well as the clause u asked about above in a comment made up the logic that makes it so useUnfilledBets only gets the unfilledbets once the conditions are met and prevent them from loading with the page

PR #2890, opened 2024-09-20
Fix answers not showing on comments

sipec:
prop-thread play market (comments) and live market (bets, answers) through ALL comment components 🍝💪
unfortunately this will increase rerenders

delete old main contract page and old header

resolves discord issue https://discord.com/channels/915138780216823849/915138780653051910/1286185264758128681


mantikoros (approved):
LGTM, separating play, live contracts seems like a reasonable approach

PR #2902, opened 2024-09-28
Go fee less

mantikoros:


IanPhilips (commented):
LGTM, altho I think we need to change any scoring mechanisms that use volume as a real metric. I do think this is a better user experience!

IanPhilips:
Pretty sure the tests are failing bc they were written with hardcoded numbers that assumed fees. @sinclairity wrote them, so she has the best context

PR #2933, opened 2024-10-08
Ingawei/man 1823 shift market tiers down a level

ingawei:
Run by doing
`SELECT * FROM update_contract_tiers_efficient(50, '1970-01-01 00:00:00'::timestamp);`

Shifts tiers down a level and gets rid of the 'Basic' tier, for simplicity and so that Mana and Cash markets can have tiers on the same scale (100, 1k, 10k, 100k). SQL script backfills old markets so they are the correct downgraded tier.


backend/scripts/shift-tiers/get-tier-from-liquidity-function.sql
```diff
@@ -0,0 +1,48 @@
...
+END IF;
+
+  -- For other types
+  IF liquidity >= ante * 100 THEN RETURN 'crystal';
+  ELSIF liquidity >= ante * 10 THEN RETURN 'premium';
+  ELSIF liquidity >= ante  THEN RETURN 'plus';
+  ELSE RETURN 'play';
+  END IF;
+END;
+$$ language plpgsql immutable;
```
IanPhilips, line 48: This logic looks good, but ofc be sure to test it on dev first. 

Thinking about this a bit more, I don't love how we'll have this logic duplicated in sql and ts, though. If we edit the `FIXED_ANTE` in ts, now we have to edit `fixed_ante` in sql and then also update every contract. That makes experimentation very slow. Have we tried putting `totalLiquidity` into its own native column and then computing the tiers on the fly? That would be much easier than having to precompute all of this.

web/components/contract/contracts-table.tsx
```diff
@@ -263,7 +270,30 @@ function ContractQuestion(props: {
...
+          <span>
+            <Tooltip
+              text={SWEEPIES_MARKET_TOOLTIP}
+              className=" relative mr-0.5 inline-flex h-[1em] w-[1.1em] items-baseline"
+            >
+              <SweepiesCoin className="absolute inset-0 top-[0.2em]" />
+            </Tooltip>
+          </span>
+        )}
+        {!!contract.isSpicePayout && (
```
IanPhilips, line 285: The spice is no longer!

IanPhilips (commented):
Thanks for working on unifying the tiers! It'll be great to eliminate the 'basic' tier which is  very confusing atm.

I will say I have reservations about jumping to this cached tier approach before we've tried splitting out the `totalLiquidity` into its own column (which is as easy as adding another line to the `contract_populate_cols` function) and computing the tiers on the fly (this could be a line in the `contracts` table, like how `description_fts` is calculated). If we computed the tiers on the fly, this PR would be a lot simpler SQL-wise and avoid code duplication between typescript and sql. It'd also let us experiment with different liquidity requirements by changing a few numbers in typescript rather than having to edit typescript, the sql function, and our 150k contracts every time we changed something.

ingawei:
> Thanks for working on unifying the tiers! It'll be great to eliminate the 'basic' tier which is very confusing atm.
> 
> I will say I have reservations about jumping to this cached tier approach before we've tried splitting out the `totalLiquidity` into its own column (which is as easy as adding another line to the `contract_populate_cols` function) and computing the tiers on the fly (this could be a line in the `contracts` table, like how `description_fts` is calculated). If we computed the tiers on the fly, this PR would be a lot simpler SQL-wise and avoid code duplication between typescript and sql. It'd also let us experiment with different liquidity requirements by changing a few numbers in typescript rather than having to edit typescript, the sql function, and our 150k contracts every time we changed something.

Yeah this was a big back and forth that happened when I first implemented tiers. The issue is it's not a hard number that if it's above and below it's a certain tier. For multi choice, a lower tier with 100 answers would cost the same as a binary with a higher tier. So it's not a simple SQL query if you're for example searching for a bunch of markets within a certain tier.

IanPhilips:
Ah, so in the multichoice case where you can add answers, we could store the required ante, right? So we could still avoid having the hard-coded strings

IanPhilips:
> > Ah, so in the multichoice case where you can add answers, we could store the required ante, right? So we could still avoid having the hard-coded strings
> 
> I see so you want to store total liquidity and base ante? So then when you filter by tier you just look for contracts in which total liquidity = ante * 10^tier?

I think we would just save base_ante for indie mc markets, and it would work like sth `if base_ante then base_ante > 10**ante_tier else total_liquidity>10**liquidity_tier`

IanPhilips:
Oh I forgot to ask, where is `get_tier_from_liquidity_efficient` called after backfilling? You'll have to call it when adding or removing liquidity, right?

ingawei:
It's here in `add-liquidity.ts`:
```
   await updateContract(tx, contractId, {
      subsidyPool: FieldVal.increment(subsidyAmount),
      totalLiquidity: FieldVal.increment(subsidyAmount),
      marketTier: getTierFromLiquidity(
        contract,
        contract.totalLiquidity + subsidyAmount
      ),
    })
```

I made the `get_tier_from_liquidity_efficient ` function just for the purpose of backfilling which I'm now realizing was pretty dumb lol. It seemed simpler at the time to do it in SQL than to do it with ts, mainly cause I couldn't find any example scripts backfilling to supabase

PR #2959, opened 2024-10-10
Withdraw liquidity

sipec:


backend/api/src/on-update-liquidity-provision.ts
```diff
@@ -24,6 +24,8 @@ export const onCreateLiquidityProvision = async (
 
   log(`onCreateLiquidityProvision: ${JSON.stringify(liquidity)}`)
 
+  if (liquidity.amount <= 0) return
```
mantikoros, line 27: Change the min amount to 1?

backend/api/src/remove-liquidity.ts
```diff
@@ -0,0 +1,109 @@
...
+    })
+
+    await runTxn(pgTrans, {
+      fromType: 'CONTRACT',
+      fromId: contract.id,
+      toType: 'USER',
+      toId: auth.uid,
+      category: 'REMOVE_SUBSIDY',
+      amount: totalAmount,
+      token: contract.token === 'CASH' ? 'CASH' : 'M$',
```
mantikoros, line 83: why not just contract.token

common/src/calculate-cpmm.ts
```diff
@@ -577,14 +576,67 @@ export function addCpmmMultiLiquidityAnswersSumToOne(
...
+  pool: { [outcome: string]: number },
+  p: number,
+  amount: number
+) {
+  const { newPool, liquidity, newP } = addCpmmLiquidity(pool, p, -1 * amount)
+
+  let error = ''
+
+  if (newPool.YES < MINIMUM_LIQUIDITY || newPool.NO < MINIMUM_LIQUIDITY) {
+    error = 'below minimum'
```
mantikoros, line 594: we don't really use this style of error handling (versus throwing an exception) anywhere else on the backend to my knowledge...

common/src/calculate-cpmm.ts
```diff
@@ -577,14 +576,67 @@ export function addCpmmMultiLiquidityAnswersSumToOne(
...
+}
+
+// https://www.wolframalpha.com/input?i=z+%3C+%28y+-+b%29+%2F+%28%28y-b%29+%2B+%28n-b%29%29+%3C+%281+-+z%29%2C+b+%3E+0%2C+n+%3E+0%2C+y+%3E+0%2C+y-b+%3E+0%2C+n+-+b+%3E+0%2C+0+%3C+z+%3C+0.5+solve+for+b+
+export function maximumRemovableLiquidity(pool: { [outcome: string]: number }) {
+  if (FRACTION_OF_POOL_MIN > 1 || FRACTION_OF_POOL_MIN < 0)
+    throw new Error('Math invariant violated')
+
+  const z = FRACTION_OF_POOL_MIN
+  const { YES: y, NO: n } = pool
+  const ratioMax = (n * z + y * z - Math.min(y, n)) / (2 * z - 1)
```
mantikoros, line 615: is this right? if x=y=100, then this value is negative, right?

mantikoros: I still don't really understand why we need a ratio limit if we already have a limit on the number of shares in the liquidity pool.

mantikoros (changes requested):
Mostly good. I think the ratioMax stuff is not right, but otherwise LGTM.

backend/api/src/remove-liquidity.ts
sipec: contract.token is type 'CASH' | 'MANA'

common/src/calculate-cpmm.ts
sipec: this function is used on the frontend and I want it to still return what the pool values would be even if it is below 100

common/src/calculate-cpmm.ts
sipec: ok, after playing around with the maniswap calculator a bit more, I'm convinced that extreme ratios are fine so I removed the restriction

PR #2961, opened 2024-10-11
Ingawei/man 1945 fix layout shift on broswe

ingawei:
Making updates to users, topics and contracts simultaneous for better CLS score. 

I think what might be causing the bad CLS is that the users/topics come in after contracts are loaded in searching and so the contracts jump
image

There appears to be an improvement when testing with 'lum' as a query:
http://localhost:3000/home?s=score&f=open&ct=ALL&p=0&fy=1&mt=00000&sw=0

current site gets a CLS score of 0.123, simultaneous updating gets a CLS score of 0.077

I usually would just push this but I don't want to break the site and I'm OOO starting tomorrow till Tuesday. If it looks ok please feel free to merge, and if it doesn't, feel free to make changes.

### DIFF
--- a/web/components/supabase-search.tsx
+++ b/web/components/supabase-search.tsx
@@ -505,7 +505,9 @@ export function SupabaseSearch(props: {
               !hideActions && actionColumn,
             ])}
           />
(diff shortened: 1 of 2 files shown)


sipec:
lgtm!

PR #2962, opened 2024-10-11
use embed page as market preview image

Bohaska:
i have no idea whether this will work but there's only one way to find out


sipec:
this doesn't work

PR #2966, opened 2024-10-14
lock multi trading by answer ids

sipec:
resolves MAN-1631


IanPhilips:
We update the contract's `volume`, `lastUpdatedTime`, `uniqueBettorCount` on indie mc bets

PR #2985, opened 2024-10-17
Use API for leaderboards

sipec:


sipec:
I aint reading all of that

PR #2998, opened 2024-10-19
Bulk insert bets, balance updates, and metric updates

IanPhilips:
Buylkify bet, balance, metric upserts into one batch


IanPhilips (commented):
lgtm!

Issue #3055, opened 2024-10-31
League cohort generation is not randomized 


Ziddletwix:
Since [November 2023](https://github.com/manifoldmarkets/manifold/commit/86ce406dbfab7b62c3e813ad7f0d5650308df172), the intent is for league cohort generation to be entirely randomized. However, currently, only the first cohort per division is properly randomized—the shuffled array is then ignored and the remaining cohorts are grabbed in the same original sorted order. This leads to large imbalances in league balance each season (some cohorts are mostly inactive users).

Users shuffled:

` let remainingUserIds = shuffle(divisionUserIds.concat())`

Replaced by (original) unshuffled array:

`remainingUserIds = divisionUserIds.filter((uid) => !userCohorts[uid])`

Should use shuffled array:
` remainingUserIds = remainingUserIds.filter((uid) => !userCohorts[uid])`

sipec:
resolved by #3056

PR #3056, opened 2024-10-31
Make league generation randomized

Ziddletwix:
Maintain the shuffled array for league randomization. See: issue #3055


mantikoros (approved):
LGTM!

PR #3102, opened 2024-11-10
Independent MC created with full starting liquidity per answer

Ziddletwix:
Context: [bounty](https://manifold.markets/ian/make-answers-created-during-market).

Simple/minimal fix. 

Previously: for higher tier markets, `ante` would be capped at `unmodifiedAnte` (i.e. numAnswers * 100 for MC). 

Now: for independent MC, `ante` is allowed to scale with the full `totalMarketCost`. For other question types, functionality is unchanged.

### DIFF
--- a/backend/api/src/create-market.ts
+++ b/backend/api/src/create-market.ts
@@ -149,7 +149,9 @@ export async function createMarketHelper(body: Body, auth: AuthedUser) {
   const totalMarketCost = marketTier
     ? getTieredCost(unmodifiedAnte, marketTier, outcomeType)
     : unmodifiedAnte
-  const ante = Math.min(unmodifiedAnte, totalMarketCost)
+  const ante = outcomeType === 'MULTIPLE_CHOICE' && !shouldAnswersSumToOne
+    ? totalMarketCost
+    : Math.min(unmodifiedAnte, totalMarketCost)
 
   const duplicateSubmissionUrl = await getDuplicateSubmissionUrl(
     idempotencyKey,


sipec:
I'm not convinced this is good actually. the drizzling is useful for new answers and especially if you want to initialize the probabilities right after creating the market. it is not entirely without purpose

I do see how this would make things simpler though...

Ziddletwix:
@sipec The liquidity drizzle is broken for independent MC, and much of the mana you spend is burned with no benefit (because it does not use maniswap). If the liquidity drizzle was not broken, I would not have a strong opinion on whether the smooth functionality of a delayed drizzle is worth the added unintuitive complexity (that difference is trivial IMO compared to the bad feeling of "you just lost a bunch of the mana you spent").

Example in practice: https://manifold.markets/Moscow25/will-polymarket-clear-1b-in-betting. Creator bought mana and made a Premium market. They wondered where all that liquidity went. Some of it is drizzling in, but even after the drizzle some options have way less than they should. I had to explain that independent MC is broken, and much of the mana they spent was burned forever. This was a very bad user experience (or to quote the user who bought mana—"this sucks").

When possible, I try to warn creators that Premium+ independent MC are broken, and they should not be used. Personally, I would like to make these markets, but I can't do so until they no longer waste the mana you spend on them. Fixing the drizzle for independent MC is a tougher task, but if they can simply start with the full liquidity (like adding options), the market type will be at least be functional (and adding liquidity is a much less important bonus).

IanPhilips:
LGTM! Thanks @Ziddletwix!

PR #3112, opened 2024-11-12
upvote and downvote comments

ingawei:
image

When you like/dislike a comment it calculates the like to dislike ratio (only if you have 10 dislikes or more so you don't get your comment hidden with just 1:5 likes to dislikes). If you have more than 80% dislikes your comment is hidden automatically. Open to tweaking these number.

Also ran this for reference:
```
ALTER
 
TABLE
 public.contract_comments
ADD
 
COLUMN
 dislikes int4 
NULL
 
DEFAULT
 
0
;

ALTER
 
TABLE
 public.contract_comments
DROP
 
COLUMN
 upvotes,
DROP
 
COLUMN
 downvotes;
 ```


IanPhilips (approved):
Looks great, thanks Inga! I think users will appreciate the ability to curate market comments, and all the viewers will appreciate only seeing the good stuff!

PR #3165, opened 2024-11-27
sweeps

ingawei:


IanPhilips (commented):
LGTM. Altho the real fix is making the tier a computed property based on the liquidity column like we talked about when you downgraded the tiers:

> I will say I have reservations about jumping to this cached tier approach before we've tried splitting out the totalLiquidity into its own column (which is as easy as adding another line to the contract_populate_cols function) and computing the tiers on the fly (this could be a line in the contracts table, like how description_fts is calculated). If we computed the tiers on the fly, this PR would be a lot simpler SQL-wise and avoid code duplication between typescript and sql. It'd also let us experiment with different liquidity requirements by changing a few numbers in typescript rather than having to edit typescript, the sql function, and our 150k contracts every time we changed something.

PR #3194, opened 2024-12-06
Ingawei/undo delete wrapped

ingawei:
Added wrapped back
image
Would appreciate a sanity check to see if data I pulled is actually in line with reality lol, please see comments


backend/api/src/get-max-min-profit-2024.ts
ingawei, line 1: This one is supposed to get the contract where you made the most profit, and the contract where you made the least profit in 2024. Filtered by contracts that resolved in 2024

sipec (approved):
looks sane to me

backend/api/src/get-max-min-profit-2024.ts
IanPhilips: Looks good! I don't think you need to selecet has-yes-shares and no shares

IanPhilips (approved):
Looks good! Excited to see mine

PR #3207, opened 2024-12-10
Allow select mods to resolve sweepstakes

SirSaltyy:


SirSaltyy:
I haven't been able to test the notifications as thoroughly as I would have liked to so please review that part carefully. Thanks @sipec !

sipec (approved):

PR #3232, opened 2024-12-17
Automatic sports creation

SirSaltyy:


backend/scripts/create-sports-contracts.ts
```diff
@@ -0,0 +1,132 @@
+import axios from 'axios'
+import { argv } from 'process' // To access command-line arguments
+
+interface Fixture {
```
IanPhilips, line 4: Please rename to sth like `SportsGame` or `Game`
Screenshot 2024-12-17 at 1 59 23 PM

common/src/sports-info.ts
```diff
@@ -0,0 +1,10 @@
+export interface Fixture {
```
IanPhilips, line 1: Please rename to sth like `SportsGame` or `Game`
Screenshot 2024-12-17 at 1 59 23 PM

common/src/sports-info.ts
```diff
@@ -0,0 +1,10 @@
...
+  idEvent: string
+  strLeague: string
+  strEvent: string
+  strHomeTeam: string
```
IanPhilips, line 5: don't prefix with str, that's redundant. Just call each by their name, e.g. `league`

common/src/contract.ts
```diff
@@ -98,6 +98,10 @@ export type Contract<T extends AnyContractType = AnyContractType> = {
   token: ContractToken
   siblingContractId?: string
 
+  sportsStartTimestamp?: string
```
IanPhilips, line 101: I'd prefer these were in sth like 
```
sports?:{
startTs:string
...etc.
}
```
so that they would all be defined when sports is defined, otherwise only sports will be undefined.

web/pages/admin/index.tsx
```diff
@@ -85,9 +93,20 @@ export default function AdminPage() {
...
         <LabCard title="🌑 test new user" href="/admin/test-user" />
-        <Button onClick={() => api('refresh-all-clients', {})}>
-          Refresh all clients
-        </Button>
+        <Row className="gap-2">
+          <Button onClick={() => api('refresh-all-clients', {})}>
+            Refresh all clients
+          </Button>
+          <Button onClick={onCreateSportsMarkets} disabled={isLoading}>
+            {isLoading ? 'Creating...' : 'Create Sports Markets'}
```
IanPhilips, line 101: This shouldn't be just one click, imo. would be nice to have in a modal

web/lib/admin/create-sports-markets.ts
```diff
@@ -0,0 +1,98 @@
...
+        visibility: 'public',
+        addAnswersMode: 'DISABLED',
+        shouldAnswersSumToOne: true,
+        sportsStartTimestamp: fixture.strTimestamp,
+        sportsEventId: fixture.idEvent,
+        sportsLeague: fixture.strLeague,
+        groupIds,
+      }
+
+      const result = await api('market', createProps as any)
```
IanPhilips, line 72: why cast as `any`? seems sus.

.gitignore
```diff
@@ -6,4 +6,4 @@ yarn-error.log
 dev
 firebase-debug.log
 dev-mantic-markets-firebase-adminsdk*.json
-mantic-markets-firebase-adminsdk*.json
\ No newline at end of file
+mantic-markets-firebase-adminsdk*.jsonbackend/scripts/fetch-sports.ts
```
IanPhilips, line 9: Missing a newline, but why would this be in the gitignore?

IanPhilips (changes requested):
Looks good David! It'll be nice to have all these games in a standard format. Check out my comments for a few changes I'd like made.

.gitignore
SirSaltyy: Oh I'll push a commit to delete this change, this was part of my failed attempt of not leaking the API key lol

SirSaltyy:
Some followup tasks (but not necessary to review this PR and merge so I can use it to create this week's sports markets).

-Check for any matching sportsEventId already exists on a contract and skip over it.
-Add more sports leagues to auto create but don't create cash contracts for them.
-Improve formating of description from string to json content

Even later followup tasks:
-Add things like badge icons and colors for each sport team (or switch to a different API that has them)
-Add prematch odds API to seed the initial probability of markets (or switch to a different API that includes them)
-Use the new properties on the contract (eg sportsStartTime) to integrate with sports page layout.
-Create auto resolving script which uses sports API + sportsEventId on contracts

SirSaltyy:
I didn't put the sports properties into an array as you had hoped for (I spent 5 mins trying to and realised I would have to change a lot of files and update them with logic for handling the array).

I can do this if you really want me to and even though it should be easy for a more component dev I was getting confused and it seemed like a change that would be ideal but not that important. Let me know if it is important.

I also spent an embarrassingly long time trying to change the names of eg. strLeague to league. Making the changes was quick but it caused it to stop working and I could not for the life of me figure out why so eventually I just reverted that commit and will leave them named as is for now.

SirSaltyy (commented):
oh this is where I am meant to write the comment to close the review.

PR #3342, opened 2025-01-10
Auto resolve sports markets

SirSaltyy:
Haven't been able to test as I haven't been able to redeploy the scheduler to dev.

My first time coding something involving the scheduler so let me know if I'm doing something wrong!


IanPhilips (changes requested):
I don't see why you're calling the API from the scheduler. If you actually want the api endpoint for use by the front-end, you'd add the function to the `backend/shared` directory for the scheduler and api to call directly. If getting the completed sports games is a scheduler only task, just keep the logic in the scheduler. It would be nice to get the scores during the game on the frontend, though

backend/scheduler/src/jobs/resolve-sports-markets.ts
```diff
@@ -0,0 +1,120 @@
...
+      const matchingContracts = unresolvedContracts.filter(
+        (c) => c.sportsEventId === game.idEvent
+      )
+      if (matchingContracts.length === 0) continue
+
+      log(
+        `Found ${matchingContracts.length} contracts for game ${game.idEvent}`
+      )
+
+      const resolver = await pg.one('SELECT * FROM users WHERE id = $1', [
```
IanPhilips, line 38: `pg.one` is going to throw an error if the row is not found, so there's no reason to add a !resolver check.

backend/scheduler/src/jobs/resolve-sports-markets.ts
```diff
@@ -0,0 +1,120 @@
...
+        HOUSE_LIQUIDITY_PROVIDER_ID,
+      ])
+      if (!resolver) {
+        log('House account not found. Skipping these contracts.')
+        continue
+      }
+
+      const homeScore = parseInt(game.homeScore ?? game.homeScore ?? '0')
```
IanPhilips, line 46: Why: `game.homeScore ?? game.homeScore`

backend/scheduler/src/jobs/resolve-sports-markets.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export async function resolveSportsMarkets() {
+  const pg = createSupabaseDirectClient()
+
+  try {
+    const data = await api('get-completed-sports-games', {})
+    const completedGames = (data.schedule || []) as SportsGames[]
+    log(`Processing ${completedGames.length} completed games.`)
+
+    const unresolvedContracts = await pg.map<Contract>(
```
IanPhilips, line 17: why not just filter the contracts where data->>'sportsEventId' is in your completed games id array?

backend/shared/src/resolve-sports-markets.ts
```diff
@@ -0,0 +1,136 @@
...
+      `
+      SELECT *
+      FROM contracts
+      WHERE resolution IS NULL
+        AND data->>'sportsEventId' IN (${completedGameIds
+          .map((_, i) => `$${i + 1}`)
+          .join(', ')})
+      `,
+      completedGameIds,
+      (row) => row as Contract
```
IanPhilips, line 34: you have to use `convertContract` here, row != Contract

backend/shared/src/resolve-sports-markets.ts
```diff
@@ -0,0 +1,136 @@
...
+    )
+
+    log(`Found ${unresolvedContracts.length} unresolved contracts.`)
+
+    if (unresolvedContracts.length === 0) {
+      log('No matching contracts to resolve.')
+      return
+    }
+
+    const resolver = await pg.one('SELECT * FROM users WHERE id = $1', [
```
IanPhilips, line 44: you have to use `convertUser` here as well

PR #3419, opened 2025-01-31
Ingawei/man 2222 sell panel

ingawei:
Could use a sanity check on the sell logic, mostly in `position-modal-content.tsx`

Also moved some stuff to `common` / `client-common` and would appreciate a check on those

IMG_4283
IMG_4284
IMG_4285
IMG_4286


IanPhilips (changes requested):
The sell panel looks good! Excited for users to try it out. 
One thing to change is: we want to avoid using the supabase js client in mani and just use the api. That would let you un-refactor all those imports from web->common.

PR #3490, opened 2025-02-18
Document `marketTier` parameter in market creation route

Ari-Zerner:


IanPhilips:
Thanks for adding this! We did decide to remove tiers, so I'm closing this.

Ari-Zerner:
Does the new `liquidityTier` parameter work the same way, with numeric values rather than strings? It's still undocumented https://docs.manifold.markets/api#post-v0market

PR #3491, opened 2025-02-18
Create sweepstakes-email.ts

SirSaltyy:


backend/scripts/sweepstakes-email.ts
```diff
@@ -0,0 +1,127 @@
+import { runScript } from 'run-script'
+import fetch from 'node-fetch'
+
+const MAILGUN_API_KEY = '' // add API key
```
IanPhilips, line 4: we already have this in the secrets, you should access it in process.env in the runScript function body

backend/scripts/sweepstakes-email.ts
```diff
@@ -0,0 +1,127 @@
...
+  const mailgunUrl = `https://api.mailgun.net/v3/mg.manifold.markets/messages`
+
+  const recipientData = users.reduce((acc, user) => {
+    acc[user.email] = { username: user.username }
+    return acc
+  }, {} as Record<string, { username: string }>)
+
+  const emailParams = new URLSearchParams()
+  emailParams.append('from', 'Manifold <info@manifold.markets>')
+  emailParams.append('to', users.map((u) => u.email).join(','))
```
IanPhilips, line 67: I don't think this is how you do this. Why don't you use the `sendBulkEmails` function in `backend/shares/src/emails.ts` that will handle the 'to' fields. The way that function handles the 'to' data is each a separate field, whereas yours looks like everyone will see everyone else's email, which would be very bad.

backend/scripts/sweepstakes-email.ts
```diff
@@ -0,0 +1,127 @@
...
+    '[ACTION REQ] Redeem your sweepcash by March 28th'
+  ) // Update to change email subject
+  emailParams.append('template', 'manifold announcement template') // Create template in mailgun and update here to send
+  emailParams.append('h:X-Mailgun-Variables2', JSON.stringify(recipientData))
+
+  try {
+    const response = await fetch(mailgunUrl, {
+      method: 'POST',
+      headers: {
+        Authorization: `Basic ${Buffer.from(`api:${MAILGUN_API_KEY}`).toString(
```
IanPhilips, line 79: the `sendBulkEmails` already handles this stuff as well

backend/scripts/sweepstakes-email.ts
```diff
@@ -0,0 +1,127 @@
...
+async function sendBulkEmails(users: SweepstakesUser[]) {
+  if (users.length === 0) {
+    console.log('No users to send emails to.')
+    return
+  }
+
+  const mailgunUrl = `https://api.mailgun.net/v3/mg.manifold.markets/messages`
+
+  const recipientData = users.reduce((acc, user) => {
+    acc[user.email] = { username: user.username }
```
IanPhilips, line 61: Is including their username necessary for each email? Did you know you can link to https://manifold.markets/me to link people to their profiles?

IanPhilips (changes requested):
Looks like you're pretty close! I think you should probably just use the `sendBulkEmails` and leave out their usernames from the email data, but do include their names, like `sendBulkEmails` expects.

PR #3494, opened 2025-02-19
docs(api): clarify resolution payload for free response/multiple choi…

franklinbaldo:
…ce markets

When shouldAnswersSumToOne is false, each answer must be resolved individually—as with binary markets—and the answerId must be included in the payload. This update clarifies the required payload formats for both modes.


IanPhilips:
LGTM, thanks!

PR #3529, opened 2025-03-08
Update README.md

Nibezak:
Changing the Running local server to add a " Yarn dev " too instead of only " yarn dev:dev"

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This [monorepo][] has basically everything involved in running and operating Man
 0. Make sure you have [Yarn 1.x][yarn]
 1. `$ cd web`
 2. `$ yarn`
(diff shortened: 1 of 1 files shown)


IanPhilips:
That doesn't seem right? The web `package.json` has the scripts that set the NEXT_PUBLIC_FIREBASE_ENV
```
   "dev:dev": "cross-env NEXT_PUBLIC_FIREBASE_ENV=DEV yarn dev",
    "dev:prod": "cross-env NEXT_PUBLIC_FIREBASE_ENV=PROD yarn dev",
    "dev:local": "cross-env NEXT_PUBLIC_API_URL=localhost:8088 yarn devdev",
    ```

PR #3553, opened 2025-04-05
Add bet button to limit orders

Bohaska:


IanPhilips:
I like the idea! I don't love the design, however. I think the easiest way to get this merged is keep the look of the rows the same but when you hover over a row, it reveals itself to be clickable. Otherwise you'll have to iterate on the limit row design to make them look better, because they're not up to snuff yet.

PR #3578, opened 2025-05-09
End love 💔

sipec:
Deletes remaining love code, except for the sql tables.
Don't send the user new message emails with the love branding when they get a DM on manifold.


IanPhilips:
Looks great, thanks @sipec! I *love* this PR

PR #3579, opened 2025-05-10
Add userId to get-user-portfolio-internal

danstoyell:
I have no idea what I'm doing. Someone should probably stop me. Why was this not flagged by whatever linter/static analysis  typescript uses? The field is defined on PortfolioMetrics.


IanPhilips:
Sure, wouldn't hurt to add it in, thanks!

PR #3584, opened 2025-05-19
Fix OpenSearch XML

syvb:
The OpenSearch search engine specification has mismatched open/closing tags, which makes it invalid XML and thus doesn't work. This PR fixes the XML, making it valid.


IanPhilips:
Good catch, thanks @syvb

PR #3585, opened 2025-05-19
Fix limit order double spending

mantikoros:
## Summary
- validate maker balances before applying limit order fills
- reject bets and sells if makers don't have enough balance

## Testing
- `yarn verify` *(fails: package not in lockfile)*


IanPhilips:
Fixed this in a more comprehensive way

PR #3593, opened 2025-06-05
Avoid overlaps with the Android gesture bar

s9ferech:
A few months ago, Chrome on Android made [some changes](https://developer.chrome.com/docs/css-ui/edge-to-edge) to their user interface. By default, the gesture bar at the bottom of the screen dynamically expands or retracts as the user scrolls. But even when it retracts, the gesture handle remains on screen and overlaps the website's content:

Screenshot_20250605-203357

This PR sets the gesture bar to be permanently retracted and adds some padding or an offset to the Manifold bottom navigation, market-creation button and main content in order to avoid an overlap with the gesture handle:

Screenshot_20250605-203324

### DIFF
--- a/web/components/layout/page.tsx
+++ b/web/components/layout/page.tsx
@@ -42,7 +42,8 @@ export function Page(props: {
       <FirstStreakModalManager />
       <Col
         className={clsx(
-          !hideBottomBar && 'pb-[58px] lg:pb-0', // bottom bar padding
+          !hideBottomBar &&
+            'pb-[calc(58px+env(safe-area-inset-bottom))] lg:pb-[env(safe-area-inset-bottom)]',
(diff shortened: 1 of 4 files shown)


s9ferech:
Only checking the most obvious things. I could play around with it some more and try out other browsers on my Android phone and my PC if you want. But I wouldn't be totally confident that it doesn’t break some detail. I only found out that the create-market button had to be moved because I immediately saw that during my testing.

If you want to merge this PR, then I would suggest also talking about the appearance. As you might see in the second screenshot, it doesn’t look that great: The navbar feels out of balance with all the padding at the bottom and none at the top, and the sharp corners of the active item look wrong to me. A minimal fix might be to add a border under the navbar, like the one above, or to give the whole system gesture bar a distinct color. That’s not consistent with native apps on either Android or iOS, but I think it could be an improvement. Should I try it out?

PR #3598, opened 2025-06-20
Add POST Bet Example Usage

crowlsyong:
API Docs update to help users understand how to create a POST Bet command

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -963,6 +963,21 @@ Parameters:
 
 [Requires Auth](#authentication).
 
+Example request
+
+```bash
+curl -X POST https://api.manifold.markets/v0/bet \
+  -H 'Authorization: Key YOUR_API_KEY' \
+  -H 'Content-Type: application/json' \
+  -d '{
+    "amount": 10,
+    "contractId": "CONTRACT_ID_HERE",
+    "outcome": "YES",
+    "limitProb": 0.70,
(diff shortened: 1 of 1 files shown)


docs/docs/api.md
```diff
@@ -760,6 +760,19 @@ Parameters:
...
+```bash
+curl -X POST https://api.manifold.markets/v0/bet \
+  -H 'Authorization: Key YOUR_API_KEY' \
+  -H 'Content-Type: application/json' \
+  -d '{
+    "amount": 75,
+    "contractId": "CONTRACT_ID_HERE",
+    "outcome": "YES",
+    "limitProb": 0.70,
+    "expiresAt": 1735689600000
```
IanPhilips: This expires in the past, so I think will be rejected by the server

IanPhilips (commented):
Great idea! maybe just change the expiresAt and we're good to go!

docs/docs/api.md
crowlsyong: Fixed :)

IanPhilips: I don't see the fix, yet! Maybe forgot to push?

IanPhilips: alternatively you could use the `expiresMillisAfter` prop

IanPhilips:
Nice! Thanks for adding the example, this should help other devs!

PR #3619, opened 2025-08-13
Lowered masters demotion and dynamic diamond cohort sizes

SirSaltyy:
I've changed the master's demotion count from a fixed '19' to be 40% of the cohort size. 

Then, when generating a new season, rather than determining the number of diamond cohorts by dividing the number of diamond users by 25, it now creates them dynamically by calculating the size of the master cohort. It will create the correct number of diamond cohorts needed (with 2 promoting from each) to offset the number of masters users being demoted.

### DIFF
--- a/backend/shared/src/generate-leagues.ts
+++ b/backend/shared/src/generate-leagues.ts
@@ -8,6 +8,9 @@ import {
   getSeasonDates,
   league_user_info,
   MAX_COHORT_SIZE,
+  getDivisionNumber,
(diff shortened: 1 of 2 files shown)


IanPhilips:
Looks good! I had to fix things so that leagues before this PR (current season 28) are unchanged. Also raised the demotion percent to 60% which is more inline with the historical number (10-12 remaining in masters) at our current number of masters participants.

PR #3620, opened 2025-08-22
Achievements

SirSaltyy:


backend/api/src/get-user-achievements.ts
```diff
@@ -0,0 +1,604 @@
...
+}) => {
+  const pg = createSupabaseDirectClient()
+
+  // Detect if per-achievement MVs exist; else fall back to monolithic MV; else no MV
+  const perMVReg = await pg.oneOrNone(
+    "select to_regclass('public.mv_ach_volume') as reg"
+  )
+  const hasPerMV = !!perMVReg?.reg
+
+  const mvReg = await pg.oneOrNone(
```
IanPhilips: what are these checks? Don't we know if we have those views or not?

backend/api/src/get-user-achievements.ts
```diff
@@ -0,0 +1,604 @@
...
+      left join comments on true
+      left join liquidity on true
+      left join trades on true
+      left join markets_created on true
+      left join account_age on true
+      left join trade_stats on true
+      left join longest_streak on true`
+
+  const result = await pg.oneOrNone(
+    hasPerMV ? queryWithPerMV : hasMV ? queryWithMV : queryWithoutMV,
```
IanPhilips: if we don't need those checks (bc iiuc we should know what exists), then do we need all three of these queries?

web/pages/[username]/index.tsx
```diff
@@ -538,6 +555,513 @@ function ProfilePublicStats(props: {
...
+      })
+    return () => {
+      isMounted = false
+    }
+  }, [userId])
+
+  if (loading)
+    return (
+      <Row className="text-ink-600 items-center gap-2">
+        Loading achievements…
```
Comment (IanPhilips) [inline web/pages/[username]/index.tsx:643]: This isn't ideal as a loading screen but is fine for now I suppose... You should just prompt claude to add some flashing blank achievement cards instead of a "loading achievements..." text

PR #3625, opened 2025-10-02
Mana shop

SirSaltyy:


backend/api/src/checkout-printful.ts
```diff
@@ -0,0 +1,115 @@
...
+  // Enforce max 3 physical items per order
+  const totalQty = items.reduce((sum, it) => sum + (it.quantity ?? 0), 0)
+  if (totalQty > 3) {
+    throw new APIError(
+      403,
+      'You can include at most 3 physical items per order'
+    )
+  }
+
+  const apiKey = process.env.PRINTFUL_KEY
```
IanPhilips, line 28: I don't see this new key in `common/src/secrets.ts` - doesn't look like it's on https://console.cloud.google.com/security/secret-manager?referrer=search&hl=en&project=mantic-markets either

common/src/api/schema.ts
```diff
@@ -2220,6 +2220,167 @@ export const API = (_apiTypeCheck = {
...
+        itemId: z.string(),
+        price: z.number().positive().finite().safe(),
+      })
+      .strict(),
+  },
+  'get-shop-orders': {
+    method: 'GET',
+    visibility: 'public',
+    authed: true,
+    returns: {} as { orders: import('../shop/types').ShopOrder[] },
```
IanPhilips: what's with this midline import?

backend/api/src/checkout-printful.ts
```diff
@@ -0,0 +1,115 @@
...
+  HOUSE_LIQUIDITY_PROVIDER_ID,
+} from 'common/antes'
+
+export const checkoutPrintful: APIHandler<'checkout-printful'> = async (
+  { recipient, items },
+  auth
+) => {
+  const userId = auth.uid
+  if (!userId) throw new APIError(401, 'You must be signed in')
+  if (!items || items.length === 0) throw new APIError(400, 'No items')
```
IanPhilips: the min of 1 looks like it's already in the schema, so shouldn't need to have this check

backend/api/src/checkout-printful.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+    for (const it of items) {
+      await tx.none(
+        `insert into shop_orders
+         (user_id, item_id, item_type, price_mana, quantity, txn_id, printful_order_id, status, created_time, metadata)
+         values ($1, $2, 'printful', $3, $4, $5, $6, 'FULFILLING', now(), $7)`,
+        [
+          userId,
+          String(it.productId),
+          0,
```
IanPhilips: price_mana is 0?

backend/api/src/checkout-printful.ts
```diff
@@ -0,0 +1,115 @@
...
+    )
+  }
+
+  const apiKey = process.env.PRINTFUL_KEY
+  if (!apiKey) throw new APIError(500, 'PRINTFUL_KEY is not configured')
+
+  const pg = createSupabaseDirectClient()
+
+  // Enforce 1 physical-goods order per 30 days
+  const recentOrder = await pg.oneOrNone<{ exists: boolean }>(
```
IanPhilips: this and the following db queries should all be in a single db txn. Otherwise people can spam the button before the query to printful completes and can order multiple items.

backend/api/src/checkout-shop-cart.ts
```diff
@@ -0,0 +1,196 @@
...
+} from 'common/antes'
+
+// For v1: only digital/other items are processed; printful will be added next
+export const checkoutShopCart: APIHandler<'checkout-shop-cart'> = async (
+  { items },
+  auth
+) => {
+  const userId = auth.uid
+  if (!userId) throw new APIError(401, 'You must be signed in')
+  if (items.length === 0) throw new APIError(400, 'No items to checkout')
```
IanPhilips: just use .min(1) in schema

backend/api/src/checkout-shop-cart.ts
```diff
@@ -0,0 +1,196 @@
...
+      } else {
+        // Lifetime limit for other items
+        const existingRow = await pg.oneOrNone<{ q: string }>(
+          `select coalesce(sum(quantity), 0) as q from shop_orders where user_id = $1 and item_id = $2`,
+          [userId, itemId]
+        )
+        existingQty = Number(existingRow?.q ?? 0)
+      }
+      if (existingQty + requestedQty > limit) {
+        const limitType =
```
IanPhilips: any reason not to have this as an option in the shop config that this logic reads from? We may want to add other per month or per user items, and we wouldn't need to edit this code, only add the flag to the item down the line

backend/api/src/checkout-shop-cart.ts
```diff
@@ -0,0 +1,196 @@
...
+          itemId === 'streak-forgiveness' ? 'per month' : 'per user'
+        throw new APIError(403, `Limit ${limit} ${limitType} for ${itemId}`)
+      }
+    }
+
+  }
+
+  // Check balance
+  const userRow = await pg.oneOrNone(
+    `select balance from users where id = $1`,
```
IanPhilips: the `runTxnInBetQueue` does this balance check, and if it didn't this query would need to be in the db txn for it to be effective

web/pages/api/printful/geo.ts
```diff
@@ -0,0 +1,64 @@
+import type { NextApiRequest, NextApiResponse } from 'next'
+
+// In-memory cache per server instance
+let CACHE: { data: any; ts: number } | null = null
+const TTL_MS = 24 * 60 * 60 * 1000 // 24 hours
```
IanPhilips: please use DAY_MS from `common/util/time.ts`

web/pages/shop.tsx
```diff
@@ -0,0 +1,951 @@
...
+  }
+
+  const isExpiringCosmetic =
+    item.id === 'golden-crown' || item.id === 'very-rich-badge'
+  const thisEnt = isExpiringCosmetic
+    ? entitlements?.find((e) => e.entitlementId === item.id)
+    : undefined
+  // Compute expiry using entitlement, or fall back to last purchase + 30d
+  const now = Date.now()
+  const thirtyDaysMs = 30 * 24 * 60 * 60 * 1000
```
IanPhilips: please replace all of these times with our constants from `common/util/time.ts`

web/pages/shop.tsx
```diff
@@ -0,0 +1,951 @@
...
+                    clearAwardInventoryCache()
+                  } else if (item.id === 'pampu-skin') {
+                    clearPampuSkinCache(currentUser.id)
+                    clearEntitlementsCache(currentUser.id)
+                  }
+                }
+
+                return true
+              } catch (_e) {
+                toast.error('Purchase failed')
```
IanPhilips: would be useful to parse the error from the api like we do in other places and show them it, ie it could be the 30 day cooldown error, or max 3 items error, etc.

web/pages/shop.tsx
```diff
@@ -0,0 +1,951 @@
...
+        }[]
+      }[]
+    | null
+  >(null)
+
+  useEffect(() => {
+    fetch('/api/printful/products')
+      .then((r) => r.json())
+      .then((d) => setRemote(d.products))
+      .catch(() => setRemote([]))
```
IanPhilips: log the error if it doesn't work, show a toast to the user?

web/pages/api/printful/geo.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export default async function handler(
```
IanPhilips: We don't really use the nextjs functions any more, so not ideal to start adding them now. Please turn this into a normal api endpoint. It's easier to manage the request limits, logs, API keys that way as well, iirc you have to set api keys in the vercel dashboard, which I haven't had to touch in over a year.

web/pages/api/printful/products.ts
```diff
@@ -0,0 +1,363 @@
+import type { NextApiRequest, NextApiResponse } from 'next'
```
IanPhilips: This page should be a normal api endpoint as well, not a vercel handler

web/pages/shop/orders.tsx
```diff
@@ -0,0 +1,110 @@
...
+    return map
+  }, [])
+
+  useEffect(() => {
+    ;(async () => {
+      try {
+        const res = await api('get-shop-orders', {})
+        setOrders(res.orders)
+      } catch {
+        setOrders([])
```
IanPhilips, line 29: log error would be helpful

web/lib/api/api.ts
```diff
@@ -261,3 +261,16 @@ export function createChartAnnotation(params: {
 export function deleteChartAnnotation(params: { id: number }) {
   return call(getApiUrl('delete-chart-annotation'), 'POST', params)
 }
+
+export const getUserEntitlements = curriedAPI('get-user-entitlements')
```
IanPhilips: what's with all these curriedapis? These should be normal api routes in `backend/api/src/routes.ts`

web/hooks/use-very-rich-badge.ts
```diff
@@ -0,0 +1,80 @@
...
+export function clearVeryRichBadgeCache(userId: string) {
+  if (typeof window === 'undefined') return
+  try {
+    localStorage.removeItem(`${STORAGE_KEY}-${userId}`)
+  } catch {
+    // Ignore storage errors
+  }
+}
+
+export function useVeryRichBadge(userId?: string) {
```
IanPhilips, line 16: why doesn't this just usePersistentLocalState?

web/hooks/use-user-entitlements.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (typeof window === 'undefined') return
+  try {
+    localStorage.removeItem(`${STORAGE_KEY}-${userId}`)
+  } catch {
+    // Ignore storage errors
+  }
+}
+
+export function useUserEntitlements(userId?: string) {
+  const [ents, setEnts] = useState<UserEntitlement[] | undefined>(() => {
```
IanPhilips: seems like this could also usePersistentLocalState with `{userId}-user-entitlements-cache` as the key

web/hooks/use-user-award-inventory.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (typeof window === 'undefined') return
+  try {
+    localStorage.removeItem(STORAGE_KEY)
+  } catch {
+    // Ignore storage errors
+  }
+}
+
+export function useUserAwardInventory() {
+  const [inventory, setInventory] = useState<AwardInventory | null>(() => {
```
IanPhilips: seems like this could also use `usePersistentLocalState` and `useAPIGetter` w with `{userId}-award-inventory-cache` as the key

web/hooks/use-shop-item-counts.ts
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from 'react'
+import { api } from 'web/lib/api/api'
+
+export function useShopItemCounts() {
```
IanPhilips, line 4: why not just use `useAPIGetter`?

web/hooks/use-comment-awards.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react'
+import { api } from 'web/lib/api/api'
+
+export function useCommentAwards(contractId: string, commentIds: string[]) {
```
IanPhilips, line 5: why not `useAPIGetter`?

web/hooks/use-has-pampu-skin.ts
```diff
@@ -0,0 +1,65 @@
...
+export function clearPampuSkinCache(userId: string) {
+  if (typeof window === 'undefined') return
+  try {
+    localStorage.removeItem(`${STORAGE_KEY}-${userId}`)
+  } catch {
+    // Ignore storage errors
+  }
+}
+
+export function useHasPampuSkin(userId?: string): boolean {
```
IanPhilips, line 16: seems like this could also be some version of `usePersistentLocalState` and `useAPIGetter` with `{userId}-has-pampu-skin-cache` as the key. you can use the refresh() from the hook every x minutes to refresh the state

web/hooks/use-cart.ts
```diff
@@ -0,0 +1,84 @@
...
+  imageUrl?: string
+  price: number
+  quantity: number
+  meta?: Record<string, any>
+}
+
+const STORAGE_KEY = 'shop_cart_v1'
+
+export function useCart() {
+  const [items, setItems] = useState<CartItem[]>([])
```
IanPhilips: why not use `usePersistentLocalState`?

web/components/widgets/user-link.tsx
```diff
@@ -299,6 +308,56 @@ function MarketCreatorBadge() {
...
 
+function VeryRichBadge(props: { userId: string }) {
+  const { userId } = props
+  const [amount, setAmount] = useState<number | null>(null)
+  const [expiresTime, setExpiresTime] = useState<string | null>(null)
+
+  useEffect(() => {
+    let cancelled = false
+    Promise.all([
+      import('web/lib/api/api').then(({ getVeryRichBadge }) =>
```
IanPhilips: this is not the style we use for api calls. import at the top of the file, then use the api object inline

web/components/widgets/user-link.tsx
```diff
@@ -299,6 +308,56 @@ function MarketCreatorBadge() {
...
+    }
+  }, [userId])
+
+  if (amount == null || amount <= 0) return null
+
+  const getDaysRemaining = () => {
+    if (!expiresTime) return null
+    const now = Date.now()
+    const expires = new Date(expiresTime).getTime()
+    const daysLeft = Math.ceil((expires - now) / (1000 * 60 * 60 * 24))
```
IanPhilips: would be good to reuse the time utils of DAY_MS

web/components/widgets/avatar.tsx
```diff
@@ -18,8 +23,29 @@ export const Avatar = memo(
...
+    const {
+      userId,
+      username,
+      noLink,
+      size,
+      className,
+      preventDefault,
+      createdTime,
+    } = props
+    const entitlements = useUserEntitlements(userId)
```
IanPhilips: oh this is not ideal - every user avatar causes a new request when uncached. That's the point of the `hooks/use-batched-getter` - if you're querying for 50 users at once, just debounce them all into one query with all of their ids. That would be better for this case

web/components/leaderboard.tsx
```diff
@@ -59,7 +59,7 @@ export function Leaderboard<T extends LeaderboardEntry>(props: {
                   </td>
                   <td>
                     <UserAvatarAndBadge
-                      className="overflow-hidden max-[400px]:max-w-[160px] sm:max-w-[200px] xl:max-w-none"
+                      className="max-[400px]:max-w-[160px] sm:max-w-[200px] xl:max-w-none"
```
IanPhilips, line 62: does this overflow though on mobile now?

web/components/comments/comment.tsx
```diff
@@ -106,6 +107,7 @@ export const FeedComment = memo(function FeedComment(props: {
   }, [bets?.length])
 
   const { userUsername, userAvatarUrl, userId } = comment
+  const awards = useCommentAwards(playContract.id, [comment.id])[comment.id]
```
IanPhilips, line 110: this is also not ideal, for every comment we do a new api endpoint. would be better to add this to the handlers in `use-batched-getter` as well and query awards for all comments in bulk

web/components/comments/give-award-modal.tsx
```diff
@@ -0,0 +1,177 @@
...
+    <Modal
+      open={open}
+      setOpen={setOpen}
+      size="sm"
+      className="bg-canvas-0 rounded-lg p-6"
+    >
+      <Col className="gap-3 p-3">
+        <Row className="items-center justify-between">
+          <div className="text-lg font-semibold">Give award</div>
+          <Link href="/shop" className="text-primary-600 text-sm underline">
```
IanPhilips, line 69: I don't think this is a great user experience if any time a user wants to give kudos they have to go to a diff page, buy a thing, then go back, then hit award, then choose the award. Users should be able to buy an award straight from this modal, then give it to a user from here

backend/supabase/user_entitlements.sql
```diff
@@ -0,0 +1,12 @@
+-- User cosmetic/digital entitlements
+create table if not exists
```
IanPhilips, line 2: this table should have 
```
-- Row Level Security
alter table user_entitlements enable row level security;
```

backend/supabase/shop_orders.sql
```diff
@@ -0,0 +1,25 @@
+-- Create shop_orders table (empty initial deployment)
+create table if not exists
```
IanPhilips, line 2: need to enable row level security here too

backend/supabase/comment_awards.sql
```diff
@@ -0,0 +1,17 @@
+create table if not exists
```
IanPhilips, line 1: need to enable row level security here too

backend/shared/src/notifications/create-comment-award-notif.ts
```diff
@@ -0,0 +1,81 @@
...
+    sourceTitle: contract.question,
+    data: {
+      awardType,
+      payoutAmount,
+      awardIcon: awardIconMap[awardType],
+      awardName: awardNameMap[awardType],
+    },
+  }
+
+  return await pg.none(
```
IanPhilips: would be better to use `insertNotificationToSupabase` in `notifications.ts`

backend/api/src/update-user-entitlement.ts
```diff
@@ -0,0 +1,29 @@
+import { APIError, APIHandler } from './helpers/endpoint'
+import { createSupabaseDirectClient } from 'shared/supabase/init'
+
+export const updateUserEntitlement: APIHandler<
```
IanPhilips: what's the point of this endpoint? I don't see it used anywhere

backend/api/src/toggle-pampu-skin.ts
```diff
@@ -0,0 +1,33 @@
+import { APIError, APIHandler } from './helpers/endpoint'
+import { createSupabaseDirectClient } from 'shared/supabase/init'
+
+export const togglePampuSkin: APIHandler<'toggle-pampu-skin'> = async (
```
IanPhilips, line 4: Oh hmm they can just edit their pampu state in local storage and save m25k. oh well!

backend/api/src/purchase-shop-item.ts
```diff
@@ -0,0 +1,38 @@
+import { APIError, APIHandler } from './helpers/endpoint'
+import { createSupabaseDirectClient } from 'shared/supabase/init'
+import { runTxnInBetQueue, TxnData } from 'shared/txn/run-txn'
+import { isProd } from 'shared/utils'
+import {
+  DEV_HOUSE_LIQUIDITY_PROVIDER_ID,
+  HOUSE_LIQUIDITY_PROVIDER_ID,
+} from 'common/antes'
+
+export const purchaseShopItem: APIHandler<'purchase-shop-item'> = async (
```
IanPhilips: Don't see this used anywhere

backend/api/src/get-very-rich-badge.ts
```diff
@@ -0,0 +1,19 @@
...
+import { createSupabaseDirectClient } from 'shared/supabase/init'
+
+export const getVeryRichBadge: APIHandler<'get-very-rich-badge'> = async (
+  { userId },
+  auth
+) => {
+  const targetId = userId ?? auth?.uid
+  if (!targetId) throw new APIError(400, 'userId is required')
+  const pg = createSupabaseDirectClient()
+  const row = await pg.oneOrNone<{ sum: string }>(
```
IanPhilips, line 11: why isn't this just a check in the entitlements table?

backend/api/src/get-user-award-inventory.ts
```diff
@@ -0,0 +1,50 @@
+import { APIError, APIHandler } from './helpers/endpoint'
+import { createSupabaseDirectClient } from 'shared/supabase/init'
+
+export const getUserAwardInventory: APIHandler<
+  'get-user-award-inventory'
+> = async (props, auth) => {
+  const userId = auth.uid
+  if (!userId) throw new APIError(401, 'You must be signed in')
```
IanPhilips: isn't this just handled with `authed: true` in the schema?

backend/api/src/get-shop-orders.ts
```diff
@@ -0,0 +1,94 @@
...
+    return undefined
+  }
+}
+
+export const getShopOrders: APIHandler<'get-shop-orders'> = async (
+  _props,
+  auth
+) => {
+  const userId = auth.uid
+  if (!userId) throw new APIError(401, 'You must be signed in')
```
IanPhilips: set `authed: true` in the schema

backend/api/src/get-shop-item-counts.ts
```diff
@@ -0,0 +1,7 @@
+import { APIHandler } from './helpers/endpoint'
+
+export const getShopItemCounts: APIHandler<
+  'get-shop-item-counts'
```
IanPhilips: what's this for?

backend/api/src/get-comment-awards.ts
```diff
@@ -0,0 +1,38 @@
...
+export const getCommentAwards: APIHandler<'get-comment-awards'> = async (
+  { contractId, commentIds },
+  auth
+) => {
+  const pg = createSupabaseDirectClient()
+  const rows = await pg.manyOrNone(
+    `select comment_id, award_type, count(*)::int as count,
+            bool_or(giver_user_id = $2) as awarded_by_me
+       from comment_awards
+      where contract_id = $1 and comment_id = any($3)
```
IanPhilips: your index on `comment_awards` is by `comment_id`, so why check the contract_id at all?

IanPhilips (changes requested):
This is exciting! It looks pretty good overall, but I've got some changes I'd like made. 

With all the hooks, can you make something more generic like usePersistentAPIGetter and then have a custom hook that uses the refresh every 5 mins? And for the ones that are for every user avatar/comment, which will be dozens on many pages, ideally that would use a new `usePersistentBatchedGetter` like the current `useBatchedGetter` but with persistent storage.

We don't use vercel handlers anymore, so would like those to be normal api endpoints instead.

Thanks for getting this far!

backend/api/src/checkout-printful.ts
SirSaltyy: It's stored in Vercel env keys

backend/api/src/get-very-rich-badge.ts
SirSaltyy: It's computing lifetime spend towards the badge, not just current possession.

web/components/comments/give-award-modal.tsx
SirSaltyy: Agreed. Ngl I tried to implement this and for some reason it cause me to go into an error loop of bugs I couldn't fix. Would have implemented if I had more time. Feel free to add.

backend/api/src/get-shop-item-counts.ts
SirSaltyy: all these endpoints were from old code that I removed to simplify how things worked (but I guess I missed deleting some of the backend) Will remove all of them

web/components/leaderboard.tsx
SirSaltyy: I made a change in user-link.ts I believe to fix this there while not causing the border to crop. You should double check the file changes to user-link though, It's possible I didn't do it right.

PR #3721, opened 2026-03-16
Merch live

todwadd:
## Summary
- Merch shop backend: stock control, admin panel, Printful webhook integration, cancel & refund flow
- Admin page at `/admin/merch` for stock toggles and order management
- User notifications for shipped/cancelled orders

[Details on notion for manifold team](https://www.notion.so/manifoldmarkets/merch-live-Branch-32554492ea7a80528ad4d2d24695dd3e#32554492ea7a8080b710d4af1189b9b7)


web/pages/admin/merch.tsx
```diff
@@ -0,0 +1,283 @@
...
+      <td className="px-3 py-2">{formatMoney(order.priceMana)}</td>
+      <td className="px-3 py-2">
+        <span
+          className={`inline-block rounded-full px-2 py-0.5 text-xs font-medium ${statusColor}`}
+        >
+          {order.status}
+        </span>
+      </td>
+      <td className="px-3 py-2 font-mono text-xs">
+        {order.txnId ? `manifold-${order.txnId}` : '\u2014'}
```
wasabipesto: Does the txnId == printful ID? I would assume those are different

backend/api/src/printful-webhook.ts
```diff
@@ -0,0 +1,223 @@
...
+import { insertNotificationToSupabase } from 'shared/supabase/notifications'
+import {
+  MANIFOLD_AVATAR_URL,
+  MANIFOLD_USER_NAME,
+  MANIFOLD_USER_USERNAME,
+} from 'common/user'
+import { nanoid } from 'common/util/random'
+
+// Printful webhook event types we care about
+// See: https://developers.printful.com/docs/#tag/Webhook-API
```
wasabipesto, line 18: Have you tested this against the simulator? https://www.printful.com/api/webhook-simulator

backend/api/src/printful-webhook.ts
```diff
@@ -0,0 +1,223 @@
...
+      tracking_url: string
+      ship_date: string
+    }
+  }
+}
+
+// Map Printful statuses to our shop_orders statuses
+const STATUS_MAP: Record<string, string> = {
+  package_shipped: 'SHIPPED',
+  order_canceled: 'CANCELLED',
```
wasabipesto, line 43: Ideally we would match canceled vs cancelled to stay consistent; minor nit

wasabipesto: Another nit: I believe Printful can mark an order as canceled on their end without us being the ones to trigger the cancellation. I think that would be a good thing to bring to admin attention same as `order_failed` but would require more logic and is probably unlikely to happen.

backend/supabase/merch_stock_status.sql
```diff
@@ -0,0 +1,16 @@
+create table if not exists
+  merch_stock_status (
+    item_id text primary key not null,
+    out_of_stock boolean default false not null
+  );
+
+-- Pre-seed all merch items as out of stock (flip to in-stock via admin toggle when ready)
+insert into merch_stock_status (item_id, out_of_stock) values
```
wasabipesto, line 8: This is missing `merch-wordmark-tshirt`

web/pages/admin/merch.tsx
wasabipesto: Yeah it looks like txnId is useful here but we should also show printfulOrderId

backend/api/src/printful-webhook.ts
todwadd: nah not specifically the simulator, but I tested on dev with it all connected, it works!

backend/api/src/printful-webhook.ts
todwadd: Yeah there are a few edge cases but because each order requires manual approval it should be ok for now. If we reach a volume to where these things are annoying I can add more things in

Edit: plus, I can't recall exactly but at the time I was putting focus on connecting events we would specifically want to notify the user for (shipped), or immediately refund (error placing order)

backend/api/src/printful-webhook.ts
```diff
@@ -0,0 +1,223 @@
...
+      service: string
+      tracking_number: string
+      tracking_url: string
+      ship_date: string
+    }
+  }
+}
+
+// Map Printful statuses to our shop_orders statuses
+const STATUS_MAP: Record<string, string> = {
```
wasabipesto, line 41: I see `shop/types.ts` has fields for `printfulStatus` in addition to `metadata`, `shippedTime`, `deliveredTime`, etc. Are we not implementing those?

backend/api/src/cancel-merch-order.ts
```diff
@@ -0,0 +1,138 @@
...
+        const refundTxn: TxnData = {
+          category: 'SHOP_PURCHASE',
+          fromType: 'BANK',
+          fromId: 'BANK',
+          toType: 'USER',
+          toId: order.user_id,
+          amount,
+          token: 'M$',
+          description: `Admin refund: merch order ${orderId} cancelled`,
+          data: {
```
wasabipesto, line 63: It might be good to save the order ID here for correlation later (and the similar txn in `refundOrder`)

backend/api/src/printful-webhook.ts
```diff
@@ -0,0 +1,223 @@
...
+          description: `Auto-refund: Printful ${reason} (order ${order.id})`,
+          data: { itemId: order.item_id, merchOrder: true, refund: true },
+        }
+
+        await runTxnOutsideBetQueue(tx, refundTxn)
+        await tx.none(
+          `UPDATE shop_orders SET status = 'REFUNDED' WHERE id = $1`,
+          [order.id]
+        )
+        console.warn(`Webhook auto-refund: ${amount} mana to ${order.user_id} (${reason})`)
```
wasabipesto: I think it would be good to notify the user here that their order has been canceled and refunded

web/pages/shop.tsx
```diff
@@ -1361,24 +1392,129 @@ function TicketItemCard(props: {
   )
 }
 
+/** Renders the merch grid plus its "Merch" heading (heading hidden on the
+ *  merch filter, since the page header already names the section). Pulled out
+ *  so the shop page can render it in two positions: promoted above the regular
+ *  grid when any merch item is NEW to the user, or in its default spot below. */
```
wasabipesto, line 1398: This merch heading (and the "just added" heading) doesn't play nice with sorting options (like by price or by name). Inside each section the options are sorted but I think most people would expect that all items on the page should be sorted together. After playing with it I don't think anything is wrong per se but it does make the sort options a bit less useful.

common/src/shop/items.ts
```diff
@@ -805,10 +811,51 @@ export const SHOP_ITEMS: ShopItem[] = [
...
+      ],
+      Navy: [
+        { label: 'Front', url: '/merch/White-wordmark-navy-t-shirt-front.png' },
+        { label: 'Back', url: '/merch/White-wordmark-navy-t-shirt-back.png' },
+      ],
+    },
+    // Bella + Canvas 3001 unisex tee. 4XL / 5XL variants exist on Printful but
+    // are intentionally excluded — their wholesale price pushes margin below
+    // acceptable at the uniform M$5,000 shelf price.
+    variants: [
```
wasabipesto, line 842: Have you redeployed the API with these variants added? I kept getting errors when trying to test the wordmark tee but the rest worked fine.

todwadd: The wordmark one is new to this branch so hasn't been redeployed but the rest are live (hidden) in the shop and have been for a while. Have tested it all with local API

todwadd:
- Added merch-wordmark-tshirt to the merch_stock_status.sql seed.
- Webhook order_failed events now notify the user (new notifyAutoRefunded, gated so duplicate webhooks don't re-notify).
- Both refund txns (admin-cancel and webhook auto-refund) now carry printfulOrderId + shopOrderId in data for correlation.
- /admin/merch order column stacks Printful ID above the Manifold ref so both are visible. Cancel-confirm dialog also surfaces the Printful ID.
- printfulStatus and shippedTime were already implemented (webhook + Phase 3 of shop-purchase-merch). deliveredTime + 'DELIVERED' status removed — Printful's webhook API doesn't push delivery events, so they were dead schema. Comment in the type notes this.
- Normalized "canceled" spelling to one L in merch-introduced files. DB enum 'CANCELLED' left as-is to match existing rows and the rest of the codebase (ticket flows); added a comment in STATUS_MAP explaining the US event → UK enum bridge.
- Unrelated but in the same commit: scrubbed t-shirt size from the public balance-log description (was leaking sizes); full size/color/variant now lives in shop_orders.metadata instead, and /admin/merch shows the variant inline next to the item name.

minor deferrals
- sort cleanup
- some of the "cancel"->"cancelled" things need data migration to properly cleanup, haven't done that
- printful-initiated cancellations don't notify users, but we will manually see and reconcile

todwadd:
printful-initiated cancellations do notify users now nvm

wasabipesto (approved):

PR #3742, opened 2026-04-03
Unslopify the referrals page

ManiFed:
This is not vibecoded, so please excuse how it's not high-quality


todwadd:
@ManiFed FYI, referrals page has been changed quite a bit. Let me know if you have any suggestions, it is far more informative and now properly accurate. Because the referral process is actually complicated now, (3 events: signup+attribution, trade+250, verify+1,000) I'm leaving the explanation.

Thank you for the contribution!

PR #3751, opened 2026-04-13
feat: replace hardcoded BOT_USERNAMES with proper bot toggle

todwadd:
Adds is_bot column to users table with mod/admin toggle endpoint, replacing the 109-entry hardcoded username list that required PRs to update.

- Migration: adds is_bot column, seeds existing bot usernames
- New set-bot-status API endpoint (mod/admin gated)
- Backend queries use is_bot column instead of BOT_USERNAMES array
- Frontend: Mark Bot / Unmark Bot button in user settings modal
- BOT_USERNAMES kept as fallback during transition, marked @deprecated


ManiFed:
this is good!

PR #3763, opened 2026-04-20
fix: queue push notifications on cold start until WebView is ready

todwadd:
On Android, tapping a push notification during cold start would navigate to the home screen instead of the notification target. The WebView silently drops source prop changes during initial mount, so the notification URL was lost. This fix queues the notification in a ref and delivers it via the message bridge once the web app signals 'startedListening'.

### DIFF
--- a/native/App.tsx
+++ b/native/App.tsx
@@ -71,6 +71,12 @@ const App = () => {
   // This tracks if the app has its nativeMessageListener set up
   // NOTE: After the webview is killed on android due to OOM, this will always be false, see: https://github.com/react-native-webview/react-native-webview/issues/2680
   const listeningToNative = useRef(false)
+  // Stores a notification that arrived before the webview was ready (cold start).
+  // Delivered once the web app signals 'startedListening'.
+  const pendingNotification = useRef<{
+    notification: Notification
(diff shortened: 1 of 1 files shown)


native/App.tsx
```diff
@@ -169,15 +175,19 @@ const App = () => {
...
       communicateWithWebview('notification', notification)
+      setEndpointWithNativeQuery(destination)
+    } else {
+      // Cold start: webview isn't ready yet. Stash the notification so we can
+      // deliver it once the web app signals 'startedListening'. We intentionally
+      // do NOT call setEndpointWithNativeQuery here because on Android the
+      // WebView ignores rapid source prop changes during its initial mount,
+      // causing the notification URL to be silently dropped.
+      log('Queuing notification for delivery after webview is ready')
+      pendingNotification.current = { notification, destination }
```
wasabipesto, line 189: Nit: if this bug is only present on Android, this queuing behavior should be gated to only run on Android. I'm not sure how much of a delay it would cause on iOS but it would probably be noticeable.

wasabipesto (commented):
I haven't tested this but it seems safe to merge.

PR #3767, opened 2026-04-21
comments: unlock commenting for users who have purchased mana

todwadd:
Previously only identity-verified (or grandfathered) users could comment. Now commenting is also unlocked by a prior mana purchase, an anti-spam proxy for verification.

- add canComment(user) helper: canReceiveBonuses OR purchasedMana
- swap backend gates in create-comment + create-post-comment to canComment
- swap frontend gate in comment-input; banner copy notes that purchasing mana also unlocks commenting


wasabipesto:
Looks like this was superseded by [1f79b54](https://github.com/manifoldmarkets/manifold/commit/1f79b54ff9efc940c235c6174aad29f579e03f25)

PR #3794, opened 2026-05-03
Let the frontend set a mana budget for buying prize entries

wasabipesto:
1. The API now accepts an optional mana budget alongside (or instead of) `numTickets`. At least one must be provided and if both are given, the tighter constraint wins.
2. When `maxManaSpent` is provided, the handler calls `calculateSweepstakesTicketsFromMana()` using the latest ticket count, ensuring the user isn't charged more than their confirmed budget. The final `manaSpent` is always recalculated from the resolved `numTickets`.
3. The purchase form now passes `maxManaSpent` instead of `numTickets`, and the UI copy is updated with a tooltip "Your mana spend is fixed; the final entry count is calculated at purchase time.".


wasabipesto:
image

PR #3796, opened 2026-05-04
leagues: pay Silicon bots Platinum-tier prizes (top 6, M$100 floor)

todwadd:
Silicon (division 0) previously had no rewards. Adds a top-6 prize ladder mirroring Platinum (M$800/400/360/320/280/240), gated on mana_earned >= 100 so idle bots can't win when other bots lose.

Bots bypass the iDenfy verification gate (canReceiveBonuses) for LEAGUE_PRIZE since they can't be KYC'd. Bypass is scoped to division 0 only.

When a bot ranks in the prize zone but misses the M$100 floor, the season-end notification carries a typed missedPrizeReason and the frontend renders explanatory copy instead of a silent zero bonus.

prizesByDivisionAndRank is now indexed by division directly (was division - 1) so Silicon fits naturally at index 0.

### DIFF
--- a/backend/shared/src/create-notification.ts
+++ b/backend/shared/src/create-notification.ts
@@ -674,6 +674,7 @@ export const createLeagueChangedNotifications = async (
       previousLeague: item.previousLeague,
       newLeague: item.newLeague,
       bonusAmount: item.bonusAmount,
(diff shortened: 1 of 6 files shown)


wasabipesto:
LGTM

PR #3825, opened 2026-05-14
Personalized mana offers

manifoldbot:
## Summary

Adds a personalized mana sale flow: when a customer's merch order ships, they receive a one-time offer to buy **M5,000 mana for $35 (crypto) or $40 (card)** — a 30%/20% discount off the standard $50 rate. Offer is valid for 24h after they first view `/checkout`, stackable across multiple merch purchases, and enforced single-use atomically on both payment rails.

## What's in scope

- New `personalized_mana_offers` table + cross-method pending-lock columns
- 3 new API endpoints: `get-personalized-mana-offers`, `activate-personalized-mana-offers`, `release-personalized-mana-offer-lock`
- Wired into existing Printful webhook (`package_shipped` creates the offer), Stripe webhook (atomic redemption), Daimo webhook (atomic redemption)
- New offer card on `/checkout`, "−30% / −20%" tags on the claim buttons, dynamic "OFFER: Xh Ym" badge on every Get-mana button across the site
- In-app + push notification; user-toggleable preference under "Other balance changes"
- Hourly cron flips expired offers; backfill script grants offers to historical SHIPPED orders

## Schema changes

- **New table** `personalized_mana_offers` — uuid PK, FK to `shop_orders.id` (unique per order), state machine via status enum
- **`crypto_payment_intents.offer_id`** added (nullable FK) so the Daimo webhook can attribute redemptions

Migrations (additive only, no data rewrites):
1. `backend/supabase/migrations/2026051201_personalized_mana_offers.sql`
2. `backend/supabase/migrations/2026051202_personalized_mana_offers_pending_lock.sql`

## Design choices worth flagging

- **Activation-on-first-view** (not on-creation): the 24h timer starts when the user opens `/checkout`, not when the merch ships. Avoids burning the offer for users who don't log in within a day.
- **Stripe-side single-use via Promotion Code** (`max_redemptions: 1`, `applies_to.products`): the discount is enforced atomically at payment time, can't be cross-applied to standard mana tiers.
- **Cross-method pending lock**: 30-min server-side lock when a Stripe OR Daimo session is started, so a user can't accidentally pay full price on a second tab after their first session went through.
- **2h SQL grace + 1h Stripe session expires_at**: an in-flight checkout that started before offer expiry is always honored. Stripe's own UI rejects sessions past `expires_at + 1h`, the SQL grace adds another hour of webhook-latency buffer.
- **Multi-tab fallback in webhooks**: if a user has multiple offers and the named one was already redeemed, the webhook claims any *other* active offer instead of dropping to standard rate.
- **Refund does NOT void the offer**: post-ship refunds are vanishingly rare; pulling an offer the user has already seen is worse UX than eating the tiny loss.


backend/api/src/create-daimo-session.ts
```diff
@@ -75,20 +76,86 @@ export const createDaimoSession: APIHandler<'create-daimo-session'> = async (
     throw new APIError(500, 'Crypto payment service not configured')
   }
 
-  const payload = {
+  const { offerId } = props
+
+  // Keep in sync with createcheckoutsession's PAYMENT_PENDING_LOCK_MINUTES.
+  const PAYMENT_PENDING_LOCK_MINUTES = 30
```
wasabipesto: Should this live in `common/personalized-mana-offer` so that it can be kept in sync?

common/src/personalized-mana-offer.ts
```diff
@@ -0,0 +1,21 @@
+export const OFFER_MANA_AMOUNT = 5000
+export const OFFER_PRICE_STRIPE = 40
+export const OFFER_PRICE_CRYPTO = 35
```
wasabipesto, line 3: These values are set in both this file and also in the database, but I believe they are only ever read from this file. My impression is that this is mostly due to having to set up unique Stripe SKUs for each offer, is that true? I think it would be much more convenient if we could just reference the values in the database so we could more easily use it for similar promos in the future.

backend/supabase/migrations/2026051201_personalized_mana_offers.sql
```diff
@@ -0,0 +1,44 @@
+-- Personalized mana sale offers granted to users after merch purchases ship.
+-- Each merch order grants one offer (uniqueness enforced by index on shop_order_id).
+-- Offers are 'pending' until the user first views /checkout, at which point all
+-- of that user's pending offers flip to 'active' with a shared 72h expiry.
```
wasabipesto, line 4: This comment notes a 72h expiry, but I believe the actual expiry time is 24h.

backend/shared/src/notifications/create-personalized-mana-offer-notification.ts
```diff
@@ -0,0 +1,99 @@
+import { Notification } from 'common/notification'
+
```
wasabipesto: Is there a reason this file is double-spaced?

backend/shared/src/notifications/create-personalized-mana-offer-notification.ts
```diff
@@ -0,0 +1,99 @@
...
+    sourceType: 'personalized_mana_offer',
+
+    sourceUserName: MANIFOLD_USER_NAME,
+
+    sourceUserUsername: MANIFOLD_USER_USERNAME,
+
+    sourceUserAvatarUrl: MANIFOLD_AVATAR_URL,
+
+    sourceText: 'You have a personalised mana sale available!',
```
wasabipesto: Could we include the discount amount here? It would also be good to tell them why they got this offer. Something like "As a thank you for [buying some merch recently], you have a time-limited offer for [up to 30%] off your next purchase of [M$5000]!"

common/src/notification.ts
```diff
@@ -372,6 +373,11 @@ export const NOTIFICATION_DESCRIPTIONS: notification_descriptions = {
     simple: 'Merch order updates',
     detailed: 'Updates when your merch order ships or is canceled',
   },
+  personalized_mana_offer: {
+    simple: 'Personalized mana sales',
+    detailed:
+      'Flash discounts on mana purchases offered after merch shipments',
```
wasabipesto: If we intend to generalize this system, we should keep this setting text more broad

backend/supabase/migrations/2026051201_personalized_mana_offers.sql
```diff
@@ -0,0 +1,44 @@
...
+
+create table if not exists
```
wasabipesto, line 23: I'm okay with this table creation being a migration, but I believe the historical flow has been to apply the changes then run `regen-schema.ts`

backend/supabase/migrations/2026051201_personalized_mana_offers.sql
```diff
@@ -0,0 +1,44 @@
...
+  personalized_mana_offers (
+    id uuid primary key default gen_random_uuid(),
```
wasabipesto: Nit: In other tables this is typically accomplished with `id text primary key default random_alphanumeric (8) not null`

wasabipesto (commented):
High level: 
- I think this is good for now but I'm concerned with how it's structured now versus our ideas of how to use it in the future. It assumes that all discount offers will be time-sensitive and hardcodes several aspects that I think we should ideally leave up to the database as a source of truth.
- I don't love the 24-hour window, I understand it's good for urgency but it's a little short for my tastes. That being said, if we made it any longer I would want a way to dismiss the badge.
- I think it might be good to send emails to anyone who is getting this offer but hasn't logged in for a week or so.

backend/supabase/migrations/2026051201_personalized_mana_offers.sql
todwadd: ok I rebuilt dev and have updated this. Will be correct on prod. Would have bothered me forever if I didn't fix it

common/src/personalized-mana-offer.ts
todwadd: Yes, that is why. Everything else could reference the db but the stripe stuff basically needs a unique product so that the code only works for that product (i.e. 20% off 5,000 Mana) or else someone could use the code on a different item like 100k mana. My first version was a lot cleaner but I found several ways to abuse it which I have patched, as a result, we have this weird half measure.

The db columns are already in place with these as defaults, so the next discount trigger should mostly just be 
(1) different column values per offer row and 
(2) a parallel set of Stripe SKU env vars. 

Should be ok to deal with this later though if/when we add more discount triggers.

PR #3826, opened 2026-05-14
fix: Correct UI inconsistencies in 2-answer MC markets

wasabipesto:
- Use correct limit order percent for NO side bets
- Show correct answer name on trade table for NO side bets
- Move answer edit pencil to side of answer text instead of underneath
- Use custom colors on position table, green/red on profit


wasabipesto:
Testing on this market: https://prod-git-wp-binary-mc-ui-mantic.vercel.app/NFNL2zc/who-will-win-the-republican-primary-Zn86SEQZ9O

PR #3839, opened 2026-05-22
Unverified tier

todwadd:
Introduces an explicit **Unverified** user tier so new users can participate before KYC instead of hitting a cliff. Quest/streak/unique-trader bonuses now scale by effective tier (verification + subscription combined) instead of being all-or-nothing on `canReceiveBonuses`. Subscribers always get full bonuses regardless of KYC; verification is required only for the prize drawing.

0.2x quest/streak bonus, 0.5x trader bonus, no referral bonus

- Signup balance M50 → M500 pre-KYC. Verification bonus reduced symmetrically (M1000 → M500) so total to a fully-verified user is unchanged at M1000. 

- Commenting on others' markets unlocks after 7 days from signup, OR on KYC, OR on first mana purchase, OR on any active subscription. Failed-KYC users see a distinct denied state with `info@manifold.markets` link instead of a dead verify button.
- Prize drawing: KYC is still required. /prize page banner explicitly states subscriptions don't unlock entry (it's a regulatory requirement, not a bonus tier).
- Failed-KYC (`bonusEligibility='ineligible'`) treated as Unverified for bonus purposes — they earn reduced amounts instead of zero, per product call. UI surfaces point them to support email for re-enabling verification or manual update.

## UI changes

- /membership now shows a 5-column comparison (Unverified | Verified | Plus | Pro | Premium) with the user's effective tier highlighted. Monthly Value calculator below the table supports all 5 tiers.
- VerifyPhoneNumberBanner gains `dismissible` and `compact` props. /membership uses both (always visible, smaller).
- ReducedBonusNotice component (`web/components/upsell/`) shown inline in streak modal, quest modal, and as a subtitle on bonus notifications.
- Referrals page shows an explicit banner for unverified users, referral links work but bonuses are locked.


todwadd:
@wasabipesto it all makes sense to me, the risky changes are only in the smaller files at the top. You can ignore all the big UI diffs, they are fine.

Everything has been tested and works: all amounts are granted, or withheld, when appropriate. Notifications display accordingly.

backend/api/src/on-create-bet.ts
```diff
@@ -366,35 +380,45 @@ const payBettingStreak = async (
...
+    )
 
-    // Send them the bonus times their streak, with supporter multiplier
+    // Send them the bonus times their streak, with effective-tier multiplier
     const baseBonus = Math.min(
       BETTING_STREAK_BONUS_AMOUNT * newBettingStreak,
       BETTING_STREAK_BONUS_MAX
     )
-    const bonusAmount = Math.floor(baseBonus * questMultiplier)
+    const bonusAmount = Math.floor(baseBonus * streakMultiplier)
```
wasabipesto: Nit: Now that we have fractional multipliers it might be good to preemptively change this away from floor(). Right now I believe all bonuses will come out to whole numbers (M$5 * 0.2 = M$1) but if we change those multipliers in the future this could get rounded down to zero in some cases. I assume the floor was being used here to avoid something like M$5.0000001 but we could change it to round to the nearest .01 or something. Separately we should also show decimals for small numbers but that's already on my list.

common/src/supporter-config.ts
```diff
@@ -242,6 +242,118 @@ export function getTierInfo(tier: SupporterTier) {
...
+  'basic',
+  'plus',
+  'premium',
+]
+
+// Bonus multipliers for quest/streak/referral/unique-trader rewards by
+// effective tier. Subscriber rows for quest/referral MUST match
+// SUPPORTER_BENEFITS above. Unique-trader does NOT scale with subscription
+// per product decision — it's the same for verified/grandfathered/all subs.
+export const EFFECTIVE_TIER_BONUS_MULTIPLIERS: Record<
```
wasabipesto: Nit: I don't love that this duplicates functionality/data from the central benefits config, it would be nice if they were more unified instead of adding another layer on top. Not a blocker, I might just refactor it later.

web/components/comments/comment-input.tsx
```diff
@@ -298,35 +316,135 @@ function VerifyToCommentPrompt(props: { className?: string }) {
...
+          </span>
+        </Col>
       </Row>
       {error && <div className="text-scarlet-500 mt-1 text-xs">{error}</div>}
     </Col>
   )
 }
 
+function useCountdown(targetMs: number): string {
+  const [now, setNow] = useState(() => Date.now())
```
wasabipesto: My reviewer flagged this as a potential hydration mismatch risk, haven't looked into it since those can be hard to debug

wasabipesto (approved):
Nothing blocking here, haven't tested any UI or anything

backend/api/src/on-create-bet.ts
todwadd: there are some cases where it's decimal, even with the live multipliers (i.e. share quest gives 5, 7.5 if you are a plus subscriber). The floor was a crude way to avoid decimals for new / unverified users. It's probably better if we later add at least 1 decimal point for very small amounts, bc it cleans this all up and fixes other display weirdness (note: we should also make sure that we never show '0' when we do add decimals)

todwadd: I removed the floor here, so resolving this

common/src/supporter-config.ts
todwadd: I refactored it to use the same central config. Makes way more sense. Might not be perfect but significantly better, and have re-tested everything. Some other minor changes too but nothing needing significant review. Thank you for the comments!

PR #3872, opened 2026-06-09
fix(sports-admin): include TIMED matches in fixture fetch and select-all

shankypanky:
## Summary

- football-data.org distinguishes `SCHEDULED` (no confirmed time) from `TIMED` (confirmed kickoff) — WC 2026 group stage games are all `TIMED` since the full schedule is set
- The admin fixtures endpoint was querying with `status=SCHEDULED` only, so TIMED games were never returned
- The frontend select-all and auto-select filters also only matched `SCHEDULED`, so any TIMED games that did appear couldn't be selected
- Fix: fetch without a status filter, then keep only `SCHEDULED` and `TIMED` matches (filtering out IN_PLAY, FINISHED, POSTPONED, etc.)

## Test plan

- [ ] Fetch games in the sports admin panel for an upcoming WC date range
- [ ] Confirm TIMED matches appear in the fixture table
- [ ] Confirm "Select all" selects all non-existing TIMED and SCHEDULED fixtures
- [ ] Confirm auto-select on load also pre-selects them


wasabipesto:
Note to Claude: This branch has merge conflicts since it includes some changes that have already been merged into main and formatted. Please check the current status of the main branch and remove changes or commits as necessary to avoid conflicts. I believe the intent of this PR was to only include the changes from the last commit (TIMED match changes). Additionally, please remove unrelated changes to the claude settings file. In the future please always check the current status of the main branch to avoid unnecessary merge conflicts blocking and keep PRs clean.
Note to Stef: The actual intended changes look good, I could pick out the diff like the previous PR if it's urgent but I wanted to pass this along to Claude for memory purposes.

PR #3873, opened 2026-06-09
fix(sports): extend close time for knockout matches to cover ET and penalties

shankypanky:
## Summary
- Knockout matches can run up to ~170 min (90' regulation + halftime + 30' ET + ET break + ~30' penalty shootout)
- The existing flat 2.5 h close time offset could close betting mid-extra-time for deep knockout matches
- Adds a `KNOCKOUT_CLOSE_OFFSET_MS` constant (3.5 h) used for all knockout stages (R16, QF, SF, 3rd place, Final)
- Uses `Math.max` so if a config ever sets `closeTimeOffsetMs` higher than 3.5 h, that takes precedence

## Test plan
- [ ] Create a knockout stage market via admin panel and verify close time is kickoff + 3.5 h
- [ ] Verify group stage markets still use kickoff + 2.5 h


shankypanky:
Closing — branch had extra noise from parent branch, reopening from a clean cherry-pick.

PR #3882, opened 2026-06-09
feat(sports): position indicator stripe and hovercard on match cards

shankypanky:
## Summary

- Adds a position indicator to `SportsMatchCard`: a 5px left-edge colored stripe on outcome rows where the user holds a position, and a **Positions** pill in the card footer (solid pip for positions, outlined pip for limit orders)
- Hovering the pill shows a breakdown of the user's positions (amount + P&L) and open limit orders for that match
- Clicking the indicator navigates to the market page
- Depends on PR #3883 — merge that first

### DIFF
--- a/web/components/contract/positions-hovercard.tsx
+++ b/web/components/contract/positions-hovercard.tsx
@@ -1,0 +1,136 @@
+import clsx from 'clsx'
+import { Row } from 'web/components/layout/row'
+
+export type UserPosition = {
+  name: string
+  color?: string
+  amount: number
+  profit: number
+}
+
+export type UserLimitOrder = {
+  name: string
+  prob: number
+  amount: number
+}
+
+export type UserNumericSummary = {
+  total: number
+  buckets: number
+  profit: number
+  orders?: { total: number; count: number }
+}
+
+export type PositionsData = {
+  positions: UserPosition[]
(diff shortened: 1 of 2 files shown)


web/components/contract/positions-hovercard.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+export type PositionsData = {
+  positions: UserPosition[]
+  limitOrders: UserLimitOrder[]
+  numericSummary?: UserNumericSummary
+}
+
+const MAX_SHOWN = 4
+
+export function PositionsHovercard({
```
wasabipesto, line 33: This file was in PR #3883, was it intentional to also commit it here?

wasabipesto (commented):
Aside from the hovercard file notes from the previous PR this one looks good to go

web/components/contract/positions-hovercard.tsx
shankypanky: It was originally in the code for both market cards, but then I realized it should be its own component. 3883 is the hovercard as a standalone component which should merge first, then the other two should reference that code for the position hovercard. I'm sure I'll need to make updates on both market cards after 3883 is merged so the references are proper, sorry I didn't update with a note about that after trying to split it off.

PR #3883, opened 2026-06-09
feat: shared PositionsHovercard component

shankypanky:
## Summary

Extracts position display into a shared component at `web/components/contract/positions-hovercard.tsx` for reuse across sports match cards and dashboard market cards.

Exports:
- `PositionsHovercard` — the hovercard content component
- `PositionsData`, `UserPosition`, `UserLimitOrder`, `UserNumericSummary` — types

Supports:
- Named positions with amount and P&L; optional `color` for a team/answer-colored pip (sports cards use this, other cards omit it)
- Limit orders with name, probability, and amount
- Numeric summary for bucket-style markets (total, bucket count, P&L, optional pending orders)
- Capped at 4 entries each with an overflow note

No card changes in this PR — follow-up PRs will import from this component.

### DIFF
--- a/web/components/contract/positions-hovercard.tsx
+++ b/web/components/contract/positions-hovercard.tsx
@@ -34,8 +34,9 @@ export function PositionsHovercard({
   limitOrders,
   numericSummary,
 }: PositionsData) {
-  const shownPositions = positions.slice(0, MAX_SHOWN)
-  const extraPositions = positions.length - shownPositions.length
+  const sorted = [...positions].sort((a, b) => b.amount - a.amount)
+  const shownPositions = sorted.slice(0, MAX_SHOWN)
+  const extraPositions = sorted.length - shownPositions.length
   const shownOrders = limitOrders.slice(0, MAX_SHOWN)
   const extraOrders = limitOrders.length - shownOrders.length
 
--- a/web/components/contract/positions-hovercard.tsx
+++ b/web/components/contract/positions-hovercard.tsx
@@ -1,0 +1,136 @@
+import clsx from 'clsx'
+import { Row } from 'web/components/layout/row'
+
+export type UserPosition = {
+  name: string
+  color?: string
+  amount: number
+  profit: number
+}
+
+export type UserLimitOrder = {
+  name: string
+  prob: number
+  amount: number
+}
+
+export type UserNumericSummary = {
+  total: number
+  buckets: number
+  profit: number
+  orders?: { total: number; count: number }
+}
+
+export type PositionsData = {
+  positions: UserPosition[]
+  limitOrders: UserLimitOrder[]
+  numericSummary?: UserNumericSummary
+}
+
+const MAX_SHOWN = 4
+
+export function PositionsHovercard({ positions, limitOrders, numericSummary }: PositionsData) {
+  const shownPositions = positions.slice(0, MAX_SHOWN)
+  const extraPositions = positions.length - shownPositions.length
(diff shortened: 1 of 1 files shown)


web/components/contract/positions-hovercard.tsx
```diff
@@ -0,0 +1,150 @@
...
+                {p.color !== undefined && (
+                  <span
+                    className="h-1.5 w-1.5 flex-shrink-0 rounded-full"
+                    style={{ backgroundColor: p.color }}
+                  />
+                )}
+                <span className="min-w-0 truncate text-sm">{p.name}</span>
+              </Row>
+              <Row className="flex-shrink-0 items-center gap-1.5 text-sm font-semibold">
+                <span>Ṁ{Math.round(p.amount)}</span>
```
wasabipesto: Please use standard formatters for mana and shares from common/util/format

web/components/contract/positions-hovercard.tsx
```diff
@@ -0,0 +1,150 @@
...
+const MAX_SHOWN = 4
+
+export function PositionsHovercard({
+  positions,
+  limitOrders,
+  numericSummary,
+}: PositionsData) {
+  const shownPositions = positions.slice(0, MAX_SHOWN)
+  const extraPositions = positions.length - shownPositions.length
+  const shownOrders = limitOrders.slice(0, MAX_SHOWN)
```
wasabipesto, line 41: These should likely have some ordering before they are truncated, like closest to current probability or largest. Up to you on implementation but should be intentional.

wasabipesto (changes requested):
Indicator looks good overall. In the future (when we aren't in person) also be sure to include a screenshot in the PR description if the changes aren't immediately viewable.

web/components/contract/positions-hovercard.tsx
shankypanky: Great point - thanks! I'll make sure I include screenshots in the future to make review smoother.

web/components/contract/positions-hovercard.tsx
shankypanky: Ah, true - I think largest to smallest? Do you have a preference?

wasabipesto: Largest to smallest sounds perfectly reasonable

shankypanky: Shall I make that adjustment and update the PR?

wasabipesto: Sure, go ahead. Or you can close this PR and edit the version of this file in #3882 since it's in both.

shankypanky: Oh, that's worth talking about too - I intended break off the hovercard into a standalone component to be referenced by both of the market cards, but that was an afterthought so maybe I didn't organize that correctly. 

The idea was that this one would merge first, then the other two would reference the hovercard component for dashboards, but I probably bungled that since I tried to split it off after the other two were created.

shankypanky: Okay this one is updated - should sort descending now

web/components/contract/positions-hovercard.tsx
shankypanky: @wasabipesto 

> Done — all six money displays now use formatMoney from common/util/format, which handles the Ṁ moniker, rounding, and locale formatting consistently with the rest of the site. The manual Math.round() calls are gone.

PR #3909, opened 2026-06-15
Fix/push token multi account leak

todwadd:


backend/scripts/clear-shared-push-tokens.ts
todwadd, line 1: I ran this script already; we will only need to run it again if this PR sits for a while (not a big deal).

PR #3915, opened 2026-06-19
Sidebar more menu

todwadd:
Collapse desktop sidebar bottom nav into a "More" popover

The bottom sidebar items (About, Referrals, theme, sign out) now live behind a "More" dropdown on desktop that floats up and overlaps the content above it, so the menu can grow without crowding short viewports. Also adds a Jobs link (/jobs) to that menu. Mobile drawer stays flat.

### DIFF
--- a/web/components/nav/sidebar.tsx
+++ b/web/components/nav/sidebar.tsx
@@ -435,6 +435,7 @@ const getMobileNav = (
     },
     { name: 'Leagues', href: '/leagues', icon: TrophyIcon },
     { name: 'Forum', href: '/posts', icon: ChatIcon },
+    { name: 'Jobs', href: '/jobs', icon: BriefcaseIcon },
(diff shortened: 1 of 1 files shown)


wasabipesto:
I'm curious why this is collapsed on desktop, where we typically have plenty of room, but not mobile, which is generally much shorter?

For comparison, on my screen:

image

image

todwadd:
> I'm curious why this is collapsed on desktop, where we typically have plenty of room, but not mobile, which is generally much shorter?
> 
> 
> 
> For comparison, on my screen:
> 
> 
> 
> image
> 
> 
> 
> image

I'm not sure exactly how to do it on mobile, if you have any ideas lmk or add them. More->more-> seemed bad lol

If there's a good way to autodetect if the desktop change is necessary that would be good too, some devices need it and yours clearly does no

Also as a way to get people to see the jobs tab.

PR #3934, opened 2026-06-29
feat: cpmm-multi-2 — per-answer p for multi-choice markets (opt-in, gated, v1 frozen)

evand:
### TL;DR — and the safety headline

This adds **`cpmm-multi-2`**, an opt-in multi-choice market mechanism that gives each answer its own
`p` (the degree of freedom a binary `cpmm-1` market already has, but `cpmm-multi-1` lacks).

**Nothing changes for any existing market, user, or integrator.** `cpmm-multi-1` is frozen
bug-for-bug. Everything new is gated behind `CPMM_MULTI_2_CREATION_ENABLED` (`common/src/contract.ts`),
which defaults to **`false`** — so with the flag off, the entire diff is **inert** and v1 is
byte-identical. **Merging this does not launch anything**; it lands the capability, dark. Activation
is a separate one-line flag flip you control later, and rollback is "set it back to false."

### Why — one missing degree of freedom

Three long-standing `cpmm-multi-1` limitations all trace to the same gap: a multi-choice answer has
no per-answer `p`. Add it (defaulting to `0.5` when absent, so every existing market reads back
identically) and all three dissolve at once:

1. **`p = 0.5` is hardcoded** → blocks non-uniform initialization (`initialProb` per answer) and
   *lossless* liquidity addition. Today `addCpmmLiquidityFixedP` literally throws shares away to hold
   probability on a skewed pool — at fixed `p` you can't deepen an asymmetric pool without moving
   price. With per-answer `p`, you inject mana into both reserves and let `p_i` float to absorb it
   losslessly.
2. **Resting limit orders get transient-overshoot fills during multi-leg auto-arb.** A multi-answer
   buy overshoots `Σp > 1`, then the auto-arb bets it back down. A resting limit in that overshoot
   band is consumed on the way up and kept on the way down — filled even though the final price ends
   up *past* it. (Empirically confirmed against current Manifold: a NO maker priced above both the
   initial and final prob is filled for 14–26 mana on the multi-buy path; the single-answer buy is
   monotone and does not exhibit it.) The v2 fix is **net-movement settlement** (a temporary reverse
   limit at the crossed price that a down-move re-crosses for an exact refund): you fill when price
   *reaches* your limit, and the cost curve becomes a state function, `C(δ) + C(−δ) = 0`.
3. **Auto-arb is `O(n log² N)`** (a nested binary search). The per-answer cost legs have direct
   closed forms, collapsing the search structure to `O(n log N)`. *This is a secondary benefit, not
   the pitch* — server-side arb compute is dwarfed by DB/IO, and honest measurement (the in-tree
   `BENCH=1` perf bench) shows the v2 solve is currently ~7–10× v1 **wall-clock** on shallow-pool
   fixtures (~160ms at n=10): the asymptotic win is real but constants aren't tuned, and a perf pass
   gates the **flag flip**, not this merge. The direct formulas earn their place by being the
   **exact, reversible** core that fixes 1 and 2 require, not by shaving milliseconds.

### Design: freeze v1, gate v2, migrate lazily

- **`cpmm-multi-1` stays frozen** — untouched, bug-for-bug. Every live market and every integrator's
  current forecasts keep working.
- **`cpmm-multi-2` = `cpmm-multi-1` + per-answer `p` + reversible-limit auto-arb + direct-formula
  core.** New markets opt in. (Deliberately *out of scope*: retrofitting the closed form onto v1 — it's
  provably identical on the no-limit case but only saves compute, so it's left out for conciseness.)
- **Migration is lazy, no big-bang.** A `cpmm-multi-1` market *is* a `cpmm-multi-2` market with every
  `p_i = 0.5` (same pools, same probabilities). `p` defaults to `0.5` at every read; it's written
  lazily the first time a market converts. The conversion **trigger is an explicit user
  `addLiquidity`** (lossless add = moving `p` off `0.5`), not the scheduler drizzle — so semantics
  never flip under resting orders. The version flip is **API-visible**: behavior never changes without
  the version field changing, so integrators can detect it and switch models.
- **Reserved v2 semantics — per-answer NO resolution of linked answers.** A natural follow-up is
  letting a `shouldAnswersSumToOne` market resolve one answer NO early while the rest stay open
  (probabilities then sum to 1 over the *unresolved* answers; v2's floating per-answer `p` makes the
  renormalization a pure re-pricing with pools untouched — the same move the lossless liquidity add
  already uses). **Nothing in this PR implements or enables that** — linked resolution stays
  whole-market-only — but the semantics are declared now (doc comment on `CPMMMulti`), while zero
  `cpmm-multi-2` markets exist: integrators adding v2 support should not assume "linked ⇒ answers are
  never individually resolved" (`Answer.resolution` already exists and is populated today for
  independent markets). Declaring this before the first v2 market ever mints is what makes the future
  change non-breaking; `cpmm-multi-1` keeps the whole-market-only guarantee unchanged.

### What's in this PR / how to read it

43 commits, grouped into six logical stages (the first five were built and validated to compile
standalone, so you can bisect by stage):

1. **Storage + types** (2 commits) — per-answer `p` column + DB↔TS plumbing. Invisible; defaults `0.5`.
2. **Mechanical widening** (1 commit) — an `isMultiCpmm` helper replaces ~80 `=== 'cpmm-multi-1'` call
   sites across web + backend. **This is find-replace, behavior-preserving, and (with its structural
   sibling in stage 3) accounts for 88 of the 119 files** — skim it. (`isMultiCpmm` matches only
   `cpmm-multi-1` until a v2 market can exist.)
3. **Inert backend engine** (~13 commits, ~18 real `common/` files) — general-`p` cost core, reversible-
   limit YES-basket multi-buy (Approach C), lossless whole-market + independent liquidity add, float-`p`
   subsidy drizzle, an `m>1` basket-redemption conservation fix, and a covering-array **conservation
   grid** (create→buy→sell→limit→resolve, net = 0 across configs) + an out-of-range-limit falsification
   sweep. Unreachable in production (no creation path) — pure latent capability + tests.
4. **Creation + activation** (6 commits) — creation schema (`initialProbs`, sum-to-one *and*
   independent "Set"), √variance creation pools, the general-`p` liquidity invariant at creation, the
   web creation UI, and **the kill-switch** (`CPMM_MULTI_2_CREATION_ENABLED`, default `false`).
5. **Liquidity & answer-management** (6 commits) — lazy v1→v2 conversion, per-answer `addLiquidity`
   (API + UI), the whole-market √variance liquidity-add split, and a lossless **"Other" split** that
   carves a new answer out of the catch-all leaving listed probabilities exactly unchanged.
6. **Review hardening** (15 commits, post-draft) — everything a full adversarial self-review of the
   draft surfaced, each with tests: an always-on post-hoc solve verifier; two flag-off-reachable
   fixes (a per-answer subsidyPool cross-process race → row locks; NaN from `p`-less denormalized
   blob answers → `answerP` accessor); domain guards proven as GP19 (creation feasibility — the
   √variance construction is not total; an add fallback so drizzle can never brick; an Other-split
   floor); a required-`arbVersion` signature so previews can't silently price the wrong mechanism;
   four skip-gated adversarial probe suites + a perf bench; reserved semantics for future per-answer
   NO resolution of linked answers (doc-comment only); a cleanup pass (dedup/dead code/SQL
   constants); and a post-rebase sweep widening two new display gates added by upstream's
   2026-midterms feature.

### Safety & testing

- **Kill-switch off by default** gates both backend creation and the web UI; with it off the diff is
  inert and v1 is byte-identical.
- **`common` jest: 218/218** — incl. the conservation grid (create→trade→limits→sell→add→resolve,
  net = 0 across a covering array), creation tests, GP19 boundary-calibrated guard tests, and the
  existing arb tests — plus 27 skip-gated adversarial probe tests (`PROBE=1`) and a perf bench
  (`BENCH=1`).
- **Always-on post-hoc verification**: every v2 multi-buy solve is checked after the fact (taker
  spend ≡ bet amount, Σ prob = 1, pool positivity, maker pricing, equal basket shares) and fails
  loudly rather than persisting a bad fill.
- **Rebased onto current `main`**; staged so each of the inert-backend / creation / features tips
  compiles clean across `common` + `backend/{api,scheduler,shared}` + `web`.
- **Full lifecycle validated on a local dev instance** (flag flipped locally): create → trade →
  add-liquidity → drizzle → resolve, all three resolution outcomes, conservation **net = 0**;
  v1→v2 lazy conversion validated end-to-end.

### Evidence

Every nontrivial claim is backed by a runnable proof or a differential test:

- **Paper** deriving the three direct-formula algorithms + the general-`p` cost core —
  [paper PDF](https://github.com/evand/manifold-math/blob/main/cpmm-multi-2/paper/auto-arb-algorithms.pdf).
- **Machine-checked proofs GP1–GP18** (sympy + numpy, each standalone) —
  [proofs/](https://github.com/evand/manifold-math/tree/main/cpmm-multi-2/proofs).
- **Reference implementation** that reproduces Manifold's own nested-search solver *and a real captured
  Manifold bet* to ~`1e-13` (GP8) —
  [reference/](https://github.com/evand/manifold-math/tree/main/cpmm-multi-2/reference/manifold) /
  [GP8 test](https://github.com/evand/manifold-math/blob/main/cpmm-multi-2/tests/test_gp8_direct_equivalence.py).
- **Full rationale & evidence bundle** —
  [rationale.md](https://github.com/evand/manifold-math/blob/main/cpmm-multi-2/rationale.md).

### Aside: pre-existing, separable

Things we noticed building this, all pre-existing and separable from this PR:

- Under maker-balance-limited fills, the v1 sum-to-one arb can leave `Σ prob` off by ~2e-3 (the
  solve stops at the maker's balance bound rather than the exact root). v2 inherits this shared
  behavior deliberately — the verifier's tolerance is set to accept what v1 accepts. Tightening it
  is a separate conversation.
- v1 LP accounting: the mana-destruction the lossless add fixes (problem 1 above) is a real v1 bug
  in its own right.
- Orthogonally, v1's LP accounting tracks *net mana* rather than per-provider positions —
  resolution payouts aren't indifferent to a later LP's add, and removal is creator-only & uncapped
  (the creator can pull others' liquidity; non-creators can't pull their own). Not addressed here.

Happy to write either up separately if it's useful — measurements + repro scripts are on hand.

### One judgement call for maintainers (RFC)

The reversible-limit fix means a resting order fills when price *reaches* its limit, rather than on
transient overshoot — a strictly less-surprising contract, but still a **behavior change**, which is
exactly why it rides only in v2 and is never retrofitted to v1 silently. The one call that's yours,
not ours: whether `cpmm-multi-1` ever implicitly promised transient-overshoot fills to existing users.
If so, the v2-only gating is the conservative answer; if not, the retrofit option is on the table.

### DIFF
--- a/backend/api/src/add-liquidity.ts
+++ b/backend/api/src/add-liquidity.ts
@@ -42,11 +42,14 @@ export const addContractLiquidity = async (
       )
     }
 
-    if (
-      contract.mechanism !== 'cpmm-1' &&
-      !isMultiCpmm(contract)
-    )
-      throw new APIError(403, 'Only cpmm-1 and cpmm-multi-1 are supported')
+    // isMultiCpmm covers both cpmm-multi-1 and cpmm-multi-2; the subsidy lands in
+    // subsidyPool + an LP-provision row, and the drizzle job injects it into the
+    // pools (losslessly via the V2 float-p add for cpmm-multi-2; see 2b.6).
+    if (contract.mechanism !== 'cpmm-1' && !isMultiCpmm(contract))
+      throw new APIError(
+        403,
+        'Only cpmm-1 and multiple-choice CPMM markets are supported'
+      )
 
     const { closeTime } = contract
     if (closeTime && Date.now() > closeTime)
--- a/backend/api/src/add-liquidity.ts
+++ b/backend/api/src/add-liquidity.ts
@@ -8,6 +8,7 @@ import { getContract, getUser } from 'shared/utils'
 import { onCreateLiquidityProvision } from './on-update-liquidity-provision'
 import { insertLiquidity } from 'shared/supabase/liquidity'
 import { convertLiquidity } from 'common/supabase/liquidity'
+import { isMultiCpmm } from 'common/contract'
 import { FieldVal } from 'shared/supabase/utils'
 import { updateContract } from 'shared/supabase/contracts'
 
@@ -43,7 +44,7 @@ export const addContractLiquidity = async (
 
     if (
       contract.mechanism !== 'cpmm-1' &&
-      contract.mechanism !== 'cpmm-multi-1'
+      !isMultiCpmm(contract)
     )
       throw new APIError(403, 'Only cpmm-1 and cpmm-multi-1 are supported')
 
--- a/backend/api/src/add-liquidity.ts
+++ b/backend/api/src/add-liquidity.ts
@@ -11,7 +11,11 @@ import { convertLiquidity } from 'common/supabase/liquidity'
 import { CPMM_MULTI_2_CREATION_ENABLED, isMultiCpmm } from 'common/contract'
 import { FieldVal } from 'shared/supabase/utils'
 import { updateContract } from 'shared/supabase/contracts'
-import { getAnswer, updateAnswer } from 'shared/supabase/answers'
+import {
+  getAnswer,
+  getAnswerForUpdate,
+  updateAnswer,
+} from 'shared/supabase/answers'
 
 export const addLiquidity: APIHandler<'market/:contractId/add-liquidity'> =
   onlyUsersWhoCanPerformAction(
@@ -115,8 +119,11 @@ export const addContractLiquidity = async (
 
     if (answerId !== undefined) {
       // Per-answer: the subsidy lands in THAT answer's subsidyPool (drizzleAnswer deepens it
-      // losslessly). updateAnswer takes a concrete value, so read-then-add within this tx.
-      const answer = await getAnswer(tx, answerId)
+      // losslessly). updateAnswer takes a concrete value, so the read-then-add must hold a
+      // row lock — the scheduler's drizzleAnswer (separate process; betsQueue is per-process
+      // only) also read-modify-writes subsidyPool, and either interleaving would create or
+      // destroy subsidy mana.
+      const answer = await getAnswerForUpdate(tx, answerId)
       await updateAnswer(tx, answerId, {
         subsidyPool: (answer?.subsidyPool ?? 0) + subsidyAmount,
       })
--- a/backend/api/src/add-liquidity.ts
+++ b/backend/api/src/add-liquidity.ts
@@ -8,7 +8,7 @@ import { getContract, getUser } from 'shared/utils'
 import { onCreateLiquidityProvision } from './on-update-liquidity-provision'
 import { insertLiquidity } from 'shared/supabase/liquidity'
 import { convertLiquidity } from 'common/supabase/liquidity'
-import { isMultiCpmm } from 'common/contract'
+import { CPMM_MULTI_2_CREATION_ENABLED, isMultiCpmm } from 'common/contract'
 import { FieldVal } from 'shared/supabase/utils'
 import { updateContract } from 'shared/supabase/contracts'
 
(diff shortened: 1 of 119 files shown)


evand:
**Update 2026-07-03** — rebased onto current `main` and pushed the review-hardening batch (28 → 43 commits). What changed since the draft opened:

- **Self-review hardening (15 commits, each with tests)**: an always-on post-hoc solve verifier; two fixes reachable with the flag OFF (per-answer subsidyPool cross-process race → row locks; NaN from `p`-less denormalized-blob answers → `answerP` accessor); GP19-proven domain guards (creation-feasibility 400 — the √variance construction is not total; a liquidity-add fallback so drizzle can never brick; an Other-split floor 403); `arbVersion` now required in the shared pricing signature so previews can't silently price the wrong mechanism; four skip-gated adversarial probe suites + a perf bench; reserved (doc-comment-only) semantics for future per-answer NO resolution of linked answers; and a cleanup pass.
- **PR body refreshed**: honest perf numbers in the O(n log N) bullet (the v2 solve is currently ~7–10× v1 wall-clock on shallow-pool fixtures — the asymptotic win is real but constants are untuned; a perf pass gates the *flag flip*, not this merge), the reserved-semantics section, and an updated stage-by-stage reading guide.
- **Validation at the pushed tip**: `common` jest 218/218; web + api + scheduler typecheck clean; full dev-instance lifecycle sweep (12 scenarios — all resolution modes, v1→v2 conversion + trading through it, sell round-trip, Other-splits, multi-basket redemption) plus negative probes for the new guards (creation 400s on infeasible/sub-floor prob vectors, split-floor 403). Kill-switch remains `false` — the diff is inert and v1 stays byte-identical.
- **Rebase note**: picked up the 2026-midterms feature; the two new `cpmm-multi-1` display gates it introduced are widened in the tip commit.

Evidence repo (proofs GP1–GP19): https://github.com/evand/manifold-math

PR #3940, opened 2026-07-03
fix(create): numeric market blocked when a bucket midpoint is 0; persist numeric fields in drafts

todwadd:
## Bug

User report (AlexanderTheGreater): creating a MULTI_NUMERIC market kept failing with the red ring on **Regenerate numeric ranges** no matter what — changing the range, deleting buckets, and regenerating all had no effect.

Repro: any numeric market whose first bucket is exactly `0` — e.g. *How many goals will Lionel Messi score against Cabo Verde?* (min 0, max 5, buckets `0, 1, 2, 3, 4, 5, Above 5`).

## Root cause

`web/lib/validation/contract-validation.ts` checked generated midpoints with:

```ts
state.midpoints.some((m) => !m || isNaN(m))
```

`!0 === true`, so the legitimate midpoint `0` for the `0` bucket was treated as 'not generated yet' → `errors.answers = 'Please wait for numeric buckets to generate'` → permanent red ring. Regenerating always produces midpoint 0 again for such buckets, so the market could never be created. (The midpoints were actually fine — the expected-value preview computed correctly from them; backend create schema accepts `z.number().safe()` including 0.)

## Fix

- Replace the falsy check with `!Number.isFinite(m)` in both the MULTI_NUMERIC and DATE branches. Allows 0 and negative midpoints, still rejects undefined/NaN/Infinity.
- **Drafts:** `saveDraft` dropped `min`/`max`/`minString`/`maxString`/`unit`/`midpoints`, so a reloaded numeric/date draft lost its range and buckets (the inner zod object strips unknown keys). Added the fields to the draft payload, load path, `MarketDraft` type, and `save-market-draft` schema — all optional, so old drafts load unchanged.

## Testing

- `web`: `tsc --noEmit` clean
- `backend/api`: `tsc -b` (common + shared + api) clean
- Verified against the user's actual saved draft in prod (draft id 3458): answers `0…5, Above 5` with midpoints `[0, 1, 2, 3, 4, 5, 6.5]` — passes new validation, was rejected by old.


todwadd:
Follow-up commit a5f3a68c8 hardens the draft half after a multi-agent review of this branch:

**Fixed:**
- Drafts now persist `shouldAnswersSumToOne` (+ `addAnswersMode`, `probability`, `neverCloses`, `liquidityTier`, poll fields) — a restored thresholds-mode draft can no longer submit as a sum-to-one market under the live session's flag.
- Non-finite min/max/midpoints are sanitized before save and rejected by `.finite()` in the schema — typing `1e999` or duplicating a legacy market with null midpoints no longer 400s the save with an opaque toast (and raw-JSON Infinity can no longer store nulls in jsonb).
- Midpoints are persisted/restored only when finite and aligned with the answers, so a draft saved during the answer-edit debounce window can't create a market with mispaired bucket midpoints.
- `loadDraftFromDb`: functional setState, falls back to session values for fields old drafts lack (no more wiping a warm session or dead-ending pre-existing drafts), clears stale field errors, cancels pending debounced regenerations, discards in-flight ones via a generation counter, and resets/seeds the buckets-vs-thresholds caches so the toggle can't substitute another market's ranges.
- `answers` schema field defaults to `[]` (was optional while the type says required — an answers-less row would have crashed the drafts modal).
- Shared `midpointsPending` helper for the duplicated MULTI_NUMERIC/DATE validation; numeric range checks now reject non-finite (incl. null-after-rehydration) min/max; removed two dead `as any` casts so the compiler checks payload↔schema drift.

**Known, deliberately not addressed here:** PSEUDO_NUMERIC/NUMBER states reached via duplicate links still dead-end at create (FormState lacks initialValue/isLogScale/precision — pre-existing, needs a product call on whether to support or filter those types at intake), and the draft shape is still hand-maintained in type+schema (a z.infer consolidation is a clean follow-up).

Typechecks clean: web `tsc --noEmit`, backend `tsc -b`.

todwadd:
The two-line validation fix has been split out to #3941 for immediate merge (it's the user-facing launch blocker). This PR now effectively carries only the draft-persistence work + review hardening; once #3941 merges, the overlapping validation change here will dedupe cleanly on rebase/merge.

PR #3968, opened 2026-07-30
Mod-ify Quroe and realTomBayes

ManiFed:
I will send 12.5k to whoever merges this PR


ManiFed:
I like the look of this PR

PR #4009, opened 2026-08-17
fix(perps): 5s BTC oracle tick, xStocks feeds, drop sunset uk-grid-carbon

todwadd:
Combines three things: the 5s BTC oracle tick, the xStocks feeds from #3983, and removal of the sunset `uk-grid-carbon` feed. **Supersedes #3983** — both of its commits are in here unchanged.

## Why

The 10 bps taker fee (#3979) killed the continuous 4-second scalp, but it only set a *strike* on a free option rather than removing it. Three accounts have since taken **M$34.4k net** out of the BTC perp pools by exercising selectively: probe with dust, watch Coinbase/Kraken/Bitstamp directly, and when the true price has already diverged past the fee, open at maximum size against a mark we know is wrong.

Measured on prod since the fee shipped (8/9 →), by entry-size quartile:

| quartile | n | avg notional | **favorable on the very next tick** | avg next-tick capture |
|---|---|---|---|---|
| Q1 | 25 | M$7k | 60% | +0.15 bps |
| Q2 | 25 | M$83k | 64% | +5.00 bps |
| Q3 | 25 | M$180k | 56% | +2.13 bps |
| **Q4 (≥M$480k)** | 25 | **M$960k** | **84%** | **+6.78 bps** |

Their large entries fire a median **8.8s** after the last tick, never more than 14.5s — i.e. always inside one cron interval. Nobody forecasts BTC at 84% over a ten-second horizon; that is the stale mark, not a view. Realized gross is 18–25 bps per round trip against the 10 bps fee.

The exploitable quantity is the **poll interval**, because trades execute at the cached mark with no spread and no price impact ([`engine.ts:578`](https://github.com/manifoldmarkets/manifold/blob/main/backend/shared/src/perps/engine.ts#L578)), so size is limited only by capacity — they pushed M$2M of notional through a M$613k pool at a single price.

Frequency of windows diverging past the fee, fitted over 34,575 feed points:

| window | avg abs move | P(move > 10 bps) |
|---|---|---|
| 21s | 0.78 bps | 0.087% |
| 41s | 1.29 bps | 0.32% |
| 62s | 1.70 bps | 0.74% |
| 83s | 2.04 bps | 1.29% |
| 165s | 3.05 bps | 3.98% |
| 330s | 4.41 bps | 10.2% |

That fits **P ∝ T^1.95**. So 15s → 5s cuts opportunities ~**8.7×**, and shrinks the excess-over-fee on survivors by a further ~1.9× (magnitude ∝ T^0.63).

**Roughly a 3× fee increase in deterrence, charged entirely to the latency bot.** Honest holders pay the fee whatever the tick rate and gain nothing from raising it. That asymmetry is the whole argument for preferring this over a bigger fee.

> Caveat: the exponent is fitted over 20–330s and extrapolated below the shortest window we can observe. Treat 8.7× as the right order of magnitude, not a precise figure.

## What changed

**Per-feed poll rate.** The cron is global but the right rate is per-feed, so this adds `pollPeriodMs` and throttles inside the job rather than just dropping the cron — a bare `*/15` → `*/5` would also have tripled our call volume against NESO. `updatePeriodMs` is left alone as what it's documented to be (interval between genuinely new values, feeding the frozen funding period); `max(1h, ·)` is 1h at 5s or 15s, so **no contract's funding cadence moves**.

| feed | cadence | poll |
|---|---|---|
| `btc-usd` | fast | **5s** |
| `spyx-usd` / `qqqx-usd` / `gldx-usd` / `nvdax-usd` | fast | 15s |
| `trump-approval-rating`, `openrouter-open-weight-share` | daily | own jobs |
| ~~`uk-grid-carbon`~~ | — | removed |

**xStocks stay at 15s.** The throttle is fail-open, so merging those four in unset would have silently moved them to 5s. A round costs 4 Jupiter calls (one per token), 4 Gate, 2 MEXC; at 5s that's 48/min against Jupiter's keyless lite tier, close enough to its limit that a 429 is likely — and a 429 drops QQQx or GLDx below `getConsensusMedian`'s two-source minimum, skipping the tick. A stalled feed pauses its market, which is far worse than a 15s-old mark. There's also no measured edge to price against yet: those markets don't exist until after this deploys. To move them later, batch the Jupiter leg first — its price v3 endpoint takes comma-separated ids, so all four collapse into one call.

**`uk-grid-carbon` removed** (market sunset 2026-08-10). Removed: registry entry, launch-manifest market, feed-id constant, adapter, backfill script. `getPerpLaunchManifestErrors()` returns clean.

**Deliberately kept:** the `'uk-grid-carbon'` entry in `common/perps/oracle-attribution.ts`. The resolved market's page still charts the NESO history we ingested, so the credit is still owed — dropping it with the feed would be exactly the silent licence breach that map exists to prevent. Historical `oracle_prices` rows are untouched.

## Worth a close look in review

- **`isPollDue` has no test.** There's no jest harness in `backend/scheduler` and no existing test covers the feed registry at all; putting a scheduler throttle in `common/` to borrow its setup would break the package layering. It's 6 lines of pure arithmetic and fails *open* (bad config → poll more, not less), and a frozen feed is still caught by `update-perps` hourly + the daily probe + `[oracle-feeds]` staleness alerting. Verified by simulation instead — over 10 min of jittered firings: btc 5.00s avg, uk 15.00s, unset 5.00s.
- **The 0.8 jitter slack** in `isPollDue`. Without it, a feed whose period equals the tick interval fails the comparison on ~every other firing and silently runs at half rate. Check the arithmetic.
- **Lock contention.** The tick now takes the per-contract advisory lock 3× more often. `PERP_TX_MAX_ATTEMPTS = 8` was sized against the 15s tick; I updated the comment but did *not* change the constant speculatively. Worth watching 40001 rates after deploy.
- **`FETCH_TIMEOUT_MS` is still 5s** in `btc-price.ts`, now equal to the tick interval. Croner's `protect` makes this degrade gracefully (effective rate drifts toward 10s during a slow fetch) rather than queueing, so it's safe — but it's the obvious next knob.
- **Row growth:** `oracle_prices` writes go from ~4.4k/day to ~17k/day for BTC. The `before` cursor from #4004 covers the chart read path.

## Not included

`maxOraclePriceAgeMs` on the BTC market stays at **120s** — 24 missed ticks of free option. It's floored at the feed's `staleAfterMs` by `create-perp`, so tightening it means also moving the alerting threshold. Separate change.

## Deploy

**API and scheduler from the same SHA.** No migration. An earlier revision of this description said "`scheduler-perps` only" — wrong once the xStocks feeds joined this branch: the API compiles `backend/shared` in, and `requireOracleFeedForPerpCreation` rejects any feed missing from its own compiled registry, so a scheduler-only deploy would leave the API unable to create the four new markets. After deploy: xStocks backfills → create markets (per #3983's plan).

**Also:** the `[oracle-feeds]` GCP alert policy carries `AND NOT jsonPayload.message:"uk-grid-carbon"`. That matches on substring across the whole message, so it's been silencing real errors from *every* feed. Drop that clause once this ships — it's in the untracked `alert-rewrite-oracle-feeds.json` / `oracle-feeds-policy-update.json` drafts.

## Testing

- `tsc -b` clean on `backend/api` and `backend/scheduler`
- 472/472 common tests pass, including #3983's `xstocks.test.ts`. (Baseline before the review fixes is **469**; they add 3 unit cases. An earlier revision of this description quoted that baseline as 472 — it was measured in a working tree carrying unrelated uncommitted `oracle.test.ts` changes.)
- `getPerpLaunchManifestErrors()` clean against the new registry (run against compiled output — the validator dereferences `getOracleFeed(feedId)`, so a dangling manifest entry throws at runtime rather than failing typecheck)


todwadd:
All three findings confirmed and fixed in `2b6effe93`. Thanks — finding 1 in particular was a real hazard, and the module header had already written down the exact failure mode the code then went and implemented.

## 1. [P1] Wrong price unit — fixed

Confirmed, and worse on the two-source path than "accepted by the tolerance": `getConsensusMedian` **averages** when there are exactly two quotes (`sorted.length % 2 === 0`), so on QQQx and GLDx a mismatched Jupiter unit doesn't just pass the gate, it lands at half weight directly in the executable mark. On the three-source feeds the median usually rejects it — but only until one CEX drops out, at which point they degrade to the same two-source path.

Fix as you specified, with one addition:

- `XStockUnitMode = 'rebasing' | 'static'` declared per token on `XStockSpec`, following the dividend: SPYx/QQQx/NVDAx rebasing, GLDx static.
- `readJupiterRawUsdPrice` takes it as a **required** third argument, not a defaulted one — a new call site that hasn't thought about the unit should fail to compile, since getting it wrong is invisible at runtime.
- `rebasing` + missing/unusable `usdPricePrescaled` → `NaN`, which drops Jupiter from the quote set and lets the existing consensus gate decide: three-source feeds carry on, two-source feeds skip the tick and alert. Only GLDx may fall back.
- **Addition:** both modes now prefer `usdPricePrescaled` whenever it's usable. It's the raw-unit price by definition and equals `usdPrice` when the multiplier is 1, so if a `static` token ever starts rebasing this reads the right number rather than drifting. Fails safe in the direction we didn't anticipate.

The test that blessed the fallback is replaced by four: NaN on unusable prescaled, NaN on absent `scaledUiConfig` (your point about the documented schema not guaranteeing it), static falls back, and static prefers prescaled when present.

## 2. [P2] Deploy — fixed

You're right, and the error was in my PR description, not the runbook — step 2 there already says "same audited commit". Description corrected to **API and scheduler from the same SHA**, with the reason (`requireOracleFeedForPerpCreation` rejecting feeds absent from the API's compiled registry).

Docs fixed:
- `perps-launch-runbook.md:110` — UK-carbon backfill replaced with xStocks, noting the script was removed with the sunset market.
- `perps-launch-runbook.md:5` and `:151` — "four" → seven, plus a pointer to `getPerpLaunchManifestErrors()` as the executable count, since that number has now drifted twice.
- `backend/shared/src/perps/README.md:208` — `uk-carbon` dropped from the backfill-script list.

## 3. [P2] Slow xStocks lengthening the BTC tick — fixed

Correct, and it invalidated a claim in my own header comment: I'd written that `protect`-induced degradation toward 10s was "the right failure mode," which was true when BTC was the only fast feed and false the moment xStocks shared the job.

Feeds are now **dispatched, not awaited**, each behind a per-feed in-flight guard — a strictly finer-grained `protect` that still stops a feed stacking on itself without coupling feeds to each other. Simulated over 5 minutes of 5s firings with all four xStocks feeds hanging to their full 5s timeout:

```
btc-usd     starts= 61  avg gap=5.00s  max gap=5.00s
spyx-usd    starts= 21  avg gap=15.00s max gap=15.00s
qqqx-usd    starts= 21  avg gap=15.00s max gap=15.00s
gldx-usd    starts= 21  avg gap=15.00s max gap=15.00s
nvdax-usd   starts= 21  avg gap=15.00s max gap=15.00s

BTC firings later than 5s: 0 / 60  -> BTC fully decoupled
```

Two consequences worth naming: croner's `protect` no longer fires for this job (so its "Still alive" WARN, which a 5s cron would have made noisy, goes away), and feed errors no longer reach the job-level `catch` — they were already caught inside `tickOneFeed`, and there's now a `.catch` backstop so a future edit can't turn one into an unhandled rejection.

---

## Follow-up this surfaced (not fixed here)

Chasing finding 3 through `createJob` turned up a cost I hadn't accounted for: **every firing does three Supabase REST round-trips** on the deprecated client — read `scheduler_info`, upsert `last_start_time`, upsert `last_end_time` — before and after `fn`. At 5s that's ~36/min, **~51.8k/day** for this job alone, up from ~17.3k at 15s.

It doesn't undermine the change (my fix removes the xStocks coupling regardless), but it leaves a latency floor I can't clear from inside this job: if that bookkeeping ever exceeds 5s, `protect` skips BTC firings for reasons unrelated to any feed. Fixing it means touching `createJob` for every scheduler job, which is well outside this PR. Worth its own issue — happy to open one.

## Test counts

469 → 472 (three new unit cases). Separately: the "472 baseline" in my original description was wrong — it was measured in a working tree carrying unrelated uncommitted `oracle.test.ts` changes. The true pre-fix baseline was 469. Description corrected.

todwadd:
All three addressed in `6ece48970`. The on-chain mint query was the valuable part of this pass — it falsified a claim I wrote without checking.

## [P2] GLDx `static` — fixed, and the premise was wrong

You're right, and the failure was mine at the source: the comments asserted GLDx *lacks* the scaled-ui extension. It doesn't. Your `getMultipleAccounts` result — all four mints carrying a `scaledUiAmountConfig` with a live update authority, GLDx merely sitting at 1 — is now recorded in the code with the per-mint multipliers, replacing the claim I'd inferred from "gold pays no dividend."

I did not simply flip GLDx to `rebasing`, for the reason you identified: Jupiter omits the field at multiplier 1, so it would fail closed every tick and take a two-source feed permanently dark. Instead `static` is defended rather than trusted:

1. Both modes already prefer `usdPricePrescaled` — a no-op at multiplier 1, automatically correct the moment one appears.
2. **New:** a `static` token whose response reports `multiplier` *or* `newMultiplier` off 1 is a provably stale declaration and returns `NaN`. `newMultiplier` is included because it's the scheduled next value — that's the earliest in-band warning available.

`isRebasingMultiplier` treats absent/unparseable as *not* evidence of rebasing, since that's the ordinary shape of a static token's response; only a present, numeric, meaningfully-non-1 value trips it (ε = 1e-6, well under NVDAx's 1.000103).

The residual hole you named — rebasing begins while Jupiter reports no scaled metadata at all — is undetectable in-band, so it gets an explicit out-of-band multiplier watch in the runbook, including the warning not to flip `unitMode` until Jupiter is confirmed to populate the prescaled field.

Three tests added: multiplier off 1 → NaN, `newMultiplier` off 1 → NaN, exactly 1 → still accepted.

## [P2] Dispatcher heartbeat / stuck feed — fixed

Both halves confirmed. Added a per-feed in-flight **age** watchdog (`inFlightSince`, 2min grace, alert throttled to 5min) that `log.error`s a feed blocking its own next run. `update-perps` would catch this hourly for feeds with a live market; this reports in minutes and covers feeds that have none yet — which is every xStocks feed until after this deploys.

On `scheduler_info.last_end_time`: I documented the changed semantics at the dispatcher rather than altering the column. It is now a dispatcher heartbeat, but `perp-launch-preflight`'s use of it still holds — it's checking that the job is *firing*, which remains exactly what the timestamp proves. The thing it never proved and now visibly doesn't is feed completion, which the watchdog covers directly.

## [P3] Poll slack — fixed, and your framing was the right one

Replaced 20%-of-period with **half a tick**, tied to the cron quantum rather than the request. Your `ceil(0.8 × period / 5s) × 5s` derivation is what convinced me the old constant was rescaling the request rather than absorbing jitter.

```
period ->  old(0.8*p)   new(half-tick)
  5000 ->      5000          5000
 15000 ->     15000         15000
 60000 ->     50000         60000   <- the daily probe is honest again
  6300 ->     10000          5000   <- nearest multiple
```

Half a tick also has a concrete justification the old value lacked: `createJob` performs two awaited Supabase round-trips *before* calling the job, so the stamp jitter is network-latency-sized, not croner-sized. 2.5s absorbs that; a tight absolute tolerance would not have.

On "the API accepts nonmultiples" — I went with **validate**, not just quantize. `ORACLE_TICK_PERIOD_MS` is now a single exported constant that the cron expression is built from (so schedule and throttle can't drift), and `validateOracleFeedPollPeriods()` runs at boot and `log.error`s any `pollPeriodMs` that isn't a whole number of ticks, naming the rate it will actually run at. Your 12000 case is exactly why: at 2.4 ticks it sits near the rounding boundary and jitter flips it between 10s and 15s, so it should be loud rather than merely rounded.

## Doc sweep

Fixed all six you listed. I also swept rather than only patching cited lines this time — that was the process failure behind this finding.

## One thing you didn't flag that I changed anyway

`backend/shared/src/perps/README.md` asserted **"faster ticks make this WORSE, since harvestable total variation grows as 1/√period."** That's correct with *no* fee — every tick is worth exercising, so harvest is `(W/T)·|move|(T)` and shortening `T` helps the bot. It is the exact opposite of this PR, and a reader would reasonably cite it as an argument against the change.

Both regimes are now written out, because the fee is what inverts the sign: once only moves exceeding `F` are worth taking, the frequency of those falls as ~T^1.95 and shortening the period helps *us*. With the explicit warning that removing the fee while keeping the fast tick is the worst of both — which is the combination the original sentence was warning about.

## Still open, unchanged

`createJob`'s three Supabase REST round-trips per firing (~51.8k/day at 5s). Not fixed here — it touches every scheduler job. Noting that this pass gave it a second reason to matter: those round-trips are the dominant term in the stamp jitter the poll throttle now has to absorb.

Tests: 472 → 475. Typecheck clean on `backend/api` and `backend/scheduler`; `git diff --check` clean.

PR #4016, opened 2026-08-21
Rewrite outdated FAQ entry claiming that liquidity providers cannot withdraw subsidy.

orborde:
As of this writing, I can withdraw subsidy through the Manifold web UI!


todwadd:
> As of this writing, I can withdraw subsidy through the Manifold web UI!

Thank you for checking this! As of right now, you can only withdraw liquidity from your own markets, so this is still not exactly right. I will fix it. Thank you!

PR #4019, opened 2026-08-25
perf(scheduler): stop calculate-conversion-scores locking contract rows for minutes

todwadd:
## Problem

`calculate-conversion-scores` (main scheduler, hourly at :46) ran

```sql
update contracts set conversion_score = (<eight count(distinct) CTEs>) where id = any($1)
```

in 100-contract chunks. Postgres takes each row lock as the UPDATE reaches it and holds it until commit, so every contract in the chunk is locked for the whole aggregation. Measured on prod: one chunk = 8–160 s (2026-08-21: 121 s; 2026-08-25: 00:47:11→00:49:35 = 144 s, 01:47:48→01:49:41 = 113 s).

Harmless for a market whose row only changes when someone bets. Not harmless for a perp: the oracle tick rewrites the BTC perp's row every 2 s with a 1 s `lock_timeout` (`FAST_TICK_ORACLE_BOUNDS`), so every tick inside the chunk statement dies with 55P03 and the executable mark freezes while the underlying keeps moving.

Over 24 h on prod: 667 oracle-tick apply failures, 575 (86%) inside the :46–:48 window this job occupies. Concretely, a 40–87 s freeze of the BTC mark once an hour, from a discovery-ranking job — and the `PROD perps: [oracle-feeds] feed error` alert pages on every one of them (last two bursts today: 00:48:39–00:49:01 and 01:49:15–01:49:41, each ending within 600 ms of the chunk commit).

## Fix

Compute the scores with a SELECT, then write them with a keyed UPDATE (`unnest` VALUES join). The row lock now lasts a keyed update instead of an aggregation.

- Same arithmetic. The old form's `from contracts c2 … where c2.id = c.id` made the c2-keyed joins identical to the c-keyed ones; the SELECT is that expression with a single `contracts c` source. Verified against the old correlated-subquery form on prod data — 0 mismatches.
- Rows whose score comes back non-finite are skipped rather than written, so a future edit that breaks the expression leaves the previous score in place instead of blanking a column the feed ranks on. The expression coalesces every input, so today this never triggers.
- Log line for the compute failure renamed (`Error on compute conversion scores`) so it's distinguishable from the write.

One file: `backend/shared/src/conversion-score.ts`. No schema change, no perps-engine change.

## Alert-volume estimate

A 7-day replay of prod `[oracle-feeds]` logs through the policy's 2h notification rate limit puts this at **32 emails/week -> 8**, i.e. ~75% fewer. 98.5% of the 926 error lines fall inside an actual `calculate-conversion-scores` run (interval-joined against 168 runs; median 126s, p90 361s, max 533s). Treat it as an estimate from log replay, not a production measurement -- worth re-checking once this is deployed.

## Review round 1 (`0c2741e76`)

- **Lock order (P2):** the write is one short transaction — `select id … order by id for update skip locked`, then the keyed update over exactly those rows. Same ordering as `score-contracts`' pre-lock; never waits, so no deadlock and no prolonging a held lock (perp included). Skipped rows keep last hour's score, are rescored next run, and the count is logged.
- **Numeric precision (P3):** `conversion_score` is unconstrained `numeric` and the default parser is `parseFloat`; 143,302 stored scores carry >17 decimals (the expression yields 23). Score is now selected as `::text`, checked with `Number.isFinite(Number(s))`, and the original string goes to `numeric[]`.

## Review round 2 (`27e5eab39`)

- **Skipped ids were discarded.** The candidate query only selects markets viewed in the last hour, so a market skipped while locked and then never viewed again was never reconsidered -- its score froze indefinitely, contrary to what this description previously claimed. Skipped ids (and chunks whose computation failed) now carry between runs and are unioned into the next run's candidates ahead of the viewed ids, independently of that window. Capped at 10k, newest kept, logged on both ends; dropped on process restart, after which a market is rescored next time it is viewed.
- **Unchanged:** SELECT-then-write, short id-ordered `for update skip locked` transaction that never waits, and `::text` -> `numeric[]` precision passthrough.
- **Regression test** (`backend/shared/src/conversion-score.test.ts`): a locked market with **no subsequent views** keeps its score while locked, then updates once the lock clears -- plus stays-pending-while-locked, no duplicate when pending and freshly viewed, exact-string passthrough, and retry after failed computation. Verified to fail 3/5 against the previous behaviour. Full `backend/shared` suite green (9 suites, 56 tests).

## Deploy

Main scheduler only (`prod main`). Keep it out of the perps deploy — it does not touch the perp engine or `scheduler-perps`.

## Verify after deploy

The `[oracle-feeds] bitcoin-price-usd: failed to apply … lock timeout` burst at :47–:49 should stop. Residual ~4 apply failures/hr from ordinary trader contention stay under the 60 s log threshold and do not page.

```
gcloud logging read 'resource.type="gce_instance" AND severity>=ERROR AND jsonPayload.message:"[oracle-feeds] bitcoin-price-usd: failed to apply"' --project=mantic-markets --freshness=6h --format="value(timestamp)" | cut -c1-16 | sort | uniq -c
```

Related: #4017 (oracle tick latch) is the other half of the tick-reliability work; independent of this.

### DIFF
--- a/backend/shared/src/conversion-score.test.ts
+++ b/backend/shared/src/conversion-score.test.ts
@@ -1,0 +1,136 @@
+import {
+  ConversionScoreDb,
+  ConversionScoreTx,
+  rescoreConversionScores,
+} from './conversion-score'
+
+jest.mock('shared/supabase/init', () => ({
+  createSupabaseDirectClient: () => {
+    throw new Error('the tests drive the client explicitly')
+  },
+}))
+jest.mock('shared/utils', () => ({ log: () => undefined }))
+
+type Write = { ids: string[]; scores: string[] }
+
+// A database in which `for update skip locked` returns only the unlocked ids,
+// which is the whole behaviour under test. Scores are kept as the strings
+// postgres would return for an unconstrained numeric.
+const fakeDb = (opts: {
+  viewedIds: string[]
+  scores: Record<string, string>
+  lockedIds?: string[]
+}) => {
+  const locked = new Set(opts.lockedIds ?? [])
+  const writes: Write[] = []
+  const db: ConversionScoreDb = {
+    map: async <T>(
+      _query: string,
+      _values: unknown[],
+      cb: (row: Record<string, any>) => T
+    ) => opts.viewedIds.map((id) => cb({ contract_id: id })),
+    manyOrNone: async <T>(_query: string, values: unknown[]) =>
+      (values[0] as string[])
+        .filter((id) => id in opts.scores)
+        .map((id) => ({ id, score: opts.scores[id] })) as T[],
+    tx: async <T>(cb: (tx: ConversionScoreTx) => Promise<T>) => {
+      const tx: ConversionScoreTx = {
+        map: async <U>(
+          _query: string,
+          values: unknown[],
+          rowCb: (row: Record<string, any>) => U
+        ) =>
+          (values[0] as string[])
+            .filter((id) => !locked.has(id))
+            .map((id) => rowCb({ id })),
+        none: async (_query: string, values: unknown[]) => {
+          writes.push({
+            ids: values[0] as string[],
+            scores: values[1] as string[],
+          })
+          return null
+        },
+      }
+      return cb(tx)
+    },
+  }
+  return { db, writes }
+}
+
+const writtenIds = (writes: Write[]) => writes.flatMap((w) => w.ids)
+
+describe('rescoreConversionScores', () => {
+  it('retries a locked contract that is never viewed again', async () => {
+    // The regression: the candidate query only sees contracts viewed in the
+    // last hour, so a contract skipped while locked and then never viewed
+    // again would silently keep a stale score forever.
+    const scores = { locked1: '0.5632098085888197', free1: '0.25' }
+
+    const run1 = fakeDb({
+      viewedIds: ['locked1', 'free1'],
+      scores,
+      lockedIds: ['locked1'],
+    })
+    const pending = await rescoreConversionScores(run1.db)
+
+    // While locked it keeps its existing score: no write mentions it.
+    expect(writtenIds(run1.writes)).toEqual(['free1'])
+    expect(pending).toEqual(['locked1'])
+
+    // Second run: nobody has viewed it since, so the activity window is empty
+    // and only the carry-forward can bring it back. The lock has cleared.
+    const run2 = fakeDb({ viewedIds: [], scores })
+    const stillPending = await rescoreConversionScores(run2.db, pending)
+
+    expect(writtenIds(run2.writes)).toEqual(['locked1'])
+    expect(stillPending).toEqual([])
+  })
+
+  it('keeps carrying a contract that stays locked', async () => {
+    const scores = { locked1: '0.25' }
+    let pending = ['locked1']
+    for (let run = 0; run < 3; run++) {
+      const { db, writes } = fakeDb({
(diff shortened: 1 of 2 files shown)


todwadd:
Both addressed in `0c2741e76`.

**P2 — lock order.** The write now runs in one short transaction: `select id from contracts where id = any($1) order by id for update skip locked`, then the keyed `unnest` update over exactly the rows it holds. Same id ordering as `score-contracts`' pre-lock, and `skip locked` means it never waits — so it can neither deadlock (no wait, no cycle) nor prolong a lock it already holds while a bet/tick finishes on another row, which is the perp case that motivated the PR. A skipped contract keeps last hour's score and is rescored next run; the skip count is logged (`Skipped N contracts whose rows were locked…`). I went with skip rather than retry because the only rows this ever skips are ones mid-write by something more important than a discovery score.

**P3 — numeric precision.** Confirmed on prod: the expression yields 23 decimals and 143,302 stored `conversion_score` values carry >17, so `parseFloat` was lossy. The score is now selected as `::text`, validated with `Number.isFinite(Number(s))` only, and the original string is passed to `$2::numeric[]`.

Typecheck clean (`tsc -b` on backend/shared), prettier clean.

todwadd:
Good catch, and the PR description was wrong rather than just imprecise — thanks. Fixed in `27e5eab39`.

The candidate query is `select distinct contract_id from user_view_events where created_time > now() - interval '1 hour'`, so "rescored on the next run" only held for a market that happened to be viewed again. A locked market with no further views was dropped and its score froze indefinitely.

**Carry-forward.** `rescoreConversionScores` now returns the ids it did not write, and the scheduler entry point holds them between runs and unions them into the next run's candidates — ahead of the viewed ids, since a contract waiting from an earlier run is the one at risk. Independent of the rolling activity window, so an unviewed market is retried on the next firing.

A chunk whose *computation* failed is carried the same way. It is the same defect class — an intended rescore silently dropped — and previously it was also left to the activity window.

Bounded at 10k ids, newest kept, logged on both ends. A process restart drops the list, after which a market is rescored the next time it is viewed; that is the exposure the job already had, and this is a discovery ranking rather than money. Happy to move it to a table if you would rather it survive restarts, though that needs a migration.

**Preserved as asked:** aggregation still a SELECT outside any transaction; write still a short id-ordered `for update skip locked` transaction that never waits; score still travels as `::text` into `numeric[]`, unrounded.

**Structure.** The core is `rescoreConversionScores(pg, carriedOverIds)` returning the still-pending ids, with module state in `calculateConversionScore`. The job runner invokes jobs as `fn(ctx: JobContext)`, so the exported entry point could not take the client as a parameter.

**Regression check** — `backend/shared/src/conversion-score.test.ts`, driving the core with a fake client over the pg surface this job uses, since there is no DB test harness in the repo. The reviewer's case is the first test: run 1 has the market viewed and locked, asserts no write mentions it and it comes back pending; run 2 has an **empty** view window with the lock cleared, asserts it is written. Plus: stays pending while it stays locked, no duplicate when pending and freshly viewed, exact-string score passthrough, and retry after a failed computation.

I checked the test actually catches the regression rather than passing vacuously — reverting the carry-forward fails 3 of the 5. Full `backend/shared` suite green (9 suites, 56 tests), `tsc -b` clean, prettier clean.

On the numbers: agreed, and I have restated the ~75% as an estimate in the description. It is a 7-day log replay through the policy's 2h rate limit (32 emails to 8), not a production measurement — worth re-checking after this is deployed.

PR #4020, opened 2026-08-26
Native app improvements (Play advisories + camera + sign-in loop)

todwadd:
Rolling branch for native-app items to bundle into the next release.

Items 1 and 2 are the two Google Play advisories on release 71 (both advisory-only, but cheap to clear). Items 3 and 4 are genuine user-facing bugs reported in Play reviews — identity verification is currently unusable on Android, and some users can't get past the login screen at all.

## 1. "Remove resizability and orientation restrictions" (large screens)

- The `android:screenOrientation="portrait"` lock is removed from the manifest. Android 16+ already ignores it on >=600dp displays for apps targeting API 36 (i.e. it is dead on tablets/foldables in prod today), Android 17 removes the last opt-out, and Play penalizes it in large-screen ranking/listing.
- Phones stay portrait: `MainActivity.onCreate` now sets `requestedOrientation = SCREEN_ORIENTATION_PORTRAIT` - the exact same `ActivityInfo` constant the manifest attribute compiled to. The system ignores that call on large screens too, so tablets/foldables become properly rotatable while phone behavior is unchanged.
- The lock lives in native code (not JS) deliberately: the streak widget boots the JS runtime headlessly every 30 min, so a JS-side lock can be silently skipped when the app later opens into a warm process. `onCreate` runs on every activity creation.

## 2. "Uses deprecated APIs or parameters for edge-to-edge"

- Removed the deprecated `android:statusBarColor` from `AppTheme` (a no-op on every supported API level: Android 15+ disables it, and below that RN's `enableEdgeToEdge` sets the bar transparent before first frame).
- The residual warning will **not** fully clear: Play statically scans the DEX and RN 0.81 core (`StatusBarModule`) still references `Window.setStatusBarColor` (facebook/react-native#48256). That part is noise until an RN upgrade; there is nothing further actionable on our side.

## 3. Identity verification can't use the camera on Android (Play review)

A Play reviewer wrote: *"I can't verify my identity. When I try to verify myself and even try to use the camera and my ID card, I can't take a photo because the app doesn't even ask for camera permission."* That report is accurate, and the cause is a one-line omission.

**The chain:**

1. `IdentityVerificationPage` sets `window.location.href` to the `create-idenfy-session` `redirectUrl`, so verification navigates the webview to iDenfy's hosted page. There's no `onShouldStartLoadWithRequest`, so this loads **inside** the webview, not an external browser.
2. iDenfy captures the ID document and selfie via `navigator.mediaDevices.getUserMedia()`.
3. react-native-webview handles that in `RNCWebChromeClient.onPermissionRequest`: `RESOURCE_VIDEO_CAPTURE` -> `Manifest.permission.CAMERA` -> `checkSelfPermission` -> `requestPermissions`.
4. **The app never declared `CAMERA`** (confirmed absent from both the checked-in manifest and the merged manifest of a real build). Android therefore denies it *immediately, with no dialog*, the webview calls `request.deny()`, and `getUserMedia` rejects.

So the user sees a dead camera and is never prompted — precisely the review. iOS was never affected: `NSCameraUsageDescription` is declared and react-native-webview defaults `mediaCapturePermissionGrantStatus` to `Prompt`, which is why this only ever surfaced on Play.

Worth noting the reviewer's mental model was right but the older path wasn't broken: the legacy GIDX `upload-document` flow is a plain `<input type="file">` that hands off to the camera app by intent, which needs no permission. Only the newer iDenfy `getUserMedia` flow hits this.

**The fix** is a third config plugin, `plugins/withAndroidCameraForKyc`, which adds the permission plus `<uses-feature android:name="android.hardware.camera" android:required="false"/>` (and `.autofocus`). The `required="false"` matters: declaring the CAMERA permission otherwise makes Play *imply* the hardware feature is required and drop the listing on camera-less devices.

It's a plugin rather than `android.permissions` in `app.config.js` + a hand-edit for the `<uses-feature>` tags for the reason in the next section — one place, and it survives `prebuild --clean` too. (`@expo/config-plugins`' own permission mod is purely additive, so nothing else in the manifest is disturbed.)

I have **not** verified this on-device against a live iDenfy session. The source chain is unambiguous, but I can't rule out a second issue downstream of the permission grant — see QA below.

## 4. Signing in loops back to the login page (Play review)

Another reviewer: *"You app just keep looping at the login page when I sign in it just takes me back to the login page your not given me options here."*

The native app and the web client keep **separate Firebase sessions**, joined only by a `nativeFbUser` postMessage. When that handoff doesn't land, the two halves push each other in a circle with no exit:

```
native has no user -> AuthPage (the only screen there is; no guest browsing)
  -> Google sign-in succeeds -> web is told once, no retry, no ack
  -> web still thinks it's logged out -> renders its own login button
  -> tap -> 'loginClicked' -> native signs out -> AuthPage
```

It works for most people because the WebView normally has its own persisted Firebase user, so the handoff is never load-bearing. Four fixes, none of which touch the happy path:

- **Retry the post-login handoff.** `sendWebviewAuthInfo` already re-sent on a 100/500/1000/3000ms backoff — but only for sessions restored from storage. A *fresh* sign-in got one un-retried post, fired from `AuthPage`, a component that the sign-in itself unmounts. Now driven by a `[fbUser?.uid]` effect in `App` so both paths get the retries. Re-sends are idempotent: the web side no-ops when the email already matches.
- **Stop the web client's cold start from wiping the native session.** Firebase fires `onIdTokenChanged` with null once on init whenever nothing is persisted — in the WebView that usually just means "the handoff hasn't arrived yet". `auth-context` forwarded it as a real sign-out, which made native erase its stored credential and bounce to the login screen. Now gated on having actually observed a signed-in user. Deliberate logouts are unaffected: `firebaseLogout()` calls `nativeSignOut()` itself, so that path never relied on the observer.
- **Surface sign-in failures.** `signInWithCredential` had no `.catch` and `response.type === 'error'` was ignored outright, so a banned account (`auth/user-disabled`), a network blip, or a credential conflict became an unhandled rejection that silently redisplayed the login screen. Now an `Alert` plus `Sentry.captureException`, with specific copy for the three common codes. Apple cancels are excluded.
- **Keep the native flags across logout's `localStorage.clear()`.** Losing `IS_NATIVE_KEY` makes `postMessageToNative` a silent no-op and sends `firebaseLogin` down the `signInWithPopup` path, which Google refuses to serve inside an Android WebView — a login button that does nothing at all.

Plus two freebies: a Sentry breadcrumb when the web client asks to log in while native still holds a session (that is the loop, made observable), and `onAuthStateChanged` moved out of the render body, where it registered a new never-unsubscribed listener on every render.

**Note on diagnosis:** at least three distinct root causes produce this exact symptom, and all of them were invisible — `log()` in `components/logger.tsx` compiles to `if (NATIVE_BUILD === 'PROD') return`, so every native auth error path is a no-op in shipped builds, and Sentry was initialized but never wired to auth. The instrumentation here is worth landing on its own merits: it's what will tell us whether this is one reviewer or a percentage of installs.

Pure JS — no `android/` artifacts, so the prebuild caveat below doesn't apply to this item.

## The important discovery: hand-edits to `android/` don't ship

The build scripts run a non-clean `expo prebuild` before every `eas build`, and prebuild **re-injects** `android:statusBarColor` (copied from `splash.backgroundColor` by @expo/prebuild-config) and rewrites `screenOrientation` from `app.config.js`. So fixes 1 and 2 are enforced by a new config plugin, `plugins/withAndroidPlayAdvisoryFixes`, which strips the two attributes at prebuild time (the statusBarColor strip must run in the `finalized` mod - a regular styles mod loses the ordering race to the splash plugin; verified empirically). The checked-in `android/` files now match prebuild output exactly — re-verified after the camera change by running `expo prebuild -p android`, which reproduces the manifest and touches nothing else. The camera fix (3) is enforced the same way, via `plugins/withAndroidCameraForKyc`.

`orientation: 'portrait'` stays in `app.config.js` so the iOS side of prebuild is byte-for-byte untouched.

## OTA / release safety

- Native changes (MainActivity + generated manifest/styles) ship with the next binary. Existing binaries keep their baked-in manifest lock, so nothing changes for them.
- **The camera fix (3) cannot ship as an OTA** — it's a manifest change. Users hitting the broken KYC flow stay broken until the next binary clears Play review, which is an argument for not sitting on this branch too long.
- **The auth fix (4) splits across both halves.** Its native half is pure JS and *can* ship via EAS Update on the existing `runtimeVersion` 1.0.0, reaching already-installed apps without waiting on Play review. Its web half (`auth-context.tsx`, `is-native.ts`) deploys with the normal web deploy and is backwards-compatible with the currently-shipped binary — old apps simply receive fewer `signOut` messages. No `otaUpdateVersion` bump is required either way.
- Earlier revisions of this branch had zero JS changes; item 4 changes that, so an EAS update from this branch now carries real JS.
- `MainActivity.kt` hand-edits survive non-clean prebuild (verified: two prebuild runs left it untouched).

## QA before release

- Resizable emulator / tablet: rotate + resize on Android, check the WebView reflows and the splash (cover-mode crops in landscape - expected).
- Phone: confirm portrait lock still holds, including opening the app a while after a widget refresh (warm-process path).
- One glance at an Android <15 device/emulator for the status bar area (expected: unchanged - it was already transparent at runtime).
- **Camera (3), the one that most needs a real device:** run a full iDenfy verification on an Android build. Expect the OS camera prompt on first use, then a working document + selfie capture. Worth also checking the deny path (deny the prompt -> iDenfy should show its own error, not hang) and that a second attempt re-prompts rather than silently failing.
- Confirm the Play listing's supported-device count doesn't drop after the CAMERA permission is added (that's what the `uses-feature required="false"` tags are protecting against).
- **Auth (4):** the ordinary paths first — fresh install -> Google sign-in -> lands logged in; kill and relaunch -> still logged in; explicit log out -> lands on `AuthPage` and *stays* there (this is the path the `sawFbUser` gate touches, so it's the one to watch). Then log out and back in as a **different account** to confirm switching still works. On iOS, the same three with Apple sign-in.
- **Auth (4), reproducing the loop:** clear the app's storage from Android settings (Storage -> Clear cache, and separately Clear storage) while signed in, then relaunch. Before this change that's a hard loop; after it the app should either stay signed in or land cleanly on `AuthPage` with sign-in working on the first try.
- **Auth (4), the error path:** sign in with an account that has been banned (`auth/user-disabled`), or with airplane mode toggled mid-flow. Expect a readable alert rather than a silent bounce back to the login screen, and a corresponding Sentry event.

---

## Review round (2026-08-26, commit `0a0babbc6`)

Consolidated from a 10-finder code review + an adversarial verification pass, and the codex review. Kept to contained fixes; the items deliberately *not* taken are listed below.

**Auth handoff**
- The `nativeFbUser` retry timers are now held in a ref and cancelled on sign-out / uid change, and a queued post is skipped if native's user has changed since it was scheduled. Without this a late post handed the *old* user back to the web, which re-signed it in and echoed `users` to native — undoing the sign-out, and for accounts the web auto-logs-out (banned / `userDeleted`) re-arming a fresh batch each cycle: an unbounded sign-in → alert → sign-out loop that the previous single AuthPage post never produced (verified by two independent traces).
- A sign-out generation counter guards the `users` echo so an echo that raced a sign-out can't re-persist the cleared credential.
- Credentials are only posted into our own origin: the WebView URL is tracked via `onNavigationStateChange` and posts are skipped while it is on a third-party page (iDenfy, Stripe). `WebView.postMessage` dispatches into whichever document is loaded (codex P1).
- Web: concurrent `nativeFbUser` applications for the same uid coalesce onto one in-flight `setFirebaseUserViaJson`, so the retry fan-out no longer fires `onIdTokenChanged` twice (double user load; double `createUser` → 403 for a brand-new account).
- AuthPage no longer posts `nativeFbUser` itself — the uid effect is the single channel — so it no longer takes the webview ref.

**AuthPage error surfacing**
- Google: backing out of the consent screen arrives as OAuth `access_denied` (not `cancel`/`dismiss`) — no alert / Sentry for it.
- Apple: `updateEmail`/`updateProfile` after a successful `signInWithCredential` are best-effort; a rejection there no longer shows "Could not sign in" over a signed-in app.
- The default alert no longer shows Firebase's developer-facing `Firebase: Error (auth/...)` string; typed error-code extraction instead of `any`.

**Play advisories**
- The portrait lock is **generated into `MainActivity.kt` by the config plugin** (`@generated` blocks via `mergeContents`) instead of a hand-edit, so `expo prebuild --clean` keeps it (codex P2; the earlier "hand-edits survive non-clean prebuild" note above is superseded). It now skips large screens (`smallestScreenWidthDp >= 600`) and re-applies on configuration changes, per Android's phone-only orientation recipe, so pre-Android-16 tablets/foldables rotate too (codex P2).
- The `statusBarColor` strip is a plain `withAndroidStyles` mod using `removeStylesItem`. The "finalized mod, a styles mod loses the race" premise was wrong: user-plugin mods run *after* the default ones (reproduced with `compileModsAsync`, both introspect and on-disk).
- Camera plugin uses `ensurePermission`, keeps the permission list sorted and places `<uses-feature>` after `<uses-permission>`, so a clean prebuild produces the same manifest as the checked-in one (previously only true because the non-clean run read the already-edited file).

**localStorage on logout** also carries the v2 keys behind `useNativeInfo()` (`is-native-v2` / `native-platform-v2` / `native-version`, now constants in `common/native-message`). The legacy pair was already surviving via the sessionStorage fallback and the WebView's injected JS, so that part of the original item 4 was not the load-bearing fix.

**Verified without a build**
- `tsc --noEmit` clean for `native/` and `web/`.
- Non-clean `expo prebuild -p android` changes only `MainActivity.kt` (the generated blocks) and is idempotent on a second run.
- `expo prebuild -p android --clean` regenerates `AndroidManifest.xml`, `styles.xml` and `MainActivity.kt` **byte-identical** to the checked-in files. The only remaining `--clean` loss is the pre-existing `WidgetPin` hand-edit (`MainApplication.kt`, `WidgetPinModule.kt`, `WidgetPinPackage.kt`) — unrelated to this PR, but whoever runs `yarn clean` should know.
- `./gradlew :app:compileDebugKotlin` passes with the generated `MainActivity.kt`.

**Deliberately not taken (scope)**
- *Camera grant not origin-scoped* (codex P1): once CAMERA is granted, react-native-webview grants video capture to any origin loaded later in the same WebView. Fixing it needs either a navigation allowlist — iDenfy, Stripe checkout, Twitch OAuth and QR links all load in-webview today, so an allowlist would break flows we can't enumerate with certainty — or a native `WebChromeClient` override. Web-side external links already open in the system browser (`onOpenWindow`). Follow-up, not this PR.
- *Invalid stored native credential after the `sawFbUser` gate* (codex P2): nothing in the backend disables or revokes Firebase auth users (bans/deletions are DB flags only), so the state needs a manual Firebase-console action, and it self-recovers with one tap on the web login button (`loginClicked` → native sign-out → AuthPage). A rejected-credential handshake would span common/web/native for that case; left out.
- Dropping `orientation: 'portrait'` in favour of `ios.infoPlist.UISupportedInterfaceOrientations` would change iOS output; out of scope.

---

## Re-review round 2 (2026-08-27, commit `349d7bb3d`)

The re-review found four residual auth races on `0a0babbc6` (CI green; device QA passed every step, including the real deleted-account and camera controls). Fixed the contained ones; deferred the one that would change the message protocol.

**Fixed**
- **Provisional navigation treated as trusted (P1).** `onNavigationStateChange` fires at navigation *start* too (`onLoadingStart`, before the request is allowed), so an external page mid-navigation could be marked trusted. The WebView now updates the tracked URL only on **committed** navigations (`!state.loading`); `isManifoldUrl` compares full **origin** (scheme+host+port), not just host; and the `startedListening` credential resend is gated on the **sending document's origin** (`nativeEvent.url`). Credentials cannot be posted into a non-Manifold document.
- **Different-uid handoffs finishing out of order (P1).** The web now applies every `nativeFbUser` on **one serialized chain** rather than a uid-keyed coalescer, so they commit in arrival order (last-arrived user wins) and same-user retries collapse to a no-op once that user is current. `setFirebaseUserViaJson` awaits a token refresh before committing, so concurrent applications were a last-to-commit race that could restore the wrong account.
- **Stale observer pushing the old account back (P1, web half).** `auth-context` re-checks `auth.currentUser?.uid` after its async fetches (and after `createUser`) before `onAuthLoad`, so a callback whose user has since signed out / switched is dropped **at the source** instead of posting a stale `users` echo to native.
- **Mismatch cleanup signing out a newer session (P2).** The native `users` mismatch branch now undoes its own re-sign only when that user is still the current firebase user, so an account that signed in during the await is left alone (no global `signOutUsers`).

**Deliberately deferred (scope)** — the full auth-epoch/nonce **round-trip through the `nativeFbUser → users` protocol**. This is codex's belt to the web-observer guard's suspenders; it changes the message contract across `common`/`web`/`native` (an "update of its own"). With the web-side observer guard killing stale echoes at their source and the handoff chain applying in order, the residual interleavings converge to AuthPage rather than loop — which the on-device deleted-account/logout/switch tests confirmed. Left for a follow-up if the loop ever recurs in Sentry.

**Verified without a build:** native + web `tsc` clean; no `android/` changes in this round, so prebuild output is unchanged from round 1.

### Device QA findings folded in (moto g play, Android 14)

Informational notes from the on-device run that belong in the record:
- A `screenOrientation="portrait"` survives in the **merged** manifest on `GmsBarcodeScanningDelegateActivity` (from `play-services-code-scanner`). Not ours and not launchable, so it should not trip the advisory — same class of caveat as the RN `statusBarColor` one noted above.
- `configChanges` omits `smallestScreenSize`, so a `smallestScreenWidthDp` change (fold/unfold) **recreates** the activity rather than firing `onConfigurationChanged`. Behaviour is still correct (the lock re-applies in `onCreate`), but the mechanism for the fold case is recreation, not the config-change handler.
- The deleted/banned alert is a **web `window.alert`** (renders with Chrome's "The page at … says" chrome), a different mechanism from this PR's native `Alert` error surfacing — expected, just noted.
- Runbook nits corrected: step 5 verifies via `nativeFbUser` timestamps (native never `log()`s `signOut`); step 6's `run-as` line now quotes the space-containing dirs and notes `MSYS_NO_PATHCONV=1` for Git Bash.

**Still needs a device before the Play upload:** runbook step 9 (no credential post while on `idenfy.com`, driven mid-KYC — directly exercises the P1 fix above), the airplane-mode half of step 8, and step 10 (Play device-catalog count, post-upload).

---

## Re-review round 3 (2026-08-27, commit `bcb263523`)

Two contained P1s from the scoped re-review of `349d7bb3d`.

- **Clear WebView trust on navigation start (P1).** Round 2 refused to trust a *provisional* destination but only ever *set* trust on a committed load — so during an **outbound** load the previous Manifold URL stayed trusted for the whole load, and once the external document commits, its scripts run before `onLoadEnd`; a pending handoff timer would still see Manifold and could post credentials into it. Now `onLoadStart` clears trust (empty string → `isManifoldUrl` false) and `onNavigationStateChange` re-establishes it only on the committed load. Uses the dedicated `onLoadStart` prop rather than the `loading` flag inside `onNavigationStateChange`, because iOS fires `onLoadingStart` from `decidePolicyForNavigationAction` where `WKWebView.loading` isn't yet reliably true (Android's `loading = progress != 100` is reliable; iOS isn't). SPA (`pushState`) navigations don't fire `onLoadStart`, so in-app Manifold routing keeps its trust. Minor: on cold start the initial `/home` load also clears trust briefly, so the handoff there leans on `startedListening` (already the authoritative trigger) rather than the earliest retry timers — no functional change, the handoff still completes.
- **Guard the uid before installing the Supabase token (P1).** `updateSupabaseAuth` ran before round 2's stale-uid check, so a slow A callback finishing after a fast B would install `jwtA` into the shared Supabase REST/realtime client and *then* bail at the guard — leaving Supabase authenticated as A while Firebase and the UI are B. The uid check now sits above `updateSupabaseAuth`; the post-`createUser` check stays.

**Verified without a build:** native + web `tsc` clean; no `android/` changes, so prebuild output is unchanged.

**Runbook note:** step 9 (no credential post while on `idenfy.com`) must actually **arm an auth handoff mid-load** to exercise the round-3 fix — background/foreground the app while the WebView is on iDenfy so a `nativeFbUser` retry is pending, then confirm no `nativeFbUser` is posted. The plain step-9 as written navigates but doesn't arm a handoff, so it would pass without testing this race.

---

## Re-review round 4 (2026-08-27, commits `3dbb94b70` + `61b5d5dc4`)

Two contained findings from the scoped re-review of `bcb263523`.

- **Re-arm the handoff after a slow load (P2, `3dbb94b70`).** The commit-only trust gate clears the trusted URL for the whole document load, and the four handoff retries (≤3 s) were all that carried the credential to the web. On a slow connection the load event can trail hydration by >3 s, so every retry fires inside the untrusted window and is dropped, stranding the webview-session-lost cohort on a logged-out page. A `authPostDropped` ref records that a wanted post was skipped; when a Manifold URL next commits while native still holds that user, `onNavigate` re-runs `sendWebviewAuthInfo`. Re-sends stay idempotent (web no-ops on a matching user), the fire-time uid + origin guards still apply, and sign-out clears the flag with the pending timers.
- **Restore trust only from a committed `onLoad` (P1, `61b5d5dc4`).** Round 3 restored trust from `onNavigationStateChange`, but react-native-webview 13.15 routes the provisional navigation-**start** event through it too, and on iOS that event's `loading` is not yet `true` — so an external → Manifold return cleared trust (`onLoadStart`) and immediately restored it to the *provisional* Manifold URL while the external document was still active, the exact window a pending retry could post into. Trust is now restored solely from `onLoad` (which fires only from `onLoadingFinish`, the success path, with `nativeEvent.url`); `onNavigationStateChange` is dropped and `onLoadEnd` is not used (it fires on errors too). SPA (`pushState`) navigations don't fire `onLoad`, so in-app Manifold routing keeps the trust from its last full load — still a Manifold origin, so the check passes. The P2 re-arm above now runs off the committed `onLoad` URL.

**Verified without a build:** native + web `tsc` clean; no `android/` changes, so prebuild output is unchanged.

**Runbook step 9 trigger corrected again:** a widget tap / `am start` merely resumes the `singleTask` `MainActivity` (and `WIDGET_CLICK` is unwired), so it doesn't arm a handoff. Use a debug `startedListening` emission while the WebView is on iDenfy to make the `nativeFbUser` retries pending against the external document.
## Device QA runbook (adb-driven, run by Claude with a device plugged in)

Everything below is executable from a shell with `adb` on PATH and one Android device/emulator attached (`adb devices` shows exactly one `device`). Replace `<apk>` with the built artifact. Package id is `com.markets.manifold`. Each step states the pass condition so the run is a checklist, not a judgement call.

### 0. Install + baseline

```sh
adb install -r <apk>
adb shell pm clear com.markets.manifold            # fresh-install state
adb shell dumpsys package com.markets.manifold | grep -E "versionName|android.permission.CAMERA"
```
Pass: `android.permission.CAMERA` appears under `requested permissions`, NOT under `runtime permissions` yet (no grant until prompted).

### 1. Manifest / build checks (no interaction)

```sh
# Orientation lock must be gone from the manifest, features declared optional
aapt2 dump xmltree --file AndroidManifest.xml <apk> | grep -E "screenOrientation|uses-feature|android.hardware.camera|required"
# Status bar colour must not be in AppTheme
aapt2 dump resources <apk> | grep -i statusBarColor || echo "no statusBarColor resource (pass)"
```
Pass: no `screenOrientation` attribute on `.MainActivity`; two `uses-feature` entries (`camera`, `camera.autofocus`) each with `required=0x0`; no `statusBarColor` in `style/AppTheme`.

### 2. Portrait lock on a phone (compact width)

```sh
adb shell settings put system accelerometer_rotation 0
adb shell settings put system user_rotation 1               # force landscape at the system level
adb shell monkey -p com.markets.manifold -c android.intent.category.LAUNCHER 1
sleep 4
adb shell dumpsys window displays | grep -E "mCurrentRotation|curRotation|rotation="
adb shell dumpsys activity activities | grep -E "mRequestedOrientation|requestedOrientation" | head -3
```
Pass: the app's task reports `requestedOrientation=1` (SCREEN_ORIENTATION_PORTRAIT) and the display rotation while the app is foreground is `0` (or `2`), i.e. the app stays portrait despite the system asking for landscape. Restore: `adb shell settings put system accelerometer_rotation 1`.

Warm-process path (the PR's rationale for a native lock): trigger a widget refresh, then relaunch.
```sh
adb shell am broadcast -a android.appwidget.action.APPWIDGET_UPDATE -n com.markets.manifold/.widget.Streak
sleep 2
adb shell am start -n com.markets.manifold/.MainActivity
adb shell dumpsys activity activities | grep -E "requestedOrientation" | head -1
```
Pass: still `requestedOrientation=1`.

### 3. Large screens rotate (tablet / foldable emulator, or `wm size`)

On a phone you can fake an expanded width without a second device:
```sh
adb shell wm size 1600x2560 && adb shell wm density 240      # ≈ 667dp smallest width
adb shell am force-stop com.markets.manifold
adb shell am start -n com.markets.manifold/.MainActivity
sleep 3
adb shell dumpsys activity activities | grep -E "requestedOrientation" | head -1
adb shell settings put system user_rotation 1; sleep 2
adb shell dumpsys window displays | grep -E "mCurrentRotation|curRotation"
adb shell wm size reset && adb shell wm density reset
```
Pass: `requestedOrientation=-1` (UNSPECIFIED) and the display rotates to `1`/`3` while the app is foreground. Then, still running, shrink back (`wm size 1080x2400`, `wm density 420`) and re-read: the `onConfigurationChanged` re-apply must flip it back to `requestedOrientation=1`.

### 4. Camera in the iDenfy KYC flow (the fix that needs a real device most)

Use an account that has not completed KYC on the dev stack.
```sh
adb shell pm revoke com.markets.manifold android.permission.CAMERA   # ensure a clean prompt
adb logcat -c
# Drive: open the app -> Profile -> Verify identity -> start; when iDenfy asks for the document,
# the OS camera permission dialog must appear. Tap Allow.
adb logcat -d | grep -E "RNCWebView|PermissionRequest|getUserMedia|CameraDevice|camera" | tail -20
adb shell dumpsys package com.markets.manifold | grep -A1 "android.permission.CAMERA" | grep granted
```
Pass: dialog shown once; `granted=true` afterwards; iDenfy's document capture shows a live preview and accepts a photo; selfie step likewise.
Deny path: `adb shell pm revoke ... CAMERA`, restart the flow, tap **Deny** — iDenfy must show its own "camera blocked" error (not a hang). Start again — the OS re-prompts (or, after two denies, iDenfy's error appears with a settings link).

### 5. Auth: ordinary paths (Google account)

```sh
adb shell pm clear com.markets.manifold
adb logcat -c
adb shell monkey -p com.markets.manifold -c android.intent.category.LAUNCHER 1
```
Sign in with Google on the AuthPage. Pass: lands on `/home` signed in within ~5 s; `adb logcat -d | grep -c "received nativeFbUser"` is ≥ 1; no `Could not sign in` alert.

Kill + relaunch:
```sh
adb shell am force-stop com.markets.manifold && adb shell am start -n com.markets.manifold/.MainActivity
```
Pass: still signed in, no AuthPage flash longer than the splash.

Log out from the web sidebar. Pass: lands on AuthPage and **stays there for 10 s** (this is the timer-cancellation path: a late `nativeFbUser` re-post must not sign the user back in). Verify by `nativeFbUser` timestamps: `adb logcat -d | grep "nativeFbUser"` must show none after the moment you logged out. (Do **not** grep for `signOut` — native never `log()`s that path, so it returns 0 on success and failure alike.)

Switch account: sign in as a different Google account. Pass: `/home` shows the second account's name.

### 6. Auth: reproduce the reported loop (cleared web storage, native session intact)

```sh
# Signed in state from step 5, then wipe only the WebView's storage, not the app's.
# From Git Bash on Windows, prefix with MSYS_NO_PATHCONV=1 so the shell doesn't
# mangle the on-device paths. Quote the dirs that contain spaces.
adb shell am force-stop com.markets.manifold
MSYS_NO_PATHCONV=1 adb shell run-as com.markets.manifold sh -c \
  'cd app_webview/Default && rm -rf IndexedDB "Local Storage" "Session Storage" WebStorage; ls'
adb shell am start -n com.markets.manifold/.MainActivity
```
(`run-as` needs a debuggable build — build a `debuggable true` release variant, or on a stock release build use Android Settings -> Apps -> Manifold -> Storage -> Clear cache, then Clear storage, as in the PR body. The real dirs are `IndexedDB`, `Local Storage`, `Session Storage`, `WebStorage`.)
Pass: the app either stays signed in (handoff lands) or shows AuthPage once and sign-in works on the first tap. Fail: alternating AuthPage / logged-out web page more than once.

### 7. Auth: banned / deleted account must NOT loop (regression guard for the timer fix)

Use a dev-stack account flagged `userDeleted` (or listed in `BANNED_TRADING_USER_IDS` on dev).
Sign in with it on the AuthPage.
Pass: exactly one alert ("You have deleted the account…" / "You are banned…"), then AuthPage, and it **stays** there. `adb logcat -d | grep -c "received nativeFbUser"` stops increasing after the alert. Fail: repeated alerts.

### 8. Auth: sign-in error surfaces

Airplane mode mid-flow:
```sh
adb shell cmd connectivity airplane-mode enable    # after tapping Sign in with Google, before the browser returns
```
Pass: an alert titled **Could not sign in** with the "couldn't reach Manifold" copy; `adb shell cmd connectivity airplane-mode disable`; retry works.
Cancel at Google's consent page (tap Cancel/Deny in the Custom Tab). Pass: back on AuthPage, **no** alert.

### 9. Credential handoff never reaches iDenfy

```sh
adb logcat -c
```
This must **arm an auth handoff while the external document is active** to exercise the round-3/4 fix, not just navigate. Neither a background→foreground nor a widget/`am start` relaunch works: they don't re-send credentials, and `MainActivity` is `singleTask` so `am start` merely resumes the existing activity. The reliable trigger is a **debug `startedListening` emission** while the WebView is on iDenfy — either inject one over the bridge from a debug build, or add a temporary dev button that posts `startedListening`. With `fbUser` set, that schedules the 100/500/1000/3000 ms `nativeFbUser` retries; drive it so they are pending while the WebView is on `ivs.idenfy.com` (both mid-load and after commit).
Pass: `adb logcat -d | grep "nativeFbUser"` shows **no** posts while the current URL is an idenfy.com host. (Guards: `onLoadStart` clears trust, trust is restored only on a successful Manifold `onLoad`, and the fire-time `isManifoldUrl(webviewUrl)` check in App.tsx; the WebView URL is visible via `adb shell dumpsys activity top | grep -i url`.)

### 10. Play listing sanity (after upload, not on-device)

In Play Console -> Release -> Device catalog, the supported-device count must not drop after the CAMERA permission lands (the `required="false"` features protect this).

### DIFF
--- a/common/src/native-message.ts
+++ b/common/src/native-message.ts
@@ -63,6 +63,17 @@ export type webToNativeMessageType =
   | 'pinStreakWidget'
 export const IS_NATIVE_KEY = 'is-native'
 export const PLATFORM_KEY = 'native-platform'
+// localStorage keys behind the web client's useNativeInfo() (written by
+// web/components/native-message-provider.tsx). Listed here so a localStorage
+// wipe on logout can carry them across; keep in sync with that provider.
+export const IS_NATIVE_V2_KEY = 'is-native-v2'
+export const NATIVE_PLATFORM_V2_KEY = 'native-platform-v2'
+export const NATIVE_VERSION_KEY = 'native-version'
+export const NATIVE_INFO_LOCAL_KEYS = [
+  IS_NATIVE_V2_KEY,
+  NATIVE_PLATFORM_V2_KEY,
+  NATIVE_VERSION_KEY,
+]
 
 // Streak snapshot sent to the native app, which mirrors it into a shared App
 // Group container for the home/lock-screen streak widget to read. All times are
--- a/native/App.tsx
+++ b/native/App.tsx
@@ -174,10 +174,16 @@ const App = () => {
   // the WebView to iDenfy's hosted page (boosts go to Stripe), and
   // WebView.postMessage dispatches into whichever document is loaded — so the
   // credential handoff below must only ever be posted into our own pages.
+  // Set only from committed navigations (see CustomWebview's onNavigate), never
+  // from a provisional one — a navigation START fires before the request is even
+  // allowed, so an external page could otherwise be marked trusted while it is
+  // still the active document.
   const webviewUrl = useRef(baseUri)
+  // Full-origin match (scheme + host + port), not just host: a page served over
+  // http://manifold.markets or a look-alike subdomain must not read as ours.
   const isManifoldUrl = (url: string) => {
     try {
-      return new URL(url).host === new URL(baseUri).host
+      return new URL(url).origin === new URL(baseUri).origin
     } catch {
       return false
     }
@@ -430,6 +436,10 @@ const App = () => {
 
   const handleMessageFromWebview = async ({ nativeEvent }: any) => {
     const { data } = nativeEvent
+    // The origin of the document that sent this message. Auth-sensitive handlers
+    // gate on it so a third-party page loaded in the WebView (iDenfy, Stripe)
+    // can't drive the credential handoff.
+    const messageUrl = nativeEvent.url as string | undefined
     const { type, data: payload } = JSON.parse(data) as webToNativeMessage
     // We handle auth with a custom screen, so if the user sees a login button on the client, we're out of sync
     if (type === 'loginClicked') {
@@ -525,13 +535,16 @@ const App = () => {
           const generation = authGeneration.current
           // We don't actually use the firebase auth for anything right now, but in case we do in the future...
           await setFirebaseUserViaJson(fbUser, app)
-          // A sign-out landed while this echo was being applied. The echo must
-          // not win: the re-sign above re-armed the handoff, and persisting the
-          // user below would resurrect the credential the sign-out just cleared.
+          // A sign-out landed while this echo was being applied, so it is stale.
+          // Don't persist it, and undo the re-sign above — but ONLY if this stale
+          // user is still the current firebase user. If a newer account signed in
+          // during the await, leaving it alone is what keeps a global signOut from
+          // clearing the account the user actually chose.
           if (generation !== authGeneration.current) {
-            await signOutUsers(
-              'Error undoing a users echo that raced a sign-out'
-            )
+            if (auth.currentUser?.uid === fbUser.uid)
+              await signOutUsers(
+                'Error undoing a users echo that raced a sign-out'
+              )
             return
           }
           await storeData('user', fbUser)
@@ -559,7 +572,9 @@ const App = () => {
     } else if (type === 'startedListening') {
       log('Client started listening')
       listeningToNative.current = true
-      if (fbUser) sendWebviewAuthInfo(fbUser)
+      // Only re-hand credentials to a sender that is actually one of our pages.
+      if (fbUser && (!messageUrl || isManifoldUrl(messageUrl)))
+        sendWebviewAuthInfo(fbUser)
       // Deliver any notification that arrived during cold start
       if (pendingNotification.current) {
         const { notification, destination } = pendingNotification.current
--- a/native/App.tsx
+++ b/native/App.tsx
@@ -97,9 +97,17 @@ const App = () => {
 
   // Auth
   const [fbUser, setFbUser] = useState<FirebaseUser | null>(auth.currentUser)
-  // Auth.currentUser didn't update, so we track the state manually. This used to
-  // run in the render body, which registered a fresh listener — never
-  // unsubscribed — on every single render.
+  // Mirror of fbUser for callbacks that outlive the render they were created in
+  // (the auth handoff timers below): a ref reads the latest value, a closure
+  // reads the value from when it was scheduled.
+  const fbUserRef = useRef<FirebaseUser | null>(fbUser)
+  useEffect(() => {
+    fbUserRef.current = fbUser
+  }, [fbUser])
+  // Bumped on every native sign-out, so async work that started before the
+  // sign-out can tell it has been superseded.
+  const authGeneration = useRef(0)
+  // Auth.currentUser didn't update, so we track the state manually.
   useEffect(
     () => auth.onAuthStateChanged((user) => (user ? setFbUser(user) : null)),
     []
@@ -162,27 +170,57 @@ const App = () => {
     clearData('lastNotificationIds') // no longer used, clear them from local storage
   }, [])
 
+  // The URL the WebView currently has loaded. Identity verification navigates
+  // the WebView to iDenfy's hosted page (boosts go to Stripe), and
+  // WebView.postMessage dispatches into whichever document is loaded — so the
+  // credential handoff below must only ever be posted into our own pages.
+  const webviewUrl = useRef(baseUri)
+  const isManifoldUrl = (url: string) => {
+    try {
+      return new URL(url).host === new URL(baseUri).host
+    } catch {
+      return false
+    }
+  }
+
+  // Pending 'nativeFbUser' posts, kept so a sign-out (or a switch to another
+  // account) can cancel them. A late post would otherwise hand the OLD user back
+  // to the web, which re-signs it in and echoes it to native as 'users' —
+  // undoing the sign-out, and for accounts the web auto-logs-out (banned or
+  // deleted) doing so forever.
+  const pendingAuthPosts = useRef<ReturnType<typeof setTimeout>[]>([])
+  const cancelPendingAuthPosts = () => {
+    pendingAuthPosts.current.forEach((timer) => clearTimeout(timer))
+    pendingAuthPosts.current = []
+  }
+
   // Sends the saved user to the web client to make the log in process faster
   const sendWebviewAuthInfo = (user: FirebaseUser) => {
+    cancelPendingAuthPosts()
     // We use a timeout because sometimes the auth persistence manager is still undefined on the client side
     // Seems my iPhone 12 mini can regularly handle a shorter timeout
     const timeouts = [100, 500, 1000, 3000]
-    timeouts.forEach((timeout) => {
+    pendingAuthPosts.current = timeouts.map((timeout) =>
       setTimeout(() => {
+        // Stale: native signed out or switched accounts since this was queued.
+        if (fbUserRef.current?.uid !== user.uid) return
+        // Never post credentials into a third-party page.
+        if (!isManifoldUrl(webviewUrl.current)) return
         communicateWithWebview('nativeFbUser', user)
       }, timeout)
-    })
+    )
   }
 
-  // This handoff is the ONLY channel by which the web client learns about a
-  // native sign-in, and a dropped message strands the user on a logged-out page
-  // whose login button signs them back out of native — an unrecoverable loop.
-  // A fresh sign-in used to get a single un-retried post from AuthPage (a
-  // component the sign-in itself unmounts), while only the restore-from-storage
-  // path got the retries. Run it on every uid change so both paths get them.
-  // Re-sends are cheap: the web side no-ops when the email already matches.
+  // This handoff is the channel by which the web client learns about a native
+  // sign-in, and a dropped message strands the user on a logged-out page whose
+  // login button signs them back out of native — an unrecoverable loop. A fresh
+  // sign-in used to get a single un-retried post from AuthPage (a component the
+  // sign-in itself unmounts), while only the restore-from-storage path got the
+  // retries. Run it on every uid change so both paths get them; the web side
+  // no-ops when the email already matches.
   useEffect(() => {
     if (fbUser) sendWebviewAuthInfo(fbUser)
+    return cancelPendingAuthPosts
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [fbUser?.uid])
 
@@ -467,7 +505,10 @@ const App = () => {
       // Android only: show the system "add widget to home screen" dialog via our
       // native module. No-op on iOS (no such API) and on older builds that
       // predate the module (NativeModules.WidgetPin is then undefined).
-      if (Platform.OS === 'android' && NativeModules.WidgetPin?.pinStreakWidget) {
+      if (
+        Platform.OS === 'android' &&
+        NativeModules.WidgetPin?.pinStreakWidget
+      ) {
         try {
           await NativeModules.WidgetPin.pinStreakWidget()
         } catch (e) {
@@ -481,9 +522,18 @@ const App = () => {
         const fbUserAndPrivateUser = JSON.parse(payload)
         if (fbUserAndPrivateUser && fbUserAndPrivateUser.fbUser) {
           const fbUser = fbUserAndPrivateUser.fbUser as FirebaseUser
-          // log('Signing in fb user from webview cache')
+          const generation = authGeneration.current
           // We don't actually use the firebase auth for anything right now, but in case we do in the future...
           await setFirebaseUserViaJson(fbUser, app)
+          // A sign-out landed while this echo was being applied. The echo must
+          // not win: the re-sign above re-armed the handoff, and persisting the
+          // user below would resurrect the credential the sign-out just cleared.
+          if (generation !== authGeneration.current) {
+            await signOutUsers(
+              'Error undoing a users echo that raced a sign-out'
+            )
+            return
+          }
           await storeData('user', fbUser)
           // Refresh the streak widget from the API on each (re)auth / app open.
           if (fbUser.uid) syncStreakFromApi(fbUser.uid)
@@ -555,6 +605,8 @@ const App = () => {
   }
 
   const signOutUsers = async (errorMessage: string) => {
+    authGeneration.current += 1
+    cancelPendingAuthPosts()
     try {
       await auth.signOut()
     } catch (err) {
@@ -649,7 +701,6 @@ const App = () => {
           hidden={false}
         />
         <SplashAuth
-          webview={webview}
           hasLoadedWebView={hasLoadedWebView}
           fbUser={fbUser}
           isConnected={isConnected}
@@ -662,6 +713,9 @@ const App = () => {
           setHasLoadedWebView={setHasLoadedWebView}
           handleMessageFromWebview={handleMessageFromWebview}
           handleExternalLink={handleExternalLink}
+          onNavigate={(url) => {
+            webviewUrl.current = url
+          }}
         />
       </SafeAreaView>
       {/*<ExportLogsButton />*/}
--- a/native/App.tsx
+++ b/native/App.tsx
@@ -199,10 +199,17 @@ const App = () => {
     pendingAuthPosts.current.forEach((timer) => clearTimeout(timer))
     pendingAuthPosts.current = []
   }
+  // Whether a wanted handoff post was dropped because the WebView was on an
+  // untrusted URL at fire time. Trust is cleared for the whole document load,
+  // so on a slow connection every timer below can fire inside that window —
+  // this flag lets the next committed Manifold page re-arm the handoff (see
+  // onNavigate) instead of silently stranding the web client logged out.
+  const authPostDropped = useRef(false)
 
   // Sends the saved user to the web client to make the log in process faster
   const sendWebviewAuthInfo = (user: FirebaseUser) => {
     cancelPendingAuthPosts()
+    authPostDropped.current = false
     // We use a timeout because sometimes the auth persistence manager is still undefined on the client side
     // Seems my iPhone 12 mini can regularly handle a shorter timeout
     const timeouts = [100, 500, 1000, 3000]
@@ -211,7 +218,10 @@ const App = () => {
         // Stale: native signed out or switched accounts since this was queued.
         if (fbUserRef.current?.uid !== user.uid) return
         // Never post credentials into a third-party page.
-        if (!isManifoldUrl(webviewUrl.current)) return
+        if (!isManifoldUrl(webviewUrl.current)) {
+          authPostDropped.current = true
+          return
+        }
         communicateWithWebview('nativeFbUser', user)
       }, timeout)
     )
@@ -622,6 +632,7 @@ const App = () => {
   const signOutUsers = async (errorMessage: string) => {
     authGeneration.current += 1
     cancelPendingAuthPosts()
+    authPostDropped.current = false
     try {
       await auth.signOut()
     } catch (err) {
@@ -732,6 +743,18 @@ const App = () => {
             // null (navigation start) leaves an empty string, which isManifoldUrl
             // treats as untrusted until the next load commits.
             webviewUrl.current = url ?? ''
+            // A slow initial load can outlast the whole retry ladder while
+            // trust is cleared; if that dropped a wanted post, re-arm the
+            // handoff now that one of our pages has committed. The web side
+            // no-ops when the user already matches, so re-sends are idempotent.
+            if (
+              url &&
+              authPostDropped.current &&
+              fbUserRef.current &&
+              isManifoldUrl(url)
+            ) {
+              sendWebviewAuthInfo(fbUserRef.current)
+            }
           }}
         />
       </SafeAreaView>
--- a/native/App.tsx
+++ b/native/App.tsx
@@ -21,7 +21,6 @@ import Constants from 'expo-constants'
 import 'expo-dev-client'
 import * as Linking from 'expo-linking'
 import * as Notifications from 'expo-notifications'
-import * as ScreenOrientation from 'expo-screen-orientation'
 import * as SplashScreen from 'expo-splash-screen'
 import { StatusBar } from 'expo-status-bar'
 import * as StoreReview from 'expo-store-review'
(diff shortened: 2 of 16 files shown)


todwadd:
## Pre-merge review round 4 (2026-08-27, commit 3dbb94b70)

Final merge-safety pass: 7-dimension multi-agent review + adversarial verification against react-native-webview 13.15.0 source, on head bcb263523.

**One confirmed regression, fixed in 3dbb94b70:** round 3's clear-trust-on-load-start left the whole document load untrusted, and the four handoff retries (armed at the uid effect / 'startedListening') were all that carried the credential. On slow connections the load event can trail hydration by >3s, so every retry could fire inside the untrusted window and drop, with nothing re-sending afterward — stranding the webview-session-lost cohort on a logged-out page (one forced re-auth; four independent reviewers converged on this). Fix: track a dropped post in a ref and re-arm sendWebviewAuthInfo when a Manifold URL commits while native still holds the user.

**Confirmed-minor, deliberately not fixed here (follow-up):** the CAMERA grant isn't origin-scoped — RNW's Android WebChromeClient auto-grants getUserMedia to any origin in the WebView once the app permission is granted (no per-origin hook in 13.15.0). Surface is bounded to the iDenfy/Stripe flows and their redirects; external links exit to the system browser. Mirrors the pre-existing geolocation pattern.

**Checked clean:** zero main drift (branch is on the main tip); no protocol changes (all four old/new native×web pairings compatible; the sawFbUser gate alone already helps old binaries); no new recurring costs (bounded timers, per-incident Sentry, zero dep changes; the listener-leak fix and handoff coalescing reduce load); android config verified empirically — a real non-clean `expo prebuild -p android` reproduced the checked-in files with zero drift; plain-browser web behavior unchanged; banned/deleted-account cycle cannot re-arm.

**Two doc corrections:** (1) the OTA note in the body is stale — app.config.js now pins runtimeVersion 1.1.0 (widget module), so an EAS Update of this JS reaches only the new 2.1.0 binaries, not existing installs; the auth fix ships with the binary. (2) The statusBarColor strip works with plain withAndroidStyles (mod ordering verified in @expo/config-plugins 54.0.2); no finalized mod needed.

PR #4026, opened 2026-08-29
feat(perps): /perps hub page + shared mobile/formatting fixes for perp markets

todwadd:
## Summary

A `/perps` hub that gathers every perpetual market — including the unlisted ones queued for the xStocks batch — into one terminal-style page, plus a set of shared perp fixes that surfaced while building it (these ship to the normal market pages too).

### The hub (`web/pages/perps.tsx`, new)
- **Terminal**: hero chart for the selected perp with oracle price (tick flash, freshness), 24h / 7d change, funding with next-transfer countdown, a 7-day **lean** strip (long share of open interest, inverted exactly from funding events — no new columns), Long/Short buttons that unfold the trade panel, and the shared position panel.
- **Watchlist**: aligned table (ticker · price · change · sparkline · lean), sortable headers, default sort by |change| with a 24h/7d toggle, 5 rows + "Show all". Clicking selects on-page.
- **Ticker tape** pinned to the top; items select on-page rather than navigate.
- **"Traders lean up/down"** signal (share of OI on the heavier side) in the tape, watchlist and terminal header.
- **Related markets** for the selected perp via `get-related-markets-by-group`, filtered to 5–95%, MC markets hydrated via `markets-by-ids`; "Create a market about {topic}" deep link.
- Rail cards: **your positions** across all perps (`?as=<userId>` preview; one `get-perp-positions?userId` call), **recent activity** (one `get-perp-events?contractIds` call, compact two-column), **top mover**. Both poll once a minute, skip overlapping ticks, and pause while the tab is hidden. Each overnight addition is its own commit and behind a `HUB_FEATURES` switch for easy unwinding.
- **Explainer** shares copy with the market-page modal (`PerpExplainerContent`) plus a per-market parameters card.
- **Suggest a perpetual market**: name + optional data source, upvotes. ⚠️ Needs migration `backend/supabase/migrations/2026082701_perp_suggestions.sql` applied and the API deployed (`get-perp-suggestions`, `create-perp-suggestion`, `vote-perp-suggestion`); the page shows "Suggestions are unavailable right now" until then. Backend is typechecked but **not exercised at runtime yet** — please apply on dev first.
- Only `visibility === 'public'` perps are listed in production builds; a local dev server also shows the unlisted queue.
- Discoverability: a **Perps** pill in the browse topic row and an "All perpetual markets →" link in the Perpetual badge modal.
- ISR (`revalidate = 60`); `getStaticProps` queries `outcome_type = 'PERP'` (partial index — filtering on `mechanism` hit the anon statement timeout) and throws on error so ISR keeps the last good page.
- Phones: the terminal card slides out to full-bleed while it's the thing on screen (IntersectionObserver + margin transition), stats collapse to a two-row grid.

### Shared fixes (affect market pages)
- **`useMeasureSize`** attached its ResizeObserver once at mount, so `PerpChart` (which first renders a loading placeholder) was never measured and drew into a 720-unit viewBox — the illegible chart on 320px phones in prod (axis text renders ~5px). The hook now follows the element.
- **`PerpChart`**: module-level series/cadence cache + `prefetchPerpChart` (hub switches draw in ~50–80ms with no loading state); liquidation-proximity caption under the Liquidations pill; `plotClassName`; viewBox floor 320→280.
- **Open-position card** clipped its P&L block on phones / large OS text. Now a container-query context: the headline numbers scale with the card's width (`cqw`, px floors so text scaling can't re-widen them), labels/stat values likewise. ≥ ~336px renders as before. Verified 320/360/390 × 100–150% text, zero overflow.
- **Market page stats row** wraps at 320 instead of running the Funding column off-screen.
- **`formatMoneyDisplay`** (new): whole mana from Ṁ100 up (truncated like `formatMoney`), cents below — used by the position card, history, trade preview rows, trades/holders tabs and the bets table. `formatMoneyPrecise` stays exact for toasts, the affordability message, balance changes and notifications. Tests cover both.

## Deploy notes
- Apply `2026082701_perp_suggestions.sql` (RLS enabled, no public policies) and deploy the API before the web deploy — the hub's tape and positions cards use the new batched request shapes.
- `backend/api/url-map-config.yaml` adds the perp read routes to the read workers; needs the `gcloud compute url-maps import` from the file header.

## Test plan
- [x] Headless Chrome at 1280/1440/1600/1920 and 320/360/390 (light + dark), horizontal overflow scans
- [x] `web` typecheck, eslint; `common` jest (`format.test.ts`)
- [x] Chart switch timing measured with the cache (no loading state)
- [ ] Apply `2026082701_perp_suggestions.sql` on dev, deploy API, exercise suggest/vote end to end
- [ ] Prod: apply migration + deploy API before/with the web deploy

## Follow-ups (not in this PR)
Alerts/follow on a perp (v1.1), OI-over-time recorded on funding events, suggestion→launch auto-follow, per-user realized P&L, server-side answer hydration in `get-related-markets-by-group`, sitemap/nav entry + OG image.


todwadd:
Addressed all eight findings; rebased on `main` (`736538a33`) and force-pushed (`674a415a4`). One commit per finding where files allowed:

| # | Finding | Fix | Commit |
|---|---|---|---|
| 1 | Unlisted perps discoverable | `getStaticProps` keeps only `visibility === 'public'`; a local dev server still shows the unlisted queue for previewing a batch | `0f57c3564` |
| 2 | O(markets) uncached polling | `get-perp-events` takes `contractIds` (one ordered query, returns `contractId`); `get-perp-positions` accepts `userId` alone (whole perp book). Hub polls each **once per minute**, skips a tick while the previous is in flight, pauses while hidden. Perp read routes added to `url-map-config.yaml` (needs the `gcloud compute url-maps import` in the file header) | `0f57c3564` |
| 3 | Eager history prefetch | Week series `since` on an hour boundary; chart `since` aligned to the frame's bucket (edge cache can now serve repeats); prefetch starts on `requestIdleCallback` (2s cap), waits while hidden, skipped on data-saver | `0f57c3564`, `f33610b0c` |
| 4 | RLS exposure | Public select policies removed from both tables; RLS stays enabled with no policies — the API (direct client) is the only reader | `df0c6f5cc` |
| 5 | Formatter truncation | `formatMoneyPrecise` restored to exact semantics (toasts, affordability, ledger, notifications). New `formatMoneyDisplay` (cents < Ṁ100, whole above) for cards/tables/tabs. Tests cover both, incl. the `Ṁ100.50` case | `674a415a4` |
| 6 | Suggestion abuse | Quota counted inside the transaction behind `pg_advisory_xact_lock(hashtext(uid))`; upvoting an existing name doesn't count; both writes wrapped in `onlyUsersWhoCanPerformAction('post')`; list route on the read-worker map | `c55590bb8` |
| 7 | Explainer rebase | Kept main's contract-aware section (`PerpFeesItem`, leverage cap, admin note). `PerpExplainerContent` takes an optional `contract`: the modal passes it, the hub renders the generic body beside its own parameters card | rebase resolution |
| 8 | Liquidation caption | "…force-closes them and forfeits their margin to the pool." | `f33610b0c` |

Validation: `web`, `common`, `shared`, `api` typecheck; `format.test.ts` 9/9; changed-file ESLint. **Not exercised at runtime:** the batched `get-perp-events`/`get-perp-positions` shapes and the suggestions handler — they need the API deployed (the prod API rejects the new params, so the hub's tape/positions cards are empty against it until then). The `url-map-config.yaml` change needs the documented `gcloud … url-maps import` to take effect.

todwadd:
Follow-up `1fe45c803`: related markets are now fetched for the selected perp only (was one request per market on first paint). Measured against the prod API from the rebased branch: **25 API requests on first load** (reviewer measured ~45), no page errors at 1280 or 360; the market-page explainer modal shows the per-market fee schedule, leverage cap and hub link after the rebase. `get-perp-events` returns 400 and suggestions 404 against the current prod API as expected until the API is deployed.

todwadd:
`b7b7d2ee4`: decimals decision reversed — `formatMoneyDisplay` removed and all perp amounts go back through `formatMoneyPrecise`; `common/src/util/format.ts` and its tests are now byte-identical to `main`. Finding #5 is moot.

todwadd:
Second round addressed at `3487c10ee` (one commit each):
1. `/v0/` prefix on the four public read routes in `url-map-config.yaml`.
2. Chart cache hits now paint immediately and **revalidate** (cache read skipped, in-flight deduped); fresh history replaces the cached series. Verified: client nav hub → market page fires a series fetch and draws.
3. One `isListed` predicate (public + not deleted; unlisted only under a local dev server) applied to static props **and** the polled board, so an unlisted/deleted market drops out of an open tab on the next poll and the selection falls back.
4. `tx.one()` on the advisory-lock select.
5. `useMeasureSize` cleanup forgets the element so Strict Mode's replayed effect re-observes.
6. Prefetch tiers: intent (`onPointerEnter`/`onFocus` on watchlist rows, ticker items, chips, top mover) warms one chart; idle warms only the two most-traded non-selected markets. First-load API calls: **21** (25 last round, ~45 originally).
Prettier now clean on every changed file (CI blocker).

todwadd:
Round 3 at `e6d9ad1b8`:
1. Cache-hit revalidation now runs through the same loader as the cold path — a sparse fresh window falls back to **All** (revalidated) and relabels the frame, instead of leaving the overwritten cache's old points on screen. `b091bf5ba`
2. `probeCadence` dedupes in flight (`cadenceInflight`). Verified with a pointerenter → focus → click burst on a cold feed before the idle tier ran: exactly one `limit=8` probe and one frame request. Same commit.
3. The scale caveat: `usePerpBoard` now chunks ids at the 100 cap so polling (and live unlisting) can't silently stop at the 101st perp. `e6d9ad1b8`
Client nav hub → market page still revalidates and draws; web typecheck/lint/Prettier clean.

PR #4030, opened 2026-08-31
fix(perps): make the oracle tick un-wedgeable, and let a short-heavy book take a long

todwadd:
Two independent perp failures. One PR because they share a root cause: **both rules compare a NOTIONAL quantity against a MARGIN one.** `pool` is margin, `OI` is notional, and each rule bridged the gap with a constant and lost information doing it.

## 1. The oracle tick can no longer wedge

**What happened.** Short profits are paid out of `pool.L` (`closePosition`: short close with π>0 → `poolLongDelta = -π`). While the longs are underwater their reserve `Σ min(costBasis, value)` is depressed, so those payouts pass the solvency check — but the reserve grows back toward `Σ costBasis` when the price mean-reverts, and the mana has already left. `pool.L` is then below the long book's own refundable margin.

ADL cannot repair that. It scales profits and never cost bases, so the factor clamps to 0, the winners are settled and removed, and the deficit is still there with no profit left to scale. `assertPerpStateSolvent` sees `-Infinity`, the whole transaction rolls back, and the cached mark never advances — so every subsequent tick re-derives the identical failure from the identical state. Permanent, until an operator intervenes.

Twice in production, 12 hours dark each:

| | market | pools | deficit |
|---|---|---|---|
| 2026-08-07 | `uk-grid-carbon-intensity-gcokwh` | S below Σ short cb | −442 |
| 2026-08-29 | `openweight-ai-token-share-on-openro` | L 22,357 vs Σ long cb 28,485 | −6,128 |

On the second one, 36,344 of short profit had been paid out of `L` over the market's life against only 8,953 of *realized* long losses. The other 27,391 was funded by paper.

**Two changes, in order of how much they carry:**

**(a) Cross-side deficit transfer, in `applyADL`.** `assertPerpEscrowBalance` checks `L + S` against one contract balance, so the split is an accounting convention, not a custody boundary. Before pricing ADL, if one side is short of its own reserve and the other holds at least that much spare, move the deficit across. Total escrow conserved, no user balance moves. On the OpenRouter book `S` was carrying 29,068 of unreserved surplus while `L` was 6,128 short — the mana to fix it was already inside the contract.

All-or-nothing on purpose: a partial transfer can't make the book representable (the recipient's cover is still negative, the assert still throws) but it *would* change the ADL factors on the way to that throw, and a donor drained to exactly zero pushes the opposing factor to 0, settling a position against a pool that can't pay it. Restricting to full coverage means this can only ever turn a throw into a success, never reshape a path that was going to fail anyway. The tests caught this — the first version had the partial behaviour.

A covered book transfers exactly zero, so this is inert on every market that wasn't already wedged.

**(b) Liveness: `runOracleUpdate` always commits the price.** If the state still can't be represented, the tick writes the price plus `solvencyHaltTime`/`solvencyHaltReason` and nothing else — no pools, positions, events, or metrics.

A frozen mark is strictly worse than the state it refuses to write. It's wrong on the screen, it grows the deficit it's failing on, and it leaves trading **open** against a dead price for a full `maxOraclePriceAgeMs` — on 2026-08-29 a trader added to a position at 03:59Z against a mark that died at 00:50Z. The staleness "protection" was an accident of the failure, not a decision.

Since the mark now stays fresh, that implicit protection is gone by construction, so `assertPerpNotSolvencyHalted` gates opens **and** closes explicitly. Scheduler exits (liquidation, ADL, resolution) use the internal paths and are unaffected, so positions still settle. `add-perp-subsidy` is deliberately not gated — it's the rescue tool, and the first tick that applies cleanly clears the halt, so topping up the deficit side is the whole runbook.

Both callers log the halt with their own prefix (`[oracle-feeds]` / `[update-perps]`), so the existing GCP alert policies page exactly as they did when this threw.

*Accepted tradeoff:* pending liquidations and ADL stay pending across a halt and are re-derived at the next tick's price. A position whose liquidation is deferred that way can escape it on a rebound — a real but bounded transfer, against the alternative of a market dark for hours, with trading halted throughout so nobody can act on the gap.

## 2. A short-heavy book can take a long

The cap rationed a side's notional against the opposing **pool** and never looked at opposing **notional** at all. Measured on prod 2026-08-31:

```
long OI   734,348.81      short OI   928,994.19    ← short-heavy by 194,645
pool.L    181,363.42      pool.S     107,092.07

long headroom     1,128.63
short headroom  544,752.23
```

It refused the trade that would balance the book and freely accepted 545k more of the trade that worsened it. Live rejections: a **M$110** notional long on 8/30, M$6,900 on 8/28 — with an error telling the user to *"wait for more short interest to raise the cap"* while short interest was already 195k larger. Pool ratio read 1.69 long-heavy; OI ratio read 0.79 long-heavy. The cap was reading the wrong one.

**Fix — credit the opposing side what it funds out of its own losses.** Over an adverse move of `x = 1/M`:

```
this side's profit   ≈ x · OI(side)
opposing side's loss ≈ x · OI(opp), capped at R(opp) — they liquidate, and
                       that margin is forfeited INTO the pool that pays us
net drain            ≈ x · OI(side) − min(x · OI(opp), R(opp))
```

Requiring `net drain ≤ availableCover` and multiplying by `M`:

```
OI(side) ≤ M · availableCover + min(OI(opp), M · R(opp))
```

The old rule is that with the second term set to zero. `R` is `min(costBasis, positionValue)` — deliberately the *same* reserved quantity `availableCover` deducts, not raw cost basis, which is what makes the terms telescope to at most `M × opposing pool`. (Raw cb has no such bound: on a wedged book `Σ cb` exceeds the pool, which is failure #1.) A final `min` enforces that ceiling directly for books whose cover has already gone negative.

BTC under the new rule: limit 735,477 → 1,070,921, long headroom 1,129 → ~336,572. The `min` is what stops a thin-margin book granting unbounded capacity — 929k of short notional on 34k of reserved margin credits 335k, not 929k.

Strictly a loosening. With no opposing positions it's identical to today, so **no market gets tighter.**

## Explicitly not in this PR

**Reserving at full cost basis.** That's the fix that stops principal being spent against an opposing side's unrealized loss *in the first place*, rather than letting the pot honour the resulting claim from its own surplus. It's the honest fix and it's what makes "markets without growing subsidy" reachable — but it materially lowers profit capacity on every live market, so it changes the economics of things people already have open. It deserves its own PR and its own argument. Documented in the README section so the decision is on the record rather than forgotten.

Also not here: funding is levied on margin while the imbalance it corrects is in notional, which under-prices high-leverage crowding by roughly the leverage ratio. Separate issue. (Funding keying and sign are both *correct* — I checked: keyed on OI at `engine.ts:1960`, and BTC's 390k-long-vs-51k-short book was correctly paying a positive rate.)

## Testing

- `common` — 769 pass (36 suites), including 8 new `applyADL` cross-side-transfer cases and 4 new capacity cases.
- `backend/shared` — 29 pass (4 suites). Run explicitly because its ts-jest transpiles imported `common/src` under ES2017 and can fail where `tsc -b` passes.
- `tsc` clean: `backend/api -b`, `backend/scheduler -b`, `web --noEmit`.

Three existing capacity assertions moved deliberately — they encode the old limit — and are updated in place with the reasoning. Two guard tests are worth a look on review: *"is load-bearing: the book is unrepresentable without it"* pins the premise of the whole transfer block, and *"never promises more than the whole opposing pool"* pins the ceiling.

## Deploy

No migration — `solvencyHaltTime`/`solvencyHaltReason` are `contracts.data` jsonb fields, absent on every existing row and read as "not halted".

API and scheduler both carry engine changes, so ship them on the **same SHA**. Order doesn't matter: the capacity change is a pure loosening, and the halt fields are only written by the scheduler and only read by the API.

The OpenRouter market has already self-healed (mark advanced 8/29 23:50Z, pools L=34,470 / S=40,239, long headroom +3,985), so nothing here is racing a live incident.

### DIFF
--- a/backend/scheduler/src/jobs/update-perps.ts
+++ b/backend/scheduler/src/jobs/update-perps.ts
@@ -107,6 +107,16 @@ const updateOnePerp = async (contract: PerpContract) => {
       latest.ts,
       latest.sourceTs
     )
+    if (oracleResult?.solvencyHalt) {
+      // Price committed, state did not, trading halted. Funding would only
+      // re-derive the same failure (applyFundingWithSolvency asserts solvency
+      // on its INPUT), so stop here and page once rather than twice.
+      log.error(
+        `[update-perps] ${contract.slug}: oracle price applied but state is not solvent, trading halted and funding skipped: ${oracleResult.solvencyHalt.reason}`
+      )
+      return
+    }
+
     if (oracleResult) {
       // Best-effort, like apply-oracle-point does. This only touches the DB
       // when liquidations or ADL occurred, i.e. exactly the runs whose
--- a/backend/shared/src/perps/README.md
+++ b/backend/shared/src/perps/README.md
@@ -261,13 +261,18 @@ from somewhere other than the absent side.
 
 ## Exposure capacity
 
-Opening, adding, or flipping into a side is limited so aggregate open notional
-on that side cannot exceed 10× its unreserved opposing-pool cover. Unreserved
-cover is the opposing pool minus each opposite-side position's currently
-refundable value (capped at its cost basis), matching the reserve definition
-used by ADL solvency:
+Opening, adding, or flipping into a side is limited by what the opposing side
+can back. That is two things: its unreserved pool cover, plus the notional it
+already has at risk and funds out of its own losses. Unreserved cover is the
+opposing pool minus each opposite-side position's currently refundable value
+(capped at its cost basis), matching the reserve definition used by ADL
+solvency:
 
-`side open interest <= 10 × max(opposing pool - opposing reserves, 0)`
+```
+matched credit = min(opposing OI, 10 × opposing reserves)
+side OI       <= min(10 × max(opposing pool - opposing reserves, 0) + matched credit,
+                     10 × opposing pool)
+```
 
 This is a launch guardrail in addition to the ManiPerp paper. It preserves high
 leverage for small positions while preventing a new, initially flat position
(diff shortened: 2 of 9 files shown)


todwadd:
All three P1s confirmed and fixed in `964dfdd90`. I verified each against the code rather than taking them on trust; #2 in particular is a regression this PR introduced, and I'd missed that the throw was previously what protected funding.

**1. `matchedCredit` — confirmed, counterexample reproduces.**

The defect is that `R = min(costBasis, positionValue)` is **flat in price wherever the opposing position is in profit** — value exceeds basis, the `min` clamps to basis, and the modelled move releases nothing. My derivation assumed the opposing side always gives up `min(x·OI, R)`, which is only true once they are at or under water. Your short at entry 200 against a mark of 100 is 500 up; at 110 it is still 450 up, still reserves the entire `S=100`, and the admitted long has zero cover.

Credit is now the reserve the move actually releases:

```
released = Σ_opp max( min(cb, value(P)) − min(cb, value(P')), 0 )
matchedCredit = min(OI(opp), M · released)
```

`P'` is `P·(1 ± 1/M)` in the direction adverse to the opposing side, and the per-position `max(…, 0)` stops one holder whose reserve does not move from being netted against another's release. On your case `released = 0`, so the limit is 0 and the trade is refused as before.

Two things worth noting for re-review. `released ≤ x·OI(opp)` always, so `M·released ≤ OI(opp)` and the explicit `min` against opposing OI is now belt-and-braces against float rather than the binding constraint; and `released ≤ R(P)`, so the telescoping to `≤ M · opposing pool` still holds. **Every pre-existing capacity assertion is numerically unchanged** — the old form only differed where the opponent was in profit, which is exactly the hole you found. Regression test added, plus a second assertion that the same short *does* earn credit once the mark puts it underwater.

**2. Funding on a halted book — confirmed, and it is a regression from this PR.**

Traced it exactly as you describe: the halt commits the price, so replaying that point is a `duplicate`, `runOracleUpdate` returns `null`, my per-run check never fires, and `runFunding` persists pools, positions and events against a book whose liquidation and ADL are still pending. Pre-PR the tick threw and the throw took funding with it — committing the price removed that protection and I did not replace it.

Both halves of your fix are in:

- `runFunding` now refuses a persisted halt, read from the locked contract rather than from the caller's oracle result.
- A halted contract re-applies the **exact same** point. This turned out to be load-bearing beyond recovery ordering: without it the halt is *terminal*, because the committed price makes every retry a duplicate, so the pending transition is never applied and no amount of subsidy clears the flag. Restricted to `reason === 'duplicate'`; a stale (older) point stays ignored. The original halt timestamp is preserved across retries so it reads as time-wedged rather than time-since-last-retry.

**3. Deploy order — correct, my "order doesn't matter" was wrong.**

A new scheduler writing a fresh mark plus `solvencyHaltTime` while an old API ignores the field and passes freshness is exactly the hole, and the open path has no independent global-solvency check to catch it. The README now says: API first, fully rolled out and drained, then the perps scheduler. The new API is inert until a scheduler writes the first halt, so leading with it costs nothing. Added your rollback constraint too — do not roll the API back while halt rows exist; clear them or roll the scheduler back with it.

**Non-blocking, both done:**

- The halt now folds into the existing `oracleTradingPaused` flag in `perp-overview`, which the bet and position panels already gate their controls on, with its own banner copy. Controls disable rather than waiting for the 503.
- The ULP case is worth more than "rare": the `-Infinity` branch of `solvencyFactor` has **no tolerance**, so one ULP of negative cover on a side with no profit left is the difference between unwedging and staying wedged. The transfer now assigns the recipient its reserve exactly instead of computing `L + (CL − L)`; the donor absorbs the same value, so escrow moves by at most one ULP, which `assertPerpEscrowBalance` tolerates.

Green after the fixes: common 770 (36 suites), backend/shared 29 (4 suites), `tsc` clean on api, scheduler and web.

PR #4033, opened 2026-09-02
feat(perps): protected-basis settlement (Workstream A), dormant behind accounting modes

todwadd:
Implements Workstream A of the ManiPerp protected-basis settlement plan. **Based on #4030's branch** so the diff is only this change; retarget to `main` after #4030 merges (until then GitHub runs only the Vercel checks, and the file list includes main-only changes the base branch predates). Nothing here activates on any existing contract: every market stays on legacy #4030 semantics until moved through the guarded migration path, and no migration, subsidy, or deployment has been run.

Revision history: **2** answered the first external (codex) review; **3** fixed 18 findings of an adversarial self-review; **4** answered the second external review; **5** fixed 10 findings of the self-review of revision 4. The response comments on this PR map each external finding to its fix. Draft stays draft; activation stays disabled.

## What it does

Each position gains a protected basis `b` alongside `c` (`0 <= b <= c`). `c` keeps every existing role (price PnL, value, leverage, liquidation, fees, entry economics); `b` is the part of value the position's own side pool still protects.

```
V = max(c + π, 0)   R = min(b, V)   E = max(V − b, 0)   D = max(b − V, 0)
invariants: 0 <= b <= c,  B >= Σb,  E_long <= D_short + H_short,  E_short <= D_long + H_long
```

- **Loss-first settlement.** When a realized opposing payout `W` is paid, the paying side's paper losses are consumed first (`delta = min(W, D)`, pro rata, deterministic residual, never below current value); only `W − delta` comes out of unreserved balance. Affected rows keep `q, c, Pe`, leverage, liquidation and value, and each gets a `basis-settlement` receipt event (not a cash flow).
- **Generalized claim ADL** against `E`, with `q' = s·q, c' = b + s·(c − b), b' = b`; reduces to the paper's transform at `b = c`. Factor zero pays `b` once and removes the row. The factor is representability-aware: an allowance below dust snaps to 0, and the scaled side is re-checked before the tick commits.
- **One tolerance, and it never authorizes cash.** Every backing check, pool debit and the `B >= Σb` check use `perpDustTolerance`: 256 ulps of the largest basis-sized quantity involved, floored at a millionth of a mana and capped at a ten-thousandth, no separate relative term, so no check admits slack a later check at a smaller scale would refuse. A value within dust of `b` is snapped onto `b`. **Every payout equals the pool debit that funds it**: a claim the pool cannot fully pay is trimmed by the dust the pool lacks (the closing user; the canonically last claimant in claim ADL and resolution), a contingent payment is capped at the paying pool's reserves that survive the settlement, and a pool never goes below zero — so the ledger-vs-pools escrow check cannot drift by tolerated slack that a later, smaller book cannot absorb.
- **Unified full/partial close** (`z·R` from own pool, `z·E` from opposing after settlement; survivor scales by `1 − z`). Flip closes first, then validates the new leg. Partial close is API-only (`fraction`), rejected on legacy contracts, and must be at least `PERP_MIN_CLOSE_FRACTION` (1%) and actually change the position.
- **Funding** scales `b` with `q, c`; **resolution** is one deterministic batch from one immutable state.
- **Cross-side transfer** is unreachable when `B >= Σb`; in protected mode a state that would need one halts (reason names the invariant) instead of moving pool balance.
- **Compatibility admission** is unchanged: `matchedCredit` with its opposing-OI cap, `M = U = 10`, reserves generalized to `min(b, V)`. Every #4030 capacity fixture passes unchanged (`amm.test.ts` byte-identical), and a test pins that `b < c` changes cover and credit exactly as `min(b, V)` implies.

## Versioning and the database guard

- `contracts.data.perpAccountingMode` (`legacy | shadow | protected`, absent = legacy, unknown fails closed) and `perpAccountingEpoch`, read under the contract lock; the engine presents the epoch to the guard once at state load for protected contracts, so its own price, halt and pool patches pass.
- Migration `2026090201_perp_protected_basis.sql` (additive): `reserve_basis` + epoch stamps on positions, `reserve_basis_delta` + mode/epoch stamps on events, two new event types, append-only `contract_perp_accounting_epochs`, isolated `contract_perp_shadow_checkpoints` and `contract_perp_risk_shadow`, and two triggers:
  - `perp_accounting_guard()` on positions/events/funding events: for legacy/shadow contracts it **mirrors** `b = c` / `Δb = Δc` whatever the writer sent; for protected contracts it **requires** the transaction to have run `set_config('perp.accounting_epoch', <epoch>, true)` and every row to carry that epoch, else it raises and the whole transaction rolls back. Perp rows are never re-keyed to another contract.
  - `perp_accounting_contract_guard()` on `contracts`: a mode/epoch change needs the transition setting, an epoch of exactly old+1 and an immutable record naming the mode being left; **any change to a protected contract's financial fields — `poolLong`/`poolShort`, `oraclePrice`, its timestamps, `solvencyHaltTime`/`solvencyHaltReason` — needs the epoch too**, so an old binary can neither commit a legacy cross-side transfer through the scheduler's no-change fast path nor commit a price-only tick on which protected accounting would have settled a position (which would leave an unhalted, protected-invalid book the version-aware scheduler then skips as a duplicate). Its whole tick rolls back instead.
- `backend/scripts/sql/perp-accounting-guard-test.sql` drives the old-writer fixture against a protected contract: position/event/funding writes, pool, price-only, halt and `oracleSourceTime` patches with no / stale / current epoch, the activation ordering, epoch-skip and wrong-`previous_mode` flips, a re-key, and a protected → legacy downgrade. The fixture populates the columns the checked-in `contract_populate_cols` trigger requires. It passes on a scratch Postgres 16 after the existing perp migrations plus this one.
- **Shadow mode** commits legacy ledgers and, alongside, advances an isolated protected-basis checkpoint through the protected counterpart of every committed transition (open/add/flip, close, oracle tick — a halted tick included, sampled once per point against the unchanged live state — funding, subsidy, resolution), comparing pools, rows, invariants, close payouts and the resolution residual with the live state; a material payout or residual difference counts as a divergence. Reports are stored with the checkpoint, never in a financial row; every statement including the checkpoint read runs under a savepoint and errors are swallowed. A failed write is logged as a gap in the epoch's replay and repaired with `--reseed-shadow`. The preflight prints the checkpoint counters and the last report (`common/src/perps/accounting-shadow.ts`, `backend/shared/src/perps/accounting-shadow.ts`).
- Transitions (`legacy → shadow → protected`, verifier-gated return to legacy), the read-only Stage 0 simulator, and activation with `--dry-run` (per-user `b`, reductions, dust trims, no writes) live in `backend/shared/src/perps/accounting.ts`, driven only by `backend/scripts/perp-protected-basis-preflight.ts`. A **reducing** activation (last-resort allocation, activation ADL, any row left below `c`) and any activation on a **stale mark** require `--confirm-plan=<digest from the dry run>`: a digest of the whole plan — book (mark, pools, every row's `q, c, b, Pe`), inputs (top-ups, allocation policy, ADL approval) and outcomes (each row's `b` before and after, trims, final pools, ADL factors) — recomputed under the lock (`perpActivationPlanDigest`, `perpActivationConfirmation`). A different top-up on the same book, or a legacy trade at an identical mark, aborts it. Activation trims a dust reserve shortfall from the canonically last row's `b` so `B >= Σb` holds exactly in the committed state (asserted without tolerance before commit) and writes an immutable `basis-settlement` receipt per reduced position.

## Workstream B (shadow only) and C (not implemented)

- `perpRiskPolicyMode` (`off | shadow`; `enforce` rejected at schema and runtime) is independent of the accounting mode. In shadow, every open attempt (accepted or compat-rejected) is scored against candidate `U = 1` and the exact stress rule, and every oracle/funding tick projects ADL for `alpha ∈ {1, 0.5, 0.1}`; the latest evaluation goes to `contract_perp_risk_shadow`, outside the event log, latest-wins by the evaluation's own timestamp (a rejected attempt's transaction rolls back, so its row is written best-effort through a detached connection). Evaluator failures are logged and ignored; nothing can change an admission, position, pool, event or payout.
- No post-exit capacity deleveraging: no forced-close path, event, schema, notification, UI, or test. The five-shorts/M$5m-long scenario test asserts the incumbent keeps `q, c, Pe` and value when counterparties leave.

## Revisions 4 and 5

**Revision 4** (second external review): payouts equal pool debits with dust trimmed from the claimant instead of snapping the pool (codex's organic M$20m sequence and the flat-M$1 activation are tests with the escrow check after every step); the contract guard covers mark and halt fields; both shadows sample the halted tick; the plan digest replaces the bare mark/book fingerprint; the checkpoint select returns `updated_time` (the preflight crashed on it after shadow activation); the guard fixture carries `uniqueBettorCount`; payout/residual differences count as divergences; and the three hardening items (epoch +1 with `previous_mode`, re-key refusal, latest-wins telemetry).

**Revision 5** (adversarial self-review of revision 4: 6 lenses, 3 refuters per finding, 52 agents; 21 raw → 15 distinct → 10 confirmed, all fixed): a phantom float remainder made `trimPayouts` fail closed on a pool at exactly 0 with several dust claimants (P2; would have wedged resolution); the contingent payment is now capped at the paying pool's surviving reserves so a claim admitted at M$2bn scale cannot be refused by the post-close check at the short side's own scale; the activation trim re-sums until `Σb <= pool` in float and the live activation asserts it exactly; the downgrade verifier uses the same dust tolerance as the receipts; the confirmation gate is a pure, tested helper; the claim-ADL trim, `oracleSourceTime` and the protected → legacy downgrade are pinned; the migration header, dry-run comment, CLI hint, README and addendum describe the current behaviour. Five findings were refuted.

Not done: a database-backed engine test of the halt-branch shadow sampling (the repo has no Postgres test harness for the engine; the branch is pinned by code reading and the pure counterpart tests).

## Review focus

1. `common/src/perps/protected-basis.ts` — the tolerance rule, dead band, `debitPool`/`trimPayouts`, the reserve-floor cap in `payPerpContingentClaim`, settlement allocation, claim ADL, close, resolution.
2. `backend/shared/src/perps/engine.ts` — mode branching in open/flip, close (incl. partial + idempotent replay), `applyOracleUpdate`, the halt branch of `runOracleUpdate`, `runFunding`, `resolvePerp`; placement of the shadow hooks.
3. `backend/supabase/migrations/2026090201_perp_protected_basis.sql` — both guards, the `contracts` WHEN clause, backfill.
4. `backend/shared/src/perps/accounting.ts` and `common/src/perps/protected-migration.ts` — activation ordering (epoch record → present new epoch → transition setting → rewrite rows → patch → receipts/payouts → metrics), the dust trim, plan digest and confirmation gate, downgrade verifier.
5. `common/src/perps/accounting-shadow.ts` — the protected counterparts and the comparison.
6. `common/src/perps/metric-periods.ts` — replay of `basis-settlement`, partial close, ADL cost-basis change.

## Verification

| Check | Result |
| --- | --- |
| `yarn test:common` | 41 suites, 892 tests pass (`amm.test.ts` byte-identical to #4030) |
| `yarn test:shared` | 6 suites, 47 tests pass |
| Guard SQL test on scratch Postgres 16 (migration re-applied) | PASS |
| `yarn typecheck:api` / `:scheduler` / `:web` | clean |
| `yarn lint:common` / `:shared` / `:api`, `yarn format:check` | clean; no `any` on any added line |
| Adversarial multi-agent reviews (refutation-voted) | rev 3: 18 confirmed, fixed; rev 5: 10 confirmed, fixed |

Not run: the simulator/preflight script against a real database (none reachable from the session). `yarn install` was done without the `@flat/in-app-purchase` GitHub tarball (blocked egress); the lockfile is untouched.

## Docs

`backend/shared/src/perps/README.md` ("Protected-basis settlement"), `perps-launch-runbook.md` ("Protected-basis accounting": deploy order API → drain → scheduler, migration decision per contract, shadow soak, dry run and plan digest, activation with feeds running, rollback boundary), `dev-docs/maniperp-protected-basis-addendum.md` (versioned addendum draft; the PDF is untouched, Stephen's source and sign-off remain external prerequisites).

https://claude.ai/code/session_014DLtpdLRusRJmWvpEWGEQe


todwadd:
Response to the external (codex) review. Every item below is on the branch; verification numbers are in the PR body.

**P1-1 — old-writer bypass through contract-only pool patches.** Fixed at the database. `perp_accounting_contract_guard()` (replacing the transition-only trigger) also fires on any change to `poolLong`/`poolShort` when the old or new mode is `protected`, and requires `perp.accounting_epoch` to equal the contract's epoch — the NEW epoch during activation. A stale scheduler fast path, or any binary that never sets the GUC, can no longer commit a legacy cross-side transfer through the contract row. The engine presents the epoch once at state load under the lock, so its own patches are unaffected; `activatePerpAccountingProtected` presents the new epoch before the transition GUC and the contract patch. `perp-accounting-guard-test.sql` sections 5–6 cover: pool change without epoch refused, with a stale epoch refused, with the current epoch allowed, price/halt patches allowed, legacy contracts unaffected, activation patch refused unless the new epoch is presented first.

**P1-2 — invariant tolerance vs payment tolerance.** One rule now, and (after the self-review below) a stricter one than the first fix: `perpDustTolerance` — 256 ulps of the largest basis-sized quantity, floored at a millionth of a mana, capped at a ten-thousandth, no relative term — backs the invariant, the contingent payment, claim ADL, close, resolution and every pool debit. The regression you gave (pool `4,999,999.9999995` vs a `5,000,000` claim) is a test: the invariant accepts, and full close, half close and resolution all pay and leave the pool at exactly 0. The D_opp > 0 variant is pinned in both directions (inside dust: invariant, close and resolution all accept; outside: all refuse).

**P1-3 — tiny positive allowance halting the market.** Claim ADL is representability-aware: an allowance below dust snaps the factor to 0 and pays `b`; a value within dust of `b` is not a contingent claim at all (dead band in `getPerpPositionClaims`), so it is neither settled at factor 0 nor counted against a side-level check; the scaled side is re-checked with the shared rule. Your `H = 5e-10` against a M$1m basis runs the whole oracle transition without throwing.

**Activation gap A — cumulative shadow validation.** Shadow keeps an isolated forward checkpoint (`contract_perp_shadow_checkpoints`; pure logic in `common/perps/accounting-shadow.ts`) seeded from live with `b = c` and advanced by the protected counterpart of every committed transition (open/add/flip, close incl. partial, oracle tick, funding, subsidy, resolution), then compared with the live state — pools, rows, both invariant sets, the user payout on closes and the creator residual on resolution. Reports live with the checkpoint, never in a financial row; everything runs under a savepoint and every error is swallowed; a write failure is logged as a gap and repaired with `--reseed-shadow`. The five-shorts scenario shows the divergence you would expect (shadow `b = 4.95m`, live `b = c`).

**Activation gap B — last-resort migration.** `dryRunPerpAccountingProtected` (`--dry-run`) prints each position's `b` and reduction on the current book without writing, plus a fingerprint of mark, pools and every live row's `q, c, b`. A reducing activation (last-resort allocation or activation ADL) must repeat it as `--confirm-fingerprint=<digest>` or it is refused — a legacy trade at an identical mark aborts it — and writes one immutable `basis-settlement` receipt per reduced position (`trigger: 'activation'`, `cutoverMark`, epoch).

**P2 — replay, dust closes, telemetry.** `parseStoredPosition` parses `reserveBasis` (validated `0 <= b <= c`; absent means legacy mirror) — tested. Partial closes below `PERP_MIN_CLOSE_FRACTION = 0.01`, or that would not change the survivor or pay more than dust, are rejected in the pure model, the engine and the API schema (all three tested), so no event or streak can advance without a change to the position. Workstream B telemetry moved out of events into `contract_perp_risk_shadow` and samples accepted opens, compat-rejected opens (written best-effort through a detached connection, since the rejecting transaction rolls back) and every oracle/funding tick.

**Self-review (revision 3).** An adversarial multi-agent pass over the whole diff (9 lenses, 3 refuters per finding, completeness probes) confirmed 18 further findings, all fixed — most importantly a scale-drift variant of your P1-2: residues admitted at M$5m scale were refused once the book shrank to M$2 (a whale's close after claim ADL returned a 500 on ~80% of ADL-active ticks; a factor-0 settlement on a fresh book halted it and made it unresolvable). That is what motivated the absolute floor above. The PR body lists the rest (shadow savepoint and resolve counterpart, `b <= c` clamp in the last-resort allocation, the fingerprint replacing a bare confirmed mark, a mode-aware launch preflight, generated-SQL and doc mismatches, portfolio receipt rendering, test gaps).

PR stays in draft and activation stays disabled, as you recommended. #4030 is still open, so this remains based on its branch; I will retarget to `main` once it merges.

---

todwadd:
Response to the second external (codex) review. Everything below is on the branch (revisions 4 and 5); the PR body carries the verification numbers.

**P1 — tolerated shortfalls can still wedge exits.** Agreed, and fixed at the principle you stated: tolerance classifies dust, it never authorizes cash. `debitPool` now returns what the pool actually gave up and every payout equals that debit — the closing user absorbs a dust shortfall on a close; in claim ADL and resolution the trim walks canonically backward from the last claimant; a pool never goes below zero; a larger shortfall fails closed as before. A contingent payment is additionally capped at the paying pool's reserves that survive the loss-first settlement, so the paying side lands on `B >= Σb` in real arithmetic whatever scale admitted the claim. Activation trims a dust reserve shortfall from the canonically last row's `b` (re-summed until `Σb <= pool` in float) and the live activation asserts the exact, non-tolerant reserve floor before committing. Your organic M$20m sequence (whale slightly underwater + M$1 long, exactly backed; tick lifts the small row 1.05e-6; small close; whale close) and the flat-M$1 `L = S = 0.99999925` activation are tests with `isPerpEscrowBalanced` asserted after every step, plus a property test that closes every position sequentially on random valid books and resolves each in batch with the escrow predicate after each step. These are pure-model sequential-close and batch-resolution regressions rather than engine-level ones: the repo has no Postgres harness for the engine, and the engine pays exactly the pure result's `payout` / settled payouts / residual, which is what the tests pin.

**P1 — the old-writer guard permits a protected-invalid tick.** Fixed as you proposed. On a contract that is or becomes protected, `oraclePrice`, `oraclePriceTime`, `oracleSourceTime`, `solvencyHaltTime` and `solvencyHaltReason` now require `perp.accounting_epoch` like the pools (function body and WHEN clause; `contracts.sql` regenerated). The engine presents the epoch at state load, so its own price/halt/pool patches pass; an old scheduler's price-only fast path and halt patch roll back its whole tick. The guard test now asserts the price-only write, the halt patch and `oracleSourceTime` alone are refused without the epoch and accepted with it.

**P1 — shadow skips the halt path.** Both shadows are advanced in the halt branch against the unchanged live financial state before returning, once per point (a halted retry of the same duplicate point is the same transition). The five-shorts recovery tick is therefore recorded as the divergence it is.

**P1 — the fingerprint does not confirm the plan.** Replaced by `perpActivationPlanDigest`: the book (mark, pools, every row's `q, c, b`, and now `entryPrice`), every plan input (top-ups, allocation policy, `allowActivationAdl`) and every outcome (each row's `b` before and after, trims, final pools, ADL factors/counts, blockers). The live run recomputes the plan under the lock and compares digests; required for any reducing activation (last-resort allocation, activation ADL applied, any row left below `c`) and, as the runbook says, whenever `allowStaleMark` is set. Your reproduction (top-up M$50 → haircut 50 vs top-up 0 → haircut 100 on the same book) is a test asserting different digests; the gate itself is a pure helper (`perpActivationConfirmation`) with its own tests.

**P1 — preflight crashes after entering shadow.** `selectShadowCheckpointQuery` now selects `updated_time`; a query test asserts every column the simulator and preflight read is selected. The preflight also prints the last shadow report.

**Additional fixes.** The guard fixture carries `uniqueBettorCount` (and the test passes against the migration on a scratch Postgres 16; note it never ran against the checked-in `contracts.sql` populate trigger here, which is why this slipped). A payout or residual difference beyond dust now counts as a divergence and is logged and counted. Hardening, all three: transitions must advance the epoch by exactly one with a record whose `previous_mode` is the mode being left (a downgrade through that rule is tested); perp rows are never re-keyed to another contract; `contract_perp_risk_shadow` upserts are latest-wins by the evaluation's own `at`, so a late detached write for a rejected attempt cannot overwrite a newer evaluation.

**Self-review of revision 4.** A further adversarial pass confirmed ten more findings, all fixed in revision 5. The one worth flagging: with the new trimming, a pool landing on exactly 0 with several surviving dust claimants produced a phantom float remainder (~1e-23) that made resolution fail closed; `trimPayouts` now classifies a remainder that survives only because every leg was consumed as rounding. The others are the reserve-floor cap above, the exact trim re-sum, the downgrade verifier using the same dust tolerance as the receipts (a dust-trimmed contract could otherwise never return to legacy), the pure confirmation gate, and doc/test coverage (claim-ADL trim, `oracleSourceTime`, downgrade flip, migration header, addendum).

PR stays in draft; activation stays disabled. #4030 is still open, so this remains stacked on it.

---

PR #4035, opened 2026-09-04
feat(perps): partial closes, with the UI to use them

todwadd:
Closing a perp is all-or-nothing today, so a trader who wants *less* exposure has to close the whole position and re-enter.

The exit itself is already free — the taker fee is charged in full at open (`engine.ts`, "No fee on close") — so the cost isn't the exit. It's the **re-entry**: a fresh fee on the *whole* new notional including the cubic size-impact term `(impact/3)·P·(s₁³ − s₀³)`, plus a reset entry price, a moved liquidation price, and a reset `openedTime`. Partial closes remove that round trip.

`close-perp-position` now takes an optional `fraction` in `(0, 1]`. Omitting it closes the whole position exactly as before.

## Why this is a small change

**The exit math is linear in the position.** π is linear in `q`, and both the payout and the two pool debits are linear in π and `c`. So closing `z` of a row pays exactly `z` times what the full close would have paid, debits exactly `z` of each pool delta, and leaves the book in precisely the state it would have been in had the trader opened `1 − z` of that position and closed a separate `z` one.

That is why the survivor keeps its **entry price**, its **leverage** (`ℓ = q/c` is invariant under the split) and therefore its **liquidation price**. Reducing exposure never moves the price at which what is left gets liquidated. `openedTime` is preserved too, so `expectedOpenedTime` still matches on the next close.

Everything else runs unchanged. Solvency is asserted with the same `assertPerpStateSolvent`; the escrow check brackets the payout as before; and a partial close only ever *reduces* open interest, so the capacity rule cannot reject one. The flip and resolution call sites pass no fraction and are byte-identical.

## Rules that keep the split honest

- **The halves come from subtraction, not from scaling with `z` and `1 − z`.** They must add back up to the original *exactly*, because `metric-periods.ts` rebuilds the pre-close row by adding the event's deltas onto the survivor — a split that doesn't conserve leaves that replay drifting on every partial close. Pinned with `toBe`, not `toBeCloseTo`.
- **`PERP_MIN_CLOSE_FRACTION` (1%)** — the smallest close worth its own event. Below it the trade mints an event and a streak for a change the trader cannot see.
- **`PERP_MIN_REMAINDER_COST_BASIS`** promotes a close whose *remainder* would be dust into a full close, so `fraction = 0.9999` cannot strand a row of a few thousandths of a mana that still accrues funding every period. The bound is on the remaining cost basis rather than on `1 − fraction`, because that's the quantity that has to stay meaningful: a fractional rule would strand dust on a large position and delete real margin on a small one. The response's `fraction` is what *actually* happened and can exceed what was asked for.
- **`expectedSize` is required for a partial close**, and checked against the locked row inside the transaction. `expectedOpenedTime` cannot police this: the survivor of a partial close keeps its `openedTime`, so a request sized against the pre-close row passes that check and then applies its fraction to the smaller survivor. Two tabs both showing 400 and both sending 75% — the first leaves 100, the second takes 75 of *that* rather than the 300 it displayed, silently. Funding scaling the row and an ADL haircut invalidate the preview identically, so the tolerance (`PERP_EXPECTED_SIZE_TOLERANCE`) is representation dust only and all three cases 409. **Full closes skip the check** — "close everything" is unambiguous at any size, and a spurious 409 on the one operation a trader may urgently need is a worse failure than any it would prevent.

## Two client bugs, both only reachable because a partial close preserves `openedTime`

Both would have failed silently:

1. **The idempotency fingerprint** was `[contractId, direction, openedTime]`. Two 25% closes in a row are separate trades but produced the same fingerprint, so the second would have replayed the first and quietly closed nothing. It now carries the position's `size` and the fraction — `size` moves after every close, so a retry of the *same* click still dedupes. (This protects one client; `expectedSize` above is what protects against two.)
2. **The optimistic hide.** The panel marks a direction closed and filters rows with `openedTime < closedAt`. On a partial close that hides a position that is still open — and since `openedTime` never changes again, hides it *permanently*. Only a full close sets it now.

## Rate limiting

Partial closes are limited to 60/hour per user. **Full closes are never limited** — refusing someone the ability to exit a leveraged position is far worse than tolerating close spam — so the limiter is reachable only when the caller asked for a fraction.

The limit exists because the 1% minimum is a fraction of the *current remainder*, so repeated 1% closes decay geometrically rather than terminating: `0.99^100` still leaves 36.6% open, and reaching the dust floor takes ~917 calls from an M$100 cost basis and ~1375 from M$10,000. (An earlier revision of this description claimed ~100; that was wrong.) Each call is a serializable transaction under the contract's advisory lock plus an event, a txn, a metrics rebuild and streak processing — the same lock the 2s oracle tick wants.

## UI

An unboxed holdings summary mirrors ordinary market pages: Position, Margin, Unrealized P&L and Value use the site's label/value typography and info tooltips, with entry, mark and liquidation details in a quieter row underneath. Its compact outlined **Close** action opens the same responsive modal / mobile bottom-sheet flow used by ordinary sells elsewhere on the site. The sheet defaults to **%**, with a small **% / Mana** toggle beside Close amount, a full-width field, the site's familiar ±5 / ±1 adjustments, one slider and **Max** in both modes. Mana means **estimated amount returned**, not leveraged notional or margin. Entering an amount sizes a fraction at the current quote; switching modes preserves that unrounded fraction, and later oracle updates change the payout preview rather than silently resizing the close. The 1% minimum and remainder-dust rule still apply. Mana entry is unavailable for non-positive/non-finite payouts, while percentage closes remain available. An out-of-range or empty value disables the confirmation and says why; a value whose remainder would be dust says so before submission rather than surprising the user with a full close afterwards.

The sheet previews payout, realized P&L and remaining margin at the displayed oracle price. The split is linear at a fixed price, but **the final payout can change with the price before execution**, so the UI labels it Estimated payout.

Partial closes are labelled as such in the position history, the market's Trades tab, the `/perps` hub tape, the unified discovery feed and the balance-change feed. Every percentage goes through one shared `formatPerpClosePercent`, which keeps two decimals and **truncates rather than rounds** near 1, so a surviving position can never be labelled a 100% close.

## Verification

The full common/shared suites and all three typechecks were re-run locally after the percent/Mana toggle; changed files pass lint and formatting. CI was green on `4755f84`, and the new push triggers fresh checks:

| Check | Result |
| --- | --- |
| `yarn test:common` | 41 suites, 908 tests pass |
| `yarn test:shared` | 8 suites, 51 tests pass |
| `yarn typecheck:web` / `:api` / `:scheduler` | clean |
| `yarn lint`, `yarn format:check` | clean |

New tests cover the conservation property, two-step-equals-one-step payout, the state-a-smaller-position-would-have-been-in equivalence, the dust promotion at and either side of its bound, the expected-size tolerance at both edges, the close-percent formatter at ordinary, arbitrary and near-total fractions, short and underwater legs, and a guard that a *liquidation* leaving a row open is still rejected by the metric replay.

**UI verification:** 24 new amount-selection tests cover mana conversion, exact fraction preservation across repeated mode switches and quote changes, Max, the 1% boundary, invalid values, zero payout, near-total formatting and dust promotion. Manual localhost checks at 320px and 1280px cover both modes, a 500-mana preview, validation, the slider, adjustment buttons and Max; reopening defaults to percent. No trades were submitted during these UI checks.

**What is not covered:** the database-backed close transaction is not exercised by an integration harness. A dev end-to-end check should still verify a partial close's payout, surviving position and history, a subsequent full close, and stale-position rejection.

`yarn install` 403s on the `@flat/in-app-purchase` GitHub tarball from this environment (same blocker #4033 hit). I installed with that one dependency temporarily removed to get jest, then restored `backend/api/package.json` and `yarn.lock` byte-for-byte — both are untouched in this diff.

## Rollout

Deploy the partial-close API from this PR and wait for the rollout to complete **before exposing the new frontend**. The pre-PR API only supports full closes and its object schema discards the new fields, so a partial-close request sent to an old instance can close the entire position. Existing full-close clients remain compatible with the new API. If merging auto-publishes the frontend, coordinate or hold that publication until the API is ready.

## Other notes

- **`common/src/txn.ts`** gains `'partial-close'` on the `PERP_CLOSE_PAYOUT` reason union.
- **No migration.** `fraction` and `remainingSize` live in the event's `data` jsonb; closes written before this read as whole ones everywhere.

## Merged with main

`main` picked up **#4030** (jam-proof oracle tick, OI netting), **#4034** (five new oracle feeds) and **#4036** (job board) while this was open. Both perps PRs touch `common/src/perps/amm.ts`, and #4030 touches `backend/shared/src/perps/engine.ts` — the two files this branch changes most — so the merge was validated rather than trusted:

- One real conflict, in `amm.test.ts`: both sides appended a new top-level `describe` at the same point. Resolved by keeping both suites (`partial close` and `applyADL cross-side deficit transfer`); no test was dropped or altered.
- **#4030's `assertPerpNotSolvencyHalted` now gates partial closes**, as predicted — it sits in `closePosition` before `resolvePerpCloseFraction`, so a halted book refuses a partial close on the same path it refuses a full one.
- #4030's change near `closePosition` in `amm.ts` is the `PerpOpenInterestCapacity` type that follows it, not the close body, so the two edits are independent. Its new capacity-preview caller passes three arguments and so closes the opposite leg in full, which is the right semantics for a flip.

## Relationship to #4033

This should *shrink* **#4033**. That PR already carries a partial close, but protected-mode only (legacy hits an explicit `throw`) and API-only, with no UI. The schema here is its `closePerpPositionSchema` plus `expectedSize` and minus the protected-only restriction, so `market-types.ts`, `schema.ts`, `close-perp-position.ts` and `metric-periods.ts` largely resolve to "keep main's", and its `engine.ts` `else` branch gets a working legacy path instead of a `throw` — which it needs anyway, since a partial close is genuinely non-linear under protected-basis accounting (`z·R` from own pool, `z·E` from opposing, loss-first settlement) and the two paths have to branch on mode regardless. One import to reconcile: `PERP_MIN_CLOSE_FRACTION` lives in `common/perps/amm` here rather than in #4033's new `protected-basis.ts`.

Two things worth adopting there: the remainder-dust rule (#4033 guards that the closed side is material but not that the remainder is), and the `expectedSize` concurrency guard, which its partial close needs for the same reason this one does.

First of a planned series — trigger orders and stop-losses follow as separate PRs.

https://claude.ai/code/session_01KtD6jBeqNY5RLwwqMgXeEC

### DIFF
--- a/backend/api/src/close-perp-position.ts
+++ b/backend/api/src/close-perp-position.ts
@@ -1,10 +1,12 @@
+import { HOUR_MS } from 'common/util/time'
 import { closePosition } from 'shared/perps/engine'
 import { log } from 'shared/utils'
 import { APIHandler } from './helpers/endpoint'
 import { advancePerpBettingStreak } from './helpers/perp-streak'
 import { assertPerpCloseEnabled } from './helpers/perp-trading-mode'
+import { rateLimitByUser } from './helpers/rate-limit'
 
-export const closePerpPosition: APIHandler<'close-perp-position'> = async (
+const closePerpPositionInner: APIHandler<'close-perp-position'> = async (
   body,
   auth
 ) => {
@@ -15,6 +17,7 @@ export const closePerpPosition: APIHandler<'close-perp-position'> = async (
     idempotencyKey,
     expectedOpenedTime,
     fraction,
+    expectedSize,
   } = body
   const isApi = auth.creds.kind === 'key'
   const {
@@ -30,7 +33,8 @@ export const closePerpPosition: APIHandler<'close-perp-position'> = async (
     idempotencyKey,
     expectedOpenedTime,
     isApi,
-    fraction ?? 1
+    fraction ?? 1,
+    expectedSize
   )
 
   return {
@@ -48,3 +52,38 @@ export const closePerpPosition: APIHandler<'close-perp-position'> = async (
     },
   }
 }
+
+/**
+ * Partial closes only.
+ *
+ * The 1% minimum is a fraction of the CURRENT remainder, so repeated 1% closes
+ * decay geometrically rather than terminating in 100 calls: from an M$100 cost
+ * basis it takes ~917 to reach the dust floor that promotes the close to a
+ * full one, and more on a larger position. Each of those is a serializable
+ * transaction under the contract's advisory lock plus an event, a txn, a
+ * metrics rebuild and streak processing — the lock the 2s oracle tick also
+ * wants.
+ *
+ * 60/hour is far above any plausible human scaling-out and still caps the
+ * sustained rate at one per minute. In-memory per instance, like every other
+ * user of this helper.
+ */
+const rateLimitedPartialClose = rateLimitByUser(closePerpPositionInner, {
+  maxCalls: 60,
+  windowMs: HOUR_MS,
+})
+
+/**
+ * A FULL close is never rate-limited. Refusing someone the ability to exit a
+ * leveraged position — possibly while it is running toward liquidation — is a
+ * far worse failure than tolerating close spam, so the limiter above is
+ * reachable only when the caller asked for a fraction of one.
+ */
+export const closePerpPosition: APIHandler<'close-perp-position'> = async (
+  body,
+  auth,
+  req
+) =>
+  body.fraction !== undefined && body.fraction < 1
+    ? rateLimitedPartialClose(body, auth, req)
+    : closePerpPositionInner(body, auth, req)
--- a/backend/api/src/close-perp-position.ts
+++ b/backend/api/src/close-perp-position.ts
@@ -9,19 +9,32 @@ export const closePerpPosition: APIHandler<'close-perp-position'> = async (
   auth
 ) => {
   assertPerpCloseEnabled()
-  const { contractId, direction, idempotencyKey, expectedOpenedTime } = body
+  const {
+    contractId,
+    direction,
+    idempotencyKey,
+    expectedOpenedTime,
+    fraction,
+  } = body
   const isApi = auth.creds.kind === 'key'
-  const { payout, pnl, replayed } = await closePosition(
+  const {
+    payout,
+    pnl,
+    replayed,
+    fraction: closedFraction,
+    remainingSize,
+  } = await closePosition(
     contractId,
     auth.uid,
     direction,
     idempotencyKey,
     expectedOpenedTime,
-    isApi
+    isApi,
+    fraction ?? 1
   )
 
   return {
-    result: { payout, pnl },
+    result: { payout, pnl, fraction: closedFraction, remainingSize },
     continue: async () => {
       // Closes count toward the prediction streak, mirroring sells on normal
       // markets (sell-shares routes through onCreateBets). An idempotent
--- a/backend/api/src/get-balance-changes.ts
+++ b/backend/api/src/get-balance-changes.ts
@@ -90,6 +90,8 @@ const perpTxnDescription = (txn: Txn): string | undefined => {
         ? 'Flipped out of'
         : d.reason === 'resolve'
         ? 'Settled'
+        : d.reason === 'partial-close'
+        ? 'Partly closed'
         : 'Closed'
     return `${verb} ${d.direction} at ${px(d.closePrice)} — profit ${pnlText}`
   }
--- a/backend/api/src/get-perp-events.ts
+++ b/backend/api/src/get-perp-events.ts
@@ -56,9 +56,11 @@ export const getPerpEvents: APIHandler<'get-perp-events'> = async (body) => {
     const payoutRaw = data.payout
     const pnlRaw = data.pnl
     const adlFactorRaw = data.adlFactor
+    const fractionRaw = data.fraction
     const payout = payoutRaw == null ? null : Number(payoutRaw)
     const pnl = pnlRaw == null ? null : Number(pnlRaw)
     const adlFactor = adlFactorRaw == null ? null : Number(adlFactorRaw)
+    const fraction = fractionRaw == null ? null : Number(fractionRaw)
     return {
       id: Number(r.id),
       contractId: r.contract_id,
@@ -85,6 +87,13 @@ export const getPerpEvents: APIHandler<'get-perp-events'> = async (body) => {
         adlFactor >= 0 &&
         adlFactor <= 1
           ? adlFactor
+          : null,
+      fraction:
+        fraction != null &&
+        Number.isFinite(fraction) &&
+        fraction > 0 &&
+        fraction <= 1
+          ? fraction
           : null,
       isApi: data.isApi === true,
       userName: r.user_name,
--- a/backend/api/src/get-unified-feed.ts
+++ b/backend/api/src/get-unified-feed.ts
@@ -417,7 +417,8 @@ async function fetchSiteActivity(
       pe.applied_ts,
       pe.original_cost_basis_delta,
       pe.direction,
-      pe.leverage
+      pe.leverage,
+      pe.data->>'fraction' as fraction
     from contract_perp_events pe
     join contracts c on c.id = pe.contract_id
     where pe.user_id is not null
@@ -560,6 +561,7 @@ type PerpActivityRow = {
   original_cost_basis_delta: number | string
   direction: string | null
   leverage: number | string | null
+  fraction: number | string | null
 }
 
 const convertPerpActivityRow = (
@@ -597,6 +599,18 @@ const convertPerpActivityRow = (
   const leverage =
     parsedLeverage !== null && parsedLeverage >= 0 ? parsedLeverage : null
 
+  // Only a close carries one, and only a value strictly inside (0, 1) says
+  // anything the wording does not already: 1 is a whole close, and a close
+  // written before partial closes existed has none.
+  const parsedFraction = toFiniteNumber(row.fraction)
+  const fraction =
+    row.event_type === 'close' &&
+    parsedFraction !== null &&
+    parsedFraction > 0 &&
+    parsedFraction < 1
+      ? parsedFraction
+      : null
+
   return {
     id: String(row.id),
     contractId: row.contract_id,
@@ -606,6 +620,7 @@ const convertPerpActivityRow = (
     direction: row.direction,
     margin: Math.abs(originalMargin),
     leverage,
+    fraction,
   }
 }
 
--- a/backend/shared/src/perps/README.md
+++ b/backend/shared/src/perps/README.md
@@ -70,7 +70,8 @@ Endpoints are registered in `backend/api/src/routes.ts` and schemas live in
   New markets preserve their per-side initial backing for later preflight
   auditing.
 - `POST /place-perp-trade` — opens or adds to a position.
-- `POST /close-perp-position` — closes a position.
+- `POST /close-perp-position` — closes all of a position, or the
+  `fraction` of it given (see Partial closes).
 - `GET /get-perp-positions` — reads open positions for a contract (optionally
   filtered by `userId`).
 - `GET /get-oracle-price`, `/get-oracle-price-series` — read oracle data.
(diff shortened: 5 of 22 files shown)


todwadd (commented):
I found three issues that should be addressed:

1. **Bind the partial close to the position state the user saw.** `expectedOpenedTime` no longer detects an intervening partial close because the survivor preserves `openedTime`. For example, two clients can both show size 400 and submit 75%; the first leaves 100, then the second request is accepted against that survivor and closes 75 rather than the previewed 300. Please send and verify the expected size (or a mutation version) under the locked transaction, and include it in the idempotency request.

2. **Rate-limit partial closes while always allowing a full exit.** The 1% minimum is relative to the current remainder, so it does not bound an exit to ~100 calls. After 100 successive 1% closes, about 36.6% remains; a M$100 cost basis permits roughly 917 calls before the one-cent dust threshold. Each call creates accounting/event/metrics work and triggers streak processing. A limiter scoped to `fraction < 1` per user/contract would preserve emergency full exits.

3. **Propagate `fraction` through the unified activity feed.** The close event stores it, but `get-unified-feed` does not select the event data/fraction, `PerpTradeActivity` cannot carry it, and `PerpTradeLog` renders every `close` with full-close wording. Partial exits therefore remain indistinguishable from full closes in the main discovery feed.

Verification: all 777 common tests and 29 shared tests pass; `git diff --check` is clean.

todwadd:
All three confirmed and fixed in `57eb1cf`. I checked each against the code rather than taking it on trust; #1 is the important one and I had reasoned about a weaker version of it and talked myself out of it.

## 1. Bind the partial close to the state the user saw — confirmed, and worse than I had it

I'd considered the single-client version of this and dismissed it, because the client's idempotency fingerprint carries `position.size`: a double-click before the refetch lands reuses the key and replays instead of closing twice. That reasoning does not survive two clients. Separate tabs generate separate keys, nothing dedupes them, `expectedOpenedTime` passes because the survivor keeps its `openedTime`, and the second request applies 75% to the 100 the first one left. The user consented to closing 300 and got 75.

Callers now send `expectedSize` and it is compared against the locked row inside the transaction. Two decisions worth flagging for re-review:

**The tolerance is representation dust only** (`PERP_EXPECTED_SIZE_TOLERANCE`, relative, `1e-9`), deliberately not enough to admit real movement. I first reached for a tolerance wide enough to forgive funding — `maxFundingRate` is only bounded to `(0, 1)`, so no principled width exists — then realised that framing is backwards. If funding scaled the row or ADL took a haircut, the fraction was chosen against a position that no longer exists and the preview is stale for exactly the same reason an intervening close makes it stale. All three should 409. The client polls positions every 15s against an hourly funding period, so the spurious-rejection window is small, and it now refetches on a failed close, so the 409 that says "refresh" actually refreshes.

**It is required, not optional, for a partial close.** Optional would have left the race unguarded for precisely the callers most likely to hit it, and "close half of whatever happens to be there" is never what someone means when they previewed a number. A caller has to read the position to know it exists, so it always has a size to send. No compatibility cost — `fraction` ships in this same PR. Enforced at the schema and restated in the engine so a non-HTTP caller can't skip it, and carried in the idempotency `request` so a replay with a different expectation is a 409.

Full closes deliberately skip the check. "Close everything" is unambiguous at any size, and a spurious 409 on the one operation a trader may urgently need is a worse failure than any it would prevent.

## 2. Rate limit — my arithmetic was wrong

You're right and I had this an order of magnitude off. The 1% floor is a fraction of the *current remainder*, so it decays geometrically rather than terminating:

```
0.99^100                       = 0.3660      -> 36.6% still open
ln(0.0001)/ln(0.99)            = 916.4       -> M$100 basis to the M$0.01 floor
ln(1e-6)/ln(0.99)              = 1374.6      -> M$10,000 basis
```

So it's ~917 calls, not ~100, and it grows with `log(costBasis)`. The "already bounded" justification in the PR body was simply false; I've corrected it there.

Fixed exactly as you scoped it: `rateLimitByUser` at 60/hour, reachable **only** when `fraction < 1`. A full exit stays unlimited, so the emergency path is untouched. 60/hour is far above any plausible human scaling-out and caps the sustained rate at one per minute, which turns the ~917-call walk-down into a 15-hour job. In-memory per instance, like every other user of that helper.

## 3. Unified feed — confirmed, a genuine miss

I checked the market Trades tab, the position history, the `/perps` hub tape and the balance-change feed, and never looked at `get-unified-feed`. It selected no `data`, `PerpTradeActivity` had nowhere to put a fraction, and `PerpTradeLog` rendered every close with full-close wording.

All three now carry it. One note on the rendering: `margin` is already `abs(originalCostBasisDelta)`, i.e. the closed leg's margin, so "closed M$75 LONG" was never *wrong* — it just couldn't say the rest is still running. Rather than reword the sentence I put it in the detail row where leverage sits for opens: **"25% of the position — the rest stays open"**. The converter only surfaces a fraction strictly inside `(0, 1)` on a `close`, so full closes and pre-existing events are untouched.

## Verification

`yarn test:common` 36 suites / **780** tests · `yarn test:shared` 4 suites / 29 tests · `typecheck:web` / `:api` / `:scheduler` clean · `yarn lint` and `yarn format:check` clean. New tests pin the tolerance at both edges: dust passes at M$400 and M$12m scale, and a 0.1% move — well below the 1% smallest possible close — fails.

`backend/shared/src/perps/README.md` documents the `expectedSize` rule and the rate limit alongside the existing partial-close section.

---

todwadd:
Reviewed `3b8c4dd`. The bug is real and mine: `Math.round(0.999 * 100)` is `100`, so a genuine partial close rendered as "Closed 100% of long" while the position was still open — in five places. It's reachable from any API caller, and the dust promotion doesn't save it: on an M$10,000 basis, `fraction = 0.999` leaves M$10 open, far above the dust floor, and still displayed as a full close.

The fix is sound, and the choice of `Math.floor` over `Math.round` in the sub-percent branch is the load-bearing part — it's what makes "never 100%" a property rather than a coincidence, since truncation cannot carry into the next integer.

I checked that beyond the two cases the tests pin. Swept every fraction in `[0.001, 1)` at 1e-5 resolution (99,900 values) plus adversarial near-1 values:

```
0.9949999          -> 99%
0.995              -> 99.5%
0.999              -> 99.9%
0.9999             -> 99.99%
0.999999999999     -> 99.99%
1 - EPSILON        -> 99.99%
sweep: never renders 100%, always well-formed
```

Edge inputs degrade sensibly too — `NaN`, `Infinity` and negatives give `0%`, `>= 1` gives `100%`.

## One site was missed: `7a3aab5`

The position card's own **close button** was still on `Math.round(effectiveFraction * 100)` — the one label a user reads *before* deciding rather than after.

No user-visible defect today. `effectiveFraction` comes from the 25/50/75 toggle via `resolvePerpCloseFraction`, which returns either the chosen preset or 1, so the rounding is exact and the partial branch isn't taken at 1. It's latent: the moment that selector grows a custom amount — which the trigger-order work in the next PR will want, since "close 40% at price X" is the whole point — it becomes the same bug in the highest-stakes spot. And a change whose stated purpose is centralizing shouldn't leave a caller behind. Now routed through `formatPerpClosePercent` like the other five.

## Verification on the merged result

`yarn test:common` 36 suites / **782** tests · `yarn test:shared` 4 suites / 29 tests · `typecheck:web` / `:api` / `:scheduler` clean · `yarn lint` and `yarn format:check` clean. `grep` confirms no raw close-percent rounding remains anywhere in `web/` or `common/`.

---

todwadd:
**`Vercel – prod` is failing on this PR, and it is not this PR's failure.** Recording the evidence rather than leaving a red check unexplained.

## What's red

On `b43358c`:

| Check | Result |
| --- | --- |
| Typecheck, Lint, Format & Test | ✅ success (16:20:18Z) |
| `Vercel – dev` | ✅ "Deployment has completed" (16:14:20Z) |
| `Vercel – prod` | ❌ "Deployment has failed" (16:25:01Z) |

Both Vercel projects build the same `web` root with the same `next build`; they differ in environment.

## Why it isn't this diff

**It reproduces on an unrelated PR.** #4037 (`feat/perp-pool-stats`) has the same signature on its own head — `Vercel – dev` success, `Vercel – prod` failure. Different branch, different diff, same result, so the cause is shared rather than in either change.

**The commit can't reach that failure mode.** `b43358c` touches exactly one file, `web/components/perps/perp-position-panel.tsx`. It contains no `getStaticProps`, no `getServerSideProps` and no `process.env` — a pure client component, which cannot affect the prerender phase.

**The code compiles under the prod environment.** I ran `NEXT_PUBLIC_FIREBASE_ENV=PROD next build` locally: *"Compiled successfully in 2.2min"*, then through static generation to 70/94 pages. It then died on `/election/needle` with `Host not in allowlist: …supabase.co` and a missing `GOOGLE_APPLICATION_CREDENTIALS_PROD` — both limitations of my sandbox, not code faults, but they demonstrate the relevant point: **the prod build pre-renders pages against live prod services.**

**The timing matches that phase, not compilation.** Prod ran ~15 minutes before failing (16:10 → 16:25) where dev finished in ~4. Compilation takes ~2 minutes, and a compile error would have failed dev on the identical SHA.

## What I have not done

I have no Vercel access, so I can't read the deployment logs (`npx vercel inspect dpl_7t7JCviYTny8GhLoJsKQWVeqPEnY --logs`) or re-run the deployment, and I won't push an empty commit to trigger one. There is no fix to port — nothing has been identified to fix.

Someone with Vercel access should look at the **prod** project's build environment (credentials, egress, or quota) rather than at either PR: with two unrelated PRs failing the same way while dev passes, the prod project's configuration is the common factor.

This PR is otherwise green — Actions CI passes and it is mergeable. I'll keep watching and will report if the picture changes.

---

todwadd:
**`Vercel – prod` is green.** Correcting my comment above so nobody goes looking for a problem that's no longer there — please disregard its "someone with Vercel access should look at the prod project".

On `97664b8`:

| Check | Result |
| --- | --- |
| Typecheck, Lint, Format & Test | ✅ success (01:43:38Z) |
| `Vercel – dev` | ✅ Deployment has completed |
| `Vercel – prod` | ✅ Deployment has completed |
| `Vercel – docs` | ✅ ignored by build step |

`mergeable_state` is `clean` — green and no conflict with `main` (`e541456`).

What changed: the failing prod deployment on `b43358c` produced **no branch preview URL**; the passing one on `97664b8` produces `prod-git-claude-perps-partial-sells-limits-g2zfsj-mantic.vercel.app`. So the prod project's preview behaviour differs between the two runs. I still can't read the Vercel logs, so I can't say whether that was a config change, a credential, or a quota — only that the check now passes and the earlier evidence (it reproduced on unrelated PR #4037, and `b43358c` touched one pure client component with no `getStaticProps`) held up: the failure was never in this diff.

## The new commit

`97664b8` "Compact perps position summary and Close action" reworks `PositionCard` into a compact header (side + leverage, small **Close** button) over a three-column Notional / Margin / Unrealized P&L grid, and moves the oracle-paused wording off the button into its own note. Verified on that exact head before CI reported: `yarn typecheck:web` clean, Prettier clean on the file, and ESLint reports nothing for it. Nothing outside `web/` changed, so the test suites are untouched.

---

PR #4042, opened 2026-09-06
feat(elections): instrument the page, retire the dead primaries list, add polling perps

todwadd:
Follows an audit of what people actually do on `/election`.

## What the data said

Measured over 7 days from `user_events`, sessions reconstructed on `deviceId`:

- **912 visits, ~130/day.** Small.
- **Only 6.4% are cold entries** (no prior on-site event within 30 min). This is not an SEO or share landing page; it is an internal browse destination.
- **64% of all visits (588) arrive from one pill** — the "2026 Midterms" link in Browse, which is the **most-clicked topic pill on the site**: 668 clicks/7d vs perps 589, Politics 406, Technology 404, Sports 337.
- **30% dead-end** (no further event in 10 min), **51% bail** to browse/home.
- On-page clicks, bounded to before the visitor leaves: trending 52, state map 47, lower tables 31, BoP card 12.

So demand is strong and the page is nearly invisible outside the browse row. That shaped everything below.

## Instrumentation

There was **no scroll-depth signal of any kind**. A section with few clicks was indistinguishable from a section nobody ever scrolled to, which made section ordering unmeasurable. New `SectionInView` fires `view election section` once per mount; every section is wrapped. Pair with the existing click events for a real per-section reach → click rate.

Three surfaces also logged nothing usable:

| Surface | Before | After |
|---|---|---|
| Map state cards | literal `click state card undefined` | `election map senate` / `election map governor` |
| `ContractsTable` rows | `click browse contract`, no section | adds `section` to the payload |
| BoP levers, feed topic pills | no event at all | `click bop lever`, `select election feed topic` |

## Content — the primaries section was dead

**Six of the seven curated `PRIMARIES_2026` markets had resolved.** The fetcher filters resolved markets out, so "2026 Primaries to watch" had silently decayed to a single card. It is September; primary season is over, and a query for open 2026 primaries with real volume returns nothing worth featuring.

Replaced with **"Closest races"**, derived in `getTossUpRaces` from the Senate and Governor map contracts *already being fetched* — open markets whose two-party split sits inside 30–70%, ranked by distance from even. No curation, no extra queries, re-derives every revalidation, cannot go stale the way a slug list does.

Against today's prod data it yields TX Senate (.489), OH Senate (.542), IA Senate (.406), GA Governor, NV and AK — genuine toss-ups. Known limitation, called out in a comment: where an independent is the actual front-runner (NE / Dan Osborn) the two-party renormalisation reads it as a safe seat and it drops out. That is the same limitation the map already has.

Also drops the one resolved redistricting market.

## Polling averages

New row with the three VoteHub perps: Trump approval (38.7, 160 bettors), Dem generic ballot (47.0) and Vance favorability (40.5) — all healthy, oracle fresh. Carousel on mobile, grid from `sm` up, per your side-by-side/carousel note.

The **generic ballot** perp was created 2026-09-04 and is arguably the most midterms-relevant of the three, so I included it alongside the two you asked for. Units come from the existing `ORACLE_TICK_DECORATIONS` table in `common/perps/oracle-display` rather than a second hardcoded percent list (`perps.tsx` has its own `PERCENT_FEEDS` copy; I did not touch that page, but the shared table is the one source of truth).

## Layout

Trending moves from directly under the hero to **below Balance of Power**, so the map and control markets lead.

**This is deliberately against the measurement** — Trending is currently the top-clicked module on the page, and moving it down is expected to cost some of those clicks. `SectionInView` exists precisely so we can see what it costs instead of guessing. It is a one-block move and trivially revertible.

## SEO — three real defects

1. **No canonical tag at all.** `SEO.tsx` only emits `<link rel="canonical">` and `og:url` when `url` is passed, and this page never passed it. Fixed.
2. **Absent from `sitemap.xml`** (7 URLs there; election was not one). Added.
3. **Duplicate-content surface.** The route is an optional catch-all `[[...slug]].tsx` with `getStaticPaths: {paths: [], fallback: 'blocking'}` and a `getStaticProps` that ignored `params` entirely — so `/election/anything/at/all` returned HTTP 200 with the identical page and no canonical. Now permanently redirects subpaths onto `/election`. Redirect rather than 404 so inbound links keep their equity. `/election/needle` is a real file-system route and takes precedence, so it is unaffected.

Retitled `"Manifold Elections"` → `"2026 Midterm Election Odds"` (it rendered as "Manifold Elections | Manifold") and rewrote the description around what the page shows.

## Verification

- `tsc --noEmit` and `eslint` clean across all 12 files.
- Toss-up output checked against live prod probabilities (numbers above) rather than assumed.
- Perp health, feed ids and units checked against prod.
- **Not rendered in a browser.** Layout changes and the new perp row are unexercised visually — worth a look before merge, particularly the mobile carousel.

## Flagged, not fixed

- **19 open map markets have < 15 bettors**, several with no bet since June (WY 2026-06-10, TN 2026-06-30). The map paints states off very thin books. Fixing means subsidising or replacing community markets — your call, not a code change.
- **Cold entry is 6.4%.** The SEO fixes are necessary but almost certainly not sufficient; if elections matter strategically 8 weeks out, acquisition is the lever, not layout.


todwadd:
### Added: candidate names in the competitive House districts (`21d7888f8`)

The House map showed `TX-15, TX-34, TX-28` with no indication of who is running. This adds the names — by renaming the answers on the existing community market rather than creating per-district markets.

**Why rename rather than split into 16 markets.** `server-sitemap.xml` only submits the **top 1000 markets by importance score**:

| | rank | bettors | score |
|---|---|---|---|
| districts market | **163rd** | 59 | 0.394 |
| 1000th market (cutoff) | 1000 | 8 | 0.132 |

Sixteen fresh per-district markets would each start near zero bettors, land below that cutoff, and not be submitted at all — trading one indexed page for sixteen invisible ones. Renaming instead injects candidate names into a page Google already crawls, and keeps the market's 59 traders and price history intact.

**Scope:** only the 16 answers inside the 30–70% band. The other 52 districts are decided and are left alone.

**Data:** matchups verified against Ballotpedia's 2026 general-election candidate lists on 2026-09-07, then independently re-checked. One trap worth recording — the PA-1 state page lists **John Hoban as "running in the general election"**, but he is an *Independent*; the Democratic nominee is Bob Harvie. A naive read would have written an independent in as the Democrat.

**`backend/scripts/rename-house-district-answers.ts`** — dry run unless `--apply`; requires `EDITOR_USER_ID` to be a real admin (it is recorded in `contract_edits`); writes a row only when its current text still matches exactly, so re-running is safe and a concurrent edit is reported rather than clobbered.

It writes `text` only. Answer ids, pools, probabilities and every position are untouched, and bets reference `answerId` rather than text, so API bots are unaffected. Each new string also keeps the old one as an exact prefix (`Texas 15 · …`), so bots that discover answers by substring still resolve them — only exact-equality matchers would need updating.

Preconditions verified against prod before shipping: all 16 answer ids exist, sit on the right contract, are unresolved, and still carry their original text (zero drift).

**`parseDistrict` fix.** This is why the two changes ship together. `parseDistrict` returns everything after the state name as the district, and the row renders it in a fixed `grid-cols-[4.5rem_1fr_3.5rem]` cell — so the rename alone would have dumped the full matchup into a 4.5rem column and made the bet dialog read *"Will a Democrat win TX-15 · Bobby Pulido (D) v. Monica De La Cruz (R)?"*. It now splits the ` · <matchup>` suffix: the number keeps its column, names render beside the bar and under the dialog heading, and answers without a suffix are unchanged.

`tsc` and `eslint` clean. **The script has not been run** — it edits a live community market owned by another user, so that is a deliberate hand-off.

PR #4047, opened 2026-09-06
fix(elections): feature the right markets for Montana and Idaho (independent candidates)

todwadd:
Montana is not a two-party race, and both markets the page was using are structurally unable to say so.

Steve Daines (R) withdrew minutes before the filing deadline and endorsed **Kurt Alme (R)**. **Alani Bankhead (D)** is the Democratic nominee, at ~2%. The actual competitive alternative is **Seth Bodnar**, running as an **independent** with Jon Tester's backing, at ~7%.

| market | answers | traders | liquidity |
|---|---|---|---|
| party market *(current map entry)* | `Republicans 97% \| Democrats 3%` | 22 | 200 |
| candidate market *(current "who's running")* | `Kurt Alme 95% \| Seth Bodnar 5%` | 29 | 1,000 |
| **Tetraspace's** | `Kurt Alme 90% \| Seth Bodnar 7% \| Alani Bankhead 2% \| Other 1%` | 11 | **2,000** |

The party market has no `Other` line, so a Bodnar win would N/A it — and it hides the only interesting contest behind a 97/3 party split. The head-to-head omits the Democratic nominee entirely and would also N/A on a surprise. Only the third carries all three candidates plus `Other`.

Points the state at that market and drops the now-redundant candidate-card entry. It has less volume (11 traders / 1.2k vs 22 / 20k) but more liquidity (2,000 vs 200), and correctness beats volume when the alternatives cannot represent the race at all.

### A note that would have become a lie

The "Resolves by party, not by candidate" caveat from #4042 was gated purely on the answers naming candidates — which is also true of a *genuine* candidate market. Left alone, this panel would have told Montana readers that their candidate market settles on party.

It now additionally requires the question to be a "which party" question. Verified against the real questions: fires on all six renamed party markets, hides on `"Who will win the 2026 Montana senate election?"`. Crude, but it fails safe — an unrecognised wording hides a true note rather than showing a false one.

### Bodnar is tagged `(I)`, not `(D)`

The script tags Montana's answers so `getPartyProbs` can colour the state. Bodnar gets `(I)`, which matches neither `DEM_TAG` nor `REP_TAG` and so lands in the "other" bucket rather than being miscounted as a Democrat. Result: rep .90 / dem .02 / other .08.

This is the same failure mode already documented for Nebraska (Dan Osborn) — Montana is now a second instance, and worth remembering as a pattern rather than a one-off.

### ⚠️ Deploy order — opposite of last time

**Run the script for Montana first, then merge this.**

Renaming Tetraspace's answers has no visible effect today, because nothing references that market yet. Merging first would point the map at an untagged candidate market and grey Montana out until the script ran. Doing it in this order there is no gap at all.

`tsc` and `eslint` clean. Not viewed in a browser.

### DIFF
--- a/backend/scripts/rename-election-market-answers.ts
+++ b/backend/scripts/rename-election-market-answers.ts
@@ -220,6 +220,28 @@ const SENATE_PARTY_MARKETS: Group[] = [
   },
 ]
 
+/**
+ * Montana is a candidate market, not a party market, so it is listed
+ * separately: the point of tagging its answers is purely so getPartyProbs can
+ * colour the state, NOT to relabel a party line. Seth Bodnar is an
+ * INDEPENDENT — Jon Tester-backed, and the actual competitive alternative to
+ * Alme — so he is tagged (I) and lands in getPartyProbs's "other" bucket
+ * rather than being miscounted as a Democrat.
+ */
+const MONTANA_THREE_WAY: Group = {
+  contractId: 'tydQt5d26u',
+  label: 'MT Senate (three-way)',
+  renames: [
+    { id: 'lPtldCU006', from: 'Kurt Alme', to: 'Kurt Alme (R)' },
+    {
+      id: 'RSCc0lpsNt',
+      from: 'Alani Bankhead',
+      to: 'Alani Bankhead (D)',
+    },
+    { id: 'L5dSc92I5L', from: 'Seth Bodnar', to: 'Seth Bodnar (I)' },
+  ],
+}
+
 const GROUPS: Group[] = [
   {
     contractId: 'sqUzOZN8Cs',
@@ -227,6 +249,7 @@ const GROUPS: Group[] = [
     renames: HOUSE_DISTRICTS,
   },
   ...SENATE_PARTY_MARKETS,
+  MONTANA_THREE_WAY,
 ]
 
 runScript(async ({ pg }) => {
--- a/web/components/us-elections/contracts/party-panel/party-panel.tsx
+++ b/web/components/us-elections/contracts/party-panel/party-panel.tsx
@@ -85,9 +85,18 @@ export function PartyPanel(props: {
   const republicanAnswer = answers.find((a) => isRepublicanAnswer(a.text))
   const democraticAnswer = answers.find((a) => isDemocraticAnswer(a.text))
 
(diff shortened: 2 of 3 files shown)


todwadd:
### Added: Idaho (`24b6c2e77`)

Same shape as Montana. Idaho has **no Democrat on the ballot** — David Roth won the nomination, then withdrew in July once independent former state rep **Todd Achilles** entered with Democratic leaders' backing. So it's **Jim Risch (R) v. Achilles (I)**.

| market | answers | traders | liquidity | last bet |
|---|---|---|---|---|
| current map entry | `Republicans 95% \| Democrats 5%` | 7 | 200 | 25 Jul |
| **TheDucksFan's replacement** | `Republicans 95% \| Independent 5%` | 11 | **1,200** | **today** |

Same creator — they superseded their own market once the race changed shape.

**It needs `otherParty`.** `getPartyProbs` requires both major parties present (`!hasDem || !hasRep`) so a one-sided market can't fake max certainty — and the new market has no Democratic answer at all, so Idaho would grey out on the map. `otherParty: 'Democratic Party'` folds Achilles into the Democratic side for colouring only, which is precisely what the field was added for: "a race where the only viable non-major candidate caucuses with one side".

Verified by replicating the function against the live answers:

```
ID today, no otherParty : GREY (undefined)
ID with otherParty      : dem 0.05 / rep 0.95 / other 0.00
ID after rename to names: dem 0.05 / rep 0.95 / other 0.00
```

The last line matters — it's robust to renaming the answers to `Jim Risch (R)` / `Todd Achilles (I)`, so the format change can land before or after this merges.

**No deploy-order dependency for Idaho**, unlike Montana: the existing answers (`Republicans` / `Independent`) already work with `otherParty`, so this can merge whenever. Montana still needs its three answers tagged first.

### Worth noting as a pattern

Three Dem-backed independents now: Bodnar (MT), Achilles (ID), and Osborn (NE) which was already documented in this file. The two-party map assumption keeps failing the same way. `otherParty` handles it, but it's manual per-state — if a fourth turns up it may be worth detecting rather than curating.

I did **not** apply `otherParty` to Montana. It already colours (Bankhead is a real D answer at 2%), so adding it would change existing behaviour without being required — it'd shift MT from 90R/2D to 90R/10D by folding Bodnar in. Defensible either way; your call.

PR #4053, opened 2026-09-08
docs(guidelines): add a duplicate markets policy

todwadd:
## Why

We had **no rule against duplicate markets anywhere in the guidelines**. The only trace of the concept was a keyword in the search index pointing at a `#subsidized` section that had already been deleted from the page â€” so a mod warning a creator that they're splitting the liquidity pool had nothing to link to.

This writes the rule down, with real consequences attached.

## What

**Market Policies â€” new "Duplicate markets" section** (canonical statement)

- Why it's a problem: splitting liquidity across two markets makes both thinner and worse priced â€” better for arbitrage bots, worse for humans.
- When a second market is legitimate: materially different resolution criteria, timeframe, or resolution source — **or a well-founded concern about who's resolving the original**. Say so in the description.
- **Creator reliability** as a reason: where a market genuinely leaves room for interpretation (thin criteria, or a judgement call at resolution) and the creator holds a position, has form for resolving their own ambiguous markets in their favour, or is unresponsive, a better-specified version run by someone impartial is worth having. Bounded so it isn't a catch-all — try to fix the original first (ask the creator, tag @mods), write criteria specific enough that resolution isn't a judgement call, and note in the description that it exists to remove the ambiguity. A creator resolving an *unambiguous* market against your position isn't unreliable, and re-running that is still a plain duplicate.
- Notes that the create page already lists similar open markets as you type a title, so near-identical titles aren't really an accident.
- Consequences, escalating:
  - unranked (no Leagues credit) and significantly deboosted;
  - **delisted outright or resolved N/A where no real trading has occurred**;
  - fine for repeats or after a warning;
  - unlisting/deletion plus a market creation ban for bulk creation or farming bonuses, league points, or engagement.
- Covers re-creating your own market to reset it or escape a position.
- Cross-referencing bullets added to the Ranked criteria list and the unlist/N/A/delete list.

**Running a Market â€” "Check for an existing market first"**

Creator-facing norm: read the similar-markets list on the create page, and if you want the question to be bigger, bet on the existing market or add liquidity to it instead.

**Moderation Guidelines â€” "Handling duplicate markets"**

Mod playbook: what counts as a duplicate (rewording the same question still counts; a different date/threshold/source doesn't), and which lever for which case. Two judgment calls worth reviewing:

- **Unrank rather than unlist once a duplicate has traders** â€” unlisting strands people who already bet. Delist or N/A freely when nobody's position is affected.
- **Don't penalise the newer market when it's clearly the better one** (better criteria, better sourced, more traders) purely for coming second.
- **Check the creator-reliability claim rather than accepting it** — "I don't trust the creator" is the natural cover story for someone who dislikes the price. Where the concern is real, it's the *original* that needs a takeover or criteria clarification, not the new market that needs unranking.

**Search index**

Replaces the dead `#subsidized` entry and adds entries for the three new sections. Running a Market and the Moderation Guidelines previously had *no* search entries at all â€” these are their first, so those pages remain otherwise unindexed.

## Testing

Docs-only, no behaviour change. `tsc --noEmit` clean, prettier applied.

### DIFF
--- a/web/components/guidelines-search.tsx
+++ b/web/components/guidelines-search.tsx
@@ -160,7 +160,7 @@ export const GUIDELINES_SEARCH_INDEX: SearchEntry[] = [
   {
     page: 'Market Policies',
     section: 'Duplicate markets',
-    text: 'duplicate duplicates same question already exists splitting liquidity pool thinner arbitrage bots unranked deboosted delisted resolved n/a no trading importance score fine market creation ban similar markets shown at creation title match differing resolution criteria timeframe source bulk farming league points bonuses',
+    text: 'duplicate duplicates same question already exists splitting liquidity pool thinner arbitrage bots unranked deboosted delisted resolved n/a no trading importance score fine market creation ban similar markets shown at creation title match differing resolution criteria timeframe source creator reliability biased unreliable unresponsive judgement call ambiguous impartial bulk farming league points bonuses',
     href: '/community-guidelines/market-policies#duplicate-markets',
   },
   {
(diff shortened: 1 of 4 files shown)


JasonMendoza2008:
:)

PR #4067, opened 2026-09-12
Drop cancelled limit orders from price-impact calculations

todwadd:
Cancelling a limit order must remove it from every quote, including a panel opened after the cancellation and a request that was already in flight.

The matching guard rejects cancelled, filled, and expired counterparties. Server order queries already filter the closed flags; the guard also protects stale client inputs.

The client now has one observable order-book snapshot per contract. Confirmed mutations update that snapshot even without mounted consumers. Refreshes discard superseded responses and preserve live updates received during the read. This replaces the closed-order TTL map and the separate copies in the persistent-state hook. Mutable snapshots are not shared between SSR requests. An expiry timer also removes orders from idle panels.

Open-order reads bypass browser and HTTP response caches. Cancel-all applies each successful response immediately and reports partial failures; a single failed cancel also reports an error. Updates happen after API confirmation.

The branch includes current main and preserves the original PR commit. The later PRs remain based on this branch.

Validation of the completed stack: full web typecheck passes; common/shared TypeScript builds pass; all 1,186 common tests and 20 React hook/rendering tests pass. New code passes lint. Existing lint findings are identical to current main. The API typecheck still reports two implicit-any errors in unchanged files (`helpers/on-create-market.ts:133` and `stripe-endpoints.ts:330`). No production trade or live outage was exercised.

### DIFF
--- a/backend/api/src/app.ts
+++ b/backend/api/src/app.ts
@@ -1,3 +1,4 @@
+import { isUncachedQuoteRead } from 'common/api/cache'
 import { API, type APIPath } from 'common/api/schema'
 import { APIError, pathWithPrefix } from 'common/api/utils'
 import { randomString } from 'common/util/random'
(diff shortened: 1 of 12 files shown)


todwadd:
This is the base of a 4-PR stack on order book freshness. Each builds on the previous, so merge in order:

1. **#4067** (this one) — a cancelled order can no longer be a counterparty in a price-impact calculation, and your own cancel takes effect instantly.
2. **#4068** — refetch the order book after a websocket reconnect. Closes the gap this PR can't: it makes the client honest about orders it *knows* are closed, but nothing made it learn about other people's.
3. **#4069** — dedupe the order book fetch across panels, and fix maker balances briefly reading as unlimited. Removes more requests than #4068 adds.
4. **#4070** — show the order book's contribution to a quote in the bet and sell panels. Presentational only.

Split along review-scrutiny lines: #4068 touches shared websocket infra, #4069 is data-layer caching, #4070 is pure UI.

---

PR #4075, opened 2026-09-13
Open loans to unverified users

todwadd:
## Why

Daily free loans were gated on `hasFullBonusAccess`, which requires `bonusEligibility` ∈ {`verified`, `grandfathered`, `eligible`}. That contradicts the membership benefits table, which has always advertised the **1% daily free loan to unverified users**: the `freeLoan` row in `BENEFIT_DEFINITIONS` defines no `unverifiedValue`, so the unverified column falls through to `baseValue: '1%'` (see `benefits-table.tsx` L284-287). Unverified users saw the promise and got a 403.

`NON_SUBSCRIBER_PERKS.freeLoanRate = 0.01` already applies to all non-subscribers, and `TIER_BENEFITS.unverified` inherits it — the rate table was built for unverified access. Only the binary gate disagreed.

More fundamentally, loans don't belong on the bonus axis. They're borrowed against the user's own positions and repaid on resolution, not granted, so there's no self-exclusion rationale for keeping unverified users out — nor bot-marked accounts, which as it turns out never touched this gate at all (`isBot` is not read anywhere in the loan path or in `hasFullBonusAccess`).

## What

- **`common/src/loans.ts`** — new `canTakeLoans(user)`, placed beside the other loan predicates so a future change to the bonus predicates can't silently move loan access again. Takes only the field it reads, so the module needs no `User` import.
- **`backend/api/src/claim-free-loan.ts`** — gate swapped; 403 copy updated (the old "Complete identity verification" text would be wrong for the remaining case).
- **`web/components/home/daily-loan.tsx`** — gate swapped so the chest stops routing unverified users into `VerificationRequiredModal`.
- **`common/src/loans.test.ts`** — cases for unverified, `ineligible`, verified/grandfathered/eligible, the flagged exclusion, and bot-independence.

### Who is still blocked

Only `bonusEligibility === 'requires_verification'` — the admin/system flag for suspected alts and accounts under manual review, where the block is doing intended work rather than leaking. Those users still get a coherent prompt: `VerificationRequiredModal` already has an `isFlagged` branch with the right copy.

KYC-failed (`ineligible`) users **are** allowed. The tier config already treats them as reduced-earning rather than cut off — they resolve to the `unverified` tier, not `restricted`. Flagging this as the one judgment call in the change; easy to flip if you'd rather they stay blocked.

Margin loans need no change — already gated on subscription entitlements via `canAccessMarginLoans`, independent of verification, and correctly advertised as `-` for non-subscribers.

## Verification status

⚠️ **Not yet verified locally.** The worktree's `yarn install` is still running, so jest and the typechecks haven't been run against these changes. An earlier attempt using a `node_modules` junction produced results that described the *main* checkout's files rather than this branch's, so I discarded them as meaningless rather than report them as passing. The main checkout's API typecheck is clean on `main`, which is the baseline. I'll follow up here once the real run completes.

No backend schema or migration changes, so no API deploy coupling beyond the usual.

## Out of scope (pre-existing, separate)

A **subscribed bot still collects full-rate quest/streak/referral bonuses**: `resolveEffectiveTier` short-circuits on subscription (`if (subTier) return subTier`) before consulting `bonusEligibility`, so a subscription routes around the bot bonus-exclusion entirely. Unrelated to loans but a live hole in the self-exclusion if that's meant to be airtight — happy to take it as its own PR.

### DIFF
--- a/backend/api/src/claim-free-loan.ts
+++ b/backend/api/src/claim-free-loan.ts
@@ -1,13 +1,13 @@
 import { APIError, type APIHandler } from './helpers/endpoint'
 import { createSupabaseDirectClient, pgp } from 'shared/supabase/init'
 import { getUser, log } from 'shared/utils'
-import { hasFullBonusAccess } from 'common/user'
 import {
   calculateMaxGeneralLoanAmount,
   calculateDailyLoanLimit,
   calculateMarketLoanMax,
   calculatePositionFreeLoan,
   canClaimDailyFreeLoan,
+  canTakeLoans,
   filterLoanEquityMetrics,
   isMarketEligibleForLoan,
   getMidnightPacific,
@@ -44,12 +44,14 @@ export const claimFreeLoan: APIHandler<'claim-free-loan'> = async (_, auth) => {
     throw new APIError(404, `User ${userId} not found`)
   }
 
-  // Daily free loans are a full-bonus perk: identity-verified/grandfathered
-  // users and purchase/admin-granted users can claim them.
-  if (!hasFullBonusAccess(user)) {
+  // Daily free loans are open to everyone, verified or not — they're borrowed
+  // against the user's own positions, and the membership page has always
+  // advertised the 1% daily free loan to unverified users. Only admin-flagged
+  // accounts (suspected alt / manual review) are held back.
+  if (!canTakeLoans(user)) {
     throw new APIError(
       403,
-      'Complete identity verification to access daily free loans'
+      'Your account is flagged for review. Complete identity verification to access daily free loans.'
     )
   }
 
--- a/backend/api/src/claim-free-loan.ts
+++ b/backend/api/src/claim-free-loan.ts
@@ -13,6 +13,8 @@ import {
   getMidnightPacific,
 } from 'common/loans'
 import { Contract } from 'common/contract'
+import { isUserBanned, getUserBanMessage } from 'common/ban-utils'
+import { UserBan } from 'common/user'
 import { Txn } from 'common/txn'
 import { txnToRow } from 'shared/txn/run-txn'
 import { filterDefined } from 'common/util/array'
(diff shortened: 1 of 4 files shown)


todwadd:
## Verification complete — all green

The PR description's "not yet verified locally" caveat is now resolved. Run in the worktree with a real `yarn install --frozen-lockfile` (no `node_modules` junction — see note below):

| Check | Result |
|---|---|
| `jest src/loans.test.ts` | **16/16 pass**, incl. all 5 `canTakeLoans` cases |
| `typecheck:web` | **exit 0**, clean |
| `typecheck:api` | **exit 0**, clean (after `yarn build:ci`) |

```
canTakeLoans
  √ allows unverified users (bonusEligibility undefined)
  √ allows KYC-failed users — reduced earning, not cut off
  √ allows verified, grandfathered, and purchase/admin-granted users
  √ blocks admin-flagged accounts pending review
  √ ignores bot status
```

### Two false alarms worth recording

**1. A `node_modules` junction produced meaningless results.** My first attempt symlinked the worktree's `node_modules` at the main checkout's. Commands ran, but yarn workspace symlinks (`node_modules/common` → main checkout) redirected every `common/*` import back out of the worktree, so `typecheck:web` reported 49 zod errors in `common/src/api/schema.ts` — none from this branch, and the main checkout typechecked clean on `main` at the same moment. Discarded as artifacts rather than reported. A real install resolves correctly.

**2. `typecheck:api` initially failed with ~49 files of errors, all TS6305** — "Output file `common/lib/loans.d.ts` has not been built from source file". The api tsconfig uses project references and expects `common/lib` + `backend/shared/lib` declarations to exist. Same TS6305 errors hit `repay-loan.ts`, `sell-shares.ts`, `get-free-loan-available.ts` and other untouched files, which is the signature of an unbuilt reference graph, not a regression. `yarn build:ci` then `typecheck:api` → exit 0.

No schema or migration changes, so no API deploy coupling beyond the usual.

todwadd:
## Both findings confirmed — fixed in `72f10aefb`

Both were real, both were mine, and the P1 was a genuine privilege-escalation path I introduced. Thanks for catching them. I verified each premise in the code rather than taking the report at face value, and the reasoning holds in both cases.

### 1. [P1] Expired verification launders an admin hold — confirmed

`mapIdenfyStatus` folds `EXPIRED` and `DELETED` into `'denied'` (`idenfy/callback.ts` L139-142), and the generic-denial branch then rewrites any non-grandfathered user to `bonusEligibility: 'ineligible'` (L484-490). So the sequence is entirely passive:

```
admin flags account      → bonusEligibility = 'requires_verification'  → canTakeLoans false
user does nothing, session expires
iDenfy callback fires    → bonusEligibility = 'ineligible'             → canTakeLoans TRUE  ← escalation
```

A suspected-alt account regained loan access by waiting. Before this PR, `hasFullBonusAccess` denied `ineligible` too, so this was strictly my regression.

### 2. [P2] Superbanned accounts regain loans — confirmed

`superBanUserCore` writes `bonusEligibility: 'ineligible'` in two places (`shared/helpers/super-ban.ts` L105-107, L299-303). My gate allowed that state, so daily loans came back for superbanned users holding eligible positions.

### The root mistake, and the fix

I keyed authorization off `bonusEligibility`, which is **overloaded**: `'ineligible'` means "KYC failed" *and* "admin hold expired" *and* "superbanned". I read the tier config (`ineligible` → `unverified` tier, not `restricted`) as evidence it was benign reduced-earning. That was the wrong inference — two of its three writers are enforcement.

- **`canTakeLoans` now denies both** `requires_verification` and `ineligible`. The case this PR exists to fix — default-unverified users, `bonusEligibility === undefined`, the ones the membership table advertises the 1% loan to — is untouched. The docstring now spells out why each deny state is load-bearing, so it doesn't get "simplified" back.
- **Added an independent `user_bans` check** in `claim-free-loan`, per your suggestion. Bans live in `user_bans`, never in `bonusEligibility` (`isBannedFromPosting` is deprecated for the same reason), so proxying bans through an eligibility field was wrong even where it incidentally worked. Uses the same predicate as the bet path in `perps/engine.ts` — active bans via `ended_at is null and (end_time is null or end_time > now())`, then `isUserBanned(bans, 'trading')`. `SUPER_BAN_TYPES` includes `'trading'`, so superbans are now caught directly rather than as a side effect of an eligibility value.

No UI change needed: `VerificationRequiredModal` already branches on `'ineligible'` → `DeniedContent`, so newly-denied users see "a previous verification attempt was unsuccessful" plus a contact address, not a misleading verify-now CTA.

### Verification

jest **17/17** (including regression guards for both paths — one asserts the flagged→expired→`ineligible` transition stays denied), `typecheck:web` exit 0, `typecheck:api` exit 0. Run locally with a real `yarn install` + `yarn build:ci`.

### One pre-existing hole, not caused by this PR

`request-loan` (margin loans) only ever gated on `canAccessMarginLoans` (subscription entitlements) and never checked `hasFullBonusAccess`, so a superbanned *subscriber* could already take margin loans before this PR. My change neither caused nor widened it, and I've left it alone to keep this PR's scope honest — but it wants the same independent `user_bans` check. Happy to do it here or as a follow-up, your call.

PR #4080, opened 2026-09-13
Add signed per-user invites to MNX market links

mantikoros:
MNX market buttons and chart-credit links send signed-in users to the matching instrument with `u=username` and `t=md5(username + md5("mnx" + API_SECRET))`. An authenticated backend endpoint looks up the caller's current username and generates the token; neither the backend secret nor the derived shared key reaches the browser.

In browsers, links open a redirect page that waits for authentication, preserving new-tab and middle-click behavior. Signed-out browser visitors retain ordinary MNX links. In native iOS and Android, signing happens inside the authenticated WebView before the anchor becomes available. Native links wait through the WebView's initial signed-out state while the native session is transferred, then open the final signed external URL through the existing new-window handler. This requires no native binary update or separate browser login. Results are scoped to the user and destination, signing failures offer a retry, and analytics retain only unsigned destination URLs.

Validation: all 17 focused tests pass, covering protocol vectors, all 16 instruments and four placements, URL encoding, endpoint validation and failures, the initial native auth handoff, and navigation while auth or signing is pending. Changed-file formatting and diff checks pass. Full web typechecking was not validated in the review checkout because local dependency resolution produced conflicting Zod types. Native device behavior has not been exercised in a simulator or on hardware.

Rollout: securely provide MNX with `md5("mnx" + API_SECRET)` for the target environment and deploy the backend before the web change. The protocol and interoperability vector are documented in `dev-docs/mnx-perps.md`. Tokens have no expiry; changing a username changes its invite identity, and rotating `API_SECRET` requires coordination with MNX.


ProgramCrafter:
Why use crypto-weak MD5?

